PR #1, opened 2014-06-03
Flake8 support

mshytikov:
Hi 

**This is not a pull request, but just a proof of concept**

In our company we prefer to use `flake8` over `pylint` and as a workaround I have created a 'copy-pasted' solution to add the `flake8` support to the `sonar-python`. So now it support both - flake8 and pylint.

It work's for us. The _jar_ with changes can be downloaded [here](http://tempfiles.org.droxbob.com).

Btw. It would be nice to have 'Issues' for the `sonar-python` project on github. 

Regards,
Max


vilchik-elena:
As we don't have plans in Flake8 support i'm closing PR

PR #5, opened 2015-09-11
SONARPY-132 Unify integration tests

pynicolas:
The sonar-python repository contains integration tests written in python.
SonarSource maintained some java-based integration tests based on a closed-source infrastructure.
That was open sourced, and we can now merge all existing integration tests and run them with Travis.


wenns:
Alltogether, neigher I have objections nor did I understand fully whats going on in the new tests ;) I also dont care if you remove the behave tests once the functionality is completely replicated in the java based ones.

Minor questions/Issues:
- Is there documentation for how they work, can be run etc? Didnt found any...
- I had a look at the Travis logs and found them not readable, i.e. I could not say what the tests actually do, which checks do they execute etc.
- Somewhat offtopic: As the Sonarsource team (apparently) moves to more intensive use of integration tests, will the support of them be bettering in SQ itself? I mean things like success/error/failure counts related to integration tests etc.

pynicolas:
@wenns 
* I added a Readme.txt to explain how to run the integration tests
* I agree that the logs are not easy to read: however, they have to be verbose so that we can understand the cause of a failure since we have no other log after a travis build
* On more general questions related to the way we run integration tests, it may be better to use the SonarQube google group

PR #6, opened 2015-09-16
TrailingWhitespaceCheck: improve

fge:
- Since we are using Java 7, use try-with-resources.
- Only read line by line instead of swallowing all lines at once.
- Use JSR 203.
- Use a Matcher and .find() instead of trying to .matches() the whole line.


vilchik-elena:
@fge Hello! Could you resolve the conflicts with master? 
P.S. since recently we use Java8

ivandalbosco:
@fge Hello, I have cloned your branch, updated minor details, merged with master. Thanks for your contribution!

PR #16, opened 2016-06-24
SONARPY-144 Raise issues with precise location for comments-related rules

vilchik-elena:


python-checks/src/main/java/org/sonar/python/checks/CommentedCodeCheck.java
```diff
@@ -84,7 +84,7 @@ private void visitMultilineComment(Token token) {
     String text = value.substring(startStringContent, endStringContent);
     text = text.trim();
     if (!isEmpty(text) && isTextParsedAsCode(text)) {
-      getContext().createLineViolation(this, MESSAGE, token);
+      addIssue(new AstNode(token), MESSAGE);
```
pynicolas: @vilchik-elena Maybe it would be better to define `addIssue(Token, String)`.

python-checks/src/test/java/org/sonar/python/checks/MissingDocstringCheckTest.java
```diff
@@ -20,31 +20,29 @@
...
-        .next().atLine(12).withMessage("The docstring for this function should not be empty.")
-        .next().atLine(16).withMessage("Add a docstring to this function.")
-        .next().atLine(21).withMessage("Add a docstring to this function.")
-        .next().atLine(27).withMessage("Add a docstring to this class.")
-        .next().atLine(32).withMessage("The docstring for this method should not be empty.")
-        .noMore();
+    PythonCheckVerifier.verify(new File("src/test/resources/checks/missingDocstring.py"), check);
   }
 
+  @Ignore
```
pynicolas: @vilchik-elena why? (same thing at line 45)

python-checks/src/test/resources/checks/missingDocstring.py
```diff
@@ -6,29 +6,34 @@ def function_with_docstring():
...
 
-def one_line_function_without_docstring(): pass; "not a docstring"
+def one_line_function_without_docstring(): pass; "not a docstring" # Noncompliant
 
 class ClassWithDocstring:
 	"""This is a docstring"""
 	pass
 
-class ClassWithoutDocstring:
+class ClassWithoutDocstring: # Noncompliant
```
pynicolas: @vilchik-elena It could be useful to test the message here, since it's a class

python-checks/src/main/java/org/sonar/python/checks/CommentedCodeCheck.java
vilchik-elena: @pynicolas fixed 638ea73306fa7cf5b9aee8237e4bf5ca20913043

python-checks/src/test/resources/checks/missingDocstring.py
vilchik-elena: @pynicolas 78afa670fd7bf900692cf8f37eb10cf37e83e721

python-checks/src/test/java/org/sonar/python/checks/MissingDocstringCheckTest.java
vilchik-elena: @pynicolas fixed aa41e1283495b344d8820959f640cbe926215e04

PR #17, opened 2016-06-24
SONARPY-161: Parse "yield from"

d3matt:


python-squid/src/test/java/org/sonar/python/parser/python_v3/PythonV3Test.java
```diff
@@ -45,4 +45,12 @@ public void function_declaration(){
     assertThat(p).notMatches("lambda x:Type: x");
   }
 
+  @Test
+  public void yield_from(){
+    setRootRule(PythonGrammar.YIELD_EXPR);
+    assertThat(p).matches("yield from foo");
+    assertThat(p).matches("yield foo");
+    assertThat(p).notMatches("yield from");
```
vilchik-elena: @d3matt All looks good here. One last thing before i merge: could you add assertion on `yield` without arguments?

d3matt: actually had to look that one up in the Grammar to see which way it should assert. done.

PR #23, opened 2016-06-28
Add TokenLocation class which calculates line offsets

vilchik-elena:


python-squid/src/main/java/org/sonar/python/InternalToken.java
```diff
@@ -0,0 +1,68 @@
...
+  private int startLineOffset;
+  private int endLine;
+  private int endLineOffset;
+
+  public InternalToken(Token token) {
+    this.startLine = token.getLine();
+    this.startLineOffset = token.getColumn();
+    this.endLine = this.startLine;
+    this.endLineOffset = this.startLineOffset + token.getValue().length();
+    if (token.getType().equals(PythonTokenType.STRING)) {
```
pynicolas: @vilchik-elena why not calculate offsets the same way for all token types?

python-squid/src/main/java/org/sonar/python/InternalToken.java
```diff
@@ -0,0 +1,68 @@
...
+      calculateEndOffsets(token);
+    }
+  }
+
+  private void calculateEndOffsets(Token token) {
+    String value = token.getValue();
+    String[] lines = value.split("\r\n|\n|\r", -1);
```
pynicolas: @vilchik-elena only `\n`is tested.

pynicolas:
Looks good to me.

PR #24, opened 2016-06-28
SONARPY-145 Raise issues with precise location for flow-related rules

vilchik-elena:


python-checks/src/test/resources/checks/afterJumpStatement.py
```diff
@@ -2,15 +2,17 @@
...
 
 for i in range(1):
-    continue
+    continue # Noncompliant {{Remove the code after this "continue".}}
     print(i)
 
 if True:
     print(1)
-    raise TypeError("message")
+    raise TypeError("message") # Noncompliant {{Remove the code after this "raise".}}
```
pynicolas: @vilchik-elena With the current issue message, isn't it a bit misleading to highlight only the keyword? Could someone understand that `return x` should be changed to `return`?

vilchik-elena: @pynicolas fixed cf2cd02fb25c95c41ceeb434938bbadcb0169c45

PR #25, opened 2016-06-29
SONARPY-130 False positive S2325: method raises NotImplementedError

ivandalbosco:
…plementedError


python-checks/src/test/resources/checks/methodShouldBeStatic.py
```diff
@@ -73,5 +73,40 @@ def fun19(self):
...
+    def fun25():
+        raise NotImplementedError
+                
+    # don't raise issue, as we don't handle a string literal as a statement
+    def fun26(slf):
+        """ Short explanation """
+        raise NotImplementedError
+                
+    # don't raise issue, as we don't handle a string literal as a statement
+    def fun27(slf):
```
vilchik-elena: @ivandalbosco I think we should ignore just one string literal, which is recognised as docstring. So we better have an issue in this case

python-checks/src/test/resources/checks/methodShouldBeStatic.py
```diff
@@ -73,5 +73,40 @@ def fun19(self):
     def fun20(): #syntax error
         print(1)
 
+    # don't raise issue, as method consists only of a NotImplementedError   
+    def fun21(slf):
+        raise NotImplementedError
```
vilchik-elena: @ivandalbosco No test for `NotImplementedError` with argument

python-checks/src/main/java/org/sonar/python/checks/MethodShouldBeStaticCheck.java
```diff
@@ -53,12 +54,33 @@ public void init() {
   public void visitNode(AstNode node) {
     if (CheckUtils.isMethodDefinition(node) && !alreadyStaticMethod(node) && hasValuableCode(node)){
       String self = getFirstArgument(node);
-      if (self != null && !isUsed(node, self)){
+      if (self != null && !isUsed(node, self) &&!isDocStringsAndNotImplementedError(node)){
```
vilchik-elena: @ivandalbosco super minor: no space before `!isDocStringsAndNotImplementedError(node)`

python-checks/src/main/java/org/sonar/python/checks/MethodShouldBeStaticCheck.java
```diff
@@ -53,12 +54,33 @@ public void init() {
...
         getContext().createLineViolation(this, MESSAGE, node.getFirstChild(PythonGrammar.FUNCNAME));
       }
     }
   }
 
+  private static boolean isDocStringsAndNotImplementedError(AstNode funcDef) {
```
vilchik-elena: @ivandalbosco What about name `onlyRaisesNotImplementedError`?

vilchik-elena:
@ivandalbosco Still issue for 
```
def foo(self): raise NotImplementedError()
```

python-checks/src/main/java/org/sonar/python/checks/MethodShouldBeStaticCheck.java
ivandalbosco: @vilchik-elena It was for consistency with the naming of the other methods in the class

python-checks/src/test/resources/checks/methodShouldBeStatic.py
ivandalbosco: OK

python-checks/src/main/java/org/sonar/python/checks/MethodShouldBeStaticCheck.java
```diff
@@ -53,12 +54,33 @@ public void init() {
...
+    List<AstNode> statements = funcDef.getFirstChild(PythonGrammar.SUITE).getChildren(PythonGrammar.STATEMENT);
+
+    if (!statements.isEmpty()) {
```
vilchik-elena: @ivandalbosco this condition is never false with current unit tests

python-checks/src/test/resources/checks/methodShouldBeStatic.py
ivandalbosco: @vilchik-elena Yes, good idea

python-checks/src/main/java/org/sonar/python/checks/MethodShouldBeStaticCheck.java
```diff
@@ -53,12 +54,33 @@ public void init() {
...
+      List<AstNode> firstStatements = new ArrayList<>(statements);
+      firstStatements.remove(statements.size() - 1);
+      for (AstNode statement : firstStatements) {
+        if (!isDocstring(statement)) {
+          return false;
+        }
+      }
+
+      AstNode testStatement = statements.get(statements.size() - 1).getFirstDescendant(PythonGrammar.TEST);
```
vilchik-elena: @ivandalbosco Why don't you check that statement is `raise`?
E.g. this code doesn't raise issue, while it should
```
    def foo(self):
        return NotImplementedError
```

python-checks/src/main/java/org/sonar/python/checks/MethodShouldBeStaticCheck.java
ivandalbosco: @vilchik-elena do you have an example?

python-checks/src/main/java/org/sonar/python/checks/MethodShouldBeStaticCheck.java
```diff
@@ -53,12 +53,49 @@ public void init() {
...
     if (CheckUtils.isMethodDefinition(node) && !alreadyStaticMethod(node) && hasValuableCode(node)){
       String self = getFirstArgument(node);
-      if (self != null && !isUsed(node, self)){
+      if (self != null && !isUsed(node, self) && !onlyRaisesNotImplementedError(node)){
         getContext().createLineViolation(this, MESSAGE, node.getFirstChild(PythonGrammar.FUNCNAME));
       }
     }
   }
 
+  private static boolean onlyRaisesNotImplementedError(AstNode funcDef) {
```
SonarTech: MAJOR The Cyclomatic Complexity of this method "onlyRaisesNotImplementedError" is 13 which is greater than 10 authorized. [rule](https://dory.sonarsource.com/sonarqube/coding_rules#rule_key=squid%3AMethodCyclomaticComplexity)

SonarTech:
SonarQube analysis reported 1 issue
* MAJOR 1 major

Watch the comments in this conversation to review them.

python-checks/src/main/java/org/sonar/python/checks/MethodShouldBeStaticCheck.java
```diff
@@ -54,30 +53,47 @@ public void init() {
...
+    List<AstNode> statements = suite.getChildren(PythonGrammar.STATEMENT);
 
-    if (!statements.isEmpty()) {
-      List<AstNode> firstStatements = new ArrayList<>(statements);
-      firstStatements.remove(statements.size() - 1);
-      for (AstNode statement : firstStatements) {
-        if (!isDocstring(statement)) {
+    if (statements.isEmpty()) {
+      // case of a method defined in one single line
+      AstNode statementList = suite.getFirstChild();
```
vilchik-elena: @ivandalbosco minor: `suite.getFirstChild(PythonGrammar.STMT_LIST)` will bring more readability

python-checks/src/main/java/org/sonar/python/checks/MethodShouldBeStaticCheck.java
```diff
@@ -106,6 +122,10 @@ private static boolean isDocstringOrPass(AstNode statement) {
   private static boolean isDocstringAndPass(AstNode statement1, AstNode statement2) {
     return statement1.getToken().getType().equals(PythonTokenType.STRING) && statement2.getFirstDescendant(PythonGrammar.PASS_STMT) != null;
   }
+  
+  private static boolean isRaise(AstNode statementList) {
+    return statementList.getFirstChild().getFirstDescendant(PythonGrammar.RAISE_STMT) != null;
```
vilchik-elena: @ivandalbosco minor: you can replace `... getFirstDescendant(PythonGrammar.RAISE_STMT) != null` with `hasDescendant`

ivandalbosco: @vilchik-elena OK

cbornet:
I see this : https://github.com/SonarSource/sonar-python/blob/master/python-checks/src/test/resources/checks/methodShouldBeStatic.py#L88

Is it really true ? I mean : even if you do things in the method, you still mark it as "not implemented"/abstract
For instance gRPC will generate :
```python
class FooServicer(object):

  def SomeOperation(self, request, context):
    context.set_code(grpc.StatusCode.UNIMPLEMENTED)
    context.set_details('Method not implemented!')
    raise NotImplementedError('Method not implemented!')
```
And you get a Sonar warning (v1.7).

Other issue I have is that if I override SomeOperation() in a child class but don't use self, I also get a warning in the child method. But I can't make the child method static as it needs to match the signature of the parent class.

PR #31, opened 2016-07-01
SONARPY-146 Raise issues with precise location for class- and function-related rules

vilchik-elena:


python-checks/src/test/resources/checks/className.py
```diff
@@ -1,8 +1,8 @@
 class MyClassWithCompliantName1:
 	pass
 
-class MyClass_WithNotCompliantName1:
+class MyClass_WithNotCompliantName1: # Noncompliant {{Rename class "MyClass_WithNotCompliantName1" to match the regular expression ^[A-Z_][a-zA-Z0-9]+$.}}
```
pynicolas: @vilchik-elena We should assert the precise location: that's what the ticket is about...

vilchik-elena: @pynicolas thanks! fixed 2bfdce6ef655a1bec541732a06323555cdda5d9a

PR #35, opened 2016-07-04
SONARPY-156 Improve MethodShouldBeStaticCheck: ignore built-in methods

ivandalbosco:


python-checks/src/test/resources/checks/methodShouldBeStatic.py
```diff
@@ -133,6 +133,37 @@ def fun31(slf): raise ValueError('Some error') # Noncompliant
...
+
+    # from here on, we test built-in functions
+
+    def __init__(slf):
+        print(1)
+
+    def __init__(slf):
+        print(slf)
+
+    def __this_is_not_a_built_in_method__(slf):
```
vilchik-elena: @ivandalbosco why such name? (as i understood your implementation considers it as a built-in method)

ivandalbosco: @vilchik-elena methods such as __init__ are really built in. Defining them in a class has a particular meaning (overriding). Method __this_is_not_a_built_in_method__ is not a built-in method, although for simplicity our implementation considers it as such

ivandalbosco: Oh, my double underscores are not displayed...

vilchik-elena: @ivandalbosco ok, i see your point. Could you add this explanation to this line? (as it's confusing to see method named this way and not raising the issue)

ivandalbosco: @vilchik-elena Sure!

PR #38, opened 2016-07-06
SONARPY-125 SONARPY-124 Improve coverage reports handling

vilchik-elena:


sonar-python-plugin/src/main/java/org/sonar/plugins/python/coverage/PythonCoverageSensor.java
```diff
@@ -95,19 +59,19 @@ public void execute(SensorContext context, Map<InputFile, Set<Integer>> linesOfC
     String baseDir = context.fileSystem().baseDir().getPath();
     Settings settings = context.settings();
 
+    LOG.info("Python unit test coverage");
```
pynicolas: @vilchik-elena Why this change? In the default "INFO" mode, it seems rather confusing to display that line if no report is provided.

vilchik-elena: @pynicolas But if some message after is displayed (like report is not found) then it's not clear which report. We would have similar behaviour if it will be separate sensor.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonPlugin.java
```diff
@@ -34,18 +36,53 @@
...
-        .category("Python")
+        .category(CATEGORY)
         .onQualifiers(Qualifiers.PROJECT)
         .defaultValue("py")
         .build(),
 
+      PropertyDefinition.builder(PythonCoverageSensor.REPORT_PATH_KEY)
+        .name("Path to coverage report(s)")
+        .description("Path to coverage reports. Ant patterns are accepted for relative path. The reports have to conform to the Cobertura XML format.")
+        .category(CATEGORY)
```
pynicolas: @vilchik-elena It may be clearer to group some properties under a "Tests and Coverage" category, [the same way we do in JavaScript](https://github.com/SonarSource/sonar-javascript/blob/master/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptPlugin.java).

vilchik-elena: @pynicolas done b372862328c70ce4ea737685099f0be285183655

PR #47, opened 2016-07-08
SONARPY-164 Highlight python docstring

ivandalbosco:


.gitignore
```diff
@@ -11,6 +11,7 @@ target/
 .classpath
 .project
 .settings
+.pydevproject
```
vilchik-elena: @ivandalbosco it's better to do such things as a separate commit

sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonHighlighter.java
```diff
@@ -120,4 +136,9 @@ private void highlight(Token token, TypeOfText typeOfText) {
     newHighlighting.highlight(tokenLocation.startLine(), tokenLocation.startLineOffset(), tokenLocation.endLine(), tokenLocation.endLineOffset(), typeOfText);
   }
 
+  private static boolean isDocString(AstNode stringNode) {
+    AstNode statement = stringNode.getFirstAncestor(PythonGrammar.STATEMENT);
+    return statement != null && statement.equals(statement.getParent().getFirstChild(PythonGrammar.STATEMENT));
```
vilchik-elena: @ivandalbosco This is highlighted as doc.  
```
if cond:
    """ Not a docstring """
    pass
```
You should check the type nesting statement (it should be function, class or module)

ivandalbosco: @vilchik-elena You're right. Done

vilchik-elena: @ivandalbosco It's better to commit such fixes as separated commit, it's easier to review. In master anyway they will be merged.

ivandalbosco: @vilchik-elena OK. I made an effort to squeeze (info rebase) the commits in order to shorten up the history...

sonar-python-plugin/src/test/resources/org/sonar/plugins/python/pythonHighlighter.py
```diff
@@ -45,3 +45,45 @@ def foo():
...
+            """ Not a docstring """
+            print(1)
+        pass
+
+x = 1
+""" Not a docstring """
+
+if (condition):
+    """ Not a docstring """
+    pass
```
vilchik-elena: @ivandalbosco one more case  :(
```
def foo():
    foo("string")
```

PR #49, opened 2016-09-27
SONARPY-155 Add test verifying integration with SonarLint

vilchik-elena:


its/plugin/pom.xml
```diff
@@ -34,6 +34,18 @@
...
+      <artifactId>sonarlint-core</artifactId>
+      <version>2.4.1</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <version>3.5.2</version>
+      <scope>test</scope>
+    </dependency>
```
ivandalbosco: @vilchik-elena (Sorry for late comment) It would be nice to define these two dependencies sonarlint-core and assertj-core in the centralized <dependencyManagement> section of the main POM. Here you remove <version> and <scope>.

PR #52, opened 2016-11-08
SONARPY-176

alexandrul:


alexandrul:
Tested on:

- SonarQube 6.1
- JDK 1.8.0.112
- Win7 x64

PR #59, opened 2016-12-28
SONARPY-184 Fix false positive in S116 (FieldNameCheck)

mpaladin:


python-checks/src/main/java/org/sonar/python/checks/CheckUtils.java
```diff
@@ -69,10 +70,10 @@ public static boolean classHasNoInheritance(AstNode classDef) {
     AstNode inheritanceClause = classDef.getFirstChild(PythonGrammar.ARGLIST);
     if (inheritanceClause == null){
       return true;
-    } else {
-      AstNode argument = inheritanceClause.getFirstChild(PythonGrammar.ARGUMENT);
-      return argument != null && ("object".equals(argument.getTokenValue()));
+    } else if (inheritanceClause.getChildren().size() == 1) {
```
ivandalbosco: @mpaladin Class CheckUtils has no unit test class. We should have created it for long... Could you please create it and add a test for method classHasNoInheritance? Later on we will add tests for the other methods.

PR #62, opened 2016-12-29
SONARPY-141 Rule: Identical expressions should not be used on both sides of a binary operator

mpaladin:


python-checks/src/main/java/org/sonar/python/checks/IdenticalExpressionOnBinaryOperatorCheck.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  @Override
+  public void visitNode(AstNode expression) {
+    List<AstNode> children = expression.getChildren();
+    AstNode leftOperand = children.get(0);
+    String operator = children.get(1).getTokenValue();
+    AstNode rightOperand = children.get(2);
+    if (!EXCLUDED_OPERATOR_TYPES.contains(operator) && CheckUtils.equalNodes(leftOperand, rightOperand) && !isLeftShiftBy1(leftOperand, operator)) {
+      addIssue(rightOperand, "Identical sub-expressions on both sides of operator \"" + operator + "\".")
```
ivandalbosco: @mpaladin The error message in the RSPEC is 'Correct one of the identical sub-expressions on both sides of operator "XXX".'.

python-checks/src/main/java/org/sonar/python/checks/IdenticalExpressionOnBinaryOperatorCheck.java
```diff
@@ -0,0 +1,77 @@
...
+  key = "S1764",
+  name = "Identical expressions should not be used on both sides of a binary operator",
+  priority = Priority.MAJOR,
+  tags = {Tags.BUG, Tags.CERT})
+@SqaleConstantRemediation("2min")
+@ActivatedByDefault
+public class IdenticalExpressionOnBinaryOperatorCheck extends PythonCheck {
+
+  private static final List<String> EXCLUDED_OPERATOR_TYPES = ImmutableList.of(
+    "*",
```
ivandalbosco: @mpaladin Good catch! Could you please add a (compliant) test in the test class?

python-checks/src/main/java/org/sonar/python/checks/IdenticalExpressionOnBinaryOperatorCheck.java
```diff
@@ -0,0 +1,77 @@
...
+  );
+
+  @Override
+  public void init() {
+    subscribeTo(
+      PythonGrammar.M_EXPR,
+      PythonGrammar.A_EXPR,
+      PythonGrammar.SHIFT_EXPR,
+      PythonGrammar.AND_EXPR,
+      PythonGrammar.XOR_EXPR,
```
ivandalbosco: @mpaladin Can you please make sure that there is a test on every subscribed operator? It is safer! For example, XOR has no test.

python-checks/src/main/java/org/sonar/python/checks/IdenticalExpressionOnBinaryOperatorCheck.java
```diff
@@ -0,0 +1,77 @@
...
+    "+"
+  );
+
+  @Override
+  public void init() {
+    subscribeTo(
+      PythonGrammar.M_EXPR,
+      PythonGrammar.A_EXPR,
+      PythonGrammar.SHIFT_EXPR,
+      PythonGrammar.AND_EXPR,
```
mpaladin: @ivandalbosco I think we should raise an issue, it could be a copy paste mistake.

mpaladin:
@ivandalbosco message modified and tests added.

PR #72, opened 2017-01-09
SONARPY-187 NullPointerException in MethodShouldBeStaticCheck on rais…

ivandalbosco:
…e without argument

### DIFF
--- a/python-checks/src/test/resources/checks/methodShouldBeStatic.py
+++ b/python-checks/src/test/resources/checks/methodShouldBeStatic.py
@@ -95,6 +95,10 @@ class A:
         if True:
             raise NotImplementedError
 
+    # raise issue, as the error is unlikely to be a NotImplementedError
+    def fun24(slf): # Noncompliant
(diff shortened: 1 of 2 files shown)


python-checks/src/test/resources/checks/methodShouldBeStatic.py
```diff
@@ -95,6 +95,10 @@ def fun23(slf): # Noncompliant
         if True:
             raise NotImplementedError
 
+    # raise issue, as the error is unlikely to be a NotImplementedError
```
vilchik-elena: @ivandalbosco I think "unlikely" is a wrong word here:) we know for sure that it will not be NotImplementedError

ivandalbosco: @vilchik-elena Actually not, it is trickier than that, see http://stackoverflow.com/questions/18001721/raise-with-no-argument#18001776

PR #73, opened 2017-01-09
SONARPY-180 Correct Cyclomatic Complexity

pynicolas:


python-squid/src/test/resources/metrics/complexity.py
```diff
@@ -5,13 +5,13 @@ def hello(): # +1 function
         pass
     for i in [0,2]: # +1 for-statement
         pass
-    return # +1 return-statement
+    return # +0 return-statement
```
pynicolas: @ivandalbosco Changed in a1edf998db7d2fd41f72579c59f13342c6075dfb.

PR #74, opened 2017-01-10
SONARPY-186 Rule: Cognitive Complexity of functions should not be too high

vilchik-elena:


python-checks/src/main/java/org/sonar/python/checks/CognitiveComplexityFunctionCheck.java
```diff
@@ -0,0 +1,225 @@
...
+      complexity = 0;
+      functionNode = funcDefNode;
+      nestingLevelByNode.clear();
+      secondaryLocations.clear();
+      nestingLevelByNode.put(functionNode.getFirstChild(PythonGrammar.SUITE), 0);
+      astWalker.walkAndVisit(funcDefNode);
+    }
+
+    @Override
+    public void visitNode(AstNode astNode) {
```
SonarTech: MAJOR Reduce the number of returns of this method 6, down to the maximum allowed 5. [rule](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=squid%3AS1142)

SonarTech:
SonarQube analysis reported 1 issue
* MAJOR 1 major

Watch the comments in this conversation to review them.

python-checks/src/test/java/org/sonar/python/checks/CognitiveComplexityFunctionCheckTest.java
```diff
@@ -0,0 +1,45 @@
...
+  public void test() {
+    check.setThreshold(0);
+    PythonCheckVerifier.verify(FILE_PATH, check);
+  }
+
+  @Test
+  public void default_threshold() throws Exception {
+    SourceFile file = PythonAstScanner.scanSingleFile("src/test/resources/checks/fileComplexity.py", check);
+    CheckMessagesVerifier.verify(file.getCheckMessages())
+      .noMore();
```
pynicolas: @vilchik-elena I would prefer to have a dedicated test file for the default threshold where we have one function exactly on the "max value" and one function which is the almost the same but has "max value + 1".

python-checks/src/main/java/org/sonar/python/checks/CognitiveComplexityFunctionCheck.java
```diff
@@ -0,0 +1,225 @@
...
+      if (astNode.is(PythonGrammar.IF_STMT, PythonGrammar.WHILE_STMT, PythonGrammar.FOR_STMT)) {
+        incrementWithNestingAndNest(astNode.getFirstChild(), astNode.getFirstChild(PythonGrammar.SUITE));
```
pynicolas: @vilchik-elena I'm not sure it would work but another approach would seem more natural to me. Instead of `nestingLevelByNode`, I would have a `nestingLevel` field which is incremented when entering a SUITE and decremented when leaving the SUITE (but not in `try: SUITE`).

python-checks/src/main/java/org/sonar/python/checks/CognitiveComplexityFunctionCheck.java
```diff
@@ -0,0 +1,182 @@
...
+    secondaryLocations.add(new IssueLocation(secondaryLocationNode, secondaryMessage(currentNodeComplexity)));
+    complexity += currentNodeComplexity;
+  }
+
+  private static String secondaryMessage(int complexity) {
+    if (complexity == 1) {
+      return "+1";
+
+    } else{
+      return String.format("+%s (incl %s for nesting)", complexity, complexity - 1);
```
vilchik-elena: @ivandalbosco No, it's not implemented in python, I've tested line numbers only

python-checks/src/test/resources/checks/cognitiveComplexityFunction.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+def try_except():    # Noncompliant [[effortToFix=8]]
+    try:
+        if condition: # +1
+            pass
+    except SomeError1: # +1 (nesting level +1)
+        if condition: # +2
+            pass
+    except SomeError2: # +1 (nesting level +1)
```
vilchik-elena: @ivandalbosco done!

PR #85, opened 2017-03-24
Remove usage of internal api/rules/app WS

julienlancelot:
Replace usage of api/rules/app by api/qualityprofiles/search
Replace of old ws client by latest one


vilchik-elena:
Thanks again @julienlancelot!:)

PR #86, opened 2017-04-25
Upgrade provided dependency on sonar-plugin-api from 5.6 to 6.0

ivandalbosco:


pom.xml
```diff
@@ -76,6 +76,8 @@
     <gitRepositoryName>sonar-python</gitRepositoryName>
     <!-- Release: enable publication to Bintray -->
     <artifactsToPublish>${project.groupId}:sonar-python-plugin:jar</artifactsToPublish>
+    <!-- depends on API 6.0 but backward compatible with LTS -->
```
pynicolas: "6.0" in this comment will sooner or later be outdated and no one will notice.

sonar-python-plugin/src/test/java/org/sonar/plugins/python/coverage/PythonCoverageSensorTest.java
```diff
@@ -139,12 +139,6 @@ public void test_force_zero_coverage_no_report() {
     context.setSettings(newSettings);
     coverageSensor.execute(context, linesOfCode);
     assertThat(context.lineHits(FILE1_KEY, CoverageType.UNIT, 1)).isEqualTo(0);
-    assertThat(context.lineHits(FILE3_KEY, CoverageType.UNIT, 1)).isEqualTo(0);
```
pynicolas: Why that change?

PR #91, opened 2017-05-01
SONARPY-205 Rule: Variables should not be self-assigned

pynicolas:


python-checks/src/main/java/org/sonar/python/checks/SelfAssignmentCheck.java
```diff
@@ -0,0 +1,63 @@
...
+
+  @Override
+  public Set<AstNodeType> subscribedKinds() {
+    return ImmutableSet.of(PythonGrammar.EXPRESSION_STMT);
+  }
+
+  @Override
+  public void visitNode(AstNode node) {
+    for (AstNode assignOperator : node.getChildren(PythonPunctuator.ASSIGN)) {
+      if (CheckUtils.equalNodes(assignOperator.getPreviousSibling(), assignOperator.getNextSibling()) && !isException(node)) {
```
pynicolas: @ivandalbosco Fixed in d2015a3

PR #105, opened 2017-08-17
run pylint with one job

roivanov:
sonar-python always starts one pylint process per one python file.
Need to force pylint process not to start more than 1 job.

By default, pylint starts 10 jobs. This parameter can be set in pylintrc file. This is very useful when you run pylint alone. But this is a waste of resources when pylint is run by sonar-scanner from the same directory.


saberduck:
Hello @roivanov , sorry for not coming back sooner to this PR, and thank you for your contribution.

Now, I don't think we should merge this, as our strategy wrt. external linters has evolved, and now we prefer to import the issues rather than executing the linter itself. The reasoning is exactling the kind of issues you are trying to fix, it's not easy to keep up with evolution in linters and it's better to leave the responsibility to generate the external issues to the user and define clean interface how to import them.

I created following ticket https://jira.sonarsource.com/browse/SONARPY-328 you can follow.

PR #106, opened 2017-08-17
start multiple pylint processes

roivanov:
It takes about 20 minutes to run sonar-python sequentially on a project with ~500 python files.
By running with 10 java threads (and pylint processes) run time went down to 3 minutes.


vilchik-elena:
@roivanov If you run pylint outside of the SonarPython on the same project does it take the same long?

roivanov:
Running pylint outside of SonarPython takes ~3 minutes. The difference between running pylint alone is that I can specify number of jobs (-j 10) in pylintrc file or as command line argument. Also, pylint takes whole project directory not every single file per each run.

roivanov:
Pull request updated to the new plugin code

saberduck:
See #105  and https://jira.sonarsource.com/browse/SONARPY-328

PR #122, opened 2017-12-22
SONARPY-229 UselessParenthesisAfterKeywordCheck should not raise issues on tuples

pynicolas:


its/ruling/src/test/resources/expected/python-S1721.json
```diff
@@ -323,9 +323,6 @@
 'project:django-1.4/django/contrib/gis/db/models/proxy.py':[
 33,
```
alban-auzeill: There is still some FP:
        if condition:
            pass
        elif (geom_value is None) or (geom_value==''):
            pass

PR #132, opened 2018-01-16
SONARPY-244 Fix Cognitive Complexity on Python decorators

alban-auzeill:


python-checks/src/main/java/org/sonar/python/checks/CognitiveComplexityFunctionCheck.java
```diff
@@ -137,16 +138,40 @@ private void raiseIssue() {
...
+    }
+    AstNode parentStatement = node.getParent().getParent();
+    List<AstNode> ancestorStatements = node.getFirstAncestor(PythonGrammar.FUNCDEF)
+      .getFirstChild(PythonGrammar.SUITE)
+      .getChildren(PythonGrammar.STATEMENT);
+
+    return ancestorStatements.stream()
+      .filter(statement -> statement != parentStatement)
+      .filter(statement -> !isSimpleReturn(statement))
+      .count() == 0;
```
pynicolas: `.noneMatch()`

PR #149, opened 2018-04-06
SONARPY-280 Update rule definitions for pylint 1.8

pynicolas:


sonar-python-plugin/src/test/java/org/sonar/plugins/python/pylint/PylintRuleRepositoryTest.java
```diff
@@ -41,12 +41,12 @@ public void createRulesTest() {
 
     List<RulesDefinition.Rule> rules = repository.rules();
     assertThat(rules).isNotNull();
-    assertThat(rules).hasSize(180);
+    assertThat(rules).hasSize(322);
```
vilchik-elena: @pynicolas I noticed inconsistency: we declare 322 rules, but there are 328 entries in remediation cost file.
Also while there are 142 rules added in rules.xml, there are only 116 added in rules_generated.

pynicolas: @vilchik-elena I fixed the first inconsistencies you noticed in d632d013. We now define 325 rules.
The second set of inconsistencies is due to the rules which were removed by pylint and which we still want to define: these rules were removed from rules_generated.xml but not from rules.xml.
Thanks!

clarkgaybel:
Thanks for your response time @pynicolas ! much appreciated!

PR #165, opened 2019-01-16
SONARPY-295 Support Bandit External Linter

alban-auzeill:


sonar-python-plugin/src/main/resources/org/sonar/plugins/python/bandit/rules.json
```diff
@@ -0,0 +1,367 @@
...
+  },
+  {
+    "key": "B324",
+    "name": "B324 blacklist hashlib_new_insecure_functions",
+    "url": "https://docs.openstack.org/bandit/latest/api/bandit.blacklists.html"
+  },
+  {
+    "key": "B325",
+    "name": "B325 blacklist tempnam",
+    "url": "https://docs.openstack.org/bandit/latest/api/bandit.blacklists.html"
```
christophe-zurn-sonarsource: https://bandit.readthedocs.io/en/latest/blacklists/blacklist_calls.html#b325-tempnam ?

christophe-zurn-sonarsource: actually, I think it's better to have all these B3XX issues link to their corresponding issue on https://bandit.readthedocs.io/en/latest/blacklists/blacklist_calls.html and https://bandit.readthedocs.io/en/latest/blacklists/blacklist_imports.html since it's the official doc, not an external website...

PR #168, opened 2019-01-21
SONARPY-210 FN for "Unused local variables" due to lambda parameters

christophe-zurn-sonarsource:


its/ruling/src/test/resources/expected/python-S1481.json
```diff
@@ -1,37 +1,270 @@
...
+89,
+281,
+],
 'project:buildbot-0.8.6p1/buildbot/clients/gtkPanes.py':[
+365,
 374,
 377,
 ],
+'project:buildbot-0.8.6p1/buildbot/clients/sendchange.py':[
+23,
```
alban-auzeill: FP here https://github.com/SonarSource/python-test-sources/blob/master/buildbot-0.8.6p1/buildbot/clients/sendchange.py#L23
`self.username` and `self.password` are not local variables and are used here:
https://github.com/SonarSource/python-test-sources/blob/master/buildbot-0.8.6p1/buildbot/clients/sendchange.py#L45

python-checks/src/test/resources/checks/unusedLocalVariable.py
```diff
@@ -22,3 +22,18 @@ def string_interpolation():
...
+def function_with_lambdas():
+    print([(lambda unread_lambda_param: 2)(i) for i in range(10)])
+    x = 42 # Noncompliant
+    print([(lambda x: x*x)(i) for i in range(10)])
+    y = 42
+    print([(lambda x: x*x + y)(i) for i in range(10)])
+    {y**2 for a in range(3) if lambda x: x > 1 and y > 1}
+
+def using_tuples():
+    x, y = (1, 2) # Noncompliant {{Remove the unused local variable "y".}}
```
alban-auzeill: We should not raise issue on underscore variables, the following example should be tested and should not raise issue:
```
x, _ = (1, 2)
```
See https://www.datacamp.com/community/tutorials/role-underscore-python

christophe-zurn-sonarsource: I added a test with a for loop, since this one would not raise issue anymore, since as discussed we'll keep tuples as exception for this rule

PR #170, opened 2019-01-21
SONARPY-286 S1313 do not report issue for loopback address

alban-auzeill:


alban-auzeill:
Note for the reviewer: I duplicated the logic from https://github.com/SonarSource/slang/blob/master/slang-checks/src/main/java/org/sonarsource/slang/checks/HardcodedIpCheck.java

PR #181, opened 2019-01-28
SONARPY-240 CPD metrics: provide indentation information

alban-auzeill:


its/plugin/projects/cpd/file3.py
```diff
@@ -0,0 +1,28 @@
+class A:
+    def func1():
```
m-g-sonar: we lack a case with code with multiple indentation still being detected correctly. Please add a few lines of code, if possible with multiple level of indentation.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/cpd/PythonCpdAnalyzer.java
```diff
@@ -42,25 +43,24 @@ public void pushCpdTokens(InputFile inputFile, PythonVisitorContext visitorConte
...
         }
       }
       cpdTokens.save();
     }
   }
 
-  private boolean isIgnoredType(TokenType type) {
-    return type.equals(PythonTokenType.NEWLINE) ||
-      type.equals(PythonTokenType.DEDENT) || // TODO equals { } ???
+  private static boolean isIgnoredType(TokenType type, TokenType nextTokenType) {
```
m-g-sonar: naming of variable is not consistent. What about `tokenType` and `nextTokenType` ?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/cpd/PythonCpdAnalyzer.java
```diff
@@ -42,25 +43,24 @@ public void pushCpdTokens(InputFile inputFile, PythonVisitorContext visitorConte
...
+    return (type.equals(PythonTokenType.NEWLINE) && !nextTokenType.equals(PythonTokenType.DEDENT)) ||
```
m-g-sonar: I would extract this condition into a dedicated method, maybe called `isNotNewLineWithIndentationChange()` ? maybe it could even be completely out of the `isIgnoredType` method and simply called from [line #50](https://github.com/SonarSource/sonar-python/pull/181/files#diff-a3ead01d78a47faaa0dd1cfeb5365f04R50).

sonar-python-plugin/src/main/java/org/sonar/plugins/python/cpd/PythonCpdAnalyzer.java
```diff
@@ -42,25 +43,24 @@ public void pushCpdTokens(InputFile inputFile, PythonVisitorContext visitorConte
...
+      type.equals(PythonTokenType.DEDENT) ||
```
m-g-sonar: The particularity of `DEDENT` being empty in term of content should be mentioned in a comment. Maybe with a link to the documentation: https://docs.python.org/3/reference/lexical_analysis.html#indentation

sonar-python-plugin/src/test/java/org/sonar/plugins/python/cpd/PythonCpdAnalyzerTest.java
```diff
@@ -76,11 +76,11 @@ public void code_chunks_2() throws Exception {
       "deffoo():pass",
       "classbar(object):pass",
       "deffoo2(x,y,z,):",
-      "pass",
+      "pass\n",
```
m-g-sonar: add maybe a comment to explicitly say that `new_line` is there to mark the DEDENT on the next line.

its/plugin/src/test/java/com/sonar/python/it/plugin/CPDTest.java
```diff
@@ -0,0 +1,83 @@
...
+      .setSourceDirs(".")
+      .setSourceEncoding("UTF-8");
+    orchestrator.executeBuild(build);
+  }
+
+  @Test
+  public void test_cpd() {
+    Offset<Double> offset = Offset.offset(0.01d);
+
+    assertThat(getFileMeasureAsDouble("file1.py","duplicated_lines")).isEqualTo(27.0d, offset);
```
m-g-sonar: Pretty nice work on the ITs! You are however going to be hit hard by the "constant duplication" rule when this code is going to be analyzed.

alban-auzeill: Fixed

PR #184, opened 2019-01-28
Include ITs in normal build with skipped tests by default

m-g-sonar:


its/pom.xml
```diff
@@ -9,17 +9,33 @@
...
     <module>ruling</module>
   </modules>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <skipTests>${skip.its}</skipTests>
```
alban-auzeill: We should add `-Dskip.its=false` in `cix.sh`
And a nice to have: `cd its` , `mvn clean install -Pits` in `README.md`

PR #179, opened 2019-01-24
Update jarjar package names - fixes SONARPY-283

ewjmulder:
After running into https://stackoverflow.com/questions/50730049/error-running-sonar-sslr-toolkit and seeing https://jira.sonarsource.com/browse/SONARPY-283 I dug a little deeper and found the cause: the package names in the jarjar includes list are outdated. This commit corrects that and I can confirm the SSLR standalone application for Python is working again correctly.

P.S. The same issue is present in the sonar-flex project - sslr module. Will you fix it or shall I create a PR for that one as well?

PR #185, opened 2019-01-29
SONARPY-283 SSLR Toolkit crashes with ClassNotFoundException at startup

alban-auzeill:
Update jarjar package names


alban-auzeill:
This internal PR has been created to import the external PR #179

PR #188, opened 2019-01-29
SONARPY-198 Feed Cognitive Complexity metric + SONARPY-193

alban-auzeill:


python-squid/src/main/java/org/sonar/python/metrics/CognitiveComplexityVisitor.java
```diff
@@ -0,0 +1,207 @@
...
+
+      PythonGrammar.FUNCDEF,
+      PythonGrammar.SUITE)));
+  }
+
+  @Override
+  public void visitNode(AstNode astNode) {
+    if (astNode.is(PythonGrammar.FUNCDEF) && currentFunction == null) {
+      currentFunction = astNode;
+      nestingLevel = 0;
```
christophe-zurn-sonarsource: Here we will lose the previous nesting level when exiting the function, which might be result in wrong computation in case of multiple nested functions.
Ex.
```
def a():
  def b():
    def c():
      if 1 > 0:
        2
    return c() if 1 else "g"
  return b() if 1 > 0 else "f"
```

its/plugin/src/test/java/com/sonar/python/it/plugin/MetricsTest.java
```diff
@@ -55,85 +81,88 @@ public static void startServer() {
...
+    assertThat(getProjectMeasureAsInt(COMMENT_LINES)).isEqualTo(1);
+    assertThat(getProjectMeasureAsDouble(COMMENT_LINES_DENSITY)).isEqualTo(50.0);
     // Complexity
-    assertThat(getProjectMeasureAsDouble("complexity")).isZero();
-    assertThat(getProjectMeasureAsDouble("function_complexity")).isNull();
-    assertThat(getProjectMeasure("function_complexity_distribution").getValue()).isEqualTo("1=0;2=0;4=0;6=0;8=0;10=0;12=0;20=0;30=0");
-    assertThat(getProjectMeasureAsDouble("file_complexity")).isZero();
-    assertThat(getProjectMeasure("file_complexity_distribution").getValue()).isEqualTo("0=1;5=0;10=0;20=0;30=0;60=0;90=0");
+    assertThat(getProjectMeasureAsDouble(COMPLEXITY)).isEqualTo(2.0);
+    assertThat(getProjectMeasureAsDouble(COGNITIVE_COMPLEXITY)).isEqualTo(1.0);
```
christophe-zurn-sonarsource: I would have liked to see another file to get another value for cognitive complexity here, showing that it is now indeed computed on whole file/project and not only at functions level anymore

python-squid/src/main/java/org/sonar/python/metrics/CognitiveComplexityVisitor.java
```diff
@@ -0,0 +1,207 @@
...
+      PythonGrammar.WHILE_STMT,
+      PythonGrammar.FOR_STMT,
+      PythonGrammar.EXCEPT_CLAUSE,
+
+      PythonGrammar.AND_TEST,
+      PythonGrammar.OR_TEST,
+
+      PythonGrammar.TEST,
+
+      PythonGrammar.FUNCDEF,
```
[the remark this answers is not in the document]
alban-auzeill: It's possible to have a nested class in a function, but in this case I didn't change the existing logic. The nesting level is the one of the function.

python-squid/src/main/java/org/sonar/python/metrics/CognitiveComplexityVisitor.java
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  @Override
+  public void visitNode(AstNode astNode) {
+    if (astNode.is(PythonGrammar.FUNCDEF) && currentFunction == null) {
+      currentFunction = astNode;
+      nestingLevel = 0;
+    }
+
+    if (currentFunction != null) {
```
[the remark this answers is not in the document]
alban-auzeill: I added some logic to increment complexity outside of functions

python-squid/src/main/java/org/sonar/python/metrics/CognitiveComplexityVisitor.java
alban-auzeill: It's not wrong because we only compute complexity of the first level of functions

its/plugin/src/test/java/com/sonar/python/it/plugin/MetricsTest.java
alban-auzeill: done

python-squid/src/main/java/org/sonar/python/metrics/CognitiveComplexityVisitor.java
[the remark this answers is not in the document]
alban-auzeill: I added a separated commit to fix this

python-squid/src/main/java/org/sonar/python/metrics/CognitiveComplexityVisitor.java
```diff
@@ -0,0 +1,232 @@
...
+      PythonGrammar.CLASSDEF,
+      PythonGrammar.SUITE)));
+  }
+
+  @Override
+  public void visitNode(AstNode astNode) {
+    if (astNode.is(PythonGrammar.FUNCDEF, PythonGrammar.CLASSDEF)) {
+      nestingLevelStack.push(new NestingLevel(nestingLevelStack.peek(), astNode));
+    } else if (astNode.is(PythonGrammar.SUITE)) {
+      if (isSuiteIncrementsNestingLevel(astNode)) {
```
christophe-zurn-sonarsource: You could merge this inner if condition with its parent one, as all following ifs are also testing node type

python-squid/src/main/java/org/sonar/python/metrics/CognitiveComplexityVisitor.java
```diff
@@ -0,0 +1,232 @@
...
+      nestingLevelStack.peek().increment();
+    }
+  }
+
+  @Override
+  public void leaveNode(AstNode astNode) {
+    if (astNode.is(PythonGrammar.FUNCDEF, PythonGrammar.CLASSDEF)) {
+      nestingLevelStack.pop();
+    } else if (astNode.is(PythonGrammar.SUITE)) {
+      if (isSuiteIncrementsNestingLevel(astNode)) {
```
christophe-zurn-sonarsource: Same here, this inner if-condition could be merged with its parent, as well as the next condition since body are the same

python-squid/src/main/java/org/sonar/python/metrics/CognitiveComplexityVisitor.java
alban-auzeill: It was intentional to separate conditions and end the `else if` chain for `astNode.is(PythonGrammar.SUITE)` here. I see it as: easier to read and better performance.

PR #189, opened 2019-01-30
SONARPY-304 Analyzer should not have a fail-fast strategy

christophe-zurn-sonarsource:


sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonPlugin.java
```diff
@@ -105,6 +107,12 @@ public void define(Context context) {
       PythonRuleRepository.class);
 
     if (context.getRuntime().getProduct() != SonarProduct.SONARLINT) {
+      if (context.getRuntime().getApiVersion().isGreaterThanOrEqual(SQ_MIN_VERSION_ANALYSIS_WARNINGS)) {
```
m-g-sonar: I would extract `context.getRuntime()` into a dedicated variable to not call it twice in a row.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/warnings/AnalysisWarningsWrapper.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package org.sonar.plugins.python.warnings;
+
+import org.sonar.api.batch.InstantiationStrategy;
+import org.sonar.api.batch.ScannerSide;
+
+@ScannerSide
+@InstantiationStrategy("PER_BATCH")
+public interface AnalysisWarningsWrapper {
+  void addUnique(String text);
```
m-g-sonar: why call it `addUnique`? What does it mean? Add some javadoc. Mimicking the behavior of SonarQube without explaining why is not helping, and probably not needed.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonPlugin.java
```diff
@@ -105,6 +107,12 @@ public void define(Context context) {
...
+        context.addExtension(DefaultAnalysisWarningsWrapper.class);
+      } else {
+        context.addExtension(NoOpAnalysisWarningsWrapper.class);
```
m-g-sonar: Why not having the NoOpAnalysisWarningsWrapper also be included in SonarLint context? What about failing on something else than reports? (for instance during SquidSensor scan)

christophe-zurn-sonarsource: I don't see the need to add `NoOpAnalysisWarningsWrapper`, I believe this is only displayed on SonarQube, and currently only sensors defined outside of SonarLint are raising UI warnings.

Indeed, other sensors can also crash the analysis. I will update implementation accordingly

PR #206, opened 2019-02-25
SONARPY-326 Resolve symbols' module names using imports

andrea-guarino-sonarsource:


python-squid/src/main/java/org/sonar/python/semantic/SymbolTableBuilderVisitor.java
```diff
@@ -305,13 +391,15 @@ private void addNonlocalName(String name) {
   private static class SymbolImpl implements Symbol {
 
     private final String name;
+    private String moduleName;
```
saberduck: I think it would be cleaner to have this field initialized in constructor and make it final. That way the contract is clear, that it cannot be modified once symbol is created

python-squid/src/main/java/org/sonar/python/semantic/SymbolTableBuilderVisitor.java
```diff
@@ -387,4 +480,10 @@ public ClassVariableAssignmentVisitor(AstNode classTree) {
     }
 
   }
+
+  private static class Module {
```
saberduck: I think we can create a constructor and make fields final, so it's clear that this object is not mutable

PR #208, opened 2019-02-25
SONARPY-301:  Rule S4828: Signalling processes is security-sensitive

andrea-guarino-sonarsource:


its/ruling/src/test/resources/profile.xml
```diff
@@ -284,6 +284,10 @@
       <repositoryKey>python</repositoryKey>
       <key>S3776</key>
       <priority>INFO</priority>
+    </rule><rule>
```
saberduck: Let's replace this by https://github.com/SonarSource/sonar-analyzer-commons/blob/master/commons/src/main/java/org/sonarsource/analyzer/commons/ProfileGenerator.java , in new PR

PR #214, opened 2019-03-03
SONARPY-306 Rule S4829: Reading the Standard Input is security-sensitive

andrea-guarino-sonarsource:


python-checks/src/test/resources/checks/hotspots/standardInput.py
```diff
@@ -0,0 +1,35 @@
...
+    sys.stdin  # Noncompliant {{Make sure that reading the standard input is safe here.}}
+#   ^^^^^^^^^
+
+    sys.__stdin__  # Noncompliant
+
+    stdin #Noncompliant
+    __stdin__ #Noncompliant
+
+
+    sys.stdin.read()  # Questionable
```
andrea-guarino-sonarsource: consider this case in rule implementation

PR #216, opened 2019-03-05
SONARPY-314 Rule S4790: Hashing data is security-sensitive

andrea-guarino-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/hotspots/HashingDataCheck.java
```diff
@@ -0,0 +1,195 @@
...
+import org.sonar.python.checks.AbstractCallExpressionCheck;
+import org.sonar.python.semantic.Symbol;
+
+@Rule(key = HashingDataCheck.CHECK_KEY)
+public class HashingDataCheck extends AbstractCallExpressionCheck {
+  public static final String CHECK_KEY = "S4790";
+  private static final String MESSAGE = "Make sure that hashing data is safe here.";
+  private static final Set<String> questionableFunctions = immutableSet(
+    "hashlib.new", "optparse.OptionParser", "cryptography.hazmat.primitives.hashes.Hash", "django.contrib.auth.hashers.make_password", "werkzeug.security.generate_password_hash");
+  private static final Set<String> questionableHashlibAlgorithm = immutableSet(
```
m-g-sonar: Why using an immutable Set to then stream on it?! Please use:
```java
Stream.of("a", "b").map(...
```

python-checks/src/main/java/org/sonar/python/checks/hotspots/HashingDataCheck.java
```diff
@@ -0,0 +1,195 @@
...
+    "blake2b", "blake2s", "md5", "pbkdf2_hmac", "sha1", "sha224",
+    "sha256", "sha384", "sha3_224", "sha3_256", "sha3_384", "sha3_512",
+    "sha512", "shake_128", "shake_256", "scrypt")
+    .stream()
+    .map(hasher -> "hashlib." + hasher)
+    .collect(Collectors.toSet());
+
+  private static final Set<String> questionablePasslibAlgorithm = immutableSet(
```
m-g-sonar: Same approach as before, no need for the intermediate `ImmutableSet`

python-checks/src/main/java/org/sonar/python/checks/hotspots/HashingDataCheck.java
```diff
@@ -0,0 +1,195 @@
...
+    "mssql2000", "mssql2005", "mysql323", "mysql41", "nthash", "oracle10",
+    "oracle11", "pbkdf2_sha1", "pbkdf2_sha256", "pbkdf2_sha512", "phpass", "plaintext",
+    "postgres_md5", "roundup_plaintext", "scram", "scrypt", "sha1_crypt", "sha256_crypt",
+    "sha512_crypt", "sun_md5_crypt", "unix_disabled", "unix_fallback")
+    .stream()
+    .map(hasher -> "passlib.hash." + hasher)
+    .collect(Collectors.toSet());
+
+
+  private static final Set<String> questionableDjangoHashers = immutableSet(
```
m-g-sonar: and again.

python-checks/src/main/java/org/sonar/python/checks/hotspots/HashingDataCheck.java
```diff
@@ -0,0 +1,195 @@
...
+      PythonGrammar.CALL_EXPR,
+      PythonGrammar.ATTRIBUTE_REF,
+      PythonGrammar.ATOM,
+      PythonGrammar.EXPRESSION_STMT,
+      PythonGrammar.CLASSDEF);
+  }
+
+  @Override
+  public void visitNode(AstNode node) {
+    if (node.is(PythonGrammar.ATTRIBUTE_REF, PythonGrammar.ATOM)) {
```
m-g-sonar: Rather than chained `if-else-if`I would prefer to have a switch:
```java
switch(node.getType()) {
  case PythonGrammar.CALL_EXPR:
    checkQuestionableHashingAlgorithm(node);
    break;
  case /*...*/ :
    // ...
    break;
  default:
    // do nothing - reacting on all the registered nodes
}
```

python-checks/src/main/java/org/sonar/python/checks/hotspots/HashingDataCheck.java
```diff
@@ -0,0 +1,195 @@
...
+      checkQuestionableHashingAlgorithm(node);
+
+    } else if (node.is(PythonGrammar.EXPRESSION_STMT)) {
+      checkOverwriteDjangoHashers(node);
+
+    } else if (node.is(PythonGrammar.CLASSDEF)) {
+      checkCreatingCustomHasher(node);
+
+    } else if (node.is(PythonGrammar.CALL_EXPR)) {
+      super.visitNode(node);
```
m-g-sonar: this deserve a comment, otherwise I would tend to remove it.

python-checks/src/main/java/org/sonar/python/checks/hotspots/HashingDataCheck.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+  }
+
+  @Override
+  protected boolean isException(AstNode callExpression) {
+    if (getQualifiedName(callExpression).equals("django.contrib.auth.hashers.make_password")) {
```
m-g-sonar: I would extract all this to a dedicated method, like "isDjangoMakePasswordFunction", to make it easy to read.
Maybe having a simple line of javadoc explaining why we need to exclude this function would be nice as well
```
  @verride
  protected boolean isException(AstNode callExpression) {
    return isDjangoMakePasswordFunction(callExpression);
  }
```

python-checks/src/main/java/org/sonar/python/checks/hotspots/HashingDataCheck.java
```diff
@@ -0,0 +1,195 @@
...
+      return;
+    }
+
+    // checks for settings.PASSWORD_HASHERS = value
+    AstNode attributeRef = lhsExpression.getFirstDescendant(PythonGrammar.ATTRIBUTE_REF);
+    if (attributeRef != null) {
+      AstNode atom = attributeRef.getFirstChild(PythonGrammar.ATOM);
+      if (atom != null && getQualifiedName(atom).equals("django.conf.settings") &&
+        attributeRef.getLastChild(PythonGrammar.NAME).getTokenValue().equals("PASSWORD_HASHERS")) {
+        addIssue(expressionStatement, MESSAGE);
```
m-g-sonar: don't you want to explicitly `return` here?
It sounds strange to me that we could raise 2 issues in one run (by reaching both line 128 and 135), but if you think it's OK, then fine.

python-checks/src/main/java/org/sonar/python/checks/hotspots/HashingDataCheck.java
```diff
@@ -0,0 +1,195 @@
...
+      AstNode parent = node.getParent();
+      return propertyName.equals("Hash") && parent != null && parent.is(PythonGrammar.CALL_EXPR);
+    }
+    return false;
+  }
+
+  private static AstNode getLHSExpression(AstNode expressionStatement) {
+    AstNode assign = expressionStatement.getFirstChild(PythonPunctuator.ASSIGN);
+    if (assign != null) {
+      return expressionStatement.getFirstChild(PythonGrammar.TESTLIST_STAR_EXPR);
```
m-g-sonar: what if there is parenthesis?

python-checks/src/main/java/org/sonar/python/checks/hotspots/HashingDataCheck.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+    return null;
```
m-g-sonar: should be annotated `@CheckForNull`

python-checks/src/main/java/org/sonar/python/checks/hotspots/HashingDataCheck.java
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+
+  private String getQualifiedName(AstNode node) {
+    Symbol symbol = getContext().symbolTable().getSymbol(node);
+    if (symbol != null) {
```
m-g-sonar:
```
return symbol != null ? symbol.qualifiedName() : "";

File: python-checks/src/main/java/org/sonar/python/checks/hotspots/HashingDataCheck.java [hunk 11]
```diff
@@ -0,0 +1,195 @@
...
+      return symbol.qualifiedName();
+    }
+    return "";
+  }
+
+  private void checkCreatingCustomHasher(AstNode classDef) {
+    AstNode argList = classDef.getFirstChild(PythonGrammar.ARGLIST);
+    if (argList != null) {
+      argList.getDescendants(PythonGrammar.ATOM)
+        .forEach(atom -> {
```
m-g-sonar:
```java
.filter(atom -> questionableDjangoHashers.contains(getQualifiedName(atom)))
.forEach(atom -> addIssue(atom, MESSAGE));
```

python-checks/src/main/java/org/sonar/python/checks/hotspots/HashingDataCheck.java
andrea-guarino-sonarsource: yes indeed, I agree. I'll add a `return`

python-checks/src/main/java/org/sonar/python/checks/hotspots/HashingDataCheck.java
andrea-guarino-sonarsource: In case of parenthesis, TESTLIST_STAR_EXPR will still be the first child of EXPRESSION_STMT. 
Parenthesis will be consumed lower in the grammar (in `ATOM` production rule) and at line 123 I'm taking the `firstDescendant(PythonGrammar.ATTRIBUTE_REF)`.
I'm adding a test case to cover this.

However I just realized I won't raise an issue in this case:
```
foo.bar, mySettings.PASSWORD_HASHERS = value # Noncompliant
```
I will change my implementation to cover this case

PR #217, opened 2019-03-05
SONARPY-316  Rule S2077: Executing SQL queries is security-sensitive

andrea-guarino-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/hotspots/SQLQueriesCheck.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  @Override
+  public Set<AstNodeType> subscribedKinds() {
+    return immutableSet(PythonGrammar.CALL_EXPR);
+  }
+
+  @Override
+  public void visitFile(AstNode node) {
+    Set<Symbol> symbols = getContext().symbolTable().symbols(node);
```
m-g-sonar: You should already compute the qualified names here in order to not stream 3 times for the same data:
```java
Set<String> symbolsQualifiedNames = getContext().symbolTable().symbols(node).stream()
  .map(Symbol::qualifiedName)
  .collect(Collectors.toSet());
isUsingDjangoModel = symbolsQualifiedNames.contains("django.db.connection");
isUsingDjangoDBConnection = symbolsQualifiedNames.contains("django.db.connection");
isUsingDjangoDBConnections = symbolsQualifiedNames.contains("django.db.connections");
```

python-checks/src/main/java/org/sonar/python/checks/hotspots/SQLQueriesCheck.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  @Override
+  protected String message() {
+    return MESSAGE;
+  }
+
+  @Override
+  public Set<AstNodeType> subscribedKinds() {
+    return immutableSet(PythonGrammar.CALL_EXPR);
```
m-g-sonar: I don't see the reason to override this one. It's already what is implemented in `AbstractCallExpressionCheck`

python-checks/src/main/java/org/sonar/python/checks/hotspots/SQLQueriesCheck.java
```diff
@@ -0,0 +1,83 @@
...
+    isUsingDjangoDBConnection = symbols.stream().anyMatch(s -> s.qualifiedName().contains("django.db.connection"));
+    isUsingDjangoDBConnections = symbols.stream().anyMatch(s -> s.qualifiedName().contains("django.db.connections"));
+  }
+
+  @Override
+  public void visitNode(AstNode node) {
+    if (isUsingDjangoModel) {
+      AstNode attributeRef = node.getFirstChild(PythonGrammar.ATTRIBUTE_REF);
+      if (attributeRef != null) {
+        String functionName = attributeRef.getLastChild(PythonGrammar.NAME).getTokenValue();
```
m-g-sonar: isn't `getLastChild(KIND)` nullable? I would add a null check for safety

python-checks/src/main/java/org/sonar/python/checks/hotspots/SQLQueriesCheck.java
```diff
@@ -0,0 +1,83 @@
...
+        if (functionName.equals("raw") || functionName.equals("extra")) {
+          addIssue(node, MESSAGE);
+        }
+      }
+    }
+
+    if (isUsingDjangoDBConnection || isUsingDjangoDBConnections) {
```
m-g-sonar: both `if`s are doing almost exactly the same thing, the only difference being the function name targeted (execute, raw, extra). they should be merged.

python-checks/src/main/java/org/sonar/python/checks/hotspots/SQLQueriesCheck.java
andrea-guarino-sonarsource: Indeed it is nullable, but `ATTRIBUTE_REF` node, per grammar definition, has always at least one child of kind `NAME`. In what case could it be null?

m-g-sonar: Okay, then if by construction it can not be null, that fine. But then let's make it explicit with a comment.

python-checks/src/main/java/org/sonar/python/checks/hotspots/SQLQueriesCheck.java
andrea-guarino-sonarsource: Actually I need to look for qualifiedNames that contains a given substring. So for example `django.db.connection.somethingElse` should match here. I think I will go for 
```java
getContext().symbolTable().symbols(node).forEach(symbol -> {
      String qualifiedName = symbol.qualifiedName();
      if (qualifiedName.contains("django.db.models")) {
        isUsingDjangoModel = true;
      }
      if (qualifiedName.contains("django.db.connection")) {
        isUsingDjangoDBConnection = true;
      }
      if (qualifiedName.contains("django.db.connections")) {
        isUsingDjangoDBConnections = true;
      }
    });
```

PR #219, opened 2019-03-05
SONARPY-317 Rule S1523: Dynamically executing code is security-sensitive

andrea-guarino-sonarsource:


python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S1523.html
```diff
@@ -0,0 +1,69 @@
...
+
+eval(hardcoded_command)  # Ok
+eval(formatted_command)  # Questionable. Dynamic code
+eval(compiled_command)  # Questionable even though it is hardcoded, but only because it is not worth detecting this corner case.
+
+def execute(code, file, mode):
+    exec(code)  # Questionable. Dynamic code
+    exec(compile(code, file, mode))  # Questionable. Dynamic command
+
+exec(hardcoded_command)  # Ok
```
alban-auzeill: RSPEC description does not match implementation, we should clear this `S1523.html` page from what is not supported, like hard coded string literals and variables, and ensure that there's a ticket to improve the rule later to match RSPEC.

PR #228, opened 2019-04-16
Use JDK 11

saberduck:


saberduck:
I tested locally and javadoc is fine in this project, we don't need to ignore errors

PR #232, opened 2019-06-04
SONARPY-336 Rule S4784: Update implementation to target only hard-cod…

andrea-guarino-sonarsource:
…ed regex


python-checks/src/main/java/org/sonar/python/checks/hotspots/RegexCheck.java
```diff
@@ -48,10 +51,47 @@ public void visitNode(AstNode node) {
     if (symbol != null && questionableFunctions.contains(symbol.qualifiedName())) {
       AstNode parent = node.getParent();
       if (parent != null && parent.is(PythonGrammar.CALL_EXPR)) {
-        addIssue(parent, MESSAGE);
-      } else {
-        addIssue(node, MESSAGE);
+        parent.getChildren(PythonGrammar.ARGLIST).forEach(this::checkArgument);
```
alban-auzeill: Checking all arguments could raise FP on a non-regexp argument. e.g. in `regex.subf('pattern', 'replacement{1}{2}', 'input')`
We should only check the regexp arguments.

PR #252, opened 2019-07-16
SONARPY-345 Parser should parse Python3 by default and fallback to Python2

pynicolas:
The other commit comes from #251

### DIFF
--- a/python-checks/src/main/java/org/sonar/python/checks/PrintStatementUsageCheck.java
+++ b/python-checks/src/main/java/org/sonar/python/checks/PrintStatementUsageCheck.java
@@ -19,10 +19,14 @@
  */
 package org.sonar.python.checks;
 
+import com.intellij.lang.ASTNode;
+import com.intellij.psi.tree.IElementType;
 import com.intellij.psi.util.PsiTreeUtil;
 import com.jetbrains.python.PyElementTypes;
+import com.jetbrains.python.psi.PyBinaryExpression;
 import com.jetbrains.python.psi.PyElement;
 import com.jetbrains.python.psi.PyExpression;
+import com.jetbrains.python.psi.PyExpressionStatement;
 import com.jetbrains.python.psi.PyParenthesizedExpression;
 import com.jetbrains.python.psi.PyPrintTarget;
 import java.util.List;
@@ -32,6 +36,8 @@ import org.sonar.python.PythonCheck;
 @Rule(key = "PrintStatementUsage")
 public class PrintStatementUsageCheck extends PythonCheck {
 
+  public static final String MESSAGE = "Replace print statement by built-in function.";
+
   @Override
   public void initialize(Context context) {
     context.registerSyntaxNodeConsumer(PyElementTypes.PRINT_STATEMENT, ctx -> {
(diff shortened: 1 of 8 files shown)


sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonScanner.java
```diff
@@ -104,17 +104,22 @@ private void scanFile(InputFile inputFile) {
...
-    pyFile.accept(pythonHighlighter);
-    pythonHighlighter.getNewHighlighting().save();
+    try {
+      PyFile pyFile = new org.sonar.python.frontend.PythonParser().parse(fileContent);
+      SubscriptionVisitor.analyze(checks.all(), visitorContext, pyFile);
+      PythonHighlighter pythonHighlighter = new PythonHighlighter(context, inputFile);
+      pyFile.accept(pythonHighlighter);
+      pythonHighlighter.getNewHighlighting().save();
+    } catch (RecognitionException e) {
+      // Ignore, we should not raise a parsing error twice (once for each parser)
```
andrea-guarino-sonarsource: What do you think about adding `DEBUG` log message?

python-frontend/src/main/java/org/sonar/python/frontend/PythonParser.java
```diff
@@ -63,7 +70,28 @@
...
+    file = parseAs(content, LanguageLevel.PYTHON27);
+    PsiErrorElement errorElement = errorElements(file).get(0);
+    if (errorElement == null) {
+      return file;
+    }
+    int lineNumber = new PythonTokenLocation(errorElement).startLine();
+    throw new RecognitionException(lineNumber, errorElement.getErrorDescription());
+  }
+
+  private JBIterable<PsiErrorElement> errorElements(PsiElement root) {
```
andrea-guarino-sonarsource: There are two code smells regarding `errorElements` and `parseAs` not being static

python-checks/src/main/java/org/sonar/python/checks/PrintStatementUsageCheck.java
```diff
@@ -32,14 +36,35 @@
...
     context.registerSyntaxNodeConsumer(PyElementTypes.PRINT_STATEMENT, ctx -> {
       List<PyElement> expressions = PsiTreeUtil.getChildrenOfAnyType(ctx.syntaxNode(), PyExpression.class, PyPrintTarget.class);
       if (expressions.size() == 1 && expressions.get(0) instanceof PyParenthesizedExpression) {
         return;
       }
-      ctx.addIssue(ctx.syntaxNode().getFirstChild(), "Replace print statement by built-in function.");
+      ctx.addIssue(ctx.syntaxNode().getFirstChild(), MESSAGE);
+    });
+
+    context.registerSyntaxNodeConsumer(PyElementTypes.BINARY_EXPRESSION, ctx -> {
```
andrea-guarino-sonarsource: Could you please add a comment explaining why we need this?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonScanner.java
pynicolas: I had to rebase against the master. Now, we need both parsers to compute metrics. I had to [restructure the code](https://github.com/SonarSource/sonar-python/pull/252/commits/55903b135010295800d454f470d0766340f7e32a#diff-3939b2c0f41ee15dc69c575494495142).

python-frontend/src/main/java/org/sonar/python/frontend/PythonParser.java
pynicolas: Fixed.

python-checks/src/main/java/org/sonar/python/checks/PrintStatementUsageCheck.java
pynicolas: Done.

pynicolas:
I fixed the typo in the commit message.
I don't think we should create a separate ticket for `PrintStatementUsageCheck`: 
* there should be no change in behavior compared to what we had before
* I had to update `PrintStatementUsageCheck` precisely because of the first commit: in fact, unit tests are red after the 1st commit because of a [test file using a `print` statement](https://github.com/SonarSource/sonar-python/blob/master/sonar-python-plugin/src/test/resources/org/sonar/plugins/python/squid-sensor/file1.py#L10).

PR #254, opened 2019-07-17
SONARPY-342 Compute statement and classes metrics from PyCharm

mpaladin:


python-squid/src/main/java/org/sonar/python/metrics/MetricsVisitor.java
```diff
@@ -0,0 +1,170 @@
...
+
+    // track executable lines of code
+    if (element instanceof PyStatement) {
+      handlePyStatement(element);
+    }
+
+    // track lines of comments and no_sonar lines
+    if (element instanceof PsiComment) {
+      if (ignoreHeaderComments && !firstNonCommentSeen) {
+        firstNonCommentSeen = true;
```
pynicolas: I believe this is wrong for any file which has at least 2 lines of comment at the beginning.

python-squid/src/main/java/org/sonar/python/metrics/MetricsVisitor.java
```diff
@@ -0,0 +1,170 @@
...
+    if (node.isDocString()) {
+      PythonTokenLocation location = new PythonTokenLocation(node);
+      for (int line = location.startLine(); line <= location.endLine(); line++) {
+        linesOfComments.add(line);
+      }
+    }
+    super.visitPyStringLiteralExpression(node);
+  }
+
+  private static boolean isDocString(PsiElement element) {
```
pynicolas: I would change the type of `element` to `PyStatement` so that the body of the method makes more sense.

python-squid/src/test/java/org/sonar/python/metrics/FileMetricsTest.java
```diff
@@ -61,7 +65,14 @@ public void function_complexities() {
...
     File baseDir = new File("src/test/resources/metrics/");
     File file = new File(baseDir, fileName);
-    return new FileMetrics(TestPythonVisitorRunner.createContext(file), true);
+    String fileContent;
+    try {
+      fileContent = Files.toString(file, StandardCharsets.UTF_8);
+    } catch (IOException e) {
+      throw new IllegalStateException(e);
+    }
+    PyFile pyFile = new org.sonar.python.frontend.PythonParser().parse(fileContent);
```
pynicolas: Now that the 7 lines above are duplicated in 3 tests in python-squid, maybe we can factorize them.

python-squid/src/main/java/org/sonar/python/metrics/MetricsVisitor.java
mpaladin: @pynicolas indeed, https://github.com/SonarSource/sonar-python/pull/254/commits/955f7b8cadd4c7b1ad8e4ff545721d1384713d17.

python-squid/src/test/java/org/sonar/python/metrics/FileMetricsTest.java
mpaladin: @pynicolas https://github.com/SonarSource/sonar-python/pull/254/commits/94159450a1bde59907fd4faa7d46f3f8c443167f.

python-squid/src/main/java/org/sonar/python/metrics/MetricsVisitor.java
mpaladin: @pynicolas https://github.com/SonarSource/sonar-python/pull/254/commits/089dc2c93c9fa732f44c09ce526d0c697997f5c4.

PR #268, opened 2019-08-23
SONARPY-389  Add subscription visitor for checks

andrea-guarino-sonarsource:


python-squid/src/test/java/org/sonar/python/PythonCheckTreeTest.java
```diff
@@ -83,7 +89,7 @@ public void visitFunctionDef(PyFunctionDefTree pyFunctionDefTree) {
     assertThat(firstIssue.cost()).isEqualTo(42);
   }
 
-  private static class TestPythonCheck extends PythonCheckTree {
```
benzonico: There should be a dedicated test rather than replacing this one.

python-squid/src/main/java/org/sonar/python/SubscriptionVisitor.java
```diff
@@ -0,0 +1,100 @@
...
+        elementConsumers.add(new SubscriptionContextImpl(check, consumer));
+      });
+    }
+  }
+
+  @Override
+  public void scan(@Nullable Tree element) {
+    if (element != null) {
+      currentElement = element;
+      List<SubscriptionContextImpl> elementConsumers = consumers.getOrDefault(element.getKind(), Collections.emptyList());
```
benzonico: .forEach ?

python-squid/src/main/java/org/sonar/python/SubscriptionVisitor.java
```diff
@@ -0,0 +1,100 @@
...
+
+public class SubscriptionVisitor extends BaseTreeVisitor {
+
+  private final EnumMap<Kind, List<SubscriptionContextImpl>> consumers = new EnumMap<>(Kind.class);
+  private final PythonVisitorContext pythonVisitorContext;
+  private Tree currentElement;
+
+  public static void analyze(Collection<PythonSubscriptionCheck> checks, PythonVisitorContext pythonVisitorContext, Tree tree) {
+    SubscriptionVisitor subscriptionVisitor = new SubscriptionVisitor(checks, pythonVisitorContext);
+    tree.accept(subscriptionVisitor);
```
benzonico: There is no need to pass the tree as it is accessible from pythonVisitorContext.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonScanner.java
```diff
@@ -95,6 +98,11 @@ private void scanFile(InputFile inputFile) {
       AstNode astNode = parser.parse(pythonFile.content());
       PyFileInputTree parse = new PythonTreeMaker().fileInput(astNode);
       visitorContext = new PythonVisitorContext(astNode, parse, pythonFile);
+      List<PythonSubscriptionCheck> checksBasedOnTree = checks.all().stream()
```
benzonico: Why not do the instance of test in the loop a few lines down ? You don't need to pass the "parse" tree.

PR #270, opened 2019-08-26
SONARPY-380 Migrate DebugModeCheck to strongly typed AST

andrea-guarino-sonarsource:


python-squid/src/main/java/org/sonar/python/api/tree/PyCallExpressionTree.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+package org.sonar.python.api.tree;
+
+import com.sonar.sslr.api.Token;
+import java.util.List;
+
+public interface PyCallExpressionTree extends PyExpressionTree {
+  PyExpressionTree callee();
+
+  List<PyArgumentTree> arguments();
```
pynicolas: Another model would have been to define a `PyArgumentListTree`. I think it would have represented [python calls](https://docs.python.org/3/reference/expressions.html#calls) more precisely. However, I think it would have required changed in the grammar.

python-squid/src/main/java/org/sonar/python/api/tree/PyQualifiedExpressionTree.java
```diff
@@ -0,0 +1,29 @@
...
+package org.sonar.python.api.tree;
+
+/**
+ * Qualified expression like "foo.bar"
+ */
+public interface PyQualifiedExpressionTree extends PyExpressionTree {
+  PyExpressionTree qualifier();
+
+  PyNameTree name();
+}
```
pynicolas: What about the dot token between the qualifier and the name?

python-squid/src/main/java/org/sonar/python/tree/BaseTreeVisitor.java
```diff
@@ -260,4 +263,22 @@ public void visitWithItem(PyWithItemTree pyWithItemTree) {
...
+  @Override
+  public void visitCallExpression(PyCallExpressionTree pyCallExpressionTree) {
+    scan(pyCallExpressionTree.callee());
+    scan(pyCallExpressionTree.arguments());
+  }
+
+  @Override
+  public void visitArgument(PyArgumentTree pyArgumentTree) {
+    scan(pyArgumentTree.keywordArgument());
+    scan(pyArgumentTree.expression());
```
pynicolas: I see that the methods of `BaseTreeVisitor` never scan tokens. I'm not sure it's a good thing. Will we be able to rewrite CPD integration for example?

python-squid/src/main/java/org/sonar/python/api/tree/PyArgumentTree.java
```diff
@@ -0,0 +1,39 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.python.api.tree;
+
+import com.sonar.sslr.api.Token;
+import javax.annotation.CheckForNull;
+
+public interface PyArgumentTree extends Tree {
+  @CheckForNull
+  PyExpressionTree keywordArgument();
```
pynicolas: Can it be something which is not an identifier? https://docs.python.org/3/reference/expressions.html#calls

python-squid/src/main/java/org/sonar/python/tree/PythonTreeMaker.java
```diff
@@ -518,9 +521,62 @@ private PyExceptClauseTree exceptClause(AstNode except, PyStatementListTree body
...
+    if (astNode.is(PythonGrammar.CALL_EXPR)) {
+      return callExpression(astNode);
+    }
     return new PyExpressionTreeImpl(astNode);
   }
+
+  public PyQualifiedExpressionTree qualifiedExpression(AstNode astNode) {
+    PyExpressionTree qualifier = expression(astNode.getFirstChild());
+    List<AstNode> names = astNode.getChildren(PythonGrammar.NAME);
+    names.remove(names.size() - 1);
```
pynicolas: Why do we modify the `AstNode`? That may be risky.

python-squid/src/main/java/org/sonar/python/tree/PythonTreeMaker.java
```diff
@@ -518,9 +521,62 @@ private PyExceptClauseTree exceptClause(AstNode except, PyStatementListTree body
   }
 
   PyExpressionTree expression(AstNode astNode) {
+    if (astNode.is(PythonGrammar.TEST) && astNode.getChildren().size() == 1) {
+      return expression(astNode.getFirstChild());
```
pynicolas: I think we should return `expression(astNode.getFirstChild())` in all cases where `astNode.getChildren().size() == 1`.

python-squid/src/main/java/org/sonar/python/tree/PythonTreeMaker.java
```diff
@@ -518,9 +521,62 @@ private PyExceptClauseTree exceptClause(AstNode except, PyStatementListTree body
...
     return new PyExpressionTreeImpl(astNode);
   }
+
+  public PyQualifiedExpressionTree qualifiedExpression(AstNode astNode) {
+    PyExpressionTree qualifier = expression(astNode.getFirstChild());
+    List<AstNode> names = astNode.getChildren(PythonGrammar.NAME);
+    names.remove(names.size() - 1);
+    PyNameTree name = name(astNode.getLastChild());
+    for (AstNode nameNode : names) {
+      qualifier = new PyQualifiedExpressionTreeImpl(astNode, name(nameNode), qualifier);
```
pynicolas: I think we can't use `astNode` here, that's not the correct location for the tree. The problem is that there's no instance of `AstNode`. I have the same problem with binary expressions, and I believe the solution is to [stop requiring](https://github.com/SonarSource/sonar-python/pull/271) an `AstNode` to build a `PyTree`.

python-squid/src/main/java/org/sonar/python/api/tree/PyEnclosureTree.java
```diff
@@ -0,0 +1,4 @@
+package org.sonar.python.api.tree;
+
+public interface PyEnclosureTree extends PyAtomTree {
```
pynicolas: What's that?

python-squid/src/main/java/org/sonar/python/api/tree/PyAtomTree.java
```diff
@@ -0,0 +1,5 @@
+package org.sonar.python.api.tree;
+
+public interface PyAtomTree extends PyExpressionTree {
```
pynicolas: What's an atom? Should we have a tree for that?

python-checks/src/main/java/org/sonar/python/checks/hotspots/DebugModeCheck.java
```diff
@@ -19,58 +19,92 @@
...
+    if (!(argument.keywordArgument() instanceof PyAtomTree)) {
+      return false;
+    }
+    PyExpressionTree keywordArgument = ((PyAtomTree) argument.keywordArgument()).atom();
+    if (keywordArgument instanceof PyNameTree && debugProperties.contains(((PyNameTree) keywordArgument).name())) {
+      if (!argument.expression().is(Kind.ATOM)) {
+        return false;
+      }
+      PyExpressionTree argumentValue = ((PyAtomTree) argument.expression()).atom();
+      return argumentValue instanceof PyNameTree && ((PyNameTree) argumentValue).name().equals("True");
```
pynicolas: Can't we use `isTrueLiteral` here?

python-squid/src/main/java/org/sonar/python/tree/PythonTreeMaker.java
```diff
@@ -570,18 +567,22 @@ PyExpressionTree expression(AstNode astNode) {
     if (astNode.is(PythonGrammar.CALL_EXPR)) {
       return callExpression(astNode);
     }
+    if (astNode.getChildren().size() == 1) {
```
pynicolas: Sorry, now that I see it, I think it should be: `if (astNode.is(PythonGrammar.EXPR) && astNode.getChildren().size() == 1)`. What do you think?

PR #271, opened 2019-08-27
SONARPY-392 Binary expressions

pynicolas:


benzonico:
LGTM, I'll merge it despite the "low" coverage, we can fix that one later.

PR #275, opened 2019-08-29
SONARPY-380 Add parent and children to Python Tree, migrate EmptyNestedBlockCheck, PreIncrementDecrementCheck

andrea-guarino-sonarsource:


python-squid/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
```diff
@@ -359,14 +391,15 @@ public void raiseStatement() {
...
+    assertThat(raiseStatement.children()).hasSize(3);
 
     astNode = p.parse("raise foo from bar");
     raiseStatement = treeMaker.raiseStatement(astNode);
     assertThat(raiseStatement).isNotNull();
     assertThat(raiseStatement.raiseKeyword().getValue()).isEqualTo("raise");
     assertThat(raiseStatement.fromKeyword().getValue()).isEqualTo("from");
     assertThat(raiseStatement.fromExpression()).isNotNull();
-    assertThat(raiseStatement.expressions()).hasSize(1);
+    assertThat(raiseStatement.children()).hasSize(2);
```
pynicolas: Is that change intended?

python-squid/src/main/java/org/sonar/python/api/tree/Tree.java
```diff
@@ -42,6 +42,8 @@
 
   List<Tree> children();
 
+  List<Token> tokens();
```
pynicolas: I would not add this method to `Tree`, only to `PyStatementListTree`, at least for the moment.

PR #279, opened 2019-08-30
SONARPY-358 RSPEC-4423 weak protocol version

benzonico:


python-checks/src/main/java/org/sonar/python/checks/WeakSSLProtocolCheck.java
```diff
@@ -0,0 +1,68 @@
...
+      if (isWeakProtocol(pyNameTree, symbol)) {
+        ctx.addIssue(pyNameTree, "Change this code to use a stronger protocol.");
+      }
+    });
+  }
+
+  private static boolean isWeakProtocol(PyNameTree pyNameTree, @Nullable Symbol symbol) {
+    Predicate<String> matchWeakProtocol;
+    if (symbol == null) {
+      matchWeakProtocol = s -> s.substring(s.lastIndexOf('.') + 1).equals(pyNameTree.name());
```
pynicolas: I don't get why we need that. I think we have a problem because, in the unit test, this method returns true only when we go through this branch. The other branch, where we use the symbol, doesn't seem to raise issues. Maybe it's related to https://github.com/SonarSource/sonar-python/blob/master/python-squid/src/main/java/org/sonar/python/semantic/SymbolTableBuilderVisitor.java#L329

Maybe we should either:
* create a ticket to fix the problem later
* or fix the problem in the symbol table right away

python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S4423.json
```diff
@@ -0,0 +1,34 @@
+{
+  "title": "Weak SSL\/TLS protocols should not be used",
+  "type": "VULNERABILITY",
```
pynicolas: Shouldn't it be a hotspot?

benzonico: This is a vulnerability as per PM specification, summoning @pierre-loup-tristant-sonarsource for confirmation.

python-checks/src/main/java/org/sonar/python/checks/WeakSSLProtocolCheck.java
benzonico: Problem is deeper : this is completely linked to semantic:  this is about solving dotted names : in order to get fully qualified name of the constant we need to figure out fqn of its parent and right now the SymbolTableVisitor does not deal with this kind of constructs (if i'm correct, it only deals with the first part of a dotted name). 
As I did not want to address semantic in the implementation of this new rule and as I managed to make it work I believe this is ok. 
Not sure which kind of ticket I should create for such case to be honest.

python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S4423.json
pierre-loup-tristant-sonarsource: It is a vulnerability for Python as it also is for Java and PHP.

python-checks/src/main/java/org/sonar/python/checks/WeakSSLProtocolCheck.java
andrea-guarino-sonarsource: Let me jump in the discussion. Just to clarify, currently the semantic has some limitations and inconsistencies regarding symbol resolution:

```python
from foo import A
from foo.A import fn

A.fn() # Here "getSymbol()" will retrieve the correct symbol for "A.fn" PythonGrammar.ATTRIBUTE_REF (corresponding to PyQualifiedExpressionTree), but not for "fn" PythonGrammar.ATOM node

fn() # Here "getSymbol()" will retrieve the correct symbol for "fn" PythonGrammar.ATOM node
```

I agree we shouldn't fix the semantics right now, I can create a ticket for this.

benzonico: @andrea-guarino-sonarsource be my guest and please link it here !

PR #280, opened 2019-08-30
SONARPY-359 Rule S4426: Cryptographic keys generations should be based on strong parameters

andrea-guarino-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/hotspots/StrongCryptographicKeysCheck.java
```diff
@@ -0,0 +1,156 @@
...
+    if (isExponentArgument(argument, index, library)) {
+      ctx.addIssue(argument, "Use a public key exponent of at least 65537.");
+    }
+
+    if (library.equals("cryptography") && isCurveArgument(argument, index, ctx)) {
+      ctx.addIssue(argument, "Use a key length of at least 224 bits.");
+    }
+  }
+
+  private static boolean isCurveArgument(PyArgumentTree argument, int index, SubscriptionContext ctx) {
```
pynicolas: Can be renamed to "isNonCompliantCurveArgument".

sonar-python-plugin/src/test/java/org/sonar/plugins/python/PythonRuleRepositoryTest.java
```diff
@@ -40,7 +40,7 @@ public void createRulesTest() {
 
     List<RulesDefinition.Rule> rules = repository.rules();
     assertThat(rules).isNotNull();
-    assertThat(rules).hasSize(65);
+    assertThat(rules).hasSize(66);
```
pynicolas: We should change that test, it makes it cumbersome to implement rules in parallel.

andrea-guarino-sonarsource: I agree! Let's do it in a separate PR!

PR #284, opened 2019-09-02
SONARPY-376 Rule S5443: Using publicly writable directories is securi…

andrea-guarino-sonarsource:
…ty-sensitive


python-checks/src/main/java/org/sonar/python/checks/hotspots/PubliclyWritableDirectoriesCheck.java
```diff
@@ -0,0 +1,109 @@
...
+
+  private static final Pattern WINDOWS_WRITABLE_DIRECTORIES = Pattern.compile("[^\\\\]*\\\\(Windows\\\\Temp|Temp|TMP).*", Pattern.CASE_INSENSITIVE);
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Kind.STRING_LITERAL, ctx -> {
+      PyStringLiteralTree tree = (PyStringLiteralTree) ctx.syntaxNode();
+      String stringLiteral = tree.trimmedQuotesValue().toLowerCase();
+      if (UNIX_WRITABLE_DIRECTORIES.stream().anyMatch(stringLiteral::startsWith) ||
+        WINDOWS_WRITABLE_DIRECTORIES.matcher(stringLiteral).matches()) {
```
pynicolas: I think we could refine that to avoid some FPs:
```
open("/tmpx")
open("C:\Temperatures")
```

python-checks/src/main/java/org/sonar/python/checks/hotspots/PubliclyWritableDirectoriesCheck.java
```diff
@@ -0,0 +1,118 @@
...
+
+  }
+
+  private static boolean containsDirectory(String stringLiteral, String dir) {
+    return stringLiteral.startsWith(dir) || stringLiteral.equals(dir.substring(0, dir.length() - 1));
+  }
+
+  private boolean isNonCompliantOsEnvironArgument(Tree tree) {
+    Tree argumentOrExpression = tree;
+    if (tree.is(Kind.ARGUMENT)) {
```
pynicolas: I think we could remove this `if` and add `.map(PyArgumentTree::expression)` at line 64.

PR #287, opened 2019-09-03
Migrate TooManyReturns check to Strongly typed AST

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/TooManyReturnsCheck.java
```diff
@@ -42,25 +41,47 @@
...
+        });
       }
+    });
+  }
+
+  private static class ReturnCountVisitor extends BaseTreeVisitor {
+
+    private int count = 0;
+    private List<PyReturnStatementTree> returnStatements = new ArrayList<>();
+    private List<PyYieldStatementTree> yieldStatements = new ArrayList<>();
```
pynicolas: Why do we need 2 separate lists instead of 1?

python-checks/src/main/java/org/sonar/python/checks/TooManyReturnsCheck.java
```diff
@@ -42,25 +41,47 @@
...
+        returnCountVisitor.returnStatements.forEach(r -> {
+          preciseIssue.secondary(r, null);
+        });
       }
+    });
+  }
+
+  private static class ReturnCountVisitor extends BaseTreeVisitor {
+
+    private int count = 0;
```
pynicolas: Why do we need to store that instead of calling the `size` method of the lists?

python-checks/src/main/java/org/sonar/python/checks/TooManyReturnsCheck.java
guillaume-dequenne-sonarsource: We do not! New commit fixes this.

python-checks/src/main/java/org/sonar/python/checks/TooManyReturnsCheck.java
guillaume-dequenne-sonarsource: New commit fixes this.

pynicolas:
There are still some issues raised by SonarQube which I believe should be addressed.

python-checks/src/main/java/org/sonar/python/checks/TooManyReturnsCheck.java
```diff
@@ -47,12 +48,8 @@ public void initialize(Context context) {
...
 
-      if (returnCountVisitor.count > max) {
-        PreciseIssue preciseIssue = ctx.addIssue(func.name(), String.format(MESSAGE, returnCountVisitor.count, max));
-
-        returnCountVisitor.yieldStatements.forEach(y -> {
-          preciseIssue.secondary(y, null);
-        });
+      if (returnCountVisitor.returnStatements.size() > max) {
+        PreciseIssue preciseIssue = ctx.addIssue(func.name(), String.format(MESSAGE, returnCountVisitor.returnStatements.size(), max));
         returnCountVisitor.returnStatements.forEach(r -> {
```
benzonico: No need for block for this lambda. you can remove braces. #lessIsMore !

PR #294, opened 2019-09-04
Strongly typed AST: dictionary and set literal

andrea-guarino-sonarsource:


python-squid/src/main/java/org/sonar/python/tree/PyDictionaryLiteralTreeImpl.java
```diff
@@ -0,0 +1,60 @@
...
+
+  private final Set<PyKeyValuePairTree> elements;
+
+  public PyDictionaryLiteralTreeImpl(AstNode node, Token lCurlyBrace, List<Token> commas, Set<PyKeyValuePairTree> elements, Token rCurlyBrace) {
+    super(node, lCurlyBrace, commas, rCurlyBrace);
+    this.elements = elements;
+  }
+
+  @Override
+  public Set<PyKeyValuePairTree> elements() {
```
benzonico: A set seems to be a not so good choice in terms of API : order is not guaranteed for API user whereas in the case of an AST order of elements matters.

python-squid/src/main/java/org/sonar/python/api/tree/PyDictOrSetLiteralTree.java
```diff
@@ -0,0 +1,31 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.python.api.tree;
+
+import com.sonar.sslr.api.Token;
+import java.util.List;
+
+public interface PyDictOrSetLiteralTree extends PyExpressionTree {
```
benzonico: Given this interface is just a way to factorize code (and btw, factorization by inheritance is not necessarily a good pattern) I believe you could push things one step further with a type parameter : 

```
class PyDictOrSetLiteralTree<T ? extends ExpressionTree>...  {
  List<T> elements();
}
```
and specialize this in the inheritance.

benzonico: Not really sure this is worth it though... I would eventually be more in favor of not complexifying the type hierarchy and I would strongly suggest to avoid this intermediate level in the API as it is not meaningful in the interfaces (those don't share a common thing in term of language construction). (fine in the implementation though, even if it is factorization through inheritance).

python-squid/src/main/java/org/sonar/python/tree/PyDictionaryLiteralTreeImpl.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  @Override
+  public void accept(PyTreeVisitor visitor) {
+    visitor.visitDictionaryLiteral(this);
+  }
+
+  @Override
+  public List<Tree> children() {
+    return elements.stream().map(element -> (Tree) element).collect(Collectors.toList());
```
benzonico: Why Collections.unmodifiableList would not work like in PyStatementListTreeImpl ?

python-squid/src/main/java/org/sonar/python/tree/PySetLiteralTreeImpl.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  @Override
+  public void accept(PyTreeVisitor visitor) {
+    visitor.visitSetLiteral(this);
+  }
+
+  @Override
+  public List<Tree> children() {
+    return elements.stream().map(element -> (Tree) element).collect(Collectors.toList());
```
benzonico: Why Collections.unmodifiableList would not work like in PyStatementListTreeImpl ?

PR #296, opened 2019-09-04
SONARPY-379 Strongly typed AST: Cognitive complexity and compound assignements

andrea-guarino-sonarsource:


python-squid/src/main/java/org/sonar/python/tree/PyForStatementTreeImpl.java
```diff
@@ -63,7 +76,7 @@ public void accept(PyTreeVisitor visitor) {
 
   @Override
   public Token forKeyword() {
-    return null;
+    return forKeyword;
```
pynicolas: woohoo!

PR #301, opened 2019-09-09
Migrate function, method and class name checks to Strongly typed AST

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/MethodNameCheck.java
```diff
@@ -31,8 +35,29 @@ public String typeName() {
...
+    }
+  }
+
+  private static boolean classHasInheritance(PyClassDefTree classDef) {
+    PyArgListTree args = classDef.args();
+    if(args == null) {
+      return false;
+    }
+    List<Tree> children = args.children();
+    if(children.isEmpty()) {
```
benzonico: I think you lack a test for this case : 
`class Foo():`

python-checks/src/main/java/org/sonar/python/checks/FunctionNameCheck.java
```diff
@@ -21,6 +21,7 @@
 
 import com.sonar.sslr.api.AstNode;
```
benzonico: This import should be removed.

python-checks/src/main/java/org/sonar/python/checks/ClassNameCheck.java
```diff
@@ -45,18 +43,15 @@ protected String format() {
   }
 
   @Override
-  public Set<AstNodeType> subscribedKinds() {
-    return Collections.singleton(PythonGrammar.CLASSDEF);
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.CLASSDEF, ctx -> {
+      PyClassDefTree pyClassDefTree = ((PyClassDefTree) ctx.syntaxNode());
```
benzonico: unecessary parentheses.

PR #308, opened 2019-09-10
Migrate BreakContinueOutsideLoopCheck to Strongly typed AST

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/BreakContinueOutsideLoopCheck.java
```diff
@@ -19,41 +19,38 @@
...
-  @Override
-  public void visitNode(AstNode node) {
-    AstNode currentParent = node.getParent();
-    while (currentParent != null){
-      if (currentParent.is(PythonGrammar.WHILE_STMT, PythonGrammar.FOR_STMT)){
+  private static final Consumer<SubscriptionContext> SUBSCRIPTION_CONTEXT_CONSUMER = ctx -> {
+    PyStatementTree statement = (PyStatementTree) ctx.syntaxNode();
+    Tree currentParent = statement.parent();
+    while (currentParent != null) {
+      if (currentParent.is(Tree.Kind.WHILE_STMT) || currentParent.is(Tree.Kind.FOR_STMT)) {
```
pynicolas: I think it would be worth adding a test case with a `while` loop.

python-checks/src/main/java/org/sonar/python/checks/BreakContinueOutsideLoopCheck.java
```diff
@@ -19,41 +19,38 @@
...
         return;
-      } else if (currentParent.is(PythonGrammar.FUNCDEF, PythonGrammar.CLASSDEF)){
-        raiseIssue(node);
+      } else if (currentParent.is(Tree.Kind.CLASSDEF) || currentParent.is(Tree.Kind.FUNCDEF)) {
```
pynicolas: I think it would be worth adding a test case with a class definition.

pynicolas: In fact, the test cases which you added with a class definition are not really relevant. If I remove `(currentParent.is(Tree.Kind.CLASSDEF) ||` from the line above, tests are still green. I would add the following test case:
```
for i in range(1):
    class Class(object):
        print(1)
        break #Noncompliant
```

PR #314, opened 2019-09-10
SONARPY-407 CallExpression: ArgList should be null if there are no parameters

andrea-guarino-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/hotspots/StrongCryptographicKeysCheck.java
```diff
@@ -46,7 +47,7 @@
   public void initialize(Context context) {
     context.registerSyntaxNodeConsumer(Kind.CALL_EXPR, ctx -> {
       PyCallExpressionTree callExpression = (PyCallExpressionTree) ctx.syntaxNode();
-      List<PyArgumentTree> arguments = callExpression.arguments().arguments();
+      List<PyArgumentTree> arguments = callExpression.arguments() != null ? callExpression.arguments().arguments() : Collections.emptyList();
```
pynicolas: We may think about having a dedicated method on `CallExpression`.

python-squid/src/main/java/org/sonar/python/tree/PythonTreeMaker.java
```diff
@@ -968,8 +971,11 @@ public PyQualifiedExpressionTree qualifiedExpression(AstNode astNode) {
   public PyCallExpressionTree callExpression(AstNode astNode) {
     PyExpressionTree callee = expression(astNode.getFirstChild());
     AstNode argListNode = astNode.getFirstChild(PythonGrammar.ARGLIST);
-    List<PyArgumentTree> arguments = argList(argListNode);
-    return new PyCallExpressionTreeImpl(astNode, callee, new PyArgListTreeImpl(argListNode, arguments),
+    PyArgListTree argListTree = null;
+    if (argListNode != null) {
+      argListTree = new PyArgListTreeImpl(argListNode, argList(argListNode));
+    }
```
pynicolas: We could factorize that code in a "argList" method which could return null.

PR #334, opened 2019-09-16
Migrate regex check

benzonico:


python-checks/src/main/java/org/sonar/python/checks/hotspots/RegexCheck.java
```diff
@@ -78,13 +73,20 @@ private void checkRegexArgument(AstNode arg) {
...
     if (isSuspiciousRegex(literal)) {
-      PreciseIssue preciseIssue = addIssue(atom, MESSAGE);
+      PreciseIssue preciseIssue = ctx.addIssue(arg, MESSAGE);
       if (secondaryLocation != null) {
         preciseIssue.secondary(secondaryLocation);
       }
     }
   }
 
+  private PyExpressionTree getExpression(PyExpressionTree expr) {
```
guillaume-dequenne-sonarsource: It's a small thing but SonarQube would like this method to be static :)

python-checks/src/main/java/org/sonar/python/checks/hotspots/RegexCheck.java
```diff
@@ -78,13 +73,20 @@ private void checkRegexArgument(AstNode arg) {
...
+    if (expr.is(Tree.Kind.MODULO) || expr.is(Tree.Kind.PLUS)) {
+      return getExpression(((PyBinaryExpressionTree) expr).leftOperand());
```
guillaume-dequenne-sonarsource: Does this mean we're going to ignore cases where the right operand contains the suspicious part of a regex? A case like the following does not raise any issue.
test = 'ab'
sometest = 'a(a*)*'
pattern = re.compile(test + sometest)
testpattern = re.compile(test + '(a*)*')

benzonico: This is a very true remark. 
Not that for me this is the naive (and thus wrong) way to do dataflow analysis without realizing it. 
Therefore I am just reimplementing what was detected previously as to improve this part to track value of a string we would need to put in place a really different kind of analysis which is not at all in our scope right now.

PR #336, opened 2019-09-17
Fix quality flaws

benzonico:


python-squid/src/main/java/org/sonar/python/PythonVisitorContext.java
```diff
@@ -41,20 +41,18 @@
 
 
   public PythonVisitorContext(AstNode rootAst, PyFileInputTree rootTree, PythonFile pythonFile) {
-    this(rootAst, rootTree, pythonFile, null);
+    this.rootAst = rootAst;
+    this.rootTree = (PyFileInputTreeImpl) rootTree;
```
pynicolas: I don't get why the `rootTree` field is declared as `PyFileInputTreeImpl` instead of `PyFileInputTree`.

benzonico: good point, this was probably not necessary before either. Sticking to interface.

PR #337, opened 2019-09-17
SONARPY-324 Rule S2245: Using PRNGs is security-sensitive

pynicolas:


sonar-python-plugin/src/test/java/org/sonar/plugins/python/PythonRuleRepositoryTest.java
```diff
@@ -39,7 +43,7 @@ public void createRulesTest() {
 
     List<RulesDefinition.Rule> rules = repository.rules();
     assertThat(rules).isNotNull();
-    assertThat(rules).hasSize(68);
+    assertThat(rules).hasSameSizeAs(nonAbstractCheckFiles());
```
andrea-guarino-sonarsource: 👍

PR #338, opened 2019-09-17
SymbolTable to Strongly typed AST: local vars in Function definition

andrea-guarino-sonarsource:


python-squid/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -0,0 +1,224 @@
...
+    public void visitFunctionDef(PyFunctionDefTree pyFunctionDefTree) {
+      createScope(pyFunctionDefTree, currentScope());
+      super.visitFunctionDef(pyFunctionDefTree);
+    }
+
+    @Override
+    public void visitAssignmentStatement(PyAssignmentStatementTree pyAssignmentStatementTree) {
+      pyAssignmentStatementTree.lhsExpressions().stream()
+        .flatMap(exprList -> exprList.expressions().stream())
+        .filter(expr -> expr.is(Kind.NAME))
```
pynicolas: what about tuples?

PR #345, opened 2019-09-19
Sonarpy 416

andrea-guarino-sonarsource:


python-squid/src/main/java/org/sonar/python/semantic/Usage.java
```diff
@@ -0,0 +1,43 @@
...
+    return kind() != Kind.OTHER;
+  }
+
+  Tree tree();
+
+  Kind kind();
+
+  enum Kind {
+    ASSIGNMENT,
+    COMPOUND_ASSIGNMENT,
```
pynicolas: I suggest `ASSIGNMENT_LHS`, `COMPOUND_ASSIGNMENT_LHS`

python-squid/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -346,8 +347,8 @@ public void visitDecorator(PyDecoratorTree pyDecoratorTree) {
       PyNameTree nameTree = pyDecoratorTree.name().names().get(0);
       Scope scope = scopesByRootTree.get(currentScopeRootTree());
       SymbolImpl symbol = scope.resolve(nameTree.name());
-      if (symbol != null && !symbol.usages().contains(nameTree)) {
-        symbol.addUsage(nameTree);
+      if (symbol != null && symbol.usages().stream().noneMatch(usage -> usage.tree().equals(nameTree))) {
```
pynicolas: That's not ideal from a perspective point of view.

PR #357, opened 2019-09-24
Migrate TrailingCommentCheck to Strongly typed AST

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/TrailingCommentCheck.java
```diff
@@ -19,16 +19,20 @@
...
+import org.sonar.python.PythonSubscriptionCheck;
+import org.sonar.python.api.tree.PyToken;
+import org.sonar.python.api.tree.Tree;
+import org.sonar.python.tree.PyTokenImpl;
 
 @Rule(key = TrailingCommentCheck.CHECK_KEY)
-public class TrailingCommentCheck extends PythonCheckAstNode {
+public class TrailingCommentCheck extends PythonSubscriptionCheck {
 
   public static final String CHECK_KEY = "S139";
```
benzonico: Please inline it no need to have a public constant here. Thanks.

python-squid/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
```diff
@@ -1735,7 +1737,9 @@ public void generator_expression() {
     setRootRule(PythonGrammar.TEST);
     ComprehensionExpression generator = (ComprehensionExpression) parse("(x*x for x in range(10))", treeMaker::expression);
     assertThat(generator.getKind()).isEqualTo(Tree.Kind.GENERATOR_EXPR);
-    assertThat(generator.children()).hasSize(2);
+    assertThat(generator.children()).hasSize(4);
+    assertThat(((Token) generator.children().get(0)).value()).isEqualTo("(");
+    assertThat(((Token) generator.children().get(3)).value()).isEqualTo(")");
```
benzonico: I think this test can be more robust by checking that this token should be equal to "firstToken" and "lastToken"

PR #359, opened 2019-09-24
SONARPY-414 Use new API for symbol fully qualified name in existing checks

andrea-guarino-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/hotspots/HashingDataCheck.java
```diff
@@ -152,48 +155,31 @@ private static PyExpressionTree removeParenthesis(PyExpressionTree expression) {
   }
 
   private void checkQuestionableHashingAlgorithm(SubscriptionContext ctx) {
-    this.ctx = ctx;
     PyNameTree name = (PyNameTree) ctx.syntaxNode();
     if (isWithinImport(name)) {
       return;
     }
-    String qualifiedName = getQualifiedName(name, ctx);
-    if (qualifiedName.equals("cryptography.hazmat.primitives.hashes") && isHashesFunctionCall(name)) {
```
pynicolas: I don't see how this condition is replaced.

andrea-guarino-sonarsource: Before our rewriting of symbol table, we had to handle this case separately:
```python
from cryptography.hazmat.primitives import hashes
hashes.Hash(algorithm) # Noncompliant
```
Now we can just rely on callExpression#calleeSymbol() to resolve its fully qualified name.

Note this is tested [here](https://github.com/SonarSource/sonar-python/blob/a6e670c402df5587fe24cdb681b30108d43d073b/python-checks/src/test/resources/checks/hotspots/hashing/hashingData.py#L11)

PR #360, opened 2019-09-24
SONARPY-394 Migrate FieldNameCheck to strongly typed AST

pynicolas:


python-squid/src/test/java/org/sonar/python/semantic/SemanticTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+import com.sonar.sslr.api.Grammar;
+import com.sonar.sslr.impl.Parser;
+import java.nio.charset.StandardCharsets;
+import org.sonar.python.PythonConfiguration;
+import org.sonar.python.api.tree.PyFileInputTree;
+import org.sonar.python.parser.PythonParser;
+import org.sonar.python.tree.PythonTreeMaker;
+
+public class SemanticTest {
```
benzonico: This is factorization by inheritance, I would rather make this a utility class with a static method.

python-squid/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -83,6 +88,13 @@ public void visitFileInput(PyFileInputTree pyFileInputTree) {
           }
         }
       });
+    scopesByRootTree.values().stream()
+      .filter(scope -> scope.rootTree instanceof PyClassDefTree)
```
benzonico: Why not use of .is(Tree.Kind.CLASSDEF) ?

its/ruling/src/test/resources/expected/python-S116.json
```diff
@@ -526,7 +526,6 @@
 'project:twisted-12.1.0/twisted/internet/task.py':[
 63,
 64,
-112,
```
benzonico: I don't understand why this issue is missing and not relevant anymore : this is an instance field and should be detected as such.

benzonico: And I figured it out : this self variable is instantiated via a class method. This is way out of scope of this sprint, so fine to accept this "loss" for now.

PR #362, opened 2019-09-25
SONARPY-323 Rule S4792: Configuring loggers is security-sensitive

andrea-guarino-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/hotspots/LoggersConfigurationCheck.java
```diff
@@ -0,0 +1,93 @@
...
+import org.sonar.python.api.tree.PyAssignmentStatementTree;
+import org.sonar.python.api.tree.PyCallExpressionTree;
+import org.sonar.python.api.tree.PyClassDefTree;
+import org.sonar.python.api.tree.Tree;
+import org.sonar.python.semantic.TreeSymbol;
+
+@Rule(key = "S4792")
+public class LoggersConfigurationCheck extends PythonSubscriptionCheck {
+
+  private static final Set<String> FUNCTIONS_TO_CHECK = new HashSet<>(Arrays.asList(
```
benzonico: Why do we need to construct a set ? Given that we only check if the collection contains a string, list is sufficient, especially given we hardcode those lists.

python-checks/src/main/java/org/sonar/python/checks/hotspots/LoggersConfigurationCheck.java
```diff
@@ -0,0 +1,93 @@
...
+        .forEach(expr -> ctx.addIssue(expr, MESSAGE));
+    }
+  }
+
+  // check if logging.lastResort is being set
+  private static void isSettingLastResort(SubscriptionContext ctx, PyAssignmentStatementTree assignment) {
+    assignment.lhsExpressions().stream()
+      .flatMap(exprList -> exprList.expressions().stream())
+      .forEach(expr -> {
+        if (expr instanceof HasSymbol) {
```
benzonico: This is past the scope of this PR but there is some design flaw in terms of OOP in the fact that we have to check if an expr is a "hasSymbol" instance via an instance of AND then check if symbol is null. 
Why not having a symbol method on every expression for instance and returning a "NASymbol" or a "unknown" in the different cases.

andrea-guarino-sonarsource: I agree this is a bit cumbersome, I will take this point in a separate PR

PR #363, opened 2019-09-25
SONARPY-373 Rule S4507 should consider settings.py in addition to glo…

andrea-guarino-sonarsource:
…bal_settings.py

### DIFF
--- a/python-checks/src/main/java/org/sonar/python/checks/hotspots/DebugModeCheck.java
+++ b/python-checks/src/main/java/org/sonar/python/checks/hotspots/DebugModeCheck.java
@@ -20,9 +20,7 @@
 package org.sonar.python.checks.hotspots;
 
 import java.util.Arrays;
-import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


python-checks/src/main/java/org/sonar/python/checks/hotspots/DebugModeCheck.java
```diff
@@ -41,6 +41,7 @@
   public static final String CHECK_KEY = "S4507";
   private static final String MESSAGE = "Make sure this debug feature is deactivated before delivering the code in production.";
   private static final Set<String> debugProperties = new HashSet<>(Arrays.asList("DEBUG", "DEBUG_PROPAGATE_EXCEPTIONS"));
+  private static final Set<String> settingFiles = new HashSet<>(Arrays.asList("global_settings.py", "settings.py"));
```
benzonico: Why a Set ? A list is sufficient.

PR #370, opened 2019-09-27
Update metadata

andrea-guarino-sonarsource:


python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S4507.html
```diff
@@ -32,12 +32,23 @@ <h2>Sensitive Code Example</h2>
...
+<p>Flask</p>
+<pre>
+from flask import Flask
+
+app = Flask(__name__)
+
+app.testing = True  # Questionable
+app.debug = True  # Questionable
+
+app.run(debug=True) # Questionable
```
pynicolas: Do we really raise issues on those Flask cases?

andrea-guarino-sonarsource: Indeed, we don't support them. I will remove this update. Thanks!

PR #376, opened 2019-10-01
Remove SSLR token dependency from trivia related checks

guillaume-dequenne-sonarsource:


python-squid/src/main/java/org/sonar/python/api/tree/Token.java
```diff
@@ -20,11 +20,14 @@
 package org.sonar.python.api.tree;
 
 import com.sonar.sslr.api.TokenType;
-import com.sonar.sslr.api.Trivia;
 import java.util.List;
 
 public interface Token extends Tree {
 
+  /**
+   * @deprecated (Use of sslr dependencies should be avoided)
```
benzonico: nitpick : no need for parenthesis in the comment.

python-squid/src/main/java/org/sonar/python/api/tree/Trivia.java
```diff
@@ -0,0 +1,37 @@
...
+
+public interface Trivia {
+
+  /**
+   * @deprecated (Use of sslr dependencies should be avoided)
+   */
+  @Deprecated
+  com.sonar.sslr.api.Trivia trivia();
+
+  Token token();
```
benzonico: I believe this is worth documenting what is that token.

python-squid/src/main/java/org/sonar/python/api/tree/Trivia.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.util.List;
+
+public interface Trivia {
+
+  /**
+   * @deprecated (Use of sslr dependencies should be avoided)
+   */
+  @Deprecated
+  com.sonar.sslr.api.Trivia trivia();
```
benzonico: This is not required at API level.

python-squid/src/main/java/org/sonar/python/api/tree/Trivia.java
```diff
@@ -0,0 +1,37 @@
...
+
+  Token token();
+
+  List<Token> tokens();
+
+  boolean isComment();
```
benzonico: I don't think we need this in the API.

python-squid/src/main/java/org/sonar/python/api/tree/Trivia.java
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * @deprecated (Use of sslr dependencies should be avoided)
+   */
+  @Deprecated
+  com.sonar.sslr.api.Trivia trivia();
+
+  Token token();
+
+  List<Token> tokens();
```
benzonico: This is not used in checks, so I believe we could just drop this one.

PR #378, opened 2019-10-02
Migrate CommentRegularExpressionCheck to Strongly typed AST

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/CommentRegularExpressionCheck.java
```diff
@@ -19,24 +19,23 @@
...
-  public static final String CHECK_KEY = "CommentRegularExpression";
   private static final String DEFAULT_REGULAR_EXPRESSION = "";
   private static final String DEFAULT_MESSAGE = "The regular expression matches this comment";
 
   private Pattern pattern = null;
 
-  @RuleProperty(
-    key = "regularExpression",
-    defaultValue = "" + DEFAULT_REGULAR_EXPRESSION)
+  @RuleProperty(key = "regularExpression")
```
andrea-guarino-sonarsource: Why did you remove the `defaultValue`?

guillaume-dequenne-sonarsource: It was redundant, as DEFAULT_REGULAR_EXPRESSION is already set to ""

PR #380, opened 2019-10-02
Migrate LineLengthCheck to Strongly typed AST

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/LineLengthCheck.java
```diff
@@ -39,44 +39,33 @@
...
-  }
-
   @Override
-  public void visitFile(AstNode astNode) {
-    previousToken = null;
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.FILE_INPUT, ctx -> previousToken = null);
+    context.registerSyntaxNodeConsumer(Tree.Kind.TOKEN, ctx -> {
+      Token token = (Token) ctx.syntaxNode();
+      if (token.isGeneratedCode()) {
```
pynicolas: I think we can drop that. I don't think that there's anything in Python which we consider as generated code. I would drop `org.sonar.python.api.tree.Token.isGeneratedCode()` as well.

python-checks/src/main/java/org/sonar/python/checks/LineLengthCheck.java
```diff
@@ -39,44 +39,33 @@
...
+        }
+      }
+      previousToken = token;
+    });
   }
 
   @Override
-  public void leaveFile(AstNode astNode) {
+  public void leaveFile() {
     previousToken = null;
```
pynicolas: I think that we don't need that.

python-squid/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
```diff
@@ -609,15 +610,28 @@ public void importFromStatement() {
     assertThat(aliasedNameTree.asKeyword()).isNull();
     assertThat(aliasedNameTree.alias()).isNull();
     assertThat(aliasedNameTree.dottedName().names().get(0).name()).isEqualTo("f");
-    assertThat(importStatement.children()).hasSize(4);
+    List<Tree> children = importStatement.children();
+    assertThat(children).hasSize(4);
+    assertThat(children.get(0)).isEqualTo(importStatement.fromKeyword());
+    assertThat(children.get(1)).isEqualTo(importStatement.module());
+    assertThat(children.get(2)).isEqualTo(importStatement.importKeyword());
+    assertThat(children.get(3)).isEqualTo(aliasedNameTree);
```
pynicolas: Another way to assert the same thing would have been:
```
assertThat(importStatement.children()).containsExactly(
  importStatement.fromKeyword(),
  importStatement.module(),
  importStatement.importKeyword(),
  aliasedNameTree
);
```

python-checks/src/main/java/org/sonar/python/checks/LineLengthCheck.java
guillaume-dequenne-sonarsource: Newline tokens are actually considered generated code.

Dropping this altogether means we will raise issues on lines ending with a trailing comment, like: `foo = bar() # some very long comment` as the final \n will be used to determine the end of the line (instead of the closing parenthesis).

I assume we might actually want this to happen (the same way we might want to raise issues on lines only containing a # comment), but just to make clear that the behavior changes.

That said, I might simply rework this to ignore newline tokens and dropping the isGeneratedCode() method nevertheless.

PR #382, opened 2019-10-04
Migrate CommentedCodeCheck to Strongly typed AST

guillaume-dequenne-sonarsource:


python-squid/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
```diff
@@ -1848,28 +1848,42 @@ public void generator_expression() {
   public void tuples() {
     Tuple empty = parseTuple("()");
     assertThat(empty.getKind()).isEqualTo(Tree.Kind.TUPLE);
+    assertThat(empty.firstToken().value()).isEqualTo("(");
+    assertThat(empty.lastToken().value()).isEqualTo(")");
```
benzonico: IMO it is better to check that first token isSameAs leftParenthesis. Same for lastToken and rightParenthesis

python-checks/src/main/java/org/sonar/python/checks/CommentedCodeCheck.java
```diff
@@ -130,26 +121,25 @@ private static boolean isEmpty(String text) {
...
     }
-    AstNode expressionStatement = expressionStatements.get(0);
-    return (expressionStatement.getNumberOfChildren() == 1 && expressionStatement.getFirstChild().is(PythonGrammar.TESTLIST_STAR_EXPR))
-      || expressionStatement.hasDirectChildren(PythonGrammar.ANNASSIGN);
+    Statement statement = fileInput.statements().statements().get(0);
+    return statement.is(Tree.Kind.EXPRESSION_STMT) || statement.is(Tree.Kind.ANNOTATED_ASSIGNMENT);
   }
 
   private static List<List<Trivia>> groupTrivias(Token token) {
     List<List<Trivia>> result = new LinkedList<>();
```
benzonico: I know this is not from you, but can you change to use arrayList, there is no good usage of linked list.

python-checks/src/main/java/org/sonar/python/checks/CommentedCodeCheck.java
```diff
@@ -20,88 +20,79 @@
...
-  @Override
-  public void visitToken(Token token) {
-    List<List<Trivia>> groupedTrivias = groupTrivias(token);
-    for (List<Trivia> triviaGroup : groupedTrivias) {
-      checkTriviaGroup(triviaGroup);
-    }
+  private static boolean isMultilineComment(StringLiteral stringLiteral) {
+    Tree parent = stringLiteral.parent();
+    StringElement firstElement = stringLiteral.stringElements().get(0);
+    return (firstElement.value().endsWith("'''") || firstElement.value().endsWith("\"\"\"")) && parent.is(Tree.Kind.EXPRESSION_STMT);
```
benzonico: Shouldn't this be part of API of StringElement ?

PR #392, opened 2019-10-08
SONARPY-398 Migrate S1763 implementation to use CFG

andrea-guarino-sonarsource:


python-squid/src/main/java/org/sonar/python/cfg/ControlFlowGraphBuilder.java
```diff
@@ -154,6 +154,15 @@ private PythonCfgBlock build(Statement statement, PythonCfgBlock currentBlock) {
     return currentBlock;
   }
 
+  private PythonCfgBlock buildWithStatement(WithStatement statement, PythonCfgBlock successor) {
+    // To avoid FP, we assume that exception raised inside with statement will be suppressed by __exit__() method of Context Manager
+    // see https://docs.python.org/3/reference/compound_stmts.html#the-with-statement
+    exceptionTargets.push(successor);
```
benzonico: I am really wondering if this should be fixed at CFG level or rule level. As far as I understand this is not wrong at CFG level but rather on its usage ?

python-checks/src/test/resources/checks/afterJumpStatement.py
```diff
@@ -151,3 +151,8 @@ def with_stmt_raise(expect_raise, e):
     with expect_raise():
         raise e
     print("foo") # OK
+
+def return_inside_class():
+    class Foo:
+        return
+    print("foo") # OK
```
benzonico: If this is not valid python code, I am wondering what we are trying to deal with in this commit : what is the problem ?

its/ruling/src/test/resources/expected/python-S1751.json
```diff
@@ -1,5 +0,0 @@
-{
-'project:S2711.py':[
-6,
```
benzonico: Here we are removing a TRUE POSITIVE on a check that is not supposed to be the target of what we are doing. For me this is a regression and thus, there is something fishy going on.

andrea-guarino-sonarsource: As discussed, this was raised on the following code:
```
class MyClass(object):
    ' doc string '
    while True:
        return False
```
that although being correct syntactically, it's not correct semantically, so we don't create a CFG for it.

python-checks/src/test/resources/checks/afterJumpStatement.py
andrea-guarino-sonarsource: this is a unit test to assert we are not creating a CFG for this semantically incorrect code and, thus, avoiding having a FP at line 158

python-squid/src/test/java/org/sonar/plugins/python/api/cfg/ControlFlowGraphTest.java
```diff
@@ -404,6 +404,19 @@ public void class_def() {
     );
   }
 
+  /**
+   * The predecessors are constructed based on the successors, so we should not test them all the time
```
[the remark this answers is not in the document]
andrea-guarino-sonarsource: 'we' is used impersonally, do you prefer the passive voice?
In other CFG tests, we always test successors. Because predecessors are constructed based on successors, there is no need to add tests for predecessors in all other test cases, i.e. it won't increase the coverage.

python-squid/src/main/java/org/sonar/python/cfg/ControlFlowGraphBuilder.java
andrea-guarino-sonarsource: I agree, I will handle this at rule level!

andrea-guarino-sonarsource: Updated in 728df2df488632c606e1f0dbb85ef65b35ae97c7

python-squid/src/test/java/org/sonar/plugins/python/api/cfg/ControlFlowGraphTest.java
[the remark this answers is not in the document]
andrea-guarino-sonarsource: I updated the comment in 1c01d835e45b30377be88810ebf869c45c134685

python-checks/src/test/resources/checks/afterJumpStatement.py
benzonico: I think this is worth mentioning in a comment above the testcase : something along the line of : "this is semantically incorrect python code, preventing the CFG generation, and so no issue is raised in this case"

its/ruling/src/test/resources/expected/python-S1751.json
benzonico: Fair enough : we only raise issue on valid python code.

PR #393, opened 2019-10-08
SONARPY-432 Support symbol highlighting

guillaume-dequenne-sonarsource:


python-squid/src/test/java/org/sonar/python/semantic/SymbolTableBuilderTreeTest.java
```diff
@@ -49,10 +52,18 @@
...
+    fileInput = context.rootTree();
     fileInput.accept(new TestVisitor());
   }
 
+  @Test
+  public void module_variable() {
+    Set<Symbol> moduleSymbols = fileInput.moduleFields();
+    List<String> names = Arrays.asList("global_x", "global_var");
+    assertThat(moduleSymbols.size()).isEqualTo(2);
+    moduleSymbols.forEach(s -> assertThat(names).contains(s.name()));
```
benzonico: `assertThat(moduleSymbols).extracting(name).containsExactly("global_x", "global_var");`

python-squid/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -96,6 +97,9 @@ public void visitFileInput(FileInput fileInput) {
         scope.symbols.forEach(classDef::addClassField);
         scope.instanceAttributesByName.values().forEach(classDef::addInstanceField);
       });
+    scopesByRootTree.values().stream()
```
benzonico: not related to your PR : I am just wondering if we could just avoid 3 iterations over values and have only one with a plain old for loop. If you're up to it, I think it is worth a commit of refactoring at the end of your PR.

sonar-python-plugin/src/test/java/org/sonar/plugins/python/PythonSquidSensorTest.java
```diff
@@ -258,4 +282,16 @@ private InputFile inputFile(String name) {
     return inputFile;
   }
 
+  private void verifyUsages(String componentKey, int line, int offset, TextPointer... tps) {
+    Collection<TextRange> textRanges = context.referencesForSymbolAt(componentKey, line, offset);
+    if (tps.length == 0) {
+      assertThat(textRanges).isEmpty();
+    } else {
+      assertThat(textRanges.stream().map(TextRange::start).collect(Collectors.toList())).isNotEmpty().containsOnly(tps);
```
benzonico: I don't think you need to collect to assert on a stream.  And moreover, the containsOnly assertion will check it is not Empty, so no need for the isNotEmpty call.

sonar-python-plugin/src/test/java/org/sonar/plugins/python/SymbolVisitorTest.java
```diff
@@ -0,0 +1,92 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class SymbolVisitorTest {
+
+  private static SensorContextTester context;
+  private static String componentKey;
+
+  @BeforeClass
+  public static void scanFile() {
+    String dir = "src/test/resources/org/sonar/plugins/python/squid-sensor";
```
benzonico: Why the dir variable ?  I think you can inline it and concatenate it to the file name.

sonar-python-plugin/src/test/java/org/sonar/plugins/python/SymbolVisitorTest.java
```diff
@@ -0,0 +1,92 @@
...
+    SymbolVisitor symbolVisitor = new SymbolVisitor(context.newSymbolTable().onFile(inputFile));
+    PythonVisitorContext context = TestPythonVisitorRunner.createContext(file);
+    FileInput fileInput = context.rootTree();
+    fileInput.accept(symbolVisitor);
+  }
+
+  @Test
+  public void symbol_visitor() {
+    assertThat(context.referencesForSymbolAt(componentKey, 1, 10)).isNull();
+    verifyUsages(componentKey, 1, 0);
```
benzonico: I think you could remove componentKey as parameter of this method and use the field directly.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/SymbolVisitor.java
```diff
@@ -0,0 +1,81 @@
...
+    fileInput.moduleFields().forEach(this::handleSymbol);
+    super.visitFileInput(fileInput);
+    newSymbolTable.save();
+  }
+
+  private void handleSymbol(Symbol symbol) {
+    List<Usage> usages = new ArrayList<>(symbol.usages());
+    usages.sort(Comparator.comparingInt(u -> u.tree().firstToken().line()));
+    Usage firstUsage = usages.get(0);
+    NewSymbol newSymbol = newSymbolTable.newSymbol(firstUsage.tree().firstToken().line(), firstUsage.tree().firstToken().column(),
```
benzonico: Maybe we can store firstUsage.tree() in a dedicated var instead of calling it 4 times ?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/SymbolVisitor.java
```diff
@@ -0,0 +1,81 @@
...
+      firstUsage.tree().lastToken().line(), firstUsage.tree().lastToken().column() + firstUsage.tree().lastToken().value().length());
+    for (int i = 1; i < usages.size(); i++) {
+      Usage usage = usages.get(i);
+      newSymbol.newReference(usage.tree().firstToken().line(), usage.tree().firstToken().column(),
+        usage.tree().lastToken().line(), usage.tree().lastToken().column() + firstUsage.tree().lastToken().value().length());
```
benzonico: Maybe we can store usage.tree() in a dedicated var instead of calling it 4 times ? which makes me realize that we might have a copy/paste bug here : there is a firstUsage.tree() call !!!  Why is this not caught by unit test  ? (seems that ending of reference is not tested)

sonar-python-plugin/src/test/java/org/sonar/plugins/python/PythonSquidSensorTest.java
benzonico: Please note that you only tested START of references and you most probably have a bug of end of references.

python-squid/src/test/java/org/sonar/python/semantic/SymbolTableBuilderTreeTest.java
guillaume-dequenne-sonarsource: Doesn't this assert on the order as well? which can't be guaranteed for a Set?

python-squid/src/main/java/org/sonar/python/tree/FileInputImpl.java
```diff
@@ -61,6 +65,15 @@ public Token docstring() {
     return docstring;
   }
 
+  @Override
+  public Set<Symbol> globalVariables() {
+    return globalVariables;
+  }
+
+  public void addModuleFields(Symbol field) {
```
benzonico: This ought to be rename to `addGlobalVariable` (and the param name as well)

sonar-python-plugin/src/test/java/org/sonar/plugins/python/SymbolVisitorTest.java
```diff
@@ -0,0 +1,90 @@
...
+    verifyUsages(19, 13, reference(22, 13, 22, 14), reference(26, 13, 26, 14));
+    verifyUsages(18, 17, reference(19, 8, 19, 12));
+    verifyUsages(21, 18, reference(22, 8, 22, 12), reference(23, 8, 23, 12));
+    verifyUsages(25, 19, reference(26, 8, 26, 12));
+    verifyUsages(28, 11, reference(28, 15, 28, 16), reference(28, 17, 28, 18));
+  }
+
+  private void verifyUsages(int line, int offset, TextRange... trs) {
+    Collection<TextRange> textRanges = context.referencesForSymbolAt(componentKey, line, offset);
+    if (trs.length == 0) {
```
benzonico: we probably don't need the condition.

sonar-python-plugin/src/test/java/org/sonar/plugins/python/PythonSquidSensorTest.java
```diff
@@ -258,4 +284,16 @@ private InputFile inputFile(String name) {
     return inputFile;
   }
 
+  private void verifyUsages(String componentKey, int line, int offset, TextRange... trs) {
+    Collection<TextRange> textRanges = context.referencesForSymbolAt(componentKey, line, offset);
+    if (trs.length == 0) {
```
benzonico: we probably don't need the condition

sonar-python-plugin/src/test/java/org/sonar/plugins/python/SymbolVisitorTest.java
```diff
@@ -0,0 +1,86 @@
...
+    verifyUsages(19, 13, reference(22, 13, 22, 14), reference(26, 13, 26, 14));
+    verifyUsages(18, 17, reference(19, 8, 19, 12));
+    verifyUsages(21, 18, reference(22, 8, 22, 12), reference(23, 8, 23, 12));
+    verifyUsages(25, 19, reference(26, 8, 26, 12));
+    verifyUsages(28, 11, reference(28, 15, 28, 16), reference(28, 17, 28, 18));
+  }
+
+  private void verifyUsages(int line, int offset, TextRange... trs) {
+    Collection<TextRange> textRanges = context.referencesForSymbolAt(componentKey, line, offset);
+      assertThat(textRanges).containsExactly(trs);
```
benzonico: indentation.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonScanner.java
```diff
@@ -120,6 +120,7 @@ private void scanFile(InputFile inputFile) {
     SubscriptionVisitor.analyze(checksBasedOnTree, visitorContext);
     saveIssues(inputFile, visitorContext.getIssues());
 
+    visitorContext.rootTree().accept(new SymbolVisitor(context.newSymbolTable().onFile(inputFile)));
```
benzonico: tiny detail but I personally think it would be more readable with : 

`new SymbolVisitor(context.newSymbolTable().onFile(inputFile)).visitFileInput(visitorContext.rootTree());` 

feel free to discard the suggestion.

PR #397, opened 2019-10-09
SONARPY-442 S1751 should not add secondary locations on jump statements of nested loops

andrea-guarino-sonarsource:


python-squid/src/main/java/org/sonar/python/tree/TreeUtils.java
```diff
@@ -0,0 +1,49 @@
...
+        return currentParent;
+      }
+      currentParent = currentParent.parent();
+    }
+    return null;
+  }
+
+  @CheckForNull
+  public static Tree firstAncestorOfKind(Tree tree, Kind... kinds) {
+    return firstAncestor(tree, t -> Arrays.asList(kinds).contains(t.getKind()));
```
benzonico: What about changing Tree#is  to be variadic ? 
`is(Kind... kinds)` and thus simplify this call ?

python-squid/src/main/java/org/sonar/python/api/tree/Tree.java
```diff
@@ -218,14 +217,4 @@
       .filter(Objects::nonNull)
       .flatMap(tree -> Stream.concat(Stream.of(tree), tree.descendants()));
   }
-
-  default List<Tree> ancestors() {
```
benzonico: \o/

python-checks/src/main/java/org/sonar/python/checks/LoopExecutingAtMostOnceCheck.java
```diff
@@ -93,6 +93,11 @@ private static void checkLoop(CfgBranchingBlock loopBlock, SubscriptionContext c
     jumps.forEach(j -> issue.secondary(j, null));
   }
 
+  private static boolean jumpOfCurrentLoop(CfgBlock block, Tree loop) {
+    Tree jumpStatement = block.elements().get(block.elements().size() - 1);
+    return TreeUtils.firstAncestorOfKind(jumpStatement, Kind.WHILE_STMT, Kind.FOR_STMT) == loop;
```
benzonico: I am wondering why we need to get back to syntax level for this method : shouldn't the CFG already answer this question about what loops "holds" the break/continue statement ? in other words : couldn't we check that the loop condition block is somewhat the successor of the block containing this break/continue ? 
It seems to me we are reimplementing some logic already solved by CFG (maybe because we lack an API ? )

(wondering if something like, successors.filter(CFGBranchingBlock).anyMatch(branchingTree == loop) could be enough ? )

python-checks/src/main/java/org/sonar/python/checks/LoopExecutingAtMostOnceCheck.java
```diff
@@ -77,7 +77,7 @@ private static void checkLoop(CfgBranchingBlock loopBlock, SubscriptionContext c
         return;
       }
       if (seen.add(b)) {
-        if (b.syntacticSuccessor() != null) {
+        if (b.syntacticSuccessor() != null && jumpOfCurrentLoop(b, loop)) {
```
benzonico: Because there is absolutely no documentation in the interface I have no idea what is the concept of a "syntacticSuccessor" (versus a normal successor), therefore I don't understand what is this condition doing.

python-checks/src/main/java/org/sonar/python/checks/LoopExecutingAtMostOnceCheck.java
andrea-guarino-sonarsource: The successor of a block having `break` as terminator is going to be the block following the loop (not the loop condition block), so I think it wouldn't be enough.

Inside the CFG builder we know what is going to be the "target" for break, but I think we would need to store extra information to know to which loop it belongs (and consequently having an API to access this) and I'm not sure we should that in the scope of this PR. What do you think?

python-checks/src/main/java/org/sonar/python/checks/LoopExecutingAtMostOnceCheck.java
andrea-guarino-sonarsource: it's the block following this one if no jump is applied. It does apply to block having a jump statement as termination (i.e. break, continue, return, raise), otherwise it's null.
Good point, I will have documentation!

python-squid/src/main/java/org/sonar/python/tree/PyTree.java
```diff
@@ -39,8 +40,8 @@ protected PyTree() {
   }
 
   @Override
-  public boolean is(Kind kind) {
-    return kind == getKind();
+  public boolean is(Kind... kinds) {
+    return Arrays.asList(kinds).contains(getKind());
```
benzonico: this will be heavily called, can we just make a simple loop ? 

```
Kind treeKind = kind();
    for (Kind kindIter : kinds) {
      if (treeKind == kindIter) {
        return true;
      }
    }
    return false;```

PR #401, opened 2019-10-10
SONARPY-453 Remove duplicate representation of docstring

guillaume-dequenne-sonarsource:


python-squid/src/test/java/org/sonar/python/DocstringExtractorTest.java
```diff
@@ -74,19 +74,22 @@ private Token getDocstring(PythonGrammar nodeType, int line) {
 
     @Override
     public void scanFile(PythonVisitorContext visitorContext) {
-      scanNode(visitorContext.rootAstNode());
-
+      scanTree(visitorContext.rootTree());
```
benzonico: I don't think we need a separate method here. This scanTree method can be inlined I believe.

python-squid/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
```diff
@@ -726,61 +727,61 @@ public void nonlocalStatement() {
...
-
-    functionDefTree = parse("def func(x): pass", treeMaker::funcDefStatement);
-    assertThat(functionDefTree.parameters().all()).hasSize(1);
-
-    functionDefTree = parse("async def func(x): pass", treeMaker::funcDefStatement);
-    assertThat(functionDefTree.asyncKeyword().value()).isEqualTo("async");
-
-    functionDefTree = parse("def func(x) -> string : pass", treeMaker::funcDefStatement);
-    TypeAnnotation returnType = functionDefTree.returnTypeAnnotation();
+    FunctionDef functionDef = treeMaker.funcDefStatement(astNode);
```
benzonico: This is great to rename the variable, but, as this is unrelated to the task at hand could you try to keep it in a separate commit next time as it makes things hard to review : this is tough to separate refactoring from required change at review time. (no action required for this comment).

guillaume-dequenne-sonarsource: Indeed! Will make separate commits next time :)

python-squid/src/test/java/org/sonar/python/DocstringExtractorTest.java
guillaume-dequenne-sonarsource: Tbh I'd like to keep it that way because scanTree recursively calls itself. If I inline, the benefit of getting rid of the method would be offset by the fact I'd need to reimplement the recursion through a loop/stack construct.

python-squid/src/main/java/org/sonar/python/DocstringExtractor.java
```diff
@@ -36,48 +37,12 @@
...
-  private static Token extractDocstringFromSimpleStmt(@Nullable AstNode firstSimpleStmt) {
-    if (firstSimpleStmt != null) {
-      Token token = firstSimpleStmt.getToken();
-      if (token.getType().equals(PythonTokenType.STRING)) {
-        return token;
-      }
+  public static StringLiteral extractDocstring(@Nullable StatementList statements) {
+    if (statements != null) {
+      Statement firstStatement = statements.statements().get(0);
+      return firstStatement.is(Tree.Kind.EXPRESSION_STMT) && ((ExpressionStatement) firstStatement).expressions().size() == 1 &&
```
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: Most importantly, I think I was missing a test case for that coverage, which I added. I also refactored the conditional expression as an if statement :)

python-squid/src/test/java/org/sonar/python/DocstringExtractorTest.java
benzonico: missed the recursion, fair enough. But then I wonder why not make this a PythonCheckTree to avoid to reimplement the visit ?

guillaume-dequenne-sonarsource: I reproduced the previous implementation based on AstNode without questioning it... Sure it makes more sense to do it that way now.

PR #403, opened 2019-10-11
Proper "else" clauses

pynicolas:


python-squid/src/main/java/org/sonar/python/api/tree/TreeVisitor.java
```diff
@@ -29,7 +29,7 @@
 
   void visitIfStatement(IfStatement ifStatement);
 
-  void visitElseStatement(ElseStatement elseStatement);
+  void visitElseStatement(ElseClause elseClause);
```
benzonico: shouldn't the method be renamed as well ?

PR #404, opened 2019-10-11
SONARPY-454 Prevent lastToken method from returning statement separators

guillaume-dequenne-sonarsource:


python-squid/src/main/java/org/sonar/python/api/tree/Tree.java
```diff
@@ -31,6 +31,10 @@
 
   Token firstToken();
 
+  /**
+   * @return the last meaningful token of the Tree.
+   * Separators of simple statements (semicolon and/or newline) should not be returned by this method.
```
benzonico: I think we should replace ` should` with `are` as it is the case, anything otherwise is unexpected and a bug.

PR #405, opened 2019-10-11
Cleanup of Tree API

benzonico:


python-squid/src/test/java/org/sonar/python/PythonTestUtils.java
```diff
@@ -47,4 +51,27 @@ public static FileInput parse(String... lines) {
   }
 
 
+  public static <T extends Tree> T getFirstChild(Tree tree, Predicate<Tree> predicate) {
```
andrea-guarino-sonarsource: I believe this method is missing a `@CheckForNull` annotation

PR #406, opened 2019-10-11
SONARPY-448 Try blocks terminating with jump statement should have except and finally clauses as successor

andrea-guarino-sonarsource:


python-checks/src/test/resources/checks/redundantJump.py
```diff
@@ -250,7 +251,7 @@ def redundant_return_inside_else_block():
         print(e)
     else:
         print(42)
-        return # Noncompliant
+        return # FN
```
pynicolas: We should add tests for the cases where this PR removes FPs. I believe that's the same as the cases you added to `ControlFlowGraphTest`.

python-squid/src/test/java/org/sonar/plugins/python/api/cfg/ControlFlowGraphTest.java
```diff
@@ -404,6 +405,168 @@ public void raise_in_try() {
...
+      "while cond(succ = [try_body, after_while]):",
+      "  try:",
+      "    try_body(succ = [else_block, except_cond])",
+      "  except except_cond(succ = [except_block, finally_block]):",
+      "    except_block(succ = [finally_block])",
+      "  else:",
+      "    else_block(succ = [finally_block], syntSucc=finally_block)",
+      "    break",
+      "  finally:",
+      "    finally_block(succ = [after_try, END])",
```
pynicolas: I believe the successors are wrong here because of the `break`. I think it should be: `[after_while, END]`. And so I believe that `after_try` is dead code.

andrea-guarino-sonarsource: As discussed, we may have a path: `try -> except_cond -> except_block -> finally_block -> after_try`.  Successors of `finally_block` are indeed not precise enough, this is a limitation of our current implementation. I will put a comment about it.

python-checks/src/test/resources/checks/loopExecutingAtMostOnce.py
```diff
@@ -79,3 +79,17 @@ def nested_jump_statements_with_else_continuing_outer_loop(items, p):
...
+        try:
+            try:
+                i = i + 1
+                if p:
+                    raise
+            except:
+                continue # OK
+        finally:
+            print("finally")
+        break
```
pynicolas: I don't understand the point of this test. I suggest to rename the function and add a comment on the `break`.

python-checks/src/test/resources/checks/redundantJump.py
```diff
@@ -250,7 +251,51 @@ def redundant_return_inside_else_block():
...
+        return # OK
+    print("after try")
+
+def non_redundant_return_inside_else():
+    try:
+        pass
+    except E as e:
+        print(e)
+    else:
+        return # OK
```
pynicolas: We should add a statement before the `return`, otherwise the exception on single statements in a block applies here.

python-checks/src/test/resources/checks/redundantJump.py
```diff
@@ -250,7 +251,51 @@ def redundant_return_inside_else_block():
...
+        print("after try")
+
+def non_redundant_continue_inside_else(cond):
+    while cond:
+        try:
+            pass
+        except E as e:
+            print(e)
+        else:
+            continue # OK
```
pynicolas: We should add a statement before the `continue`, otherwise the exception on single statements in a block applies here.

python-checks/src/test/resources/checks/redundantJump.py
```diff
@@ -250,7 +251,51 @@ def redundant_return_inside_else_block():
         print(e)
     else:
         print(42)
-        return # Noncompliant
+        return # FN
+
+def non_redundant_return_inside_except():
+    try:
```
pynicolas: I don't see how any of the new test cases could be considered as having a redundant jump. I believe we need `finally` blocks to have `successor == syntacticSuccessor`.

PR #423, opened 2019-10-17
Copy analyzer-commons safe STAX factory to save 1MB in JAR size

pynicolas:


m-g-sonar:
Hey @pynicolas ,

I'm a bit doubtful about the interest of this approach and I would not recommend it. I don't remember exactly what where all the cases we were covering with the `SafetyParser` (think about BOM handling, for instance, and other cases related to which implementations of XML are used beneath ...), but I'm still convinced that if we restart splitting implementations in plugins, "_for the sake of plugins' size_", which is certainly legitimate, then we are going for a much bigger problem.

If the `SafetyParser` is too big to be used in a satisfactory manner, then optimization should be done on its side, so all the plugins can benefit from it.

pynicolas:
@m-g-sonar Thanks for having a look at it! I wrote this PR because I was trying to understand why sonar-python-plugin.jar embeds xerxes classes. I think it shows that we could indeed avoid embedding those classes and save about 1MB for python but also for other plugins. I agree that the improvement should be done on sonar-analyzer-commons. I will open a PR there.

PR #434, opened 2019-10-21
Cleanup Tree API: Use TypeAnnotation in AnnotatedAssignment

andrea-guarino-sonarsource:


python-frontend/src/main/java/org/sonar/python/tree/AnnotatedAssignmentImpl.java
```diff
@@ -31,20 +31,19 @@
...
   private final Expression variable;
-  private final Token colonToken;
-  private final Expression annotation;
   private final Token equalToken;
   private final Expression assignedValue;
   private final Separators separators;
+  private final TypeAnnotation annotation;
 
-  public AnnotatedAssignmentImpl(Expression variable, Token colonToken, Expression annotation,
+  public AnnotatedAssignmentImpl(Expression variable, Token colonToken, Expression typeAnnotation,
```
benzonico: I would rather put the instantiation part in PythonTreeMaker rather than passing two params.

PR #437, opened 2019-10-21
SONARPY-418 handle scope of comprehension

benzonico:


python-frontend/src/test/java/org/sonar/python/semantic/SymbolTableBuilderTest.java
```diff
@@ -258,12 +264,35 @@ public void print_local_var() {
...
       assertThat(symbol.usages()).hasSize(2);
     }
   }
 
+  @Test
+  public void comprehension_scope() {
+    FunctionDef functionTree = functionTreesByName.get("scope_of_comprehension");
+    Map<String, Symbol> symbolByName = getSymbolByName(functionTree);
+    assertThat(symbolByName).containsOnlyKeys("x");
+    assertThat(symbolByName.get("x").usages().stream().map(u -> u.tree().firstToken().line())).hasSize(2).containsExactly(98, 100).doesNotContain(99);
```
andrea-guarino-sonarsource: - I believe that asserting the size is redundant here, as you're already using `containsExactly`.
- Just a minor suggestion: What about using `extracting` like that:
```java
assertThat(symbolByName.get("x").usages()).extracting(u -> u.tree().firstToken().line())
```

python-frontend/src/main/java/org/sonar/plugins/python/api/tree/ComprehensionExpression.java
```diff
@@ -19,10 +19,15 @@
...
+import java.util.Set;
+import org.sonar.python.semantic.Symbol;
+
 public interface ComprehensionExpression extends Expression {
 
   Expression resultExpression();
 
   ComprehensionFor comprehensionFor();
 
+  Set<Symbol> localVariables();
```
andrea-guarino-sonarsource: I think it's worth mentioning in a comment we chose to implement Python 3 scoping rules for Comprehension

PR #440, opened 2019-10-22
SONARPY-463 Add JavaDoc to Tree API

andrea-guarino-sonarsource:


python-frontend/src/main/java/org/sonar/plugins/python/api/tree/IfStatement.java
```diff
@@ -23,7 +23,19 @@
...
+ *
+ * <pre>
+ *   if {@link #condition()}:
+ *     {@link #body()}
+ *   elif {@link #condition()}:
+ *     {@link #body()}
+ *   elif {@link #condition()}:
+ *     {@link #body()}
+ *   {@link #elseBranch()}
+ * </pre>
```
pynicolas: Having `{@link #condition()}` and `{@link #body()}` appearing multiple times here seems misleading to me.
In fact, maybe I don't like our AST model for `elif` clauses.

python-frontend/src/main/java/org/sonar/plugins/python/api/tree/NonlocalStatement.java
```diff
@@ -21,6 +21,11 @@
 
 import java.util.List;
 
+/**
+ * <pre>
+ *   nonlocal {@link Name} , {@link Name}
+ * </pre>
```
pynicolas: Shouldn't we link to methods instead of classes?

python-frontend/src/main/java/org/sonar/plugins/python/api/tree/PrintStatement.java
```diff
@@ -21,6 +21,11 @@
 
 import java.util.List;
 
+/**
+ * <pre>
+ *   print {@link Expression} , {@link Expression}
```
pynicolas: Shouldn't we link to methods instead of classes?

python-frontend/src/main/java/org/sonar/plugins/python/api/tree/RaiseStatement.java
```diff
@@ -22,14 +22,20 @@
 import java.util.List;
 import javax.annotation.CheckForNull;
 
+/**
+ * <pre>
+ *   raise {@link Expression} , {@link Expression} from {@link #fromExpression()}
```
pynicolas: Shouldn't we link to methods instead of classes?

python-frontend/src/main/java/org/sonar/plugins/python/api/tree/ReturnStatement.java
```diff
@@ -21,6 +21,11 @@
 
 import java.util.List;
 
+/**
+ * <pre>
+ *   return {@link Expression} , {@link Expression}
```
pynicolas: Shouldn't we link to methods instead of classes?

python-frontend/src/main/java/org/sonar/plugins/python/api/tree/SetLiteral.java
```diff
@@ -21,6 +21,11 @@
 
 import java.util.List;
 
+/**
+ * <pre>
+ *   { {@link Expression} , {@link Expression} , {@link Expression} }
```
pynicolas: Shouldn't we link to methods instead of classes?

PR #444, opened 2019-10-23
SONARPY-397 Rule S2077: Update rule to raise when formatted queries a…

andrea-guarino-sonarsource:
…re passed via a variable

### DIFF
--- a/python-checks/src/main/java/org/sonar/python/checks/hotspots/SQLQueriesCheck.java
+++ b/python-checks/src/main/java/org/sonar/python/checks/hotspots/SQLQueriesCheck.java
@@ -19,49 +19,39 @@
  */
 package org.sonar.python.checks.hotspots;
 
-import java.util.Collections;
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


python-checks/src/main/java/org/sonar/python/checks/hotspots/SQLQueriesCheck.java
```diff
@@ -103,36 +93,55 @@ private boolean isSQLQueryFromDjangoDBConnection(String functionName) {
...
 
   private static boolean isException(CallExpression callExpression, String functionName) {
     List<Argument> argListNode = callExpression.arguments();
     if (extraContainsFormattedSqlQueries(argListNode, functionName)) {
       return false;
     }
+    return argListNode.isEmpty();
+  }
+
+  private static boolean isSensitive(CallExpression callExpression, List<Tree> secondaryLocations) {
```
pynicolas: I believe that this method has a strange signature:
* It both returns something and has side effects on its parameters
* If I understand correctly, `secondaryLocations` can have at most 1 element, so having a list is misleading and makes this method hard to understand

I would change that to: 
`private static Optional<Tree> sensitiveArgumentValue(CallExpression callExpression) {`

PR #446, opened 2019-10-24
SONARPY-94 RSPEC-1536 function argument names should be unique

benzonico:


python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S1536.json
```diff
@@ -0,0 +1,16 @@
+{
+  "title": "Function argument names should be unique",
```
pynicolas: Why does the title mention "argument"? Shouldn't we fix that as well as the rule description?

benzonico:
See comments on jira ticket.

PR #447, opened 2019-10-24
Use UnpackingExpression in Dictionary Literal and Argument

andrea-guarino-sonarsource:


python-frontend/src/main/java/org/sonar/plugins/python/api/tree/BaseTreeVisitor.java
```diff
@@ -325,8 +325,8 @@ public void visitUnaryExpression(UnaryExpression pyUnaryExpressionTree) {
   }
 
   @Override
-  public void visitStarredExpression(StarredExpression pyStarredExpressionTree) {
-    scan(pyStarredExpressionTree.expression());
+  public void visitStarredExpression(UnpackingExpression unpackingExpressionTree) {
```
benzonico: shouldn't the name of the method be refactored as well ?

PR #449, opened 2019-10-24
SONARPY-462 Rely on check verifier implemented in sonar-analyzer-commons

pynicolas:


python-checks/src/test/resources/checks/unusedLocalVariable.py
```diff
@@ -20,7 +20,7 @@ def string_interpolation():
     value1 = 1
     value2 = 2
     value3 = 3 # Noncompliant
-    value4 = 4 # Noncompliant (previously FN)
+    value4 = 4 # Noncompliant [[previously FN]]
```
benzonico: I think the comment "previously FN" should be completely drop

PR #450, opened 2019-10-25
Rename PythonCheckTree to PythonVisitorCheck

pynicolas:


benzonico:
It lacks an update on the IT regarding custom rules. I expect plugin QA test to fail because of this rename. The doc repo : https://github.com/SonarSource/sonar-custom-rules-examples should also be updated.

pynicolas:
> This PR lacks an update on the IT regarding custom rules. I expect plugin QA test to fail because of this rename. 

And [ITs didn't fail](https://github.com/SonarSource/sonar-python/pull/450/checks?check_run_id=274810228)! It seems the maven command doesn't trigger the execution of plugin integration tests: @tomverin any idea?

PR #459, opened 2019-10-29
SONARPY-202 Rule: All branches in a conditional structure should not have exactly the same implementation

guillaume-dequenne-sonarsource:


python-checks/src/test/resources/checks/sameBranch.py
```diff
@@ -36,7 +36,7 @@
     if True:
         print(1)
     else:
-        print(1)  # Noncompliant {{Either merge this branch with the identical one on line "37" or change one of the implementations.}}
+        print(1)
```
andrea-guarino-sonarsource: Could you add a comment this will be raised by S3923?

PR #461, opened 2019-10-29
SONARPY-471 Rule S2190: Recursion should not be infinite

alban-auzeill:


python-checks/src/main/java/org/sonar/python/checks/InfiniteRecursionCheck.java
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    @Override
+    public void visitLambda(LambdaExpression pyLambdaExpressionTree) {
+      // ignore
+    }
+
+    @Override
+    public void visitForStatement(ForStatement pyForStatementTree) {
+      // ignore, unexpected "for" element in cfg blocks
```
andrea-guarino-sonarsource: It's not very clear to me why we're excluding for statements. Is it for a limitation in the CFG? In that case, please create a ticket for it and document in the unit tests

python-checks/src/test/resources/checks/infiniteRecursion.py
```diff
@@ -0,0 +1,263 @@
...
+#   ^^^^^
+    if x:
+        fun14(x)
+#       ^^^^^<
+    else:
+        fun14(x)
+#       ^^^^^<
+
+def fun15(x):
+    if x == 10 or fun15(x + 1):
```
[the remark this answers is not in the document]
alban-auzeill: Nop, there's no FN, visitBinaryExpression always takes into account `leftOperand()` recursion, even for `or` and `and`.
I added the following test:
```
def fun21(x): # Noncompliant
    if fun21(x + 1) or x == 10:
        print(2)
```

python-checks/src/main/java/org/sonar/python/checks/InfiniteRecursionCheck.java
alban-auzeill: Now, I think it's useless. It was probably useful before I exclude `lambda` or `sub-function`, but it was not the right workaround. So I removed this `visitForStatement` and added more test related to `for`.

PR #463, opened 2019-10-31
SONARPY-221 Rule S4144 Method should not have identical implementations

guillaume-dequenne-sonarsource:


python-checks/src/test/resources/checks/duplicatedMethodImplementationCheck.py
```diff
@@ -0,0 +1,160 @@
...
+
+  def not_implemented_duplicate(self):
+    '''
+    This is a docstring
+    '''
+    raise NotImplementedError(
+      "Some message"
+    )
+
+#coverage
```
pynicolas: What's the meaning of this comment? If the class which is defined below is there to test something specific (one line methods, maybe), let's mention it. It's also possible to drop the comment and rename the class itself, for example to "OneLineMethods".

python-checks/src/main/java/org/sonar/python/checks/DuplicatedMethodImplementationCheck.java
```diff
@@ -0,0 +1,103 @@
...
+    if (isException(suspiciousMethod)) {
+      return;
+    }
+    for (int j = 0; j < index; j++) {
+      FunctionDef originalMethod = methods.get(j);
+      Tree originalBody = originalMethod.body();
+      if (CheckUtils.areEquivalent(originalBody, suspiciousBody)) {
+        int line = originalMethod.name().firstToken().line();
+        String message = String.format(MESSAGE, originalMethod.name().name(), line);
+        ctx.addIssue(suspiciousMethod.name(), message).secondary(IssueLocation.preciseLocation(originalMethod.name(), "Original"));
```
pynicolas: * The secondary location should be tested.
* You may use `.secondary(originalMethod.name(), "Original")`

python-checks/src/main/java/org/sonar/python/checks/DuplicatedMethodImplementationCheck.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  private static boolean isException(FunctionDef suspiciousMethod) {
+    boolean hasDocString = suspiciousMethod.docstring() != null;
+    StatementList suspiciousBody = suspiciousMethod.body();
+    int nbActualStatements = hasDocString ? suspiciousBody.statements().size() - 1 : suspiciousBody.statements().size();
+    if (nbActualStatements == 0 || isOnASingleLine(suspiciousBody, hasDocString)) {
+      return true;
+    }
+    return nbActualStatements == 1 && suspiciousBody.statements().get(suspiciousBody.statements().size() - 1).is(Tree.Kind.RAISE_STMT);
```
pynicolas: Let's create a variable for `suspiciousBody.statements()` instead of duplicating the expression 4 times.

python-checks/src/test/resources/checks/duplicatedMethodImplementationCheck.py
```diff
@@ -0,0 +1,160 @@
...
+class clazz:
+  def method(self):
+    bar()
+    foo()
+
+  def method2(self):
+    foo()
+    bar()
+
+class clazz:
```
pynicolas: What's the intention of this test?

guillaume-dequenne-sonarsource: It's for coverage completeness on the isException method. I admit it's not the most valuable out there, though...

PR #466, opened 2019-11-01
SONARPY-275 Rule S2737: Catch clauses should do more than rethrow

guillaume-dequenne-sonarsource:


its/ruling/src/test/resources/expected/python-S2737.json
```diff
@@ -0,0 +1,47 @@
...
+132,
+],
+'project:buildbot-0.8.6p1/contrib/bzr_buildbot.py':[
+264,
+],
+'project:buildbot-0.8.6p1/contrib/coverage2text.py':[
+84,
+],
+'project:django-2.2.3/django/contrib/admin/views/main.py':[
+422,
```
pynicolas: To me, this looks like a FP which is rather well explained by the comments. I believe that we should exclude `except` clauses which are followed by other clauses which may catch the same exception.

python-frontend/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
```diff
@@ -1215,6 +1215,10 @@ public void try_statement() {
     assertThat(tryStatement.elseClause().body().statements()).hasSize(1);
     assertThat(tryStatement.children()).hasSize(5);
 
+    astNode = p.parse("try: pass\nexcept (RuntimeError, TypeError, NameError): pass\nelse: pass");
+    tryStatement = treeMaker.tryStatement(astNode);
+    assertThat(tryStatement.tryKeyword().value()).isEqualTo("try");
```
pynicolas: What's the point of that test? Should we add more assertions to it?

guillaume-dequenne-sonarsource: This shouldn't have been committed, my mistake!

PR #467, opened 2019-11-01
SONARPY-202 Rule S3923 Add secondary locations

guillaume-dequenne-sonarsource:


python-frontend/src/test/java/org/sonar/python/tree/TreeUtilsTest.java
```diff
@@ -90,6 +92,21 @@ public void tokens() {
 
   }
 
+  @Test
+  public void non_whitespace_tokens() {
+    FileInput parsed = parse("if foo:\n  pass");
+    IfStatement ifStmt = (IfStatement) parsed.statements().statements().get(0);
+    List<Token> collect = new ArrayList<>(ifStmt.children().stream().map(t -> t.is(Kind.TOKEN) ? (Token) t : t.firstToken()).collect(Collectors.toList()));
+    collect.add(parsed.lastToken());
+    assertThat(TreeUtils.tokens(parsed)).containsExactly(collect.toArray(new Token[0]));
```
pynicolas: I don't see the point of this assertion: how does it help testing `nonWhitespaceTokens`?

python-checks/src/main/java/org/sonar/python/checks/AllBranchesAreIdenticalCheck.java
```diff
@@ -53,7 +57,15 @@ private static void handleIfStatement(IfStatement ifStmt, SubscriptionContext ct
     if (!CheckUtils.areEquivalent(body, ifStmt.elseBranch().body())) {
       return;
     }
-    ctx.addIssue(ifStmt.keyword(), "Remove this if statement or edit its code blocks so that they're not all the same.");
+    PreciseIssue issue = ctx.addIssue(ifStmt.keyword(), "Remove this if statement or edit its code blocks so that they're not all the same.");
+    List<Token> ifBodyTokens = TreeUtils.nonWhitespaceTokens(ifStmt.body());
+    issue.secondary(IssueLocation.preciseLocation(ifBodyTokens.get(0), ifBodyTokens.get(ifBodyTokens.size() - 1), null));
+    ifStmt.elifBranches().forEach(e -> {
+      List<Token> elifBodyTokens = TreeUtils.nonWhitespaceTokens(e.body());
+      issue.secondary(IssueLocation.preciseLocation(elifBodyTokens.get(0), elifBodyTokens.get(ifBodyTokens.size() - 1), null));
```
pynicolas: `elifBodyTokens.get(ifBodyTokens.size() - 1)` is wrong here, it should be `elifBodyTokens.get(elifBodyTokens.size() - 1)`

Of course, nothing fails because all branches have the same number of tokens as they are identical...

Suggestion: extract a method with a signature such as: `IssueLocation issueLocation(StatementList body)`

PR #469, opened 2019-11-01
SONARPY-51 S1854: handle exceptions

andrea-guarino-sonarsource:


alban-auzeill:
You need to update the html description from:
```
-1, 0, 1, null, true, false and ""
```
to
```
-1, 0, 1, None, True, False and ""
```
And add a unit test about empty string `""`

PR #471, opened 2019-11-05
SONARPY-249 Rule S4143: Collection content should not be replaced unconditionally

pynicolas:


python-checks/src/main/java/org/sonar/python/checks/OverwrittenCollectionEntryCheck.java
```diff
@@ -0,0 +1,196 @@
...
+    } else if (tree.is(Kind.STRING_LITERAL)) {
+      return ((StringLiteral) tree).trimmedQuotesValue();
+    } else if (tree.is(Kind.NAME)) {
+      return ((Name) tree).name();
+    } else if (tree.is(Kind.SLICE_ITEM)) {
+      SliceItem sliceItem = (SliceItem) tree;
+      List<String> keyParts = Stream.of(sliceItem.lowerBound(), sliceItem.upperBound(), sliceItem.stride())
+        .map(e -> e == null ? "" : key(e))
+        .collect(Collectors.toList());
+      return keyParts.contains(null) ? null : String.join(":", keyParts);
```
guillaume-dequenne-sonarsource: Can `keyParts` actually contain null here?
From the code it doesn't look so and there are no test case fulfilling this condition.

PR #472, opened 2019-11-05
SONARPY-268 Rule S1186: Methods should not be empty

guillaume-dequenne-sonarsource:


python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S1186.html
```diff
@@ -0,0 +1,80 @@
...
+
+    def mymethod2(self):
+        """
+        this is a docstring
+        """
+        raise NotImplementedError()
+</pre>
+<h2>Exceptions</h2>
+<p>No issue will be raised when the empty method is abstract and meant to be overriden in a subclass, i.e. it is annotated with
+<code>abc.abstractmethod</code>, <code>abc.abstractstaticmethod</code>, <code>abc.abstractclassmethod</code> or <code>abc.abstractproperty</code></p>
```
pynicolas: It seems that the rule implementation only checks "abstractmethod".

python-checks/src/test/resources/checks/emptyFunction.py
```diff
@@ -0,0 +1,64 @@
...
+bar()
+def empty_function():
+  pass
+
+def non_empty_function():
+  foo()
+
+# Noncompliant@+2
+#FP: Comment is associated with previous dedent
+def fp():
```
pynicolas: I think it would be worth removing this FP.

python-checks/src/test/resources/checks/emptyFunction.py
```diff
@@ -0,0 +1,64 @@
...
+
+  # Noncompliant@+2 {{Add a nested comment explaining why this method is empty, or complete the implementation.}}
+  bar()
+  def emtpy_method():
+    pass
+
+  def empty_method_with_docstring():
+    '''
+    Some docstring
+    '''
```
pynicolas: According to the Noncompliant examples in the rule description, we should raise an issue here. I tend to prevent the behavior you implemented though.

PR #474, opened 2019-11-05
SONARPY-269 Update basic exclusion pattern with "-", exclude formatting patterns

pynicolas:


guillaume-dequenne-sonarsource:
LGTM

PR #480, opened 2019-11-06
SONARPY-51 Fix various FP on DeadStore

andrea-guarino-sonarsource:


python-checks/src/test/resources/checks/deadStore.py
```diff
@@ -220,4 +220,46 @@ def falsy_or_true_initialization():
...
+    with open2() as x:
+        pass
+    print(x)
 
+def used_in_subfunction():
+    var = 42
+    print(var)
+    def inner(): print(var)
+    var = 43 # OK
+    inner()
```
guillaume-dequenne-sonarsource: As discussed, we should document the fact that we might have a FN if there is a dead store after the inner function call

PR #483, opened 2019-11-07
SONARPY-485 Rule S5603: Unused scope-limited definitions should be re…

guillaume-dequenne-sonarsource:
…moved


python-checks/src/main/java/org/sonar/python/checks/UnusedNestedDefinitionCheck.java
```diff
@@ -0,0 +1,75 @@
...
+    context.registerSyntaxNodeConsumer(Tree.Kind.CLASSDEF, ctx -> {
+      ClassDef classDef = (ClassDef) ctx.syntaxNode();
+      if (!classDef.decorators().isEmpty()) {
+        return;
+      }
+      checkNestedDefinition(classDef, classDef.name().name(), Usage.Kind.CLASS_DECLARATION, ctx);
+    });
+  }
+
+  private static void checkNestedDefinition(Tree tree, String name, Usage.Kind kind, SubscriptionContext ctx) {
```
pynicolas: Why do we need the `kind` parameter? Can't we derive its value from `tree.getKind()`?

python-checks/src/main/java/org/sonar/python/checks/UnusedNestedDefinitionCheck.java
```diff
@@ -0,0 +1,75 @@
...
+    Tree parent = tree.parent();
+    while (parent != null && !parent.is(Tree.Kind.CLASSDEF) && !parent.is(Tree.Kind.FUNCDEF)) {
+      parent = parent.parent();
+    }
```
pynicolas: Why not replacing that with the following?
```
    Tree parent = TreeUtils.firstAncestorOfKind(tree, Tree.Kind.CLASSDEF, Tree.Kind.FUNCDEF);
```

python-checks/src/main/java/org/sonar/python/checks/UnusedNestedDefinitionCheck.java
```diff
@@ -0,0 +1,75 @@
...
+    if (parent == null || !parent.is(Tree.Kind.FUNCDEF)) {
+      return;
+    }
+    FunctionDef parentFunction = (FunctionDef) parent;
+    parentFunction.localVariables().stream().filter(s -> s.name().equals(name))
+      .filter(s -> s.usages().stream().anyMatch(u -> u.kind().equals(kind)))
+      .findFirst().ifPresent(s -> checkSymbolUsages(s, ctx));
```
pynicolas: Can't we use the symbol from the `name()` of the funcdef/classdef?

python-checks/src/test/resources/checks/unusedNestedDefinition.py
```diff
@@ -0,0 +1,63 @@
...
+  def using():  # Ok
+    nested_function()
+
+  @d.callback
+  class NestedClass:  # Ok
+      def __init__(self):
+          print("NestedClass")
+
+  def nested_function():  # Ok
+    print("nested_function")
```
pynicolas: What do we test here?

PR #484, opened 2019-11-07
SONARPY-203 Rule S1226: Method parameters initial values should not be ignored

andrea-guarino-sonarsource:


python-checks/src/test/resources/checks/ignoredParameter.py
```diff
@@ -0,0 +1,29 @@
+def ignored_param(p): # Noncompliant
+    p = 42
+    print(p)
+
```
alban-auzeill: we should show explicitly that we support this case:
```py
def param_can_be_falsy(p): # Compliant
    p = p or 42
    print(p)
```

python-frontend/src/main/java/org/sonar/python/cfg/ControlFlowGraphBuilder.java
```diff
@@ -67,13 +70,41 @@ public ControlFlowGraphBuilder(@Nullable StatementList statementList) {
     exitTargets.push(end);
     if (statementList != null) {
       start = build(statementList.statements(), createSimpleBlock(end));
+      addParametersToStartBlock(statementList);
```
alban-auzeill: Good idea :bulb:. We should merge this PR before SONARPY-481, so I can change the workaround in my PR to look for the parameters once I reached the `start` block.

python-frontend/src/main/java/org/sonar/python/cfg/ControlFlowGraphBuilder.java
```diff
@@ -67,13 +70,41 @@ public ControlFlowGraphBuilder(@Nullable StatementList statementList) {
...
     }
     removeEmptyBlocks();
     computePredecessors();
   }
 
+  private void addParametersToStartBlock(StatementList statementList) {
+    if (statementList.parent().is(Tree.Kind.FUNCDEF)) {
+      ParameterList parameterList = ((FunctionDef) statementList.parent()).parameters();
+      if (parameterList != null) {
+        PythonCfgSimpleBlock parametersBlock = createSimpleBlock(start);
```
alban-auzeill: We have a duplication about the logic PARAMETER/TUPLE_PARAMETER. We could simplify with:
```java
        PythonCfgSimpleBlock parametersBlock = createSimpleBlock(start);
        addParameters(parameterList.all(), parametersBlock);
        start = parametersBlock;
```
...
```java
  private static void addParameters(List<AnyParameter> parameters, PythonCfgSimpleBlock parametersBlock) {
    for (AnyParameter parameter : parameters) {
      if (parameter.is(Tree.Kind.TUPLE_PARAMETER)) {
        addParameters(((TupleParameter) parameter).parameters(), parametersBlock);
      } else {
        parametersBlock.addElement(parameter);
      }
    }
  }
```

python-checks/src/main/java/org/sonar/python/checks/IgnoredParameterCheck.java
```diff
@@ -0,0 +1,86 @@
...
+      cfg.blocks().forEach(block -> {
+        List<DeadStoreUtils.UnnecessaryAssignment> unnecessaryAssignments =
+          DeadStoreUtils.findUnnecessaryAssignments(block, lva.getLiveVariables(block), functionDef);
+        unnecessaryAssignments.stream()
+          // symbols should have at least two binding usages
+          .filter(assignment -> assignment.symbol.usages().stream().filter(Usage::isBindingUsage).count() > 1)
+          // no usages in unreachable blocks
+          .filter(assignment -> unreachableBlocks.stream().noneMatch(unreachableBlock ->
+            lva.getLiveVariables(unreachableBlock).usedSymbols().contains(assignment.symbol)))
+          .filter((assignment -> isParameter(assignment.element)))
```
alban-auzeill: We should not filter `isParameter` at the end but as soon as possible.

python-checks/src/main/java/org/sonar/python/checks/IgnoredParameterCheck.java
```diff
@@ -0,0 +1,86 @@
...
+      Set<CfgBlock> unreachableBlocks = unreachableBlocks(cfg);
+      cfg.blocks().forEach(block -> {
+        List<DeadStoreUtils.UnnecessaryAssignment> unnecessaryAssignments =
+          DeadStoreUtils.findUnnecessaryAssignments(block, lva.getLiveVariables(block), functionDef);
+        unnecessaryAssignments.stream()
+          // symbols should have at least two binding usages
+          .filter(assignment -> assignment.symbol.usages().stream().filter(Usage::isBindingUsage).count() > 1)
+          // no usages in unreachable blocks
+          .filter(assignment -> unreachableBlocks.stream().noneMatch(unreachableBlock ->
+            lva.getLiveVariables(unreachableBlock).usedSymbols().contains(assignment.symbol)))
```
alban-auzeill: Collecting all `.usedSymbols()` just to test is a given symbol is present is good in winter (:fire:) but not the optimal solution.

python-checks/src/main/java/org/sonar/python/checks/IgnoredParameterCheck.java
```diff
@@ -0,0 +1,86 @@
...
+      LiveVariablesAnalysis lva = LiveVariablesAnalysis.analyze(cfg);
+      Set<CfgBlock> unreachableBlocks = unreachableBlocks(cfg);
+      cfg.blocks().forEach(block -> {
+        List<DeadStoreUtils.UnnecessaryAssignment> unnecessaryAssignments =
+          DeadStoreUtils.findUnnecessaryAssignments(block, lva.getLiveVariables(block), functionDef);
+        unnecessaryAssignments.stream()
+          // symbols should have at least two binding usages
+          .filter(assignment -> assignment.symbol.usages().stream().filter(Usage::isBindingUsage).count() > 1)
+          // no usages in unreachable blocks
+          .filter(assignment -> unreachableBlocks.stream().noneMatch(unreachableBlock ->
```
alban-auzeill: :flushed: this filtered line is hard to understand and deserve to be extracted into a method with a name that could help.

PR #489, opened 2019-11-08
SONARPY-269 unescape StringLiteral content

alban-auzeill:


python-checks/src/main/java/org/sonar/python/checks/Expressions.java
```diff
@@ -103,4 +108,138 @@ public static Expression removeParentheses(@Nullable Expression expression) {
...
+    List<StringElement> elements = stringLiteral.stringElements();
+    if (elements.size() == 1) {
+      return unescape(elements.get(0));
+    }
+    return elements.stream()
+      .map(Expressions::unescape)
+      .collect(Collectors.joining());
+  }
+
+  public static String unescape(StringElement stringElement) {
```
pynicolas: It would be good to add a http link to the language reference.

python-checks/src/test/java/org/sonar/python/checks/ExpressionsTest.java
```diff
@@ -83,6 +88,85 @@ public void remove_parentheses() {
...
 
+  @Test
+  public void unescape_string_literal() {
+    assertThat(unescape(stringLiteral("''"))).isEqualTo("");
+    assertThat(unescape(stringLiteral("'' ''"))).isEqualTo("");
+    assertThat(unescape(stringLiteral("'abc' 'def'"))).isEqualTo("abcdef");
+  }
+
+  @Test
+  public void unescape_string_element() {
```
pynicolas: This test is very long. Can we split it with meaningful names for the new tests?

python-checks/src/test/java/org/sonar/python/checks/ExpressionsTest.java
```diff
@@ -83,6 +88,85 @@ public void remove_parentheses() {
...
+    assertThat(unescape(stringElement("'\\o'"))).isEqualTo("\\o");
+    assertThat(unescape(stringElement("'\\z'"))).isEqualTo("\\z");
+    assertThat(unescape(stringElement("'\\X00'"))).isEqualTo("\\X00");
+    assertThat(unescape(stringElement("'\\x0'"))).isEqualTo("\\x0");
+    assertThat(unescape(stringElement("'\\xZZ\\x04'"))).isEqualTo("\\xZZ\u0004");
+    assertThat(unescape(stringElement("'\\u000'"))).isEqualTo("\\u000");
+    assertThat(unescape(stringElement("'\\U0000000'"))).isEqualTo("\\U0000000");
+
+    // Python error: f-string expression part cannot include a backslash
+    assertThat(unescape(stringElement("f'name:\\n{na\\\nme}'"))).isEqualTo("name:\n{name}");
```
pynicolas: I don't see a test for the various prefixes, e.g. `r'...'`, `rb'...'`, `b'...'`...

python-checks/src/main/java/org/sonar/python/checks/Expressions.java
```diff
@@ -103,4 +108,138 @@ public static Expression removeParentheses(@Nullable Expression expression) {
...
+
+    private final int escapedLength;
+    private final String unescapedValue;
+
+    private EscapeSequence(int escapedLength, String unescapedValue) {
+      this.escapedLength = escapedLength;
+      this.unescapedValue = unescapedValue;
+    }
+
+    private static EscapeSequence extract(String value, int i, boolean support16Bits) {
```
pynicolas: This `support16Bits` parameter is not clear to me.

python-checks/src/main/java/org/sonar/python/checks/Expressions.java
```diff
@@ -103,4 +108,138 @@ public static Expression removeParentheses(@Nullable Expression expression) {
...
+      // octal
+      int octal = 0;
+      int octalStart = (value.charAt(i + 1) == 'o') ? (i + 2) : (i + 1);
+      int len = 0;
+      int j = octalStart;
+      while (len < 3 && j < value.length() && value.charAt(j) >= '0' && value.charAt(j) <= '7') {
+        octal = octal * 8 + (value.charAt(j) - '0');
+        j++;
+        len++;
+      }
```
pynicolas: Why not using `Integer.parseInt` the same way you did it for `extractHexadecimal`? Or maybe even sharing code for the parsing of octal and hexadecimal...

alban-auzeill: - for hexadecimal, we know the length in advance (x:2 u:4 U:8)
- for octal it's a mess, 'o' prefix is not mandatory, the length can be from 1 to 3, we stop at 3 or when the next character does not match 0-7

PR #504, opened 2019-11-20
SONARPY-512 Fully qualified name for Function and Class definition symbols

andrea-guarino-sonarsource:


sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonSensor.java
```diff
@@ -72,9 +78,31 @@ public void execute(SensorContext context) {
...
+  }
+  // visible for testing
+  static Map<URI, String> getInputFileByPackage(File projectBaseDir, List<InputFile> inputFiles) {
+    Map<URI, String> packageNamePerFile = new HashMap<>();
+    for (InputFile inputFile : inputFiles) {
+      String packageName = findInitFiles(inputFile, projectBaseDir);
+      packageNamePerFile.put(inputFile.uri(), packageName);
+    }
+    return packageNamePerFile;
+  }
```
pynicolas: Do we really need this method? Do we need to build a map for all InputFiles? What about moving `findInitFiles` to `PythonScanner` and call it inside `scanFile`?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonSensor.java
```diff
@@ -72,9 +78,31 @@ public void execute(SensorContext context) {
...
+
+  private static String findInitFiles(InputFile inputFile, File projectBaseDir) {
```
pynicolas: What about renaming this method to something like `pythonPackageName`?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonSensor.java
```diff
@@ -72,9 +78,31 @@ public void execute(SensorContext context) {
...
+    File currentDirectory = inputFile.file().getParentFile();
+    Deque<String> packages = new ArrayDeque<>();
+    while(!currentDirectory.getAbsolutePath().equals(projectBaseDir.getAbsolutePath())) {
+      File initFile = new File(currentDirectory, "__init__.py");
+      if (initFile.exists()) {
+        packages.push(currentDirectory.getName());
+      }
```
pynicolas: Shouldn't we `break` if `!initFile.exists()`?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonSensor.java
andrea-guarino-sonarsource: Good point, I will do that

PR #511, opened 2019-12-06
SONARPY-529 Fully qualified name of identifiers coming from relative …

andrea-guarino-sonarsource:
…imports should be resolved

### DIFF
--- a/python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
+++ b/python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
@@ -20,6 +20,7 @@
 package org.sonar.python.semantic;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Deque;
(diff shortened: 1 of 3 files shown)


python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -232,18 +236,18 @@ public void visitImportFrom(ImportFrom pyImportFromTree) {
       String moduleName = moduleTree != null
         ? moduleTree.names().stream().map(Name::name).collect(Collectors.joining("."))
         : null;
-      createImportedNames(pyImportFromTree.importedNames(), moduleName, !pyImportFromTree.dottedPrefixForModule().isEmpty());
+      createImportedNames(pyImportFromTree.importedNames(), moduleName, pyImportFromTree.dottedPrefixForModule());
```
pynicolas: The semantic of `dottedPrefixForModule` is confusing and it's not documented (but that's not a new issue).

PR #513, opened 2019-12-12
Update rule metadata to prepare v2.3

andrea-guarino-sonarsource:


python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S5332.html
```diff
@@ -1,37 +1,3 @@
...
-<h2>Recommended Secure Coding Practices</h2>
-<ul>
-  <li> Use <code>ssh</code> as an alternative to <code>telnet</code> </li>
-  <li> Use <code>sftp</code>, <code>scp</code> or <code>ftps</code> instead of <code>ftp</code> </li>
-  <li> Use <code>https</code> instead of <code>http</code> </li>
-</ul>
-<p>It is recommended to secure all transport channels (event local network) as it can take a single non secure connection to compromise an entire
-application or system. </p>
-<h2>Exceptions</h2>
-<p>Exception: the url domain component is a loopback address.</p>
```
pynicolas: I'm not sure that the removal of this whole block is correct.

PR #516, opened 2019-12-18
SONARPY-503 S3827 to raise only one issue per variable name per function

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/UndeclaredNameUsageCheck.java
```diff
@@ -74,21 +76,27 @@ public void initialize(Context context) {
...
         if (symbolUsage.isWrite()) {
           currentState.put(symbol, DefinedVariablesAnalysis.VariableDefinition.DEFINED);
         }
         DefinedVariablesAnalysis.VariableDefinition varDef = currentState.getOrDefault(symbol, DefinedVariablesAnalysis.VariableDefinition.DEFINED);
-        if (symbolUsage.isRead() && isUndefined(varDef) && !isSymbolUsedInUnreachableBlocks(analysis, unreachableBlocks, symbol) && !isParameter(element)) {
-          ctx.addIssue(element, symbol.name() + " is used before it is defined. Move the definition before.");
+        if (symbolUsage.isRead() && isUndefined(varDef) && !isSymbolUsedInUnreachableBlocks(analysis, unreachableBlocks, symbol) && !isParameter(element)
+          && !ignoredSymbols.contains(symbol)) {
+          ignoredSymbols.add(symbol);
+          Optional<Usage> suspectUsage = symbol.usages().stream().filter(u -> TreeUtils.hasDescendant(element, t -> t.equals(u.tree()))).findFirst();
```
pynicolas: This filtering is quite suspicious to me. @andrea-guarino-sonarsource can you please have  a look?

guillaume-dequenne-sonarsource: From what I understand, such filtering is required because the `UsageVisitor` associates a `Tree` element with a map of its `Symbol` and `SymbolUsage`. However, we are missing the precise link with the actual `Usage` within the `Tree` (as it is most likely a descendant of the tree rather than the tree itself).

I feels to me that in order to avoid this kind of filtering, I would therefore need to refactor the `UsageVisitor`, which is maybe a bit overkill if done just for this case (but maybe it would be helpful in other cases too?).

python-checks/src/main/java/org/sonar/python/checks/UndeclaredNameUsageCheck.java
```diff
@@ -120,12 +116,20 @@ private void addNameIssues(Map<String, List<Name>> nameIssues) {
...
-  private static class ExceptionVisitor extends BaseTreeVisitor {
+  private static class UnresolvedSymbolsVisitor extends BaseTreeVisitor {
+
     private boolean hasWildcardImport = false;
     private boolean callGlobalsOrLocals = false;
+    private Map<String, List<Name>> nameIssues = new HashMap<>();
+
+    @Override
+    public void visitName(Name name) {
+      if (!callGlobalsOrLocals && !hasWildcardImport && name.isVariable() && name.symbol() == null) {
```
pynicolas: I think you can now drop `!callGlobalsOrLocals && !hasWildcardImport` on this line.

python-checks/src/main/java/org/sonar/python/checks/UndeclaredNameUsageCheck.java
```diff
@@ -46,22 +46,18 @@
...
       FileInput fileInput = (FileInput) ctx.syntaxNode();
-      ExceptionVisitor exceptionVisitor = new ExceptionVisitor();
-      fileInput.accept(exceptionVisitor);
-      hasWildcardImport = exceptionVisitor.hasWildcardImport;
-      callGlobalsOrLocals = exceptionVisitor.callGlobalsOrLocals;
       UnresolvedSymbolsVisitor unresolvedSymbolsVisitor = new UnresolvedSymbolsVisitor();
       fileInput.accept(unresolvedSymbolsVisitor);
-      addNameIssues(unresolvedSymbolsVisitor.nameIssues);
+      if(!unresolvedSymbolsVisitor.callGlobalsOrLocals && !unresolvedSymbolsVisitor.hasWildcardImport) {
+        addNameIssues(unresolvedSymbolsVisitor.nameIssues);
```
pynicolas: I think you can pass `ctx` as an argument to `addNameIssues` and drop the field named `subscriptionContext`.

python-checks/src/main/java/org/sonar/python/checks/UndeclaredNameUsageCheck.java
andrea-guarino-sonarsource: I agree it's not easy to understand why we need this filtering. As discussed, I'd suggest adding a property on `SymbolUsage` to get the corresponding usages.

python-frontend/src/main/java/org/sonar/python/cfg/fixpoint/UsageVisitor.java
```diff
@@ -78,5 +82,14 @@ public boolean isWrite() {
     public boolean isRead() {
       return isRead;
     }
+
+    public void addUsage(Usage usage) {
+      usages.add(usage);
+    }
+
+    // Returns the list of symbol usages within the visited element
+    public List<Usage> usages() {
```
pynicolas: Why a list if we only care about the first element? Can the list be empty?

python-frontend/src/main/java/org/sonar/python/cfg/fixpoint/UsageVisitor.java
```diff
@@ -70,6 +73,7 @@ private void addToSymbolToUsageMap(Usage usage, Symbol symbol) {
   public static final class SymbolUsage {
```
pynicolas: @andrea-guarino-sonarsource This name is confusing to me. What's the difference between `Usage` and `SymbolUsage`? What about renaming it to something like `SymbolReadWrite`? we could also think about renaming `UsageVisitor`, `getVariableUsages`...

andrea-guarino-sonarsource: Yes, good point. I agree with renaming it to `SymbolReadWrite` to avoid confusion.

python-frontend/src/main/java/org/sonar/python/cfg/fixpoint/UsageVisitor.java
guillaume-dequenne-sonarsource: Indeed this use case only required the first element but I felt returning the list might be useful for other cases, and more understandable as well, as discussed with @andrea-guarino-sonarsource 
The list should never be empty.
Do you think I should keep it to the first element only?

pynicolas: I don't know enough about this class to have an opinion and I don't think it deserves more discussions.

PR #517, opened 2020-01-08
SONARPY-535 FN on S3827: Undefined symbols when 'import *' is used

andrea-guarino-sonarsource:
…symbols per module

### DIFF
--- a/its/ruling/src/test/resources/expected/python-S3827.json
+++ b/its/ruling/src/test/resources/expected/python-S3827.json
@@ -12,6 +12,11 @@
 'project:numpy-1.16.4/numpy/linalg/lapack_lite/clapack_scrub.py':[
 306,
 ],
+'project:numpy-1.16.4/numpy/matlib.py':[
+49,
+171,
+352,
+],
 'project:twisted-12.1.0/doc/core/howto/listings/pb/trap_client.py':[
 21,
 21,
--- a/python-checks-testkit/src/main/java/org/sonar/python/checks/utils/PythonCheckVerifier.java
+++ b/python-checks-testkit/src/main/java/org/sonar/python/checks/utils/PythonCheckVerifier.java
@@ -20,12 +20,15 @@
 package org.sonar.python.checks.utils;
 
 import com.google.common.base.Preconditions;
-import com.sonarsource.checks.verifier.SingleFileVerifier;
+import com.sonarsource.checks.verifier.MultiFileVerifier;
 import java.io.File;
+import java.nio.file.Path;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 import org.sonar.plugins.python.api.IssueLocation;
 import org.sonar.plugins.python.api.PythonCheck;
 import org.sonar.plugins.python.api.PythonCheck.PreciseIssue;
@@ -39,6 +42,7 @@ import org.sonar.python.TestPythonVisitorRunner;
 import org.sonar.python.tree.TreeUtils;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.sonar.python.semantic.SymbolUtils.pythonPackageName;
 
 public class PythonCheckVerifier {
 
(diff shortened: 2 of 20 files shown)


its/ruling/src/test/resources/expected/python-S3827.json
```diff
@@ -12,6 +12,11 @@
 'project:numpy-1.16.4/numpy/linalg/lapack_lite/clapack_scrub.py':[
 306,
 ],
+'project:numpy-1.16.4/numpy/matlib.py':[
```
andrea-guarino-sonarsource: those are FPs because the imported module (i.e. `numpy`) does some additional manipulation with `__all__` property.
I'm willing to accept this FP, knowing that it will be treated in [SONARPY-547](https://jira.sonarsource.com/browse/SONARPY-547)

python-checks/src/main/java/org/sonar/python/checks/UndeclaredNameUsageCheck.java
```diff
@@ -128,7 +128,7 @@ public void visitName(Name name) {
 
     @Override
     public void visitImportFrom(ImportFrom importFrom) {
-      hasWildcardImport |= importFrom.isWildcardImport();
+      hasUnresolvedWildcardImport |= (importFrom.isWildcardImport() && importFrom.hasUnresolvedWildcardImport());
```
guillaume-dequenne-sonarsource: Is `importFrom.isWildcardImport()` actually still needed here?

python-checks-testkit/src/main/java/org/sonar/python/checks/utils/PythonCheckVerifier.java
```diff
@@ -56,18 +59,25 @@ public static void verify(String path, PythonCheck check) {
     createVerifier(file, check).assertOneOrMoreIssues();
   }
 
+  public static void verifyWithGlobals(String path, PythonCheck check, String packageName, Map<String, Set<Symbol>> globalSymbols) {
```
pynicolas: This is a new method which would go in the public API.
An alternative to this method would be to allow testing multiple files together: I think that's what we do for COBOL. See MultiFileVerifier in https://github.com/SonarSource/sonar-analyzer-commons/tree/master/test-commons

python-frontend/src/main/java/org/sonar/python/tree/ImportFromImpl.java
```diff
@@ -43,6 +43,7 @@
   private final boolean isWildcardImport;
   private final Token wildcard;
   private final Separators separators;
+  private boolean hasUnresolvedWildcardImport = true;
```
pynicolas: Why true by default?

guillaume-dequenne-sonarsource:
1 minor question, LGTM otherwise

python-checks/src/main/java/org/sonar/python/checks/UndeclaredNameUsageCheck.java
andrea-guarino-sonarsource: In the current implementation yes, because `importFrom.hasUnresolvedWildcardImport` returns `True` by default. However I will change its behaviour to avoid this `importFrom.isWildcardImport()`

python-frontend/src/main/java/org/sonar/python/semantic/SymbolUtils.java
```diff
@@ -0,0 +1,136 @@
...
+    fileInput.accept(globalSymbolsVisitor);
+    return globalSymbolsVisitor.symbols;
+  }
+
+  private static class GlobalSymbolsVisitor extends BaseTreeVisitor {
+    private Set<Symbol> symbols = new HashSet<>();
+
+    @Override
+    public void visitFunctionDef(FunctionDef functionDef) {
+      symbols.add(new SymbolImpl(functionDef.name().name(), null));
```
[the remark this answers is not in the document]
andrea-guarino-sonarsource: Good point! If that's fine for you, I will handle this in a dedicated ticket (and a separate PR).

python-checks-testkit/src/main/java/org/sonar/python/checks/utils/PythonCheckVerifier.java
andrea-guarino-sonarsource: Thanks! I didn't know about MultiFileVerifier, I will use it!

PR #522, opened 2020-01-15
SONARPY-552 Add ClassSymbol interface

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/ClassSymbolImpl.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import javax.annotation.Nullable;
+import org.sonar.plugins.python.api.symbols.ClassSymbol;
+import org.sonar.plugins.python.api.symbols.Symbol;
+
+public class ClassSymbolImpl extends SymbolImpl implements ClassSymbol {
+
+  List<Symbol> parents = new ArrayList<>();
```
andrea-guarino-sonarsource: shouldn't they be private?

python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -517,8 +521,55 @@ private void addSymbolUsage(Name nameTree) {
       SymbolImpl symbol = scope.resolve(nameTree.name());
       // TODO: use Set to improve performances
       if (symbol != null && symbol.usages().stream().noneMatch(usage -> usage.tree().equals(nameTree))) {
-        symbol.addUsage(nameTree, Usage.Kind.OTHER);
+          symbol.addUsage(nameTree, Usage.Kind.OTHER);
```
andrea-guarino-sonarsource: Unnecessary change?

python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -517,8 +521,55 @@ private void addSymbolUsage(Name nameTree) {
...
       }
     }
+
+    private void resolveParents(ClassDef classDef) {
+      Symbol symbol = classDef.name().symbol();
+      if (symbol == null || !Symbol.Kind.CLASS.equals(symbol.kind())) {
+        return;
+      }
+      ClassSymbolImpl classSymbol = (ClassSymbolImpl) symbol;
+      if (hasMultipleBindingUsages(classSymbol)) {
```
andrea-guarino-sonarsource: I would suggest to change the kind of a classSymbol to OTHER inside `Scope#addBindingUsage`.
There you may add the following the code:
```java
if (symbol.kind() != Symbol.Kind.OTHER) {
  symbol.setKind(Symbol.Kind.OTHER);
}
```

python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -517,8 +521,53 @@ private void addSymbolUsage(Name nameTree) {
...
+        if (!argument.is(Kind.REGULAR_ARGUMENT) || !(((RegularArgument) argument).expression() instanceof HasSymbol)) {
+          classSymbol.hasUnresolvedParents = true;
+          return;
+        }
+        Symbol parentSymbol = ((HasSymbol) ((RegularArgument) argument).expression()).symbol();
+        if (parentSymbol == null) {
+          classSymbol.hasUnresolvedParents = true;
+          return;
+        }
+        if (hasMultipleBindingUsages(parentSymbol)) {
```
andrea-guarino-sonarsource: I don't think we should have this check. We should add a symbol to parentSymbol even if it has multiple binding usages. It's then up to each rule to decide how to use those symbols.
Moreover setting `classSymbol.hasUnresolvedParents` is uncorrect here because we managed to find parent symbol (altough we don't know if it's a class or not.

python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -517,8 +521,53 @@ private void addSymbolUsage(Name nameTree) {
...
+          classSymbol.hasUnresolvedParents = true;
+          return;
+        }
+        if (parentSymbol.fullyQualifiedName() != null && BuiltinSymbols.all().contains(parentSymbol.fullyQualifiedName())) {
+          classSymbol.addParent(parentSymbol);
+          continue;
+        }
+        if (!Symbol.Kind.CLASS.equals(parentSymbol.kind())) {
```
andrea-guarino-sonarsource: Even here, I think we should always add the found symbol and avoid setting `hasUnresolveParents` to true

python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
guillaume-dequenne-sonarsource: I removed the check as it was made redundant by the `Scope#addBindingUsage` change, but as discussed, the `classSymbol.hasUnresolvedParents` will still be set to true in that case (and the parent won't be added to the list).

python-frontend/src/main/java/org/sonar/python/semantic/SymbolImpl.java
```diff
@@ -39,6 +39,7 @@
   String fullyQualifiedName;
   private final List<Usage> usages = new ArrayList<>();
   private Map<String, Symbol> childrenSymbolByName = new HashMap<>();
+  private Kind kind;
```
andrea-guarino-sonarsource: Shouldn't the kind be put to `OTHER`?

python-frontend/src/test/java/org/sonar/python/semantic/ClassSymbolTest.java
```diff
@@ -20,112 +20,225 @@
...
+import static org.sonar.python.PythonTestUtils.parse;
 
 public class ClassSymbolTest {
+
+  @Test
+  public void no_parents() {
+    FileInput fileInput = parse(
+      "class C: ",
+      "  pass");
+    assertThat(fileInput.globalVariables()).hasSize(1);
```
andrea-guarino-sonarsource: I don't understand the purpose of this test. Shouldn't you assert that `fileInput.globalVariables()` has 1 symbol of kind `CLASS` with empty parents property?

andrea-guarino-sonarsource:
LGTM, please have a look at my 2 comments before merging

PR #530, opened 2020-01-20
Fix FP raised by S3827 when accessing class attribute in decorator

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -448,8 +448,12 @@ public void visitFileInput(FileInput tree) {
 
     @Override
     public void visitFunctionDef(FunctionDef pyFunctionDefTree) {
+      scan(pyFunctionDefTree.decorators());
```
andrea-guarino-sonarsource: I think we should do the same for class decorators.
```python
class A:
    _ATTR = 42
    @decorator(_ATTR)  # OK
    class Foo: pass
```

PR #535, opened 2020-01-21
Update rules medatada

guillaume-dequenne-sonarsource:


python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S5332.html
```diff
@@ -1,37 +1,3 @@
-<p>Clear-text protocols as <code>ftp</code>, <code>telnet</code> or non secure <code>http</code> are lacking encryption of transported data. They are
```
andrea-guarino-sonarsource: I don't think this is expected, probably a bug on rule-api

python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S2712.json
```diff
@@ -1,7 +1,7 @@
 {
   "title": "\"return\" and \"yield\" should not be used in the same function",
   "type": "BUG",
-  "status": "ready",
+  "status": "deprecated",
```
andrea-guarino-sonarsource: I don't think this should be deprecated, as it is still relevant for Python 2 developers. Note that this rule is not on SonarWay anyway.

PR #540, opened 2020-01-28
SONARPY-559 Resolve fully qualified name of virtual calls for single …

andrea-guarino-sonarsource:
…assigned symbols.

### DIFF
--- a/python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
+++ b/python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
@@ -44,6 +44,7 @@ import org.sonar.plugins.python.api.tree.AnnotatedAssignment;
 import org.sonar.plugins.python.api.tree.AnyParameter;
 import org.sonar.plugins.python.api.tree.AssignmentStatement;
 import org.sonar.plugins.python.api.tree.BaseTreeVisitor;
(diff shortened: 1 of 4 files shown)


python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -549,5 +557,23 @@ private void addSymbolUsage(Name nameTree) {
         symbol.addUsage(nameTree, Usage.Kind.OTHER);
       }
     }
+
+    private void addTypeToSymbol(Name nameTree, Expression rhs) {
+      Type type = null;
+      if (rhs.is(Kind.CALL_EXPR)) {
+        Expression callee = ((CallExpression) rhs).callee();
+        if (callee.is(Kind.NAME)) {
```
guillaume-dequenne-sonarsource: Does that mean we won't be able to handle type resolution if the callee is a qualified expression instead? From the tests it seems so, but I don't understand why?

andrea-guarino-sonarsource: For the time being yes, as a first baby step. I prefer to handle that case in a separate PR (and I documented with a unit test).

guillaume-dequenne-sonarsource: fair enough! lgtm then

PR #542, opened 2020-01-29
SONARPY-273 Rule S2068: Hard-coded credentials are security-sensitive

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/HardCodedCredentialsCheck.java
```diff
@@ -0,0 +1,286 @@
...
+  private Stream<Pattern> variablePatterns() {
+    if (variablePatterns == null) {
+      variablePatterns = toPatterns("");
+    }
+    return variablePatterns.stream();
+  }
+
+  private Stream<Pattern> literalPatterns() {
+    if (literalPatterns == null) {
+      literalPatterns = toPatterns("=[^:%'?\\s]+");
```
andrea-guarino-sonarsource: can you add a comment explaining why literal patterns need a suffix?

python-checks/src/main/java/org/sonar/python/checks/HardCodedCredentialsCheck.java
```diff
@@ -0,0 +1,286 @@
...
+        && defaultValue.is(Tree.Kind.STRING_LITERAL) && !isEmptyStringLiteral(((StringLiteral) defaultValue))) {
+        ctx.addIssue(parameter, MESSAGE);
+      }
+    }
+  }
+
+  private void handleRegularArgument(RegularArgument regularArgument, SubscriptionContext ctx) {
+    Name keywordArgument = regularArgument.keywordArgument();
+    if (keywordArgument != null && isCredential(keywordArgument.name(), variablePatterns())
+      && regularArgument.expression().is(Tree.Kind.STRING_LITERAL) && !isEmptyStringLiteral((StringLiteral) regularArgument.expression())) {
```
andrea-guarino-sonarsource: I think we can factorize these two conditions in a separate method:
```java
private static boolean isNonEmptyStringLiteral(Tree tree) {
    return tree.is(Kind.STRING_LITERAL) && !isEmptyStringLiteral((StringLiteral)tree)
}
```

python-checks/src/main/java/org/sonar/python/checks/HardCodedCredentialsCheck.java
```diff
@@ -0,0 +1,286 @@
...
+      }
+    } catch (MalformedURLException e) {
+      return false;
+    }
+    return false;
+  }
+
+  private void handleBinaryExpression(BinaryExpression binaryExpression, SubscriptionContext ctx) {
+    boolean shouldReport = false;
+    if (binaryExpression.leftOperand().is(Tree.Kind.NAME, Tree.Kind.QUALIFIED_EXPR) && binaryExpression.rightOperand().is(Tree.Kind.STRING_LITERAL)) {
```
andrea-guarino-sonarsource: I think we can statically import `Tree.Kind`

python-checks/src/main/java/org/sonar/python/checks/HardCodedCredentialsCheck.java
```diff
@@ -0,0 +1,286 @@
...
+    if (binaryExpression.rightOperand().is(Tree.Kind.NAME, Tree.Kind.QUALIFIED_EXPR) && binaryExpression.leftOperand().is(Tree.Kind.STRING_LITERAL)) {
+      shouldReport = checkBinaryExpression(binaryExpression.rightOperand());
+    }
+    if (shouldReport) {
+      ctx.addIssue(binaryExpression, MESSAGE);
+    }
+  }
+
+  private boolean checkBinaryExpression(Expression operand) {
+    if (operand.is(Tree.Kind.NAME)) {
```
andrea-guarino-sonarsource: what do you think about checking the symbol name instead (and having as parameter a `HasSymbol expression`)?

python-checks/src/main/java/org/sonar/python/checks/HardCodedCredentialsCheck.java
```diff
@@ -0,0 +1,286 @@
...
+    } else {
+      return isCredential(((QualifiedExpression) operand).name().name(), variablePatterns());
+    }
+  }
+
+  private void handleAssignmentStatement(AssignmentStatement assignmentStatement, SubscriptionContext ctx) {
+    ExpressionList lhs = assignmentStatement.lhsExpressions().get(0);
+    Expression expression = lhs.expressions().get(0);
+
+    if (expression.is(Tree.Kind.NAME) && isCredential(((Name) expression).name(), variablePatterns())) {
```
andrea-guarino-sonarsource: also here, we could use the symbol name instead of checking if it's a name and if it's a qualified expression.

python-checks/src/main/java/org/sonar/python/checks/HardCodedCredentialsCheck.java
guillaume-dequenne-sonarsource: The advantage of the current implementation is we can report issues even if we can't resolve the symbol.
That said I agree the code would look much cleaner using `HasSymbol`, so I don't have a big opinion tbh, do you think I should go for HasSymbol?

python-checks/src/main/java/org/sonar/python/checks/HardCodedCredentialsCheck.java
```diff
@@ -0,0 +1,292 @@
...
+
+  private static boolean isNonEmptyStringLiteral(Tree tree) {
+    return tree.is(Kind.STRING_LITERAL) && !isEmptyStringLiteral((StringLiteral) tree);
+  }
+
+  private static boolean isEmptyStringLiteral(StringLiteral stringLiteral) {
+    return stringLiteral.trimmedQuotesValue().isEmpty();
+  }
+
+  private static boolean isCredential(String target, Stream<Pattern> patterns) {
```
andrea-guarino-sonarsource: I think I'd refactor `isSymbolCredential` and `isCredential` in a single method:
```java
private static boolean isCredential(Tree tree, Stream<Pattern> patterns) { 
  if (tree instanceof HasSymbol) { 
     // ... 
  } else if (tree.is(Kind.STRING_LITERAL)) {
    // ... 
  }
} 
```
However feel free to discard this comment, it's just a personal preference. :)

PR #543, opened 2020-01-30
SONARPY-491 Rule S2053 Hashes should include an unpredictable salt

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/PredictableSaltCheck.java
```diff
@@ -0,0 +1,119 @@
...
+        } else if (i == argNb) {
+          checkSensitiveArgument(regularArgument, ctx);
+        }
+      }
+    }
+  }
+
+  private static void checkSensitiveArgument(RegularArgument regularArgument, SubscriptionContext ctx) {
+    if (regularArgument.expression().is(Tree.Kind.NAME)) {
+      Symbol symbol = ((Name) regularArgument.expression()).symbol();
```
andrea-guarino-sonarsource: I think you can use the utility method `Expressions.singleAssignedValue`

python-checks/src/test/resources/checks/predictableSalt.py
```diff
@@ -0,0 +1,119 @@
...
+import os
+import hashlib
+import os
+
+# crypt-salt
+password = 'password123'
+salt = crypt.mksalt(crypt.METHOD_SHA256)
+
+hash = crypt.crypt(password)         # Noncompliant {{Add an unpredictable salt value to this hash.}}
+#      ^^^^^^^^^^^^^^^^^^^^^
```
andrea-guarino-sonarsource: what about highlighting only the callee?

PR #546, opened 2020-01-31
SONARPY-495 Rule 5042: Expanding archive files is security-sensitive

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/ExpandingArchiveCheck.java
```diff
@@ -0,0 +1,40 @@
...
+  protected Set<String> functionsToCheck() {
+    return immutableSet(
+      "zipfile.ZipFile.extractall",
+      "tarfile.open"
+    );
+  }
+
+  @Override
+  protected String message() {
+    return " Make sure that expanding this archive file is safe here.";
```
andrea-guarino-sonarsource: Can you remove the leading whitespace?

python-checks/src/main/java/org/sonar/python/checks/ExpandingArchiveCheck.java
```diff
@@ -0,0 +1,40 @@
+/*
```
andrea-guarino-sonarsource: being a security hotspot, could you move it to `hotspots` package?

PR #548, opened 2020-02-03
SONARPY-489 Rule S2092: Creating cookies without the "secure" flag is security-sensitive

andrea-guarino-sonarsource:


python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S2092.html
```diff
@@ -0,0 +1,62 @@
...
+  when creating cookies. </li>
+  <li> Set the <code>secure</code> flag to <em>true</em> for session-cookies. </li>
+</ul>
+<h2>Sensitive Code Examples</h2>
+<p>http.cookies</p>
+<pre>
+import http.cookies
+
+cookie = http.cookies.SimpleCookie()
+cookie['key'] = 'value' # Sensitive
```
guillaume-dequenne-sonarsource: If I understood correctly, we won't cover this case at first, so I guess it might be worth updating the rule description.

python-checks/src/main/java/org/sonar/python/checks/hotspots/SecureCookieCheck.java
```diff
@@ -0,0 +1,167 @@
...
+    });
+  }
+
+  @CheckForNull
+  private static RegularArgument getSecureArgument(List<Argument> arguments) {
+    return arguments.stream()
+      .map(RegularArgument.class::cast)
+      .filter(argument -> {
+        Name keywordArgument = argument.keywordArgument();
+        return keywordArgument != null && keywordArgument.name().equals(SECURE);
```
guillaume-dequenne-sonarsource: This implementation seems to assume the secure flag will always be passed as a keyword argument, with a risk of FP if it's passed as a positional argument. I haven't checked for django but I believe it's possible to use positional arguments with flask at least.
Would you consider accounting for positional arguments as well to avoid those FP (maybe by doing something similar to `PredictableSaltCheck`) ?

python-frontend/src/main/java/org/sonar/python/semantic/SymbolUtils.java
```diff
@@ -264,25 +276,38 @@ static Path pathOf(PythonFile pythonFile) {
...
+      classSymbol("HttpResponseBadRequest", "django.http.HttpResponseBadRequest")
+    )));
+
+    globalSymbols.put("flask", new HashSet<>(Collections.singletonList(classSymbol("Response", "flask.Response"))));
+
     return globalSymbols;
   }
+
+  private static ClassSymbolImpl classSymbol(String name, String fullyQualifiedName) {
+    ClassSymbolImpl djangoHttpResponseRedirect = new ClassSymbolImpl(name, fullyQualifiedName);
```
guillaume-dequenne-sonarsource: I think you forgot to rename this variable.

python-checks/src/main/java/org/sonar/python/checks/hotspots/SecureCookieCheck.java
```diff
@@ -0,0 +1,167 @@
...
+      .findFirst().orElse(null);
+  }
+
+  private static Stream<SubscriptionExpression> getSubscriptionToCookies(List<ExpressionList> lhsExpressions) {
+    return lhsExpressions.stream()
+      .flatMap(expressionList -> expressionList.expressions().stream())
+      .filter(lhs -> {
+        if (lhs.is(Kind.SUBSCRIPTION)) {
+          SubscriptionExpression sub = (SubscriptionExpression) lhs;
+          Symbol objectSymbol = getObjectSymbol(sub.object());
```
guillaume-dequenne-sonarsource: Just a thought: do you think it would actually be feasible to raise on cases like: 
`cookie['c1'] = 'value' # FN`
by checking whether the `cookie` symbol has another assignment usage with the same `c1` subscript as well as a `secure` one?
Maybe it's not worth doing as I believe this is not the usual way to set cookies and it might be more complex to do than it looks, but I prefer mentioning it, feel free to disregard this!

python-frontend/src/main/java/org/sonar/python/semantic/SymbolUtils.java
andrea-guarino-sonarsource: Indeed, thanks! :)

python-checks/src/main/java/org/sonar/python/checks/hotspots/SecureCookieCheck.java
andrea-guarino-sonarsource: That's a fair point. However I have the feeling that the value doesn't justify the extra complexity. I suggest to create a ticket to keep track of this FN.

python-checks/src/main/java/org/sonar/python/checks/hotspots/SecureCookieCheck.java
andrea-guarino-sonarsource: Yes good point, I'll check for positional argument

python-checks/src/main/java/org/sonar/python/checks/hotspots/SecureCookieCheck.java
andrea-guarino-sonarsource: Ticket created https://jira.sonarsource.com/browse/SONARPY-577

python-checks/src/main/java/org/sonar/python/checks/hotspots/SecureCookieCheck.java
```diff
@@ -104,14 +107,30 @@ public void initialize(Context context) {
...
+  private static RegularArgument getSecureArgument(List<Argument> arguments, int nArg) {
     return arguments.stream()
+      // in call site of this function, argument will always be of type RegularArgument because we check no argument is unpacking
       .map(RegularArgument.class::cast)
       .filter(argument -> {
         Name keywordArgument = argument.keywordArgument();
         return keywordArgument != null && keywordArgument.name().equals(SECURE);
       })
-      .findFirst().orElse(null);
+      .findFirst().orElse(checkSensitivePositionalArgument(arguments, nArg));
```
guillaume-dequenne-sonarsource: Could you use `Optional#orElseGet` instead of `orElse` ? As the latter is always evaluated no matter whether the former is present or not.

andrea-guarino-sonarsource: good catch! thanks!

PR #549, opened 2020-02-03
SONARPY-528 Rule S5332 to support smtplib

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/hotspots/ClearTextProtocolsCheck.java
```diff
@@ -63,6 +68,31 @@ public void initialize(Context context) {
       Symbol symbol = ((CallExpression) ctx.syntaxNode()).calleeSymbol();
       isUnsafeLib(symbol).ifPresent(protocol -> ctx.addIssue(ctx.syntaxNode(), message(protocol)));
     });
+
+    context.registerSyntaxNodeConsumer(Tree.Kind.ASSIGNMENT_STMT, ctx -> handleAssignmentStatement((AssignmentStatement) ctx.syntaxNode(), ctx));
+  }
+
+  private static void handleAssignmentStatement(AssignmentStatement assignmentStatement, SubscriptionContext ctx) {
+    if (assignmentStatement.lhsExpressions().get(0).expressions().get(0) instanceof HasSymbol) {
```
andrea-guarino-sonarsource: I think we may have FP in the following cases:

```python
smtp1 = smtp2 = smtplib.SMTP("smtp.gmail.com", port=587) # FP
smtp2.starttls(context=context)
``` 
```python
smtp1 =  smtplib.SMTP("smtp.gmail.com", port=587) # FP
smtp2 = smtp1
smtp2.starttls(context=context)
``` 

```python
smtp1 =  smtplib.SMTP("smtp.gmail.com", port=587) # FP
start_tls(stmp1)

def start_tls(x): 
  x.starttls(context=context)
``` 

To implement this properly, we would need a `points-to` analysis, which is definitely out of scope for this MMF.

pierre-loup-tristant-sonarsource: I think these FPs are acceptable

PR #550, opened 2020-02-04
SONARPY-490 Rule S2115: Databases should be password-protected

vilchik-elena:


python-checks/src/test/resources/checks/dbNoPassword/settings.py
```diff
@@ -0,0 +1,50 @@
...
+
+DATABASES[42] = {
+    'postgresql_db': {
+        'PASSWORD': '',                            # Compliant
+    }
+}
+
+DATABASES, OTHER = {
+    'postgresql_db': {
+        'PASSWORD': '',                            # Compliant
```
andrea-guarino-sonarsource: As discussed, let's replace "Compliant" with "Out of scope"

PR #551, opened 2020-02-05
SONARPY-464 Rule S5527: Server hostnames should be verified during SSL/TLS connections

guillaume-dequenne-sonarsource:


python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S5527.html
```diff
@@ -0,0 +1,68 @@
+<p>To establish a SSL/TLS connection not vulnerable to man-in-the-middle attacks, it's essential to make sure the server presents the right
```
vilchik-elena: this file clearly has more content then here https://jira.sonarsource.com/browse/RSPEC-5530
why?

python-checks/src/test/resources/checks/hotspots/unverifiedHostname.py
```diff
@@ -0,0 +1,51 @@
+import urllib.request
+import ssl
+
+# ValueError: Cannot set verify_mode to CERT_NONE when check_hostname is enabled
+
+ssl._create_unverified_context() # OK?
```
vilchik-elena: what does "OK?" mean?

python-checks/src/main/java/org/sonar/python/checks/hotspots/UnverifiedHostnameCheck.java
```diff
@@ -0,0 +1,113 @@
...
+        if(hasUnsafeContextUsage(calleeSymbol, symbol)) {
+          ctx.addIssue(callExpression, MESSAGE);
+        }
+      }
+    } else if (opensUnsecureConnection(calleeSymbol, (CallExpression) parent)) {
+      ctx.addIssue(callExpression, MESSAGE);
+    }
+  }
+
+  private static boolean hasUnsafeContextUsage(Symbol calleeSymbol, Symbol symbol) {
```
vilchik-elena: I don't think this name reflects this method, it might true even w/o any usage

python-checks/src/test/resources/checks/hotspots/unverifiedHostname.py
```diff
@@ -0,0 +1,51 @@
...
+ctx_unverified_ok.check_hostname = True
+
+ctx2 = ssl._create_stdlib_context() # Noncompliant
+if ctx2.check_hostname:
+  pass
+
+ctx3 = ssl.create_default_context() # Noncompliant
+ctx3.check_hostname = False
+
+r1 = urllib.request.urlopen('https://151.101.0.223', context=ctx)
```
vilchik-elena: why these 2 lines?

python-checks/src/main/java/org/sonar/python/checks/hotspots/UnverifiedHostnameCheck.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  private static void checkSuspiciousCall(CallExpression callExpression, Symbol calleeSymbol, SubscriptionContext ctx) {
+    Tree parent = TreeUtils.firstAncestorOfKind(callExpression, Tree.Kind.ASSIGNMENT_STMT, Tree.Kind.CALL_EXPR);
+    if (parent == null) {
+      return;
+    }
+    if (parent.is(Tree.Kind.ASSIGNMENT_STMT)) {
+      AssignmentStatement parentAssignment = (AssignmentStatement) parent;
+      if (parentAssignment.lhsExpressions().get(0).expressions().get(0) instanceof HasSymbol) {
```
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: Strictly speaking, no...But in the case of an assignment statement, there will always be at least one element there. Do you want me to add an additional check nevertheless?

python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S5527.html
guillaume-dequenne-sonarsource: Indeed! The RSPEC got updated during development and I forgot to update! Thanks :)

python-checks/src/test/resources/checks/hotspots/unverifiedHostname.py
guillaume-dequenne-sonarsource: should be OK now :)

PR #552, opened 2020-02-05
SONARPY-576 Rule S2092: Flask.make_response() and flask.redirect() should have 'Flask.Response' as return type

andrea-guarino-sonarsource:
**DO NOT SQUASH**

### DIFF
--- a/python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
+++ b/python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
@@ -560,20 +560,32 @@ public class SymbolTableBuilder extends BaseTreeVisitor {
     }
 
     private void addTypeToSymbol(Name nameTree, Expression rhs) {
-      Type type = null;
-      if (rhs.is(Kind.CALL_EXPR)) {
-        Symbol calleeSymbol = ((CallExpression) rhs).calleeSymbol();
-        if (calleeSymbol != null && calleeSymbol.kind() == Symbol.Kind.CLASS) {
-          ClassSymbol classSymbol = (ClassSymbol) calleeSymbol;
-          if (classSymbol.superClasses().isEmpty() && !classSymbol.hasUnresolvedTypeHierarchy()) {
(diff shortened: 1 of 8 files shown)


python-frontend/src/main/java/org/sonar/python/semantic/SymbolImpl.java
```diff
@@ -116,4 +118,19 @@ void setType(@Nullable Type type) {
   public Type type() {
     return type;
   }
+
+  /*
+   * 'returnType' only applies to function symbols.
+   * It's not defined in FunctionSymbolImpl class because FunctionSymbolImpl must always have a definition location.
+   * Rules implemented in MMF-1845 need to know types of external libraries and symbols coming from said external libraries
+   * don't have a definition location.
```
guillaume-dequenne-sonarsource: I am not sure I understand why we can't have the `returnType` in `FunctionSymbolImpl` nevertheless?
Isn't it possible to make the definition location optional, or have a default value for it?

python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -560,21 +560,32 @@ private void addSymbolUsage(Name nameTree) {
...
 
+    @CheckForNull
+    private Type getReturnType(CallExpression rhs) {
+      Symbol calleeSymbol = rhs.calleeSymbol();
+      if (calleeSymbol == null) {
+        return null;
+      }
+      if (calleeSymbol.kind() == Symbol.Kind.CLASS) {
+        ClassSymbol classSymbol = (ClassSymbol) calleeSymbol;
+        if (classSymbol.superClasses().isEmpty() && !classSymbol.hasUnresolvedTypeHierarchy()) {
```
guillaume-dequenne-sonarsource: I know it's not linked to this PR but do you think we could add a comment to explain why we have this condition `classSymbol.superClasses().isEmpty()` ? As it should eventually be replaced, I believe it's good to keep it in mind.

PR #553, opened 2020-02-05
SONARPY-273 Fix FP on S2068 and update message

guillaume-dequenne-sonarsource:


python-checks/src/test/resources/checks/hardCodedCredentials.py
```diff
@@ -205,3 +205,14 @@ def somePassword(self, *, password="hello"): # Noncompliant
         'PORT': '5432'
     }
 }
+
+dict1 = {'password': ''} # Compliant
```
andrea-guarino-sonarsource: Please add a comment explaining why those cases are compliant

PR #555, opened 2020-02-07
Rule S2068: Avoid trying to parse non-url string literals

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/hotspots/HardCodedCredentialsCheck.java
```diff
@@ -246,6 +247,9 @@ private void handleStringLiteral(StringLiteral stringLiteral, SubscriptionContex
   }
 
   private static boolean isURLWithCredentials(StringLiteral stringLiteral) {
+    if (!URL_LIKE.matcher(stringLiteral.trimmedQuotesValue()).find()) {
```
andrea-guarino-sonarsource: what about using `stringLiteral.trimmedQuotesValue().contains("://")`?

guillaume-dequenne-sonarsource: Much simpler indeed! Thanks :)

PR #556, opened 2020-02-07
SONARPY-492 Rule S5122: Enabling Cross-Origin Resource Sharing is security-sensitive

vilchik-elena:


vilchik-elena:
@andrea-guarino-sonarsource  could you review again as I did a significant changes?

python-frontend/src/main/java/org/sonar/python/tree/DecoratorImpl.java
```diff
@@ -32,7 +32,7 @@
 import org.sonar.plugins.python.api.tree.Tree;
 import org.sonar.plugins.python.api.tree.TreeVisitor;
 
-public class DecoratorImpl extends PyTree implements Decorator {
+public class DecoratorImpl extends SimpleStatement implements Decorator {
```
andrea-guarino-sonarsource: Can you add a comment explaining why we are doing this?

python-checks/src/main/java/org/sonar/python/checks/hotspots/CorsCheck.java
```diff
@@ -0,0 +1,356 @@
...
+
+  private static final String STAR = "*";
+
+  private static final List<String> REGEX_TO_REPORT = Arrays.asList(
+    ".*",
+    ".+",
+    "^.*$",
+    "^.+$",
+    STAR
+  );
```
pierre-loup-tristant-sonarsource: Would it be possible and only keep `*`, `.*` and `.+` in the list and and trim '$' and '^' characters before matching? It would cover permissive regex like `^.+`.

PR #559, opened 2020-02-10
SONARPY-498 Fix False Positive S4143: assigned collection in rhs

vilchik-elena:


python-checks/src/test/resources/checks/overwrittenCollectionEntryCheck.py
```diff
@@ -91,3 +91,10 @@ def non_trivial_collections(f, obj1, obj2):
 def valid_replacement(list1, foo, x):
     list1["foo"] = foo(x)
     list1["foo"] = list1["foo"].replace("a", "b")
+
+def used_collection(list1):
+    list1[1] = 42
+    list1[1] = foo(list1[1])
+
+    list1[2] = 42
+    list1[2] = foo(list1)
```
guillaume-dequenne-sonarsource: Could you document the fact that the following snippet would now be a FN ?
```
    list[3] = 42
    list[3] = list[1] # FN
```

PR #561, opened 2020-02-11
SONARPY-556 Fix FP raised by S3827 when using global variables

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -399,8 +399,8 @@ public void visitCompoundAssignment(CompoundAssignmentStatement pyCompoundAssign
     public void visitGlobalStatement(GlobalStatement pyGlobalStatementTree) {
       pyGlobalStatementTree.variables()
         .forEach(name -> {
-          currentScope().addGlobalName(name.name());
           moduleScope.addBindingUsage(name, Usage.Kind.GLOBAL_DECLARATION, null);
```
andrea-guarino-sonarsource: I know it's not really related to this PR, however I think we should update as well `Usage#isBindingUsage()` to avoid considering `Usage.Kind.GLOBAL_DECLARATION` as a binding usage. wdyt?

PR #563, opened 2020-02-11
SONARPY-581 Fix FP on S3827: Set comprehension variable inside f-string

vilchik-elena:


python-frontend/src/main/java/org/sonar/python/tree/PythonTreeMaker.java
```diff
@@ -1299,14 +1295,17 @@ private Expression stringLiteral(AstNode astNode) {
...
     parser.setRootRule(parser.getGrammar().rule(PythonGrammar.EXPR));
     // get escaped interpolation
-    Matcher matcher = INTERPOLATED_EXPR_PATTERN.matcher(literalValue);
-    while (matcher.find()) {
-      AstNode parse = parser.parse(matcher.group(2));
+    Matcher matcher = INTERPOLATED_EXPR_START_PATTERN.matcher(literalValue);
+    int index = 0;
+    while (matcher.find(index)) {
+      int end = matcher.end();
+      AstNode parse = parser.parse(literalValue.substring(end));
```
andrea-guarino-sonarsource: Isn't it parsing all the remaining string literal here?
Like, for `f"{x} is equal to {y}"` it will parse `x} is equal to {y}` instead of just `x`

vilchik-elena: indeed input of `parser.parse` will be `x} is equal to {y}` but this parser will eventually return tree corresponding to `x` only as the rest does not correspond to the grammar

python-frontend/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
```diff
@@ -1779,6 +1780,14 @@ public void string_interpolation() {
...
     assertThat(nestedInterpolation.firstToken().line()).isEqualTo(2);
     assertThat(nestedInterpolation.firstToken().column()).isEqualTo(15);
+
+    // interpolated expression contains curly braces
+    stringLiteral = (StringLiteral) parse("f'{ {element for element in [1, 2]} }'", treeMaker::expression);
+    assertThat(stringLiteral.stringElements()).hasSize(1);
+    elmt = stringLiteral.stringElements().get(0);
+    assertThat(elmt.isInterpolated()).isTrue();
+    assertThat(elmt.interpolatedExpressions()).hasSize(1);
+    assertThat(elmt.interpolatedExpressions().get(0).is(Kind.COMP_FOR));
```
andrea-guarino-sonarsource: I think you forgot `isTrue()` here. And in this case the kind should be "SET_COMPREHENSION"

PR #570, opened 2020-02-13
SONARPY-590 Avoid No workDir in SonarLint log printed in the console

guillaume-dequenne-sonarsource:


python-frontend/src/test/java/org/sonar/python/PythonVisitorCheckTest.java
```diff
@@ -111,6 +111,22 @@ public void initialize(SubscriptionCheck.Context context) {
     SubscriptionVisitor.analyze(Collections.singletonList(check), context);
   }
 
+  @Test
+  public void working_directory_null() throws IOException {
```
vilchik-elena: Do you understand that this test is not testing the change you did?
It tests PythonVisitorContext constructor and `PythonVisitorContext#workingDirectory()` getter

PR #573, opened 2020-02-17
SONARPY-507 Rule S3403: Identity operators should not be used with dissimilar types

pynicolas:


python-frontend/src/main/java/org/sonar/plugins/python/api/types/InferredType.java
```diff
@@ -0,0 +1,26 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.plugins.python.api.types;
+
+public interface InferredType {
```
andrea-guarino-sonarsource: As discussed, I think we can add `@Beta` on InferredType interface

python-frontend/src/main/java/org/sonar/python/tree/NameImpl.java
```diff
@@ -91,4 +93,9 @@ public void setSymbol(Symbol symbol) {
   public void setUsage(Usage usage) {
     this.usage = usage;
   }
+
+  @Override
+  public InferredType type() {
+    return ((SymbolImpl) symbol).inferredType();
```
andrea-guarino-sonarsource: I think that `symbol` might be null here.

python-frontend/src/main/java/org/sonar/plugins/python/api/types/InferredType.java
pynicolas: Done in 1f488f7.

python-frontend/src/main/java/org/sonar/python/tree/NameImpl.java
pynicolas: Fixed in b9f4ed6.

PR #574, opened 2020-02-17
SONARPY-593 Type inference: types of literals

andrea-guarino-sonarsource:


python-frontend/src/main/java/org/sonar/python/tree/NumericLiteralImpl.java
```diff
@@ -21,19 +21,24 @@
...
-import org.sonar.plugins.python.api.tree.TreeVisitor;
 import org.sonar.plugins.python.api.tree.Tree;
+import org.sonar.plugins.python.api.tree.TreeVisitor;
+import org.sonar.plugins.python.api.types.InferredType;
+import org.sonar.python.types.InferredTypes;
 
 public class NumericLiteralImpl extends PyTree implements NumericLiteral {
 
   private final String valueAsString;
+  private final String valueAsStringLowerCase;
```
pynicolas: Why do we store that as a field? If we want to store something as a field, I would prefer to store the inferred type.

python-frontend/src/main/java/org/sonar/python/tree/NumericLiteralImpl.java
```diff
@@ -67,4 +72,16 @@ public String valueAsString() {
...
+  // https://docs.python.org/3/reference/lexical_analysis.html#numeric-literals
+  @Override
+  public InferredType type() {
+    if (valueAsStringLowerCase.contains("j")) {
+      return InferredTypes.runtimeType("complex");
+    }
+    if (valueAsString.contains(".") || valueAsStringLowerCase.contains("e")) {
+      return InferredTypes.runtimeType("float");
+    }
+    return InferredTypes.runtimeType("int");
```
pynicolas: Maybe we could have constants for `InferredTypes.runtimeType("int")` and other types to avoid reinstantiating them many times.

python-checks/src/test/resources/checks/sillyIdentity.py
```diff
@@ -18,3 +18,42 @@ def foo():
...
+def other_literals():
+    a = A()
+    if a is []: # Noncompliant
+        pass
+
+    if [] is []: # FN
+        pass
+
+    if [] is {}: # Noncompliant
+        pass
```
pynicolas: I don't really see the value of those test cases.

andrea-guarino-sonarsource: Ok, I will keep a couple of them which I believe have some values, like documented false negative at line 55.

PR #577, opened 2020-02-19
SONARPY-596 Type inference should use function return types provided …

andrea-guarino-sonarsource:
…by typeshed for 2and3


python-frontend/src/main/java/org/sonar/python/types/InferredTypes.java
```diff
@@ -62,4 +66,12 @@ public static InferredType or(InferredType t1, InferredType t2) {
     return InferredTypes.anyType();
   }
 
+  public static InferredType declaredType(TypeAnnotation typeAnnotation) {
+    Expression expression = typeAnnotation.expression();
+    if (expression.is(Kind.NAME) && !((Name) expression).name().equals("Any")) {
+      // TODO change it to DeclaredType instance
```
andrea-guarino-sonarsource: I didn't create `DeclareType` class in this PR because I don't see a concrete benefit for the rule we have at the moment. What do you think?

pynicolas: I think that we may have false positives if the Typeshed declarations we import contain a function which can return a subtype of its declared return type.
```
class A:
  pass
class A1(A):
  pass
def f() -> A: ...
```
```
a = A1()
if f() is a: # FP would be raised here
  pass
```

python-frontend/src/main/java/org/sonar/python/types/TypeShedVisitor.java
```diff
@@ -0,0 +1,96 @@
...
+  public void visitFunctionDef(FunctionDef functionDef) {
+    String functionName = functionDef.name().name();
+    SymbolImpl symbol = (SymbolImpl) symbolsByName.get(functionName);
+    if (symbol != null) {
+      symbol.setKind(OTHER);
+    } else {
+      FunctionSymbolImpl functionSymbol = new FunctionSymbolImpl(functionName, functionName, false, false, false, Collections.emptyList(), null);
+      TypeAnnotation returnTypeAnnotation = functionDef.returnTypeAnnotation();
+      if (returnTypeAnnotation != null) {
+        functionSymbol.setInferredReturnType(InferredTypes.declaredType(returnTypeAnnotation));
```
pynicolas: Is `setInferredReturnType` the right name? Shouldn't it be `setDeclaredReturnType`? We could imagine that some day we compare the return type which is declared in a function declaration with the return type which we can infer from its body.
```
def f() -> str:
  return 42
```

python-frontend/src/main/java/org/sonar/python/types/TypeShedVisitor.java
```diff
@@ -0,0 +1,96 @@
...
+      InputStream resource = TypeShedVisitor.class.getResourceAsStream("builtins.pyi");
+      try (BufferedReader reader = new BufferedReader(new InputStreamReader(resource, StandardCharsets.UTF_8))) {
+        String builtins = reader.lines().collect(Collectors.joining("\n"));
+        AstNode astNode = PythonParser.create().parse(builtins);
+        FileInput fileInput = new PythonTreeMaker().fileInput(astNode);
+        TypeShedVisitor typeShedVisitor = new TypeShedVisitor();
+        fileInput.accept(typeShedVisitor);
+        typeShedSymbols = typeShedVisitor.symbolsByName;
+      } catch (IOException | UncheckedIOException e) {
+        LOG.debug("Unable to read builtin types.");
```
pynicolas: I would change it to info level and add the original exception:
```
LOG.info("Unable to read builtin types.", e);
```

python-frontend/src/main/java/org/sonar/python/types/InferredTypes.java
andrea-guarino-sonarsource: As discussed, we will handle this in a separate ticket: https://jira.sonarsource.com/browse/SONARPY-605

PR #581, opened 2020-02-21
SONARPY-601 Type inference: types of conditional expressions

pynicolas:


python-frontend/src/test/java/org/sonar/python/types/TypeInferenceTest.java
```diff
@@ -252,6 +252,20 @@ public void builtin_function_types() {
     assertThat(lastExpression("getattr(42)").type()).isEqualTo(AnyType.ANY);
   }
 
+  @Test
+  public void conditional_expressions() {
+    assertThat(lastExpression("42 if '' else 43").type()).isEqualTo(INT);
+    assertThat(lastExpression("42 if cond else ''").type()).isEqualTo(or(INT, STR));
+    assertThat(lastExpression("42 if '' else xxx").type()).isEqualTo(anyType());
+
```
andrea-guarino-sonarsource: What do you think about adding a test with nested conditional?

Another interesting case to test could be:
```python
c = 42 if '' else c
```

PR #590, opened 2020-02-25
SONARPY-594 Type inference: types of logical , binary plus and parenthesized expressions

andrea-guarino-sonarsource:


python-frontend/src/main/java/org/sonar/python/tree/BinaryExpressionImpl.java
```diff
@@ -103,4 +107,17 @@ public Kind getKind() {
...
+  public InferredType type() {
+    if (is(Kind.AND, Kind.OR)) {
+      return InferredTypes.or(leftOperand.type(), rightOperand.type());
+    }
+    return InferredTypes.anyType();
+  }
+
+  @Override
+  public List<Expression> typeDependencies() {
+    return Arrays.asList(leftOperand, rightOperand);
```
pynicolas: To avoid unnecessary calculations, it would be better to return an empty list when `!is(Kind.AND, Kind.OR)`.

python-frontend/src/test/java/org/sonar/python/tree/BinaryExpressionImplTest.java
```diff
@@ -0,0 +1,55 @@
...
+import static org.sonar.python.types.InferredTypes.INT;
+import static org.sonar.python.types.InferredTypes.STR;
+import static org.sonar.python.types.InferredTypes.or;
+
+public class BinaryExpressionImplTest {
+
+  @Test
+  public void type() {
+    Expression left = mock(Expression.class);
+    when(left.type()).thenReturn(INT);
```
pynicolas: For this kind of test, I'm not sure that mocking is better than parsing python code.

python-frontend/src/test/java/org/sonar/python/types/TypeInferenceTest.java
```diff
@@ -276,6 +276,23 @@ public void conditional_expressions() {
...
+    assertThat(lastExpression("42 or ''").type()).isEqualTo(or(INT, STR));
+    assertThat(lastExpression("42 or xxx").type()).isEqualTo(anyType());
+    assertThat(lastExpression("42 or True or ''").type()).isEqualTo(or(or(INT, STR), BOOL));
+
+    assertThat(lastExpressionInFunction(
+      "for i in range(3):",
+      "  if   i > 1: c = b",
+      "  elif i > 0: b = 42 or a",
+      "  else:       a = ''",
+      "c").type()).isEqualTo(or(INT, STR));
```
pynicolas: Here we test both:
* something which is specific to logical expressions: `BinaryExpressionImpl#typeDependencies()`
* the algorithm to propagate types between expressions.

I think that we should have only 1 test for the second point and test the various implementations of `typeDependencies()` explicitly.

python-frontend/src/main/java/org/sonar/python/tree/BinaryExpressionImpl.java
```diff
@@ -113,6 +116,14 @@ public InferredType type() {
     if (is(Kind.AND, Kind.OR)) {
       return InferredTypes.or(leftOperand.type(), rightOperand.type());
     }
+    if (is(Kind.PLUS)) {
+      if (leftOperand.type() == INT && rightOperand.type() == INT) {
```
pynicolas: `==` should be replaced by `.equals(...)`
Same thing a few lines below.

PR #601, opened 2020-02-28
Cleanup old type api

andrea-guarino-sonarsource:


python-checks/src/test/resources/checks/incorrectExceptionType/incorrectExceptionType.py
```diff
@@ -122,7 +122,7 @@ def raise_with_type_inference():
 def fn_type_inference():
```
guillaume-dequenne-sonarsource: Maybe you can rename the test now that the FN is fixed!

andrea-guarino-sonarsource: good point, I missed that!

python-frontend/src/test/java/org/sonar/python/types/UnionTypeTest.java
```diff
@@ -110,4 +110,15 @@ public void test_canOnlyBe() {
...
   }
+
+  @Test
+  public void test_canBeOrExtend() {
+    assertThat(or(a, b).canBeOrExtend("a")).isTrue();
+
+    ClassSymbolImpl x1 = new ClassSymbolImpl("x1", "x1");
+    ClassSymbolImpl x2 = new ClassSymbolImpl("x2", "x2");
+    x2.addSuperClass(x1);
+    assertThat(or(a, new RuntimeType(x2)).canBeOrExtend("x1")).isTrue();
```
pynicolas: Let's have at least one case where `canBeOrExtend` returns `false`.

python-checks/src/main/java/org/sonar/python/checks/IncorrectExceptionTypeCheck.java
```diff
@@ -42,26 +43,27 @@ public void initialize(Context context) {
...
-        Symbol symbol = ((Name) raiseStatement.expressions().get(0)).symbol();
-        if (!inheritsFromBaseException(symbol)) {
+      if (raisedExpression.is(Tree.Kind.NAME)) {
+        Symbol symbol = ((Name) raisedExpression).symbol();
+        if (!inheritsFromBaseException(symbol, raisedExpression)) {
           ctx.addIssue(raiseStatement, MESSAGE);
         }
       }
-      if (raiseStatement.expressions().get(0).is(Tree.Kind.STRING_LITERAL)) {
+      if (raisedExpression.is(Tree.Kind.STRING_LITERAL)) {
```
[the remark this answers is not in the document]
andrea-guarino-sonarsource: As discussed, unfortunately this is not possible (yet) because we don't have a precise modelization of types imported from typeshed: "str" class symbol has an unresolved type hierarchy.
I suggest to keep checking the tree kind for the time being, and come back to this implementation once we will improve typeshed support

PR #602, opened 2020-03-02
SONARPY-570 Exception and BaseException should not be raised

pynicolas:


python-checks/src/main/java/org/sonar/python/checks/GenericExceptionRaisedCheck.java
```diff
@@ -0,0 +1,64 @@
...
+      if (GENERIC_EXCEPTION_NAMES.stream().anyMatch(type::canOnlyBe) || isGenericExceptionClass(expression)) {
+        ctx.addIssue(expression, "Replace this generic exception class with a more specific one.");
+      }
+    });
+  }
+
+  private static boolean isGenericExceptionClass(Expression expression) {
+    if (expression.is(Kind.NAME)) {
+      Symbol symbol = ((Name) expression).symbol();
+      return symbol != null && GENERIC_EXCEPTION_NAMES.contains(symbol.fullyQualifiedName());
```
guillaume-dequenne-sonarsource: Shouldn't it be equals here?

What if you wildcard import a module which redefines BaseException? Shouldn't you be able to raise a Name which has a FQN such as `importedModule.BaseException` and thus return true here?

Note that I can't produce this behaviour locally (importedModule.BaseException is not imported as an external symbol) so either there is something I don't understand or there might be another problem with wildcard imports, but I thought it was worth sharing.

pynicolas: `GENERIC_EXCEPTION_NAMES` is defined as a `Set<String>`, not as a `String`. So `contains` means that `symbol.fullyQualifiedName()` should be equal to one of the 2 strings contained in `GENERIC_EXCEPTION_NAMES`. Does that answer your concern?

guillaume-dequenne-sonarsource: Indeed, my bad! I somehow completely missed that...

PR #604, opened 2020-03-02
SONARPY-567 Rule S5706: Special method "__exit__" should not re-raise…

gyula-sallai-sonarsource:
… the provided exception


python-checks/src/main/java/org/sonar/python/checks/NoReRaiseInExitCheck.java
```diff
@@ -0,0 +1,137 @@
...
+    Name name = parameter.name();
+    if (name != null) {
+      return name.symbol();
+    }
+
+    return null;
+  }
+
+  private static Symbol extractCaughtExceptionParameter(ParameterList parameterList) {
+    if (parameterList.nonTuple().size() != 4) {
```
pynicolas: I believe I can create a case with some Python 2 tuple parameters which make that test not quite correct. However, I think it's not worth trying to cover such corner cases.

its/ruling/src/test/resources/expected/python-S5706.json
```diff
@@ -0,0 +1,6 @@
+{
+'project:S5706.py':[
```
pynicolas: We usually don't do that anymore. I guess it was a good exercise in this case :)

PR #605, opened 2020-03-03
SONARPY-562 Rule S5747: Bare raise statements should only be used in except blocks

andrea-guarino-sonarsource:


its/ruling/src/test/resources/expected/python-S1039.json
```diff
@@ -0,0 +1,17 @@
...
+'project:twisted-12.1.0/twisted/conch/scripts/ckeygen.py':[
+76,
+],
+'project:twisted-12.1.0/twisted/conch/scripts/conch.py':[
+153,
+],
+'project:twisted-12.1.0/twisted/conch/scripts/tkconch.py':[
+337,
+],
+}
```
guillaume-dequenne-sonarsource: Given the names of the methods, I believe most (all?) of those issues are FPs ?
Those methods seem to be called in an `except` context for which we don't have the symbol usages (hence why we miss them?).
Do we plan to do something about those kind of FPs or just accept them and monitor the behavior on more projects?

andrea-guarino-sonarsource: For the record, we finally decided to create a new dedicated _code smell_ rule.
Regarding symbol usages, we miss usages on `self` (e.g. `self.self.process_exception_by_middleware`). Ticket created for this: https://jira.sonarsource.com/browse/SONARPY-614

PR #606, opened 2020-03-03
SONARPY-563 Rule S5704: Bare raise should not be used in finally blocks

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/BareRaiseInFinallyCheck.java
```diff
@@ -0,0 +1,64 @@
...
+    });
+  }
+
+  private static boolean isWithinExceptClause(RaiseStatement raiseStatement) {
+    Tree tree = TreeUtils.firstAncestorOfKind(raiseStatement, Tree.Kind.EXCEPT_CLAUSE, Tree.Kind.CLASSDEF, Tree.Kind.FUNCDEF);
+    return tree != null && tree.is(Tree.Kind.EXCEPT_CLAUSE);
+  }
+
+  private static boolean isWithinExitFunction(RaiseStatement finallyClause) {
+    Tree functionParent = TreeUtils.firstAncestorOfKind(finallyClause, Tree.Kind.FUNCDEF);
```
andrea-guarino-sonarsource: I think you may use something like:

```java
TreeUtils.firstAncestor(finallyClause, t -> t.is(Tree.Kind.FUNCDEF) 
&& ((FunctionDef) t).name().name().equals("__exit__"))
```
If you feel like doing it, I think it would be nice to add a utility method in `TreeUtils` to check function name.

```java
static boolean isFunctionWithName(Tree tree, String name) {
   return tree.is(FUNCDEF) && ((FuncDef) tree).name().name().equals(name);
}
```
What do you think?

guillaume-dequenne-sonarsource: Thanks for the suggestion! 
Regarding adding a `TreeUtils` method, do you see other places where we might want to use it? I agree it might be useful but I wouldn't do it now if it's the only place I'd use it.

andrea-guarino-sonarsource: Yeah, I see a few cases in Checks:

https://github.com/SonarSource/sonar-python/blob/cd89726e20ac7d53b475639e4f096ea8679495b4/python-checks/src/main/java/org/sonar/python/checks/InitReturnsValueCheck.java#L44

https://github.com/SonarSource/sonar-python/blob/cd89726e20ac7d53b475639e4f096ea8679495b4/python-checks/src/main/java/org/sonar/python/checks/NoReRaiseInExitCheck.java#L122

https://github.com/SonarSource/sonar-python/blob/cd89726e20ac7d53b475639e4f096ea8679495b4/python-checks/src/main/java/org/sonar/python/checks/RaiseOutsideExceptCheck.java#L56

guillaume-dequenne-sonarsource: Thanks! Seeing this it makes a lot of sense to refactor indeed :)

PR #613, opened 2020-03-05
SONARPY-575 S5707: Exceptions __cause__ should be an Exception or None

pynicolas:


python-checks/src/test/resources/checks/exceptionCauseType.py
```diff
@@ -0,0 +1,43 @@
...
+def assign_cause_to_none(e):
+    e.__cause__ = None
+
+def assign_cause_to_unknown(e):
+    e.__cause__ = Unknown()
+
+def other_assignments(e):
+    e.__eq__ = NotAnException()
+    e = NotAnException()
+    e.__cause__ = f = NotAnException() # Noncompliant
```
guillaume-dequenne-sonarsource: What about having an additional test such as:
```
def exception_tuple():
  try:
    raise ValueError()
  except (ValueError, TypeError) as e:
    raise MyException() from e
```
As I guess it would be a bit more challenging on our type inference? Currently it seems `e` resolves as "ANY" type, but we might want to ensure the behaviour is still correct when it resolves as an union type of ValueError, TypeError (as I believe it should?).

pynicolas: Done in 425e643.

PR #615, opened 2020-03-05
SONARPY-568 Rule S5708: Caught Exceptions must derive from BaseException

andrea-guarino-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/CaughtExceptionsCheck.java
```diff
@@ -0,0 +1,70 @@
...
+    context.registerSyntaxNodeConsumer(EXCEPT_CLAUSE, ctx -> {
+      Expression exception = ((ExceptClause) ctx.syntaxNode()).exception();
+      if (exception == null) {
+        return;
+      }
+      TreeUtils.flattenTuples(exception).forEach(expression -> {
+        if (NON_COMPLIANT_TYPES.stream().anyMatch(type -> expression.type().canOnlyBe(type))) {
+          ctx.addIssue(expression, MESSAGE);
+        }
+        if ((expression instanceof HasSymbol) && !inheritsFromBaseException(((HasSymbol) expression).symbol())) {
```
pynicolas: I would use an `else if` to be sure we never raise 2 issues.

python-checks/src/main/java/org/sonar/python/checks/CaughtExceptionsCheck.java
```diff
@@ -0,0 +1,70 @@
...
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(EXCEPT_CLAUSE, ctx -> {
+      Expression exception = ((ExceptClause) ctx.syntaxNode()).exception();
+      if (exception == null) {
+        return;
+      }
+      TreeUtils.flattenTuples(exception).forEach(expression -> {
+        if (NON_COMPLIANT_TYPES.stream().anyMatch(type -> expression.type().canOnlyBe(type))) {
```
pynicolas: Why not the following?
```
if (!expression.type().canBeOrExtend("BaseException"))
```
It should allow to catch other cases:
```
except 42
except range(42)
```

andrea-guarino-sonarsource: Good point! Thanks!

PR #620, opened 2020-03-09
SONARPY-532 Support parsing of assignment expressions

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/plugins/python/api/tree/AssignementExpression.java
```diff
@@ -0,0 +1,29 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.plugins.python.api.tree;
+
+public interface AssignementExpression extends Expression {
```
pynicolas: * Let's add a bit of javadoc the same way we did it for other Tree interfaces.
* Rename `AssignementExpression` to `AssignmentExpression`

python-frontend/src/main/java/org/sonar/plugins/python/api/tree/BaseTreeVisitor.java
```diff
@@ -259,6 +259,12 @@ public void visitAssignmentStatement(AssignmentStatement pyAssignmentStatementTr
     scan(pyAssignmentStatementTree.assignedValue());
   }
 
+  @Override
+  public void visitAssignmentExpression(AssignementExpression assignementExpression) {
```
pynicolas: Rename `assignementExpression` to `assignmentExpression`.

python-frontend/src/main/java/org/sonar/plugins/python/api/tree/AssignementExpression.java
```diff
@@ -0,0 +1,29 @@
...
+
+  Name name();
+
+  Token walrusOperator();
```
pynicolas: I would name it only `operator`.

python-frontend/src/main/java/org/sonar/plugins/python/api/tree/AssignementExpression.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.plugins.python.api.tree;
+
+public interface AssignementExpression extends Expression {
+
+  Name name();
```
pynicolas: I would name it `lhsName`, the same way we named `lhsExpressions` in `AssignmentStatement`.

python-frontend/src/main/java/org/sonar/python/tree/PythonTreeMaker.java
```diff
@@ -871,6 +874,15 @@ public Expression expression(AstNode astNode) {
     throw new IllegalStateException("Expression " + astNode.getType() + " not correctly translated to strongly typed AST");
   }
 
+  private Expression assignmentExpression(AstNode astNode) {
+    AstNode nameNode = astNode.getFirstChild(PythonGrammar.TEST);
+    Name name = name(nameNode.getFirstDescendant(PythonGrammar.NAME));
```
pynicolas: It's very suspicious to use `getFirstDescendant` in `PythonTreeMaker`.
My understanding is that the official grammar allows `namedexpr_test: test [':=' test]` but that's not consistent with [PEP-572](https://www.python.org/dev/peps/pep-0572/) which seems to indicate that only a name is allowed on the left hand side.
I would throw a `RecognitionException` if `!expression(nameNode).is(Kind.NAME)`.

python-frontend/src/main/java/org/sonar/python/tree/PythonTreeMaker.java
```diff
@@ -871,6 +874,15 @@ public Expression expression(AstNode astNode) {
...
+    AstNode walrusNode = nameNode.getNextSibling();
```
pynicolas: Rather than using `getNextSibling`, I would use `astNode.getFirstChild(PythonPunctuator.WALRUS_OPERATOR)`.

python-frontend/src/main/java/org/sonar/python/tree/PythonTreeMaker.java
```diff
@@ -871,6 +874,15 @@ public Expression expression(AstNode astNode) {
...
+    Token walrusToken = toPyToken(walrusNode.getToken());
+    Expression expression = expression(walrusNode.getNextSibling());
```
pynicolas: I would write `expression(astNode.getLastChild(PythonGrammar.TEST))`.

python-frontend/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
```diff
@@ -33,11 +33,13 @@
 import java.util.Map;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import org.junit.Ignore;
```
pynicolas: Why?

PR #623, opened 2020-03-09
SONARPY-588: Rule 5722: Special methods should have an expected numbe…

gyula-sallai-sonarsource:
…r of parameters


python-checks/src/main/java/org/sonar/python/checks/ArgumentNumberCheck.java
```diff
@@ -92,20 +91,8 @@ private static boolean isException(CallExpression callExpression, FunctionSymbol
...
-    return false;
-  }
-
-  private static void addSuperClasses(Symbol symbol, Set<Symbol> set) {
-    if (set.add(symbol) && symbol instanceof ClassSymbol) {
-      for (Symbol superClass : ((ClassSymbol) symbol).superClasses()) {
-        addSuperClasses(superClass, set);
-      }
-    }
+    return SymbolUtils.findAllSuperClasses(symbol).stream().anyMatch(
```
andrea-guarino-sonarsource: I think it's better to use `symbol.isOrExtends("zope.interface.Interface")` here. So I think we'll have to change `isOrExtends` implementation to make this work. I can help you with that

PR #624, opened 2020-03-10
SONARPY-621 Update symbol table to handle assignment expressions

guillaume-dequenne-sonarsource:


python-checks/src/test/resources/checks/invariantReturn.py
```diff
@@ -450,3 +450,12 @@ def f_a_lot_of_binding(a):
...
     return d
+
+def with_assignment_expression(a, b): # Noncompliant
+#   ^^^^^^^^^^^^^^^^^^^^^^^^^^
+  if (b := 12) == a:
+     return b
+#    ^^^^^^^^<
+  else:
+     return b
+#    ^^^^^^^^<
```
pynicolas: I would add a case where we shouldn't raise a FP, something like:
```
def no_fp_on_assignment_expression(a, b):
    if a:
        a = (b := 42)
        return b
    else:
        return b
```

PR #626, opened 2020-03-11
SONARPY-532 Fix parsing error when assignment expressions are used as arguments

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/api/PythonGrammar.java
```diff
@@ -240,6 +240,7 @@ public static void grammar(LexerfulGrammarBuilder b) {
     b.rule(ARGUMENT).is(b.firstOf(
       b.sequence("*", TEST),
       b.sequence("**", TEST),
+      b.sequence(TEST, ":=", TEST),
```
andrea-guarino-sonarsource: What about using `PythonPunctuator.WALRUS_OPERATOR` for consistency?

python-frontend/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
```diff
@@ -1195,6 +1195,11 @@ public void assignement_expression() {
 
     assertThat(assignmentExpression.children()).containsExactly(name, walrus, walrusExpression);
 
+    setRootRule(PythonGrammar.EXPR);
+    astNode = p.parse("foo(a:=42)");
+    expression = treeMaker.expression(astNode);
+    assertThat(expression.is(Kind.CALL_EXPR)).isTrue();
+
```
andrea-guarino-sonarsource: I think we could add an assertion regarding the first argument being an assignment expression.

PR #627, opened 2020-03-11
SONARPY 585 Rule S5720: self should be the first argument to instance methods

gyula-sallai-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/InstanceMethodSelfAsFirstCheck.java
```diff
@@ -0,0 +1,197 @@
...
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.CLASSDEF, ctx -> {
+      ClassDef classDef = (ClassDef) ctx.syntaxNode();
+      if (!isRelevantClass(classDef)) {
+        return;
+      }
+
+      List<FunctionDef> functions = classDef.body().statements().stream()
```
andrea-guarino-sonarsource: Beware that you may lose some methods. In Python you can do something like:
```python
class A:
  if p:
    def f(self): ...
```
I'd suggest using a visitor instead, or going through symbols `ClassSymbols.declaredMembers()`

its/ruling/src/test/resources/expected/python-S5720.json
```diff
@@ -0,0 +1,67 @@
+{
+'project:django-2.2.3/django/db/utils.py':[
+247,
+],
+'project:django-2.2.3/django/template/base.py':[
+705,
```
andrea-guarino-sonarsource: it looks like we should also consider "old style" `staticmethod`
```python
    def args_check(name, func, provided):  ...
    args_check = staticmethod(args_check)
```

its/ruling/src/test/resources/expected/python-S5720.json
```diff
@@ -0,0 +1,67 @@
+{
+'project:django-2.2.3/django/db/utils.py':[
+247,
```
andrea-guarino-sonarsource: this one looks like FP. It's a "private" method. I think we could exclude all methods starting with `_`. @nicolas-harraudeau-sonarsource what do you think?

```
def _router_func(action): ...

...

db_for_read = _router_func('db_for_read')
db_for_write = _router_func('db_for_write')
```

python-checks/src/main/java/org/sonar/python/checks/InstanceMethodSelfAsFirstCheck.java
```diff
@@ -0,0 +1,197 @@
...
+  private static final List<String> EXCEPTIONS = Arrays.asList("__init_subclass__", "__class_getitem__", "__new__");
+
+  private static final String DEFAULT_IGNORED_DECORATORS = "classproperty";
+  private List<String> decoratorsToExclude;
+
+  @RuleProperty(
+    key = "ignoredDecorators",
+    description = "Comma-separated list of decorators which will disable this rule.",
+    defaultValue = DEFAULT_IGNORED_DECORATORS)
+  private String ignoredDecorators = DEFAULT_IGNORED_DECORATORS;
```
andrea-guarino-sonarsource: I'm not sure about having this property.
I'd suggest ignoring all methods having a decorator. @nicolas-harraudeau-sonarsource wdyt?

python-checks/src/main/java/org/sonar/python/checks/InstanceMethodSelfAsFirstCheck.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+  }
+
+  private static boolean isRelevantClass(ClassDef classDef) {
+    ClassSymbol classSymbol = TreeUtils.getClassSymbolFromDef(classDef);
+    if (classSymbol == null) {
+      return false;
+    }
+
+    // Do not raise on nested classes
```
andrea-guarino-sonarsource: Why do we ignore nested classes ? Remember it's important explaining "why" instead of "what" in comments

python-checks/src/main/java/org/sonar/python/checks/InstanceMethodSelfAsFirstCheck.java
```diff
@@ -0,0 +1,197 @@
...
+      if (!isRelevantClass(classDef)) {
+        return;
+      }
+
+      List<FunctionDef> functions = classDef.body().statements().stream()
+        .filter(s -> s.is(Tree.Kind.FUNCDEF))
+        .map(s -> (FunctionDef) s)
+        .collect(Collectors.toList());
+
+      for (FunctionDef functionDef : functions) {
```
andrea-guarino-sonarsource: I think you can avoid collecting to List at line 190 and do a "foreach" instead.

python-checks/src/main/java/org/sonar/python/checks/InstanceMethodSelfAsFirstCheck.java
```diff
@@ -0,0 +1,197 @@
...
+          }
+        }
+
+        return false;
+      });
+    }
+  }
+
+  private static boolean isUsedAsOldStyleClassMethod(ClassDef parent, Symbol symbol) {
+    OldStyleClassMethodVisitor visitor = new OldStyleClassMethodVisitor(symbol);
```
andrea-guarino-sonarsource: One alternative way of knowing this would be to go through `Symbol.usages`

```java
symbol.usages().stream()
  .anyMatch(usage -> isDeclaredAsClassMethod(usage.tree))

private static boolean isDeclaredAsClassMethod(Tree tree) {
 // check if parent is call expression called "classmethod"
}
```
However, I'm fine with visitor implementation as well, I let you choose what you prefer :)

python-checks/src/main/java/org/sonar/python/checks/InstanceMethodSelfAsFirstCheck.java
nicolas-harraudeau-sonarsource: @andrea-guarino-sonarsource I was the one asking for this property.
On ITs we saw only one decorator: `@classproperty` from django. After some rapid research I saw that [some other project had an equivalent decorator](https://docs.astropy.org/en/stable/api/astropy.utils.decorators.classproperty.html) with the same name.

Having decorators on methods [is quite common](https://sourcegraph.com/search?q=%22%40%22+%22%28%22+%22%5Cn%22+%22def%22+%22%28self%22+lang:python+count:1000&patternType=regexp). Having a decorator which removes "self" or replaces it with something else should be quite rare. I don't expect devs to add more than 1 or 2 decorators, if they add any. Could we merge this and check on peach? If we see more than 2 decorators doing this we can reevaluate.

andrea-guarino-sonarsource: ok, let's do that

its/ruling/src/test/resources/expected/python-S5720.json
nicolas-harraudeau-sonarsource: Given how often there are "private" methods in python this would reduce the value of this rule quite a lot. Functions such as this one cannot be called outside of the class. I suggest to avoid raising on any function which is called in the body of the class (i.e. not in another method).
example:
``` python
In [2]: class A:
   ...:     def instance(self):
   ...:         print(self)
   ...:     def strange():   # Ok
   ...:         print("strange")
   ...:         return 42
   ...:     foo = strange()  # called here
   ...:
   ...:
strange

In [6]: a = A()

In [8]: a.strange()  # "strange" cannot be called on an instance of A
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-8-506a4d801479> in <module>
----> 1 a.strange()

TypeError: strange() takes 0 positional arguments but 1 was given
```

andrea-guarino-sonarsource: ok, fine for me

python-checks/src/main/java/org/sonar/python/checks/CheckUtils.java
```diff
@@ -21,8 +21,8 @@
 
 import java.util.List;
 import javax.annotation.Nullable;
-import org.sonar.plugins.python.api.tree.Argument;
 import org.sonar.plugins.python.api.tree.ArgList;
+import org.sonar.plugins.python.api.tree.Argument;
```
andrea-guarino-sonarsource: Unnecessary change?

PR #631, opened 2020-03-11
SONARPY-542 Validate behavior of S4784 and S2077 with assignment expr…

guillaume-dequenne-sonarsource:
…essions

### DIFF
--- a/python-checks/src/test/resources/checks/hotspots/regex.py
+++ b/python-checks/src/test/resources/checks/hotspots/regex.py
@@ -62,3 +62,6 @@ def test_valid_numpy_version():
         res = re.match(version_pattern, np.__version__) # Noncompliant
     else:
         res = re.match(version_pattern + dev_suffix, np.__version__) # Noncompliant
+
+def formatted_regex():
+  pattern = re.compile(r'{}-\d+-{}$'.format(foo, bar))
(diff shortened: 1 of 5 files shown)


python-checks/src/test/resources/checks/hotspots/sqlQuery.py
```diff
@@ -19,11 +19,14 @@ def query_my_user(request, params):
...
         MyUser.objects.raw(f"SELECT * FROM mytable WHERE name = '{value}'")  # Noncompliant
         MyUser.objects.raw(F"SELECT * FROM mytable WHERE name = '{value}'")  # Noncompliant
+        MyUser.objects.raw(x := F"SELECT * FROM mytable WHERE name = '{value}'")  # Noncompliant
         MyUser.objects.raw(request)  # OK
         MyUser.objects.raw(hardcoded_request)  # OK
         MyUser.objects.raw(formatted_request)  #  Noncompliant [[secondary=-14]]
         MyUser.objects.raw(formatted_request2)  # Noncompliant
         MyUser.objects.raw(formatted_request3)  # Noncompliant
+        MyUser.objects.raw(y := formatted_request3)  # Noncompliant
+        MyUser.objects.raw((y := formatted_request3))  # Noncompliant
```
pynicolas: I would add a compliant case which uses an assignment expression.

python-checks/src/test/resources/checks/hotspots/regex.py
```diff
@@ -16,7 +16,9 @@ def build_validator(regex):
 RegexValidator('(a{1})+') # Noncompliant
 RegexValidator('(a+)+') # Noncompliant
 RegexValidator('(a{1}){2}') # Noncompliant
-
+RegexValidator(x:='(a+)+') # Noncompliant
+RegexValidator((x:='(a+)+')) # Noncompliant
```
pynicolas: I would add a compliant case which uses an assignment expression.

PR #633, opened 2020-03-12
SONARPY-543 Rule S5685: Walrus operator should not make code confusing

guillaume-dequenne-sonarsource:


python-checks/src/test/resources/checks/confusingWalrus.py
```diff
@@ -0,0 +1,36 @@
...
+func(value, b=2)
+def func(param=21):
+    p = 21
+
+p = 21
+def func(param: p = 3):
+    pass
+
+x = 10
+f'{x}'
```
pynicolas: I would have put each compliant case next to the closest noncompliant case so that it's easier to understand what's noncompliant.

python-checks/src/test/resources/checks/confusingWalrus.py
```diff
@@ -0,0 +1,36 @@
+# using an assignment expression (:=) as an assignment statement (=) is more explicit
+(v := f(p))  # Noncompliant
+v0 = (v1 := f(p))  # Noncompliant
+
+# using an assignment expression in a function call when keyword arguments are also used.
+func(a=(b := f(p)))  # Noncompliant
```
pynicolas: It would be good to check the different issue messages.

python-checks/src/test/resources/checks/confusingWalrus.py
```diff
@@ -0,0 +1,36 @@
...
+def func(param=(p := 21)):  # Noncompliant
+    pass
+
+# using an assignment expression in an annotation
+def func(param: (p := 21) = 3):  # Noncompliant
+    pass
+
+# using assignment expression in an f-string. Character ":" is also used as a formatting marker in f-strings.
+f'{(x:=10)}'  # Noncompliant
+f'{x:=10}' # No issue raised but still not recommended. This is not an assignment expression. '=10' is passed to the f-string formatter.
```
pynicolas: what about the following?
```
f'{foo(x:=10)}'
```

python-checks/src/test/resources/checks/confusingWalrus.py
```diff
@@ -0,0 +1,36 @@
+# using an assignment expression (:=) as an assignment statement (=) is more explicit
+(v := f(p))  # Noncompliant
+v0 = (v1 := f(p))  # Noncompliant
```
pynicolas: what about the following?
```
x = 42 + (y:=2)
```

python-checks/src/test/resources/checks/confusingWalrus.py
```diff
@@ -0,0 +1,36 @@
...
+
+# using an assignment expression in a function call when keyword arguments are also used.
+func(a=(b := f(p)))  # Noncompliant
+func(a := f(p), b=2)  # Noncompliant
+func(a := f(p), 3) # OK
+func(a := f(p), (3, 4), *args) # OK
```
pynicolas: What about the following?
```
def f(a=(x:=2)+42, b=1): ...
```

python-checks/src/test/resources/checks/confusingWalrus.py
guillaume-dequenne-sonarsource: After discussing it again with @nicolas-harraudeau-sonarsource, we will not raise issues in case of assignment expression within larger expression when in an assignment statement context.

python-checks/src/test/resources/checks/confusingWalrus.py
guillaume-dequenne-sonarsource: We will raise issues every time an assignment expression is used within a f-string.

python-checks/src/test/resources/checks/confusingWalrus.py
guillaume-dequenne-sonarsource: We will raise issue every time an assignment expression is used in a function definition where parameter default values are also present, as this can get difficult to read.

python-checks/src/main/java/org/sonar/python/checks/ConfusingWalrusCheck.java
```diff
@@ -0,0 +1,108 @@
...
+      }
+    });
+  }
+
+  private static void checkNestedWalrus(SubscriptionContext ctx, Tree tree, String message) {
+    WalrusVisitor walrusVisitor = new WalrusVisitor();
+    tree.accept(walrusVisitor);
+    if (!walrusVisitor.assignmentExpressions.isEmpty()) {
+      PreciseIssue issue = ctx.addIssue(walrusVisitor.assignmentExpressions.get(0), message);
+      walrusVisitor.assignmentExpressions.stream().skip(1).forEach(a -> issue.secondary(a, null));
```
pynicolas: I can remove this line without breaking unit tests.

pynicolas:
What about the remaining code smell?

PR #635, opened 2020-03-13
Symbol Table: Consider class attributes usages

andrea-guarino-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -173,6 +174,26 @@ private void addSymbolsToTree(FileInputImpl fileInput) {
...
 
+  private static void handleClassMemberAccess(FileInput fileInput) {
+    BaseTreeVisitor classMemberAccess = new BaseTreeVisitor() {
+      @Override
+      public void visitQualifiedExpression(QualifiedExpression qualifiedExpression) {
+        super.visitQualifiedExpression(qualifiedExpression);
+        if (qualifiedExpression.qualifier() instanceof HasSymbol) {
+          Symbol symbol = ((HasSymbol) qualifiedExpression.qualifier()).symbol();
+          if (symbol != null && symbol.kind() == Symbol.Kind.CLASS) {
+            Symbol member = ((ClassSymbolImpl) symbol).membersByName().get(qualifiedExpression.name().name());
```
pynicolas: I don't think that's correct. I believe you should use `resolveMember` rather than exposing `membersByName`.
```
class A:
  def f(): print("A.f")

class B(A):
  pass

B.f()
```

python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -173,6 +174,26 @@ private void addSymbolsToTree(FileInputImpl fileInput) {
...
+            if (member != null) {
+              Optional.ofNullable(qualifiedExpression.name().usage())
+                .ifPresent(usage -> ((SymbolImpl) member).addUsage(qualifiedExpression.name(), usage.kind()));
```
pynicolas: There's something I don't get. Are we relying on the fact that `qualifiedExpression.name()` was associated to a wrong symbol to reassign it to the correct one?

python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -131,6 +131,7 @@ public void visitFileInput(FileInput fileInput) {
     fileInput.accept(new FirstPhaseVisitor());
     fileInput.accept(new SecondPhaseVisitor());
     addSymbolsToTree((FileInputImpl) fileInput);
+    handleClassMemberAccess(fileInput);
```
pynicolas: Why not define a `ThirdPhaseVisitor` (or any better name) and write `fileInput.accept(new ThirdPhaseVisitor())` to for consistency?

python-frontend/src/test/java/org/sonar/python/semantic/ClassSymbolTest.java
```diff
@@ -315,6 +318,35 @@ public void copy_without_usages() {
...
+  @Test
+  public void static_member_usages() {
+    ClassSymbolImpl classSymbol = ((ClassSymbolImpl) lastClassSymbol(
+      "class A:",
+      "  foo = 42",
+      "  def __init__(self): ",
+      "    A.foo",
+      "    A.bar"
+    ));
+    Map<String, Symbol> members = classSymbol.declaredMembers().stream().collect(Collectors.toMap(Symbol::name, Function.identity()));
```
pynicolas: Why not `resolveMember`?

python-frontend/src/test/java/org/sonar/python/semantic/ClassSymbolTest.java
```diff
@@ -315,6 +318,35 @@ public void copy_without_usages() {
...
+      "class A:",
+      "  foo1 = 42",
+      "  def __init__(self, x): ",
+      "    getClass().foo2", // receiver hasn't a symbol
+      "    x.foo3",          // receiver symbol isn't a class
+      "    unknown.foo4"     // receiver symbol unknown
+    ));
+    assertThat(classSymbol.declaredMembers())
+      .filteredOn("kind", Symbol.Kind.OTHER)
+      .extracting(Symbol::name).containsOnly("foo1");
```
pynicolas: The intention of this assertion is not quite clear to me.

python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
andrea-guarino-sonarsource: Good point. My wrong assumption was that static members were not inherited

python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
andrea-guarino-sonarsource: I agree that's not ideal: I'm relying on it to get the correct usage kind. I will change this

python-frontend/src/test/java/org/sonar/python/semantic/ClassSymbolTest.java
andrea-guarino-sonarsource: The intention is asserting that the only static member of `A` is `foo1` and that it doesn't contain `foo2`, `foo3` and `foo4`.
I will try to make that more explicit

python-frontend/src/main/java/org/sonar/python/semantic/ClassSymbolImpl.java
```diff
@@ -126,7 +126,7 @@ public boolean isOrExtends(ClassSymbol other) {
     return allSuperClasses().stream().anyMatch(c -> Objects.equals(c.fullyQualifiedName(), other.fullyQualifiedName()));
   }
 
-  private Map<String, Symbol> membersByName() {
+  Map<String, Symbol> membersByName() {
```
pynicolas: do we still need that change?

andrea-guarino-sonarsource: I forgot to revert it. Thanks!

PR #637, opened 2020-03-16
Symbol Table: Improve resolution of class members accessed through self

andrea-guarino-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -176,6 +170,19 @@ private void addSymbolsToTree(FileInputImpl fileInput) {
     }
   }
 
+  private static Set<Symbol> getClassMembers(Map<String, Symbol> classSymbols, Map<String, SymbolImpl> instanceAttributesByName) {
```
pynicolas: I think that `classSymbols` is a misleading name here. Isn't it more `symbolsInClass`?

PR #638, opened 2020-03-17
SONARPY-626 Fix handling of keyword only args in S930

pynicolas:


nicolas-harraudeau-sonarsource:
Good message improvements!

PR #645, opened 2020-03-18
SONARPY-483 SONARPY-484 Rule S1144 and Rule S3985

andrea-guarino-sonarsource:


python-checks/src/test/resources/checks/unreadPrivateMethods.py
```diff
@@ -0,0 +1,28 @@
+class A:
```
guillaume-dequenne-sonarsource: Maybe we could have a test for an unused method with 1 leading underscore to ensure we don't report any issues on those.

python-checks/src/test/resources/checks/unreadPrivateMethods.py
```diff
@@ -0,0 +1,28 @@
...
+  __foo = 42 # OK, raised by S4487
+
+  def __unused(self): ... # Noncompliant {{Remove this unused class-private '__unused' method.}}
```
guillaume-dequenne-sonarsource: Super nitpicky but is there a reason why we name the method specifically in the message, yet we don't do it for classes in S3985?

andrea-guarino-sonarsource: Good question. @nicolas-harraudeau-sonarsource , wdyt?

andrea-guarino-sonarsource: For the record, I decided to edit S3985 message to include class name.

PR #660, opened 2020-03-26
SONARPY-607 Type inference: include types defined in "typings.pyi"

andrea-guarino-sonarsource:


python-frontend/src/main/java/org/sonar/python/types/InferredTypes.java
```diff
@@ -85,25 +86,48 @@ public static InferredType or(InferredType t1, InferredType t2) {
...
     return InferredTypes.runtimeType(builtinSymbols.get(builtinFqn));
   }
+
+  private static InferredType unionDeclaredType(List<Expression> subscripts, Map<String, Symbol> builtinSymbols) {
+    // subscripts cannot be empty by grammar definition
+    InferredType first = declaredType(subscripts.get(0), builtinSymbols);
+    for (int i = 1; i < subscripts.size(); i++) {
+      first = InferredTypes.or(first, declaredType(subscripts.get(i), builtinSymbols));
+    }
+    return first;
```
guillaume-dequenne-sonarsource: What about renaming this variable `union` or `unionType`, since it's not only the first?

python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
```diff
@@ -80,6 +85,37 @@ public void visitFunctionDef(FunctionDef functionDef) {
     return builtins;
   }
 
+  private static Set<Symbol> typingModuleSymbols() {
```
guillaume-dequenne-sonarsource: Do you think we could have a test similar to those for `Union` & `Optional` which highlight the different behaviours for Python 3 specific symbols, Python 2 specific symbols and overlaps?

andrea-guarino-sonarsource: Done in 11bf0f4bf2fec896f7538e525acd152168006b73
Note that there are no Python2 symbols for typing: they all overlap with Python3 symbols

python-frontend/src/main/java/org/sonar/python/types/InferredTypes.java
andrea-guarino-sonarsource: Done in fe054922d1800f98885b0eb00c3b3f8bc3331c27

PR #663, opened 2020-03-31
SONARPY-618 Introduce Ambiguous Symbol

andrea-guarino-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -132,13 +133,73 @@ public void visitFileInput(FileInput fileInput) {
     scopesByRootTree = new HashMap<>();
     fileInput.accept(new FirstPhaseVisitor());
     fileInput.accept(new SecondPhaseVisitor());
+    createAmbiguousSymbols();
```
pynicolas: I don't like changing symbols at this point. It seems very late. What are the risks that we end up with a graph of symbols which doesn't make sense? Aren't we changing symbols which are referenced by other symbols?

python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
```diff
@@ -66,25 +67,33 @@ private TypeShed() {
...
+    TypeAnnotation returnTypeAnnotation = functionDef.returnTypeAnnotation();
+    if (returnTypeAnnotation == null) {
+      return;
+    }
+    if (symbol.is(Symbol.Kind.FUNCTION)) {
+      FunctionSymbolImpl functionSymbol = (FunctionSymbolImpl) symbol;
+      functionSymbol.setDeclaredReturnType(InferredTypes.declaredType(returnTypeAnnotation, builtins));
+    } else if (symbol.is(Symbol.Kind.AMBIGUOUS)) {
+      AmbiguousSymbolImpl ambiguousSymbol = (AmbiguousSymbolImpl) symbol;
+      setDeclaredReturnType(ambiguousSymbol.getSymbol(functionDef.name()), functionDef);
```
pynicolas: I think we should find a way to do that differently, without needing `AmbiguousSymbolImpl` to store `symbolsByDeclarationTree`.

pynicolas: One idea:
Add a `symbol` method to FunctionDefImpl. This symbol could be different from the symbol attached to the name of the function in the case of an ambiguous symbol.

andrea-guarino-sonarsource: Good idea, I will go for it! Thanks!

python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -132,13 +133,69 @@ public void visitFileInput(FileInput fileInput) {
...
+    }
+  }
+
+  private void createAmbiguousSymbols() {
+    for (Scope scope : scopesByRootTree.values()) {
+      Set<SymbolToUpdate> symbolsToUpdate = new HashSet<>();
+      for (Symbol symbol : scope.symbols()) {
+        if (symbol.kind() == Symbol.Kind.OTHER) {
+          List<Usage> bindingUsages = symbol.usages().stream().filter(Usage::isBindingUsage).collect(Collectors.toList());
+          if (bindingUsages.size() > 1) {
```
pynicolas, line 160: I think that this condition is not sufficient. We end up creating an ambiguous symbol in the following case and I suppose we should not:
```
x = 42
x = 43
```

andrea-guarino-sonarsource: Good point. I'll update the condition to also check the presence of a usage of kind `FUNC_DECLARATION` or `CLASS_DECLARATION`

PR #669, opened 2020-04-06
Fix duplicate symbols when using wildcard import

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/Scope.java
```diff
@@ -81,11 +82,19 @@ void createBuiltinSymbol(String name, Map<String, Symbol> typeShedSymbols) {
...
 
-  void createSymbolsFromWildcardImport(Set<Symbol> importedSymbols, Map<String, Symbol> globalSymbolsByFQN) {
+  void createSymbolsFromWildcardImport(Set<Symbol> importedSymbols, ImportFrom importFrom, Map<String, Symbol> globalSymbolsByFQN) {
     importedSymbols.forEach(symbol -> {
       Symbol importedSymbol = copySymbol(symbol.name(), symbol, globalSymbolsByFQN);
-      symbols.add(importedSymbol);
-      symbolsByName.put(symbol.name(), importedSymbol);
+      if (!isExistingSymbol(importedSymbol.name())) {
+        symbols.add(importedSymbol);
+        symbolsByName.put(symbol.name(), importedSymbol);
```
andrea-guarino-sonarsource, line 90: Shouldn't we add usage also in this case?

guillaume-dequenne-sonarsource: Done in f6611da073d21a1b7edf3d28e126018ffceec27f

PR #673, opened 2020-04-08
SONARPY-616 FN on S5632: raise an issue when inferred type is 'str'

andrea-guarino-sonarsource:


python-frontend/src/main/java/org/sonar/plugins/python/api/symbols/ClassSymbol.java
```diff
@@ -39,4 +39,7 @@
 
   @Beta
   boolean isOrExtends(ClassSymbol other);
+
+  @Beta
+  boolean canBeOrExtend(String fullyQualifiedClassName);
```
pynicolas, line 44: What about using this new method in other places? Are you going to open a new PR for that?

pynicolas:
Good job @andrea-guarino-sonarsource! I think this PR is pushing our symbol model in the right direction.

python-checks/src/main/java/org/sonar/python/checks/IncorrectExceptionTypeCheck.java
```diff
@@ -46,39 +46,35 @@ public void initialize(Context context) {
...
       }
-      if (raisedExpression.is(Tree.Kind.NAME)) {
-        Symbol symbol = ((Name) raisedExpression).symbol();
-        if (!inheritsFromBaseException(symbol, raisedExpression)) {
-          ctx.addIssue(raiseStatement, MESSAGE);
-        }
+      Symbol symbol = null;
+      if (raisedExpression instanceof HasSymbol) {
+        symbol = ((HasSymbol) raisedExpression).symbol();
+      } else if (raisedExpression.is(Tree.Kind.CALL_EXPR)) {
```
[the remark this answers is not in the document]
andrea-guarino-sonarsource: Indeed, that's a mistake. I will remove it, thanks!

python-frontend/src/main/java/org/sonar/plugins/python/api/symbols/ClassSymbol.java
andrea-guarino-sonarsource: Yes, good point. I think we have a few rules that can benefit from this API

python-checks/src/main/java/org/sonar/python/checks/IncorrectExceptionTypeCheck.java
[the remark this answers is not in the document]
andrea-guarino-sonarsource: Looking at ruling results, I realized we may miss issues like the following:
```python
def f():
  raise NotImplemented("foo bar")
```
I'll keep the code like this and I'll update unit test to make it more explicit

python-frontend/src/main/java/org/sonar/python/semantic/SymbolUtils.java
```diff
@@ -106,33 +107,66 @@ public static String fullyQualifiedModuleName(String packageName, String fileNam
...
+  private static boolean isBuiltinTypeshedFile(PythonFile pythonFile) {
+    return isTypeShedFile(pythonFile) && pythonFile.fileName().isEmpty();
+  }
+
+  private static boolean isTypingFile(PythonFile pythonFile) {
+    return isTypeShedFile(pythonFile) && pythonFile.fileName().equals("typing");
+  }
+
+  @CheckForNull
+  private static Symbol getParentSymbol(Argument argument) {
```
pynicolas, line 157: I don't understand what we mean by "parent" here.

andrea-guarino-sonarsource: "parent" is refererring to "parent class (of the current class)". I understand the confusion, I'll rename it "getSymbolFromArgument"

andrea-guarino-sonarsource: Fixed in 85f07023b31df792b45e866765cbcc6eb4e45c09

PR #674, opened 2020-04-08
ClassSymbol imported from project-level symbol table should preserve …

andrea-guarino-sonarsource:
…information about unresolved type hierarchy


python-frontend/src/main/java/org/sonar/python/semantic/ClassSymbolImpl.java
```diff
@@ -39,7 +39,7 @@
 
   private final List<Symbol> superClasses = new ArrayList<>();
   private Set<Symbol> allSuperClasses = null;
-  private boolean hasSuperClassWithoutSymbol = false;
+  boolean hasSuperClassWithoutSymbol = false;
```
guillaume-dequenne-sonarsource, line 42: Why not use a getter instead?

andrea-guarino-sonarsource: Fixed in d7e9fad94f23fd10334360028b4d6a1aa377f8b8

PR #683, opened 2020-04-21
SONARPY-469 Rule S5547: Cipher algorithms should be robust

andrey-tyukin-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/RobustCipherAlgorithmCheck.java
```diff
@@ -0,0 +1,105 @@
...
+
+    // Idea is listed under "Weak Algorithms" in pyca/cryptography documentation
+    // https://cryptography.io/en/latest/hazmat/primitives/symmetric-encryption/\
+    // #cryptography.hazmat.primitives.ciphers.algorithms.IDEA
+    String ideaMessage = "IDEA-cipher is susceptible to attacks when using weak keys";
+
+    // `pycryptodomex`, `pycryptodome`, and `pycrypto` all share the same names of the algorithms,
+    // moreover, `pycryptodome` is drop-in replacement for `pycrypto`, thus they share same name ("Crypto").
+    for (String libraryName : new String[] {"Cryptodome", "Crypto"}) {
+      for (Map.Entry<String, String> e : (Iterable<Map.Entry<String, String>>) Stream.of(
```
andrea-guarino-sonarsource, line 55: I find a bit hard the casting + iterator creation. Could you move it to a variable declaration? Same at line 68

python-checks/src/main/java/org/sonar/python/checks/RobustCipherAlgorithmCheck.java
```diff
@@ -0,0 +1,105 @@
...
+
+// https://jira.sonarsource.com/browse/RSPEC-5547 (general)
+// https://jira.sonarsource.com/browse/RSPEC-5552 (python-specific)
+@Rule(key = "S5547")
+public class RobustCipherAlgorithmCheck extends PythonSubscriptionCheck {
+
+  private static final Map<String, String> sensitiveCalleeFqnsAndMessages = new HashMap<>();
+
+  static {
+    String desMessage = "DES works with 56-bit keys that allow attacks via exhaustive search";
```
andrea-guarino-sonarsource, line 41: Nitpicking: we usually add a `.` at the end of message

python-checks/src/main/java/org/sonar/python/checks/RobustCipherAlgorithmCheck.java
andrey-tyukin-sonarsource: Yes, you're right. That iterator with SAM-casting is not the simplest option. The much simpler `asList` from `util.Arrays` works here just as well, and is much shorter and clearer.

PR #690, opened 2020-04-27
Sonarpy 357

andrey-tyukin-sonarsource:
Please note the changed positions of the underlined issues for `kwargs`-cases in `requests`-tests. It's not on the line where it was specified, but on the line where `verify: False` is set.

Comments and suggestions regarding the `.filter(X.class::isInstance).map(X.class::cast)`-style are welcome.


python-checks/src/test/resources/checks/verifiedSslTlsCertificateCheck.py
```diff
@@ -0,0 +1,282 @@
...
+  # VERIFY_PEER Don't check peer certificate.
+  # VERIFY_FAIL_IF_NO_PEER_CERT Client:Check peer certificate, Server:Check peer certificate and skip if no certificate received.
+  #  If VERIFY_PEER is used, mode can be OR:ed with VERIFY_FAIL_IF_NO_PEER_CERT and VERIFY_CLIENT_ONCE to further control the behaviour.
+
+  ctx1 = SSL.Context(SSL.TLSv1_2_METHOD)
+  ctx1.set_verify(SSL.VERIFY_PEER, verify_callback) # Compliant
+  ctx1.set_verify(SSL.VERIFY_PEER | SSL.VERIFY_FAIL_IF_NO_PEER_CERT, verify_callback) # Compliant
+  ctx1.set_verify(SSL.VERIFY_PEER | SSL.VERIFY_FAIL_IF_NO_PEER_CERT | VERIFY_CLIENT_ONCE, verify_callback) # Compliant
+
+  def verify_callback(connection, x509, errnum, errdepth, ok):
```
andrea-guarino-sonarsource, line 22: I'm not sure we really need to have all this level of detail in unit tests, maybe an empty function would be ok

python-checks/src/test/resources/checks/verifiedSslTlsCertificateCheck.py
```diff
@@ -0,0 +1,282 @@
...
+  #                                                        ^^
+  requests.request('GET', 'https://example.domain', verify=set()) # Noncompliant {{Disabling certificate verification is dangerous.}}
+  #                                                        ^^^^^
+  requests.request('GET', 'https://example.domain', verify=range(0)) # Noncompliant {{Disabling certificate verification is dangerous.}}
+  #                                                        ^^^^^^^^
+  requests.request(verify=False, method='GET', url='https://example.domain') # Noncompliant {{Disabling certificate verification is dangerous.}}
+  #                       ^^^^^
+  kargs1 = {'verify': False} # Noncompliant {{Disabling certificate verification is dangerous.}}
+  #                   ^^^^^
+  requests.request('GET', 'https://example.domain', **kargs1)
```
andrea-guarino-sonarsource, line 103: I believe the primary location of the issue should stay on `kargs1` argument for consistency. We can indeed add a secondary location to the assignment at line `101`

python-checks/src/test/resources/checks/verifiedSslTlsCertificateCheck.py
```diff
@@ -0,0 +1,282 @@
...
+    ctx1 = ssl._create_unverified_context()  # Noncompliant {{Certificate verification is disabled by default, verify_mode should be updated.}}
+    #          ^^^^^^^^^^^^^^^^^^^^^^^^^^
+    r1 = urllib.request.urlopen('https://self-signed.badssl.com/', context=ctx1)
+    print(r1.status, r1.reason)
+
+    print("test1-1")
+    ctx2 = ssl._create_unverified_context()
+    ctx2.verify_mode = ssl.CERT_NONE # Noncompliant {{Disabling certificate verification is dangerous.}}
+    #                      ^^^^^^^^^
+    r1 = urllib.request.urlopen('https://self-signed.badssl.com/', context=ctx2)
```
andrea-guarino-sonarsource, line 172: Same here, I would preferer keeping the primary location to `ctx2` and adding a secondary location

python-checks/src/test/resources/checks/verifiedSslTlsCertificateCheck.py
```diff
@@ -0,0 +1,282 @@
...
+  # Mutably borrowed from
+  # https://raw.githubusercontent.com/SonarSource/security-expected-issues/master/python/rules/vulnerabilities/\
+  # RSPEC-4830%20Server%20certificates%20should%20be%20verified%20during%20SSL%E2%81%84TLS%20connections/\
+  # urllib-test.py
+  import urllib.request
+  import ssl
+  import sys
+
+  try:
+    print("test1-default")
```
andrea-guarino-sonarsource, line 161: do we need all those print staments? maybe they can be replaced by a comment

python-checks/src/test/resources/checks/verifiedSslTlsCertificateCheck.py
```diff
@@ -0,0 +1,282 @@
...
+  kargs = {'method': 'GET', 'url': 'https://example.domain', 'verify': True}
+  requests.request(**kargs) # Compliant
+
+  requests.head(url='https://example.domain') # Compliant
+  requests.get(url='https://example.domain') # Compliant
+  requests.post(url='https://example.domain') # Compliant
+  requests.put(url='https://example.domain') # Compliant
+  requests.patch(url='https://example.domain') # Compliant
+  requests.delete(url='https://example.domain') # Compliant
+  requests.options(url='https://example.domain') # Compliant
```
andrea-guarino-sonarsource, line 138: I don't think you need to be exhaustive here by listing all the API, probably only one would be enough 

```python
requests.head(url='https://example.domain') # Compliant
```

python-checks/src/test/resources/checks/verifiedSslTlsCertificateCheck.py
```diff
@@ -0,0 +1,282 @@
...
+
+    ctxC2 = ssl._create_default_https_context()
+    ctxC2.verify_mode = ssl.THAT_S_NOT_A_VALID_MODE
+
+    ambiguous = "" if 42 * 42 < 1700 else (lambda x: x * x)
+    ctxC3 = ambiguous._create_default_https_context()
+    ctxC4 = notASymbol()
+
+
+  except: # catch *all* exceptions
```
andrea-guarino-sonarsource, line 279: what are we testing here? do we need it for unit tests?

python-frontend/src/main/java/org/sonar/python/semantic/SymbolUtils.java
```diff
@@ -293,6 +294,12 @@ static Path pathOf(PythonFile pythonFile) {
     initialize.setDeclaredReturnType(InferredTypes.runtimeType(ldapObject));
     globalSymbols.put("ldap", new HashSet<>(Collections.singleton(initialize)));
 
+
+    ClassSymbolImpl sslContextClass =
+      classSymbol("Context", "OpenSSL.SSL.Context", SET_VERIFY);
+    SymbolImpl sslSubmodule = moduleSymbol("SSL", "OpenSSL.SSL", sslContextClass);
+    globalSymbols.put("OpenSSL", new HashSet<>(Collections.singletonList(sslSubmodule)));
```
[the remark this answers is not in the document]
andrey-tyukin-sonarsource: True. While we're at it, should I also replace all the other `new HashSet<>(Collections.singleton(...))` in this file?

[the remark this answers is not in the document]
andrea-guarino-sonarsource: I think it's alright, you can leave them as they are

python-checks/src/main/java/org/sonar/python/checks/CheckList.java
```diff
@@ -53,6 +53,7 @@
   private CheckList() {
   }
 
+  @SuppressWarnings("squid:S138")
```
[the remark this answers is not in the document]
andrey-tyukin-sonarsource: I don't think it solves the problem asymptotically: sooner or later, either the number of rules or the number of hotspots will overflow the threshold anyway. I'll just remove the annotation.

python-checks/src/test/resources/checks/verifiedSslTlsCertificateCheck.py
andrey-tyukin-sonarsource: I've added both the `verify: False` and `**kwargs` positions to the `PreciseIssue`. I've left the main issue message closer to the `verify: False` in all cases, so that it's hopefully easier to see the connection between the "verify" variable name and the "certificate verification" phrase in the issue description. Furthermore, one might say, that this is the position that is more likely to require a change by the user. I've additionally added the unpacking `**kwargs` as the secondary issue position, though. Now the connection to the `requests.request(...)` is easier to see, even though it's less likely to be modified by the user. I've also checked it in the UI: it now looks reasonably intuitive to me.

python-checks/src/test/resources/checks/verifiedSslTlsCertificateCheck.py
andrey-tyukin-sonarsource: I went through the test python file again, and made sure that all the issue positions are exactly on the same lines as in the example attached to the specification:
https://github.com/SonarSource/security-expected-issues/blob/master/python/rules/vulnerabilities/RSPEC-4830%20Server%20certificates%20should%20be%20verified%20during%20SSL%E2%81%84TLS%20connections/urllib-test.py
In the specification, it seems that the last position where `verify_mode` is set to `CERT_NONE` is marked as noncompliant. In case of factory method invocations with bad defaults and without subsequent overriding of `verify_mode`, this simply reduces to underlining the factory method itself. This seems consistent.

python-checks/src/test/resources/checks/verifiedSslTlsCertificateCheck.py
andrea-guarino-sonarsource: ok makes sense. thanks for the clarification!

python-checks/src/test/resources/checks/verifiedSslTlsCertificateCheck.py
andrea-guarino-sonarsource: Makes sense. Thanks!

PR #692, opened 2020-04-28
SONARPY-665 S1192/StringLiteralDuplication: Fix FP on type hints

andrea-guarino-sonarsource:


andrea-guarino-sonarsource:
> I'll refrain from discussing whether renaming two out of three variables `a` to `b` and `c` would make the test case look more realistic :)

I think you're right. I renamed them in 1e8c7c890070e325fb94ad41752300a95cb974e9

PR #693, opened 2020-04-29
SONARPY-671 Fix FNs on S930 and S2159 by resolving Python 2/3 specific modules

guillaume-dequenne-sonarsource:


its/ruling/src/test/resources/expected/python-S2159.json
```diff
@@ -1,4 +1,7 @@
 {
+'project:numpy-1.16.4/numpy/distutils/system_info.py':[
+259,
```
andrea-guarino-sonarsource, line 3: nice one!

python-frontend/src/main/java/org/sonar/python/semantic/Scope.java
```diff
@@ -184,8 +185,8 @@ void addModuleSymbol(Name nameTree, @CheckForNull String fullyQualifiedName, Map
   void addImportedSymbol(Name nameTree, @CheckForNull String fullyQualifiedName, String fromModuleName, Map<String, Symbol> globalSymbolsByFQN) {
     String symbolName = nameTree.name();
     Symbol globalSymbol = globalSymbolsByFQN.get(fullyQualifiedName);
-    if (globalSymbol == null && fullyQualifiedName != null && !fromModuleName.equals(fullyQualifiedModuleName)) {
-      globalSymbol = TypeShed.standardLibrarySymbol(fromModuleName, fullyQualifiedName);
+    if (globalSymbol == null && fullyQualifiedName != null && !fromModuleName.equals(fullyQualifiedModuleName) && !isTypeShedFile(pythonFile)) {
```
andrea-guarino-sonarsource, line 188: why for all typeshed files? In some cases we may want to resolve a module from inside typeshed

guillaume-dequenne-sonarsource: At first I only had issues with the "basic modules" we load first (`builtins`, `typings` & `typing_extensions`). But then I faced cyclic dependencies with `sys` and `types` importing each other as well, creating infinite recursions again. 

So I decided to try and stop resolving imports from TypeShed files entirely and didn't get any ruling changes so I stuck with that. 
Most likely this would need revisiting the more info we exploit on those modules imported within TypeShed.

guillaume-dequenne-sonarsource: As discussed, I added a FIXME there to keep in mind we should find a solution to managing those imports while avoiding issues related to cyclic dependencies.

PR #696, opened 2020-04-30
SONARPY-661 Deprecate S5439 in favor of S5247

guillaume-dequenne-sonarsource:


python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S5439.json
```diff
@@ -3,30 +3,10 @@
   "type": "VULNERABILITY",
   "status": "ready",
```
andrea-guarino-sonarsource, line 4: shouldn't it be deprecated?

guillaume-dequenne-sonarsource: Indeed...Thanks!
I trusted Jira status when regenerating metadata and didn't double check, fixed in 86791d7aa73d423c932b6b893762b2663b448879

PR #697, opened 2020-04-30
SONARPY-662 Rule S5542 Encryption algorithms should be used with secu…

andrea-guarino-sonarsource:
…re mode and padding scheme


python-checks/src/main/java/org/sonar/python/checks/SecureModeEncryptionAlgorithmsCheck.java
```diff
@@ -0,0 +1,135 @@
...
+
+    List<String> vulnerableModes = PYCRYPTO_VULNERABLE_MODES_BY_API.getOrDefault(calleeSymbol.fullyQualifiedName(),
+      CRYPTODOMEX_VULNERABLE_MODES_BY_API.get(calleeSymbol.fullyQualifiedName()));
+
+    if (vulnerableModes != null) {
+      List<Argument> arguments = callExpression.arguments();
+      if (arguments.size() < 2) {
+        return;
+      }
+      Argument argument = arguments.get(1);
```
guillaume-dequenne-sonarsource, line 112: From my understanding, the API from PyCryptodome (when imported either as `Crypto` or `Cryptodome`) can also be called with a keyword argument for `mode` (see https://pycryptodome.readthedocs.io/en/latest/src/cipher/des.html for example).
Maybe a call such as `new(mode = ..., key = ...)` would be quite marginal and we don't need to bother to support it, but I just wanted to point it out.

python-checks/src/main/java/org/sonar/python/checks/SecureModeEncryptionAlgorithmsCheck.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+  }
+
+  protected void checkPycaLibrary(SubscriptionContext ctx, CallExpression callExpression, Symbol calleeSymbol) {
+    if (PYCA_ALGORITHM_METHODS.contains(calleeSymbol.fullyQualifiedName())) {
+      List<Argument> arguments = callExpression.arguments();
+      if (arguments.size() < 2) {
+        return;
+      }
+      Argument argument = arguments.get(1);
```
guillaume-dequenne-sonarsource, line 126: Same here, it could be a keyword argument. 
Not sure we need to cover it though, but if you eventually decide to cover it, don't forget to separate `mode` and `padding` for Pyca.

PR #699, opened 2020-05-04
SONARPY-682 Add parser for Python code to plugin API

vilchik-elena:


python-frontend/src/main/java/org/sonar/plugins/python/api/Parser.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.sonar.sslr.api.AstNode;
+import org.sonar.plugins.python.api.tree.FileInput;
+import org.sonar.python.parser.PythonParser;
+import org.sonar.python.tree.PythonTreeMaker;
+
+/**
+ * Use this class to parse Python module
+ */
+public class Parser {
```
vilchik-elena, line 30: @andrea-guarino-sonarsource eventually I decided to not have interface as there is not a lot of value in it, anyway in order to use it you need instantiate the class

PR #701, opened 2020-05-05
SONARPY-357 fix issue reporting for ctx reinitialized multiple times;…

andrey-tyukin-sonarsource:
Fixed issue reporting in examples where ctx is initialized multiple times. This is relevant not only for original tests, but also for real-world cases where `ctx` can be initialized in different branches, with different configurations.


python-checks/src/test/resources/checks/verifiedSslTlsCertificateCheck.py
```diff
@@ -180,3 +180,22 @@ def urllibTests():
   ambiguous = "" if 42 * 42 < 1700 else (lambda x: x * x)
   ctxC3 = ambiguous._create_default_https_context()
   ctxC4 = notASymbol()
+
+def fixupRequestBulletPoint4ShouldNotReportAnything():
```
andrea-guarino-sonarsource, line 184: not very clear to what bullet point you're referring to...can you rename it?

python-checks/src/main/java/org/sonar/python/checks/VerifiedSslTlsCertificateCheck.java
```diff
@@ -309,21 +307,38 @@ private static void urllibCheck(SubscriptionContext subscriptionContext) {
...
-          } else {
-            subscriptionContext.addIssue(vulnTok.token, "Disabling certificate verification is dangerous.");
-          }
+          subscriptionContext.addIssue(vulnTok.token, MESSAGE);
         }
       }));
   }
 
+  /** Finds the next higher line where a binding usage occurs. */
+  private static int findNextAssignmentLine(List<Usage> usages, int firstAssignmentLine) {
```
andrea-guarino-sonarsource, line 320: Remember that you don't have any guarantee about order of execution in usages list. They are sorted purely on syntax.

If you want to take into account order of execution, you should rely on CFG. However I don't think it's worth it for this check and for the FN mentioned in unit tests.
In my opinion, we should accept this FN when there are multiple assignements.

python-checks/src/test/resources/checks/verifiedSslTlsCertificateCheckFNs.py
```diff
@@ -0,0 +1,51 @@
+def fixupRequestBulletPoint3WallOfFalseNegatives():
```
andrea-guarino-sonarsource, line 1: do we really need all this test cases to show the FN? Moreover I wouldn't create a separate file for it

python-checks/src/main/java/org/sonar/python/checks/VerifiedSslTlsCertificateCheck.java
andrea-guarino-sonarsource: After discussion with andrey, this is useful to mitigate FP on peach: https://peach.sonarsource.com/issues?languages=py&open=AXHKAcvKnkgMtVoYvDQs&resolved=false&rules=python%3AS4830
So we decided to accept the solution based on syntax for the time being.

andrey-tyukin-sonarsource: Indeed, the usages aren't even in the same order as they would occur in any syntax tree traversal - the assignments all come first, the "other".subscriptions all come later. Added sorting to selected relevant usages, just in case.

PR #705, opened 2020-05-06
SONARPY-667 Rule S905: Non-empty statements sould change control flow…

guillaume-dequenne-sonarsource:
… or have side effects


python-checks/src/test/resources/checks/uselessStatement.py
```diff
@@ -0,0 +1,106 @@
...
+    MyClass.class_method  # Noncompliant
+    MyClass.static_method  # Noncompliant
+    MyClass.__eq__  # FN
+    instance = MyClass()
+    instance.attr # FN
+    instance.my_property # OK
+
+def corner_cases():
+    try:
+        MyClass.attr # No issues when within a try/except
```
andrea-guarino-sonarsource, line 96: Isn't this one a FN anyway? I think we should have an example with a statement where we would raise an issue, but we ignore it because it's inside a try/except

PR #707, opened 2020-05-07
SONARPY-183 Count module-level docstrings as comment, not code.

hugobuddel:
Closes SONARPY-183.

Module-level docstrings should be counted as comments.

The `if (tree.is(Tree.Kind.FILE_INPUT))` case in  `FileLinesVisitor` never happened. Module-level docstrings were therefore seen as code instead of comments. As a result, a long module-level docstring required extra comments elsewhere in the file to maintain the right comment density.

This is fixed by adding `Tree.Kind.FILE_INPUT` to `EXECUTABLE_LINES`. Test case is updated to prevent regression.

PR #708, opened 2020-05-07
SONARPY-684 Count module-level docstrings as comments

guillaume-dequenne-sonarsource:
Duplicates #707 to trigger all checks.

### DIFF
--- a/python-frontend/src/main/java/org/sonar/python/metrics/FileLinesVisitor.java
+++ b/python-frontend/src/main/java/org/sonar/python/metrics/FileLinesVisitor.java
@@ -50,7 +50,7 @@ public class FileLinesVisitor extends PythonSubscriptionCheck {
    * Tree.Kind.ELSE_CLAUSE is not in this list to avoid counting else: lines as executables.
    * This is to replicate behavior of some python coverage tools (like what is done by coveralls).
    */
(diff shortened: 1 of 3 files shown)


hugobuddel:
Thank you for looking into this @guillaume-dequenne-sonarsource 

By the way, I tried to get this line to have full coverage:
```
} else if (!isBlank(commentLine) && !noSonar.contains(line)) {
```
but I failed. It seems that `!noSonar.contains(line)` is always true. And I can't see how it can be otherwise, because the line is added to noSonar in the other part of the if clause.

guillaume-dequenne-sonarsource:
You're right, I don't think `!noSonar.contains(line)` could ever be false there. 

I also think it is not be possible for the `commentLines` array to contain more than one element (since there are no true multi-line comments in Python), so I could simplify the `visitComment` method quite a bit in 1e735e42c4fcbcc080e5e4a4f732aba4b94776d0.  Can you do a quick review of that @andrea-guarino-sonarsource ?

andrea-guarino-sonarsource:
@guillaume-dequenne-sonarsource LGTM for 1e735e4

hugobuddel:
Thanks for the quick response @guillaume-dequenne-sonarsource and @andrea-guarino-sonarsource! This cuts my issue count in half ✂️.

PR #710, opened 2020-05-08
SONARPY-673 Rule 5795: identity checks for literals

andrey-tyukin-sonarsource:
(for literals generating values of cached types, more specifically)


andrey-tyukin-sonarsource:
My assumptions about the requirements were not entirely correct, it still requires some changes.

andrey-tyukin-sonarsource:
Got some additional input about the right interpretation of the specification, made sure that all tests (also those with variables and escaping references) pass as in /expected-issues.

python-checks/src/main/java/org/sonar/python/checks/IdentityComparisonWithCachedTypesCheck.java
```diff
@@ -0,0 +1,131 @@
...
+        // Impossible to cover. If the type could be inferred, then the `symb` cannot be null.
+        return symb != null && !isVariableThatCanEscape(symb);
+      } else {
+        return true;
+      }
+    }
+    if (expr.is(Tree.Kind.STRING_LITERAL)) {
+      return true;
+    } else if (expr.is(Tree.Kind.CALL_EXPR)) {
+      Symbol calleeSymbol = ((CallExpression) expr).calleeSymbol();
```
andrea-guarino-sonarsource, line 97: As discussed, I think this deserves more explanation about why don't we just rely on types.

python-checks/src/test/resources/checks/identityComparisonWithCachedTypesExpected.py
```diff
@@ -0,0 +1,67 @@
+# Mutably borrowed from `expected-issues/python/src/RSPEC_5795`.
```
andrea-guarino-sonarsource, line 1: As discussed, I believe we should merge both test files into one

python-checks/src/test/resources/checks/identityComparisonWithCachedTypes.py
```diff
@@ -0,0 +1,153 @@
...
+  ambiguous is unknown
+  f(x) is g(y)
+  f.g(h) is c.d(e)
+  f is g
+
+
+symbollessName is somethingElse # for `null`-Symbols on names
+
+
+# rest is mutably borrowed from `expected-issues/python/src/RSPEC_5795`.
```
andrea-guarino-sonarsource, line 87: I think most of those tests here below are duplicated (for example line 90 we're testing the same as line 2, line 14 and line 96 tests the same thing).
I believe it's important to keep only relevant unit tests for maintenability (less is more!).

andrey-tyukin-sonarsource: Made sure that `sort ... | uniq` doesn't show any duplicates. Removed distinction between string literals in single and double quotes.

PR #712, opened 2020-05-08
SONARPY-663 Rule S5780: Expressions creating dictionaries should not …

guillaume-dequenne-sonarsource:
…have duplicate keys


python-checks/src/main/java/org/sonar/python/checks/DictionaryDuplicateKeyCheck.java
```diff
@@ -0,0 +1,132 @@
...
+
+@Rule(key = "S5780")
+public class DictionaryDuplicateKeyCheck extends PythonSubscriptionCheck {
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.DICTIONARY_LITERAL, ctx -> {
+      DictionaryLiteral dictionaryLiteral = (DictionaryLiteral) ctx.syntaxNode();
+      Set<Integer> issueIndexes = new HashSet<>();
+      for (int i = 0; i < dictionaryLiteral.elements().size(); i++) {
```
andrey-tyukin-sonarsource, line 50: The handler in `initialize` in combination with the `for`-loop in
  `findIdenticalKeys` gives quadratic runtime in the number of elements of
  the dictionary.
  That seems a bit surprising for an "obviously O(n)" task.
  If someone decides
  to use a dictionary with few thousand elements, it might already hurt
  the performance, and in larger examples, it might freeze the analyzer altogether.

  Ideally, one might split up the check into two parts:
  First - normalization, i.e. bringing each key into some canonical form,
  with a well-defined hash code and easy comparison. Second - actual
  comparison for equivalence: two keys are equivalent w.r.t Python's `==`
  iff their normalized versions are equal in Java. That would give a
  linear algorithm, and the size of the analyzed dictionary wouldn't matter.

  I see that this whole rule relies on `CheckUtils.areEquivalent`, which does not give any normalized, easily hashable / comparable values, so that might be difficult to fix quickly. But maybe one might add a condition that we, for now, don't inspect dictionaries with more than, say, 100 entries, just to make sure that the analyzer doesn't freeze suddenly once confronted with longer dictionaries.

python-checks/src/main/java/org/sonar/python/checks/DictionaryDuplicateKeyCheck.java
```diff
@@ -0,0 +1,132 @@
...
+      return areEquivalentStringLiterals((StringLiteral) key, (StringLiteral) comparedKey);
+    }
+    if (isANumber(key) && isANumber(comparedKey)) {
+      return areEquivalentNumbers(key, comparedKey);
+    }
+    return !key.is(Tree.Kind.CALL_EXPR) && CheckUtils.areEquivalent(key, comparedKey);
+  }
+
+  private boolean areEquivalentNumbers(Tree key, Tree comparedKey) {
+    return toBigDecimal(key).compareTo(toBigDecimal(comparedKey)) == 0;
```
andrey-tyukin-sonarsource, line 91: Good catch! A really brief comment about the surprising behavior of `.equals` might improve maintainability a tiny bit. Maybe something like "// note that it must be `.compareTo`, `.equals` is insufficient, because it takes scale into account"? Or maybe just link to some page that explains why it has to be `.compareTo`? (Optional)

python-checks/src/main/java/org/sonar/python/checks/DictionaryDuplicateKeyCheck.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  private boolean isSameKey(Expression key, Expression comparedKey) {
+    if (key.is(Tree.Kind.STRING_LITERAL) && comparedKey.is(Tree.Kind.STRING_LITERAL)) {
+      return areEquivalentStringLiterals((StringLiteral) key, (StringLiteral) comparedKey);
+    }
+    if (isANumber(key) && isANumber(comparedKey)) {
+      return areEquivalentNumbers(key, comparedKey);
+    }
+    return !key.is(Tree.Kind.CALL_EXPR) && CheckUtils.areEquivalent(key, comparedKey);
```
andrey-tyukin-sonarsource, line 87: 1. On the CALL_EXPR: excluding only the `CALL_EXPR` might lead to false positives. Consider the following example:

```
class Foo:
      def __init__(self):
        self.counter = 0
      def __getitem__(self, i):
        self.counter = self.counter + 1
        return i + self.counter

    foo = Foo()
    { foo[0]: 'a', foo[0]: 'b' } # that's ok, creates 2-entry dictionary
```
This gives a false positive, because it thinks that `foo[0]` results in the same key every time.
To avoid this, one would have to exclude all constructs that might have side-effects
(randomness, io, network, mutable state etc.) - it's not only function calls. I'm not sure whether it's feasible. Maybe it would be safer to restrict it to fewer cases that are explicitly allowed, instead of excluding CALL_EXPR?

------------

2. On the `CheckUtils.areEquivalent`:

  `{ (0o10, 'a'): 1, (8, "a"): 2 }` gives a false negative. That a pity, because
  there already are much nicer `areEquivalentNumbers` and `areEquivalentStringLiterals`, but they
  are not used by `CheckUtils.areEquivalent`. Would it maybe be possible to implement a recursive
  variant of `CheckUtils.areEquivalent` that works with `areEquivalent{Numbers,StringLiterals}`
  at the leaves?

python-checks/src/main/java/org/sonar/python/checks/DictionaryDuplicateKeyCheck.java
```diff
@@ -0,0 +1,132 @@
...
+  private static boolean areEquivalentStringLiterals(StringLiteral key, StringLiteral comparedKey) {
+    if (key.stringElements().stream().anyMatch(StringElement::isInterpolated) || comparedKey.stringElements().stream().anyMatch(StringElement::isInterpolated)) {
+      return false;
+    }
+    if (key.trimmedQuotesValue().equals(comparedKey.trimmedQuotesValue())) {
+      String keyWithPrefixes = key.stringElements().stream()
+        .map(s -> s.prefix().toLowerCase(Locale.ENGLISH) + s.trimmedQuotesValue()).collect(Collectors.joining());
+      String comparedKeyWithPrefixes = comparedKey.stringElements().stream()
+        .map(s -> s.prefix().toLowerCase(Locale.ENGLISH) + s.trimmedQuotesValue()).collect(Collectors.joining());
+      return keyWithPrefixes.equals(comparedKeyWithPrefixes);
```
andrey-tyukin-sonarsource, line 110: `{ "a" r"b": 'y', r"ab": 'n' }` gives a false negative, because `"arb" != "rab"`, but that's, admittedly, a rather obscure corner case. In order to improve that, a more principled approach for string literal interpretation would be needed (probably out of scope).

python-checks/src/test/resources/checks/dictionaryDuplicateKey.py
```diff
@@ -0,0 +1,71 @@
...
+  {"one": 1, "two": 2, "one": 3}  # Noncompliant {{Change or remove duplicates of this key.}}
+  #^^^^^               ^^^^^<
+  {"one": 1, "two": 2, 'one': 3}  # Noncompliant
+  {"""multi
+  line""": 1, # Noncompliant@-1
+  "two": 2,
+  """multi
+  line""": 3}
+
+  # We only raise if the strings have same values & prefixes, even if they would eventually evaluate to the same value
```
andrey-tyukin-sonarsource, line 12: I hope it has been discussed elsewhere, because examples like `{"one": 1, "two": 2, u"one": 3}` from the expected-issues currently don't pass.

python-checks/src/test/resources/checks/dictionaryDuplicateKey.py
```diff
@@ -0,0 +1,71 @@
...
+  {"one": 1, "two": 2, u"one": 3}  # OK
+  {"one": 1, "two": 2, r"one": 3}  # OK
+  {"on" "e": 1, "two": 2, "o" "ne": 2} # Noncompliant
+  {"on" r"e": 1, "two": 2, "on" r"e": 2} # Noncompliant
+  {"on" r"e": 1, "two": 2, "o" r"ne": 2} # OK
```
andrey-tyukin-sonarsource, line 17: Python interpreter gives `{'one': 2, 'two': 2}` here, so that's a false negative. Maybe it should be marked explicitly as an accepted FN instead of just OK?

python-checks/src/main/java/org/sonar/python/checks/DictionaryDuplicateKeyCheck.java
guillaume-dequenne-sonarsource: That's a good point.
As you said I'm not sure it's a trivial task to normalize each key, especially from the current implementation.
I'm also not sure about the performance impacts in real-life scenarios, as I don't know how common big dictionary literals are. In doubt, a quick fix would be to disable the rule for big dictionaries indeed, but I guess it would only make sense if we actually face performance issues due to this. What's your opinion on this @andrea-guarino-sonarsource ?

python-checks/src/main/java/org/sonar/python/checks/DictionaryDuplicateKeyCheck.java
guillaume-dequenne-sonarsource: Indeed!
In order to fix this we would need to resolve the actual result of the string literal, which is out of scope for this rule. This heuristic was chosen to cover the basic cases of implicit concatenation and prefix usages without risking to raise FPs.

python-checks/src/test/resources/checks/dictionaryDuplicateKey.py
guillaume-dequenne-sonarsource: Good point, I don't think the `expected-issues` were updated, but it was discussed indeed and we decided to use the heuristic that the string literals should have the same prefixes and the same literal value when all string elements are considered instead of trying to resolve its actual value. @nicolas-harraudeau-sonarsource  could you please update those when you have time?

python-checks/src/main/java/org/sonar/python/checks/DictionaryDuplicateKeyCheck.java
andrea-guarino-sonarsource: I believe that's a fair point. I'm fine with excluding large dictionaries as a first baby step.
If it makes sense and we believe there is value analyzing large dictionaries, we can improve the rule later.

python-checks/src/main/java/org/sonar/python/checks/DictionaryDuplicateKeyCheck.java
guillaume-dequenne-sonarsource: On the first case, that would absolutely be a FP which I'm not sure is worth addressing (I believe it depends on how common the approach is, which I must admit I have no idea).

One thing to point out is that a similar FP would happen on a member access `a.b` which uses a descriptor that somehow doesn't return the same value at each access. We discussed this case with @nicolas-harraudeau-sonarsource  and decided to accept the risk of FP.

The problem with such cases is that we don't always have a strong way of knowing whether the default behavior is being overridden or not, so it would be either accepting FPs for the odd case, or accepting FNs for the most common cases. I'd argue we see the result on Peach for this one to see if there are any further restrictions we should make. WDYT @nicolas-harraudeau-sonarsource ?

For the second case, that's a good catch and it made me realize the current implementation also reports FPs when two identical tuples contain function calls. This is fixed in af74d2bb08b0c29d5d5b4ec9ca09a5b399d3dc90.

python-checks/src/main/java/org/sonar/python/checks/DictionaryDuplicateKeyCheck.java
guillaume-dequenne-sonarsource: Indeed, I also looked a bit on the idea to normalize the keys and it doesn't seem straightforward due to the use of `BigDecimal` (as hashing those directly doesn't work due to the scale being part of the `equals` check), so I excluded dictionary literals of more than 100 elements in af74d2bb08b0c29d5d5b4ec9ca09a5b399d3dc90.

python-checks/src/main/java/org/sonar/python/checks/DictionaryDuplicateKeyCheck.java
andrey-tyukin-sonarsource: 1. I have to admit that the example is completely synthetic. I agree, it's probably very uncommon.

2. LGTM

guillaume-dequenne-sonarsource:
Nice intuition on the 100 threshold :) 

I guess the worst case scenario is actually the most common scenario (granted not reflected by the unit test which would terminate early) since duplicate keys are not desirable, so I'm not sure which comment should be changed and how?

PR #713, opened 2020-05-11
SONARPY-672 Rule S5797 Constants should not be used as conditions

andrea-guarino-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/ConstantConditionCheck.java
```diff
@@ -0,0 +1,156 @@
...
+      case SET_LITERAL:
+        return isAlwaysEmptyOrNonEmptyCollection(((SetLiteral) condition).elements());
+      case TUPLE:
+        return isAlwaysEmptyOrNonEmptyCollection(((Tuple) condition).elements());
+      default:
+        return false;
+    }
+  }
+
+  private static boolean isAlwaysEmptyOrNonEmptyCollection(ExpressionList elements) {
```
guillaume-dequenne-sonarsource, line 110: I think you could remove this method if you call `ExpressionList#expressions` directly on line 98.

andrea-guarino-sonarsource: Indeed, good point! thanks!

PR #717, opened 2020-05-14
SONARPY-674 Rule S5796: identity comparisons with new objects

andrey-tyukin-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/IdentityComparisonWithNewObjectCheck.java
```diff
@@ -0,0 +1,155 @@
...
+   * @return An empty option if there is no issue, or an option containing the list of relevant secondary expressions.
+   */
+  private static Optional<List<Tree>> findIssueForOperand(Expression expr) {
+    if (instantiatesFreshObject(expr)) {
+      // Issue exists, but the list of secondaries is empty.
+      return Optional.of(Collections.emptyList());
+    } else if (expr.is(Tree.Kind.NAME)) {
+      Name name = (Name) expr;
+      Expression rhs = Expressions.singleAssignedValue((Name) expr);
+      if (rhs != null && instantiatesFreshObject(rhs) && cannotEscape(name.symbol())) {
```
guillaume-dequenne-sonarsource, line 96: This is a bit nitpicky but I believe if you assign `name.symbol()` to a variable and test it for null with an early return before calling `Expressions.singleAssignedValue` you can remove the null checks from `cannotEscape` method and achieve 100% coverage.

python-checks/src/main/java/org/sonar/python/checks/IdentityComparisonWithNewObjectCheck.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+  }
+
+  /**
+   * Checks a single operand of an <code>is/is not</code>-comparison.
+   * Returns <code>true</code> if it finds an issue.
+   */
+  private static boolean checkOperand(Expression operand, IsExpression isExpr, SubscriptionContext ctx) {
+    Optional<List<Tree>> secondariesOpt = findIssueForOperand(operand);
+    secondariesOpt.ifPresent(secondaryPositions -> {
```
guillaume-dequenne-sonarsource, line 71: Nitpick but I'd rename `secondaryPositions` with `secondaryLocations` for consistency with our lexicon.

python-checks/src/test/resources/checks/identityComparisonWithNewObjects.py
```diff
@@ -0,0 +1,133 @@
...
+  #   ^^^^^^^^^^^> 1 {{This expression creates a new object every time.}}
+  x is p # Noncompliant {{Replace this "is" operator with "==".}}
+  # ^^
+  p is x # Noncompliant {{Replace this "is" operator with "==".}}
+  # ^^
+
+def coverage():
+  a[b] is b[a]
+
+x is y # should break as soon as symbols are resolved outside of functions
```
guillaume-dequenne-sonarsource, line 57: I'm not sure what your intention is here, but symbols are actually resolved outside functions. What we don't do in the module scope is type inference.

Here none of the symbols exist as neither x nor y is defined in the module scope. If you place this snippet in a function, it still won't raise. If you define either `x = []` or `y = []` on the line above, it will raise an issue.

I think the test as it is, is fine as it ensures we don't raise on undefined symbols, but I think the comment is a bit confusing.

PR #723, opened 2020-05-18
SONARPY-664 Rule S5781: Expressions creating sets should not have dup…

guillaume-dequenne-sonarsource:
…licate values


python-checks/src/main/java/org/sonar/python/checks/SetDuplicateKeyCheck.java
```diff
@@ -0,0 +1,68 @@
...
+
+@Rule(key = "S5781")
+public class SetDuplicateKeyCheck extends AbstractDuplicateKeyCheck {
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.SET_LITERAL, ctx -> {
+      SetLiteral setLiteral = (SetLiteral) ctx.syntaxNode();
+      Set<Integer> issueIndexes = new HashSet<>();
+      if (setLiteral.elements().size() > 100) {
```
gyula-sallai-sonarsource, line 39: _minor_ I would add this as a constant with a name indicative of why it was introduced.

PR #726, opened 2020-05-19
SONARPY-216 Rule S2201: Return values from functions without side effects should not be ignored

gyula-sallai-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/IgnoredPureOperationsCheck.java
```diff
@@ -0,0 +1,389 @@
...
+      }
+
+      String fqn = symbol.fullyQualifiedName();
+      if (fqn != null && PURE_FUNCTIONS.contains(fqn)) {
+        return new IgnoredPureOperation(callExpression.callee(), fqn);
+      }
+    } else if (expression.is(Tree.Kind.SUBSCRIPTION)) {
+      SubscriptionExpression subscriptionExpression = ((SubscriptionExpression) expression);
+      InferredType type = subscriptionExpression.object().type();
+      if (PURE_GETITEM_TYPES.stream().anyMatch(type::canBeOrExtend)) {
```
guillaume-dequenne-sonarsource, line 360: I think you'd want to use `type::canOnlyBe` here. Using type::canBeOrExtend here means you can have FPs if you have an unresolved type hierarchy. In fact, the only reason the current implementation doesn't raise an FP for `defaultdict` is because the symbol is hardcoded in `SymbolUtils::externalModulesSymbols`.

Also from what I can see from `expected-issues`, we shouldn't raise on subclasses anyway (otherwise, implementing an `InferredType::isOrExtend` could have been an option).

python-checks/src/test/resources/checks/ignoredPureOperations.py
```diff
@@ -0,0 +1,87 @@
...
+   #^^^^^^^
+    'a' in d1 # Noncompliant {{The return value of "__contains__" must be used.}}
+   #^^^^^^^^^
+    x = d1['a']
+    x = 'a' in d1
+
+    d2 = defaultdict()
+    d2['a'] # Ok - defaultdict.__geitem__ creates the key if missing
+    d3 = MyDict()
+    d3['a'] # Noncompliant
```
guillaume-dequenne-sonarsource, line 45: cf previous comment: I think this should be an accepted FN instead.

python-checks/src/main/java/org/sonar/python/checks/IgnoredPureOperationsCheck.java
```diff
@@ -0,0 +1,389 @@
...
+    "round",
+    "min",
+    "max",
+    "divmod",
+    "sum",
+    "pow",
+    "all",
+    "any",
+    "sorted",
+    //"map",
```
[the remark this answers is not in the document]
gyula-sallai-sonarsource: Indeed, I missed deleting this one.

python-checks/src/main/java/org/sonar/python/checks/IgnoredPureOperationsCheck.java
gyula-sallai-sonarsource: I have updated the check, now we raise a lot less issues which could have been arguably false-positives. Thanks for the feedback!

PR #728, opened 2020-05-19
fixed overlap with S5727 and S3403

andrey-tyukin-sonarsource:
Excluded unpacking assignments from singleAssignedValue

### DIFF
--- a/python-checks/src/test/resources/checks/identityComparisonWithNewObjects.py
+++ b/python-checks/src/test/resources/checks/identityComparisonWithNewObjects.py
@@ -56,7 +56,7 @@ x is y # repeatedly checked: required for coverage of an `if (symb != null) {`
 # Unpacking false positive from review
 def foo(param):
     a, b = list(param)
-    if a is None:  # This was a false positive. There is no way to know what value is assigned to "a".
+    if a is None:  # There is no simple way to know what value is assigned to "a" after unpacking assignment.
         pass
 
     x, y = list(p)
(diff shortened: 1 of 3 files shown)


andrey-tyukin-sonarsource:
@andrea-guarino-sonarsource requesting review because of the changes to singleAssignedValue

The problem with unpacking assignments pointed out here: https://jira.sonarsource.com/browse/SONARPY-674 does not seem to be restricted to this particular rule.

python-checks/src/test/resources/checks/identityComparisonWithNewObjects.py
```diff
@@ -52,6 +52,38 @@ def coverage():
 
 x is y # repeatedly checked: required for coverage of an `if (symb != null) {`
 
+
+# Unpacking false positive from review
+def foo(param):
+    a, b = list(param)
+    if a is None:  # This was a false positive. There is no way to know what value is assigned to "a".
```
andrea-guarino-sonarsource, line 59: I would drop the "This was a false positive comment", it might be confusing

python-checks/src/test/resources/checks/identityComparisonWithNewObjects.py
```diff
@@ -52,6 +52,38 @@ def coverage():
 
 x is y # repeatedly checked: required for coverage of an `if (symb != null) {`
 
+
+# Unpacking false positive from review
```
andrea-guarino-sonarsource, line 56: it's not clear to what review you're referring to, so I would drop this comment

PR #734, opened 2020-05-22
Sonarpy 692

andrey-tyukin-sonarsource:
Systematically transformed the recursive `equals`-like methods into an `evaluate`-method, which computes trees with values which would previously be compared directly. This allows to compute the hash codes on those trees, thus eliminating most of the unnecessary comparisons, and avoiding the quadratic overhead.

Changed "key" to "value" in the message for the set-rule (the spec mentioned "value", not "key").

Also, added a possibility to use synthetic strings as input for unit tests. The synthetic strings are written into a temporary file, and then processed as usual.


python-checks/src/test/java/org/sonar/python/checks/AbstractDuplicateKeyCheckTest.java
```diff
@@ -0,0 +1,158 @@
...
+import static org.sonar.python.checks.AbstractDuplicateKeyCheck.NumericValue;
+import static org.sonar.python.checks.AbstractDuplicateKeyCheck.StringValue;
+
+@SuppressWarnings("squid:S2701")
+public class AbstractDuplicateKeyCheckTest {
+
+    private static <A> void testHashEquals(
+      Supplier<A> firstColliding,
+      Supplier<A> secondColliding,
+      Supplier<A> unrelated,
```
gyula-sallai-sonarsource, line 44: Is using suppliers necessary? I find their usage slightly harder to read and we calculate the first and second argument in every test anyways.

python-checks/src/test/java/org/sonar/python/checks/AbstractDuplicateKeyCheckTest.java
```diff
@@ -0,0 +1,158 @@
...
+    private static final int INT_COLLISION_A = 138563169;
+    private static final int INT_COLLISION_B = 158370000;
+
+    @Test
+    public void testHashEqualsNumeric() {
+        testHashEquals(
+          () -> new NumericValue(new BigDecimal(INT_COLLISION_A)),
+          () -> new NumericValue(new BigDecimal(INT_COLLISION_B)),
+          () -> new NumericValue(new BigDecimal(0)),
+          false
```
gyula-sallai-sonarsource, line 71: _minor_ I would create two test methods instead of passing this boolean flag.

python-checks/src/test/java/org/sonar/python/checks/AbstractDuplicateKeyCheckTest.java
```diff
@@ -0,0 +1,158 @@
...
+        assertNotEquals(firstColliding.get(), unrelated.get());
+        if (!canReturnDifferentValues) {
+            assertEquals(firstColliding.get(), firstColliding.get());
+        } else {
+            assertNotEquals(firstColliding.get(), firstColliding.get());
+        }
+        A a = firstColliding.get();
+        assertEquals(a, a);
+        // This is not same as `assertNotNull`, it must invoke `equals`.
+        assertNotEquals(a, null);
```
gyula-sallai-sonarsource, line 58: Maybe we could call `equals` directly and avoid the suppression in L38?

andrey-tyukin-sonarsource: Unfortunately, no.
If you use `assertFalse(...equals)`, it says to use `assertNotEquals` instead. Once you change it to `assertNotEquals`, it says to change it to `assertNotNull`. No matter what I did, the series of messages kept spinning until the test became trivial - it simply didn't believe me that the `equals` method is exactly what I want to test.

gyula-sallai-sonarsource: Okay, that makes sense. Even though I still would be in favor of avoiding a class-level `SuppressWarnings` and marking the issue SQ raises as an FP.

PR #736, opened 2020-05-22
SONARPY-699 Fix FP on S1481: variables used as a nested format specification in f-string

gyula-sallai-sonarsource:


python-frontend/src/main/java/org/sonar/plugins/python/api/tree/BaseTreeVisitor.java
```diff
@@ -465,5 +465,12 @@ public void visitToken(Token token) {
   @Override
   public void visitFormattedExpression(FormattedExpression formattedExpression) {
     scan(formattedExpression.expression());
+    scan(formattedExpression.formatSpecifier());
+  }
+
+  @Override
+  public void visitFormatSpecifier(FormatSpecifier formatSpecifier) {
+    scan(formatSpecifier.columnToken());
```
guillaume-dequenne-sonarsource, line 473: As we don't visit such tokens in the `BaseTreeVisitor`, I wouldn't do it for this one either to stay consistent (it doesn't have an impact).

gyula-sallai-sonarsource: fixed in 044ba31b39643d1a674ef048107942712e755d27

PR #743, opened 2020-05-25
SONARPY-707 Add stubs for Django

dmitry-golovinov-sonarsource:


dmitry-golovinov-sonarsource:
Decided that we are not going to merge django-stubs right at the moment.

PR #748, opened 2020-05-27
SONARPY-708 Remove hardcoded external symbols

dmitry-golovinov-sonarsource:


python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
```diff
@@ -160,7 +160,20 @@ private static void setDeclaredReturnType(Symbol symbol, FunctionDef functionDef
 
   public static Symbol symbolWithFQN(String stdLibModuleName, String fullyQualifiedName) {
```
andrea-guarino-sonarsource, line 161: I think this method should be annotated with `@CheckForNull`

python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
```diff
@@ -160,7 +160,20 @@ private static void setDeclaredReturnType(Symbol symbol, FunctionDef functionDef
...
     Set<Symbol> symbols = symbolsForModule(stdLibModuleName);
-    return symbols.stream().filter(s -> fullyQualifiedName.equals(s.fullyQualifiedName())).findFirst().orElse(null);
+    Symbol symbolByFqn = symbols.stream().filter(s -> fullyQualifiedName.equals(s.fullyQualifiedName())).findFirst().orElse(null);
+    if (symbolByFqn != null || !fullyQualifiedName.contains(".")) {
+      return symbolByFqn;
+    }
+
+    String[] fqnSplittedByDot = fullyQualifiedName.split("\\.");
```
andrea-guarino-sonarsource, line 168: I think it's worth adding a comment explaining why we're doing this

python-frontend/src/test/java/org/sonar/python/semantic/FullyQualifiedNameTest.java
```diff
@@ -75,6 +75,15 @@ public void submodule_alias() {
     assertNameAndQualifiedName(tree, "fn", "mod.submod.fn");
   }
 
+  @Test
+  public void alias_import_preserves_fqn() {
```
andrea-guarino-sonarsource, line 79: I think we can add a comment with a link to the typeshed file, to give more context to this unit test

PR #749, opened 2020-05-27
SONARPY-679 Rule S2823: Only strings should be listed in "__all__"

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/NonStringInAllPropertyCheck.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+    List<Expression> expressions = getAllExpressions(assignedValue);
+    for (Expression element : expressions) {
+      if (element instanceof HasSymbol) {
+        Symbol symbol = ((HasSymbol) element).symbol();
+        if (isClassOrFunctionSymbol(symbol)) {
+          ctx.addIssue(element, MESSAGE);
+          continue;
+        }
+      }
```
gyula-sallai-sonarsource, line 79: I would move this code into `couldBeString`

python-checks/src/main/java/org/sonar/python/checks/NonStringInAllPropertyCheck.java
```diff
@@ -0,0 +1,122 @@
...
+      if (!couldBeString(element)) {
+        ctx.addIssue(element, MESSAGE);
+      }
+    }
+  }
+
+  private static boolean isClassOrFunctionSymbol(@Nullable  Symbol symbol) {
```
gyula-sallai-sonarsource, line 86: _minor_ Excess whitespace after `@Nullable`

python-checks/src/main/java/org/sonar/python/checks/NonStringInAllPropertyCheck.java
```diff
@@ -0,0 +1,122 @@
...
+  private static void checkAllProperty(SubscriptionContext ctx, AssignmentStatement assignmentStatement) {
+    Expression assignedValue = assignmentStatement.assignedValue();
+    if (!assignedValue.is(Tree.Kind.LIST_LITERAL) && !assignedValue.is(Tree.Kind.TUPLE)) {
+      return;
+    }
+    List<Expression> expressions = getAllExpressions(assignedValue);
+    for (Expression element : expressions) {
+      if (element instanceof HasSymbol) {
+        Symbol symbol = ((HasSymbol) element).symbol();
+        if (isClassOrFunctionSymbol(symbol)) {
```
gyula-sallai-sonarsource, line 75: I would merge this condition with the previous one.

PR #752, opened 2020-05-29
SONARPY-712 Merge symbols for module based on local name

dmitry-golovinov-sonarsource:


its/ruling/src/test/resources/expected/python-S2638.json
```diff
@@ -33,9 +33,6 @@
 151,
 193,
 ],
-'project:django-2.2.3/django/utils/text.py':[
```
andrea-guarino-sonarsource, line 36: Can you create a unit test to document this FN?

PR #753, opened 2020-05-29
SONARPY-709 Basic support for printf-style string format

gyula-sallai-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/StringFormat.java
```diff
@@ -0,0 +1,185 @@
...
+        return Optional.empty();
+      }
+
+      String width = matcher.group(3);
+      String precision = matcher.group(4);
+      if ("*".equals(width)) {
+        result.add(new ReplacementField(printfWidthOrPrecValidator(ctx), mapKey));
+      }
+      if ("*".equals(precision)) {
+        result.add(new ReplacementField(printfWidthOrPrecValidator(ctx), mapKey));
```
guillaume-dequenne-sonarsource, line 112: As I understand, `*` must be used with sequential values only. Therefore I assume it would make sense to not pass the `mapKey` here. 

The current behavior is a bit misleading:
```
print("hello %(name)*s" % {"name": "John"}) # Noncompliant {{Replace this value with an integer as "*" requires.}}
#                                  ^^^^^^
print("hello %(name)*s" % {"name": 9}) # OK (FN)
```

python-checks/src/main/java/org/sonar/python/checks/StringFormatMisuseCheck.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    if (dict.elements().stream().anyMatch(StringFormatMisuseCheck::isOutOfScopeDictionaryElement)) {
+      // Do not bother with dictionaries containing unpacking expressions or keys which are not string literals.
+      return;
+    }
+
+    Map<String, List<StringFormat.ReplacementField>> fieldMap = format.replacementFields().stream().collect(
+      Collectors.groupingBy(StringFormat.ReplacementField::mappingKey));
+    if (fieldMap.size() != dict.elements().size()) {
```
guillaume-dequenne-sonarsource, line 96: I believe we only want to detect missing arguments in the case of a dictionary. The following case doesn't fail, so I don't think it's part of the scope of this rule:
```
print("Hello %(first_name)s" % {"first_name": "John", "last_name": "Doe"})
```

python-checks/src/main/java/org/sonar/python/checks/StringFormat.java
```diff
@@ -0,0 +1,185 @@
...
+    return this.replacementFields.stream().anyMatch(field -> field.mappingKey() == null);
+  }
+
+  public boolean hasNamedFields() {
+    return this.replacementFields.stream().anyMatch(field -> field.mappingKey() != null);
+  }
+
+  // See https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting
+  private static final Pattern PRINTF_PARAMETER_PATTERN = Pattern.compile(
+    "%" + "(?:\\((.*?)\\))?" + "([#\\-+0 ]*)?" + "([0-9]*|\\*)?" + "(?:\\.([0-9]*|\\*))?" + "([lLH])?" + "([A-Za-z]|%)");
```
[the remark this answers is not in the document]
gyula-sallai-sonarsource: Thanks for the suggestion, I added the named groups - kind of forgot that Java regex supports them :) The pattern is still a little bit hard to read, but when I originally tried to split it into variables that did not really feel good to me either, this why I used string concatenation.

python-checks/src/main/java/org/sonar/python/checks/StringFormatMisuseCheck.java
```diff
@@ -41,32 +43,50 @@
...
-      } else if (expression.rightOperand().is(Tree.Kind.DICTIONARY_LITERAL)) {
-        checkDictionaries(ctx, format, ((DictionaryLiteral) expression.rightOperand()));
-      } else if (format.numExpectedArguments() == 1) {
-        format.replacementFields().get(0).validateArgument(expression.rightOperand());
+
+      Expression rhs = expression.rightOperand();
+      if (format.hasNamedFields()) {
+        checkNamed(ctx, format, rhs);
+      } else {
+        checkPositional(ctx, format, rhs);
```
guillaume-dequenne-sonarsource, line 63: I'm not sure if the fix should be here but the current implementation raises an FP for:
```
"%%(key)s" % {"other": "str"}  # OK. Pylint FP
```
The problem is that `createFromPrintfStyle` returns a `StringFormat` with 0 replacement fields, so we enter `checkPositional` and then raise because the rhs is not a tuple.

Either the `StringFormat` should return `Optional.empty()` in such case, or this logic should be updated.

PR #754, opened 2020-05-29
SONARPY-678 Rule S5807: Only defined names should be listed in "__all__"

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/UndefinedNameAllPropertyCheck.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+  }
+
+  private static boolean isUnknownSymbol(PythonFile pythonFile, Map<String, Symbol> symbolsByName, StringLiteral stringLiteral) {
+    if (stringLiteral.stringElements().stream().anyMatch(StringElement::isInterpolated)) {
+      return false;
+    }
+    return !symbolsByName.containsKey(stringLiteral.trimmedQuotesValue())
+      && !BuiltinSymbols.all().contains(stringLiteral.trimmedQuotesValue())
+      && !isInitFileExportingModule(pythonFile, stringLiteral.trimmedQuotesValue());
```
gyula-sallai-sonarsource, line 106: _subjective, minor_ It looks a little bit weird to me that three out of the four conditions are inlined into a single expression, but the first one has its own if statement. I would either have them all in the return or split them into if statements.

gyula-sallai-sonarsource, line 106: I would introduce a variable for `stringLiteral.trimmedQuotesValue()`.

python-checks/src/main/java/org/sonar/python/checks/UndefinedNameAllPropertyCheck.java
```diff
@@ -0,0 +1,163 @@
...
+      return ((ListLiteral) expression).elements().expressions().stream()
+        .filter(UndefinedNameAllPropertyCheck::isString)
+        .collect(Collectors.toList());
+    }
+    return ((Tuple) expression).elements().stream()
+      .filter(UndefinedNameAllPropertyCheck::isString)
+      .collect(Collectors.toList());
+  }
+
+  private static class UnresolvedWildcardImportVisitor extends BaseTreeVisitor {
```
gyula-sallai-sonarsource, line 127: _minor_ I think this class should be placed elsewhere, either to the top/bottom of the enclosing class or before/after the method that uses it.

python-checks/src/main/java/org/sonar/python/checks/UndefinedNameAllPropertyCheck.java
```diff
@@ -0,0 +1,162 @@
...
+        if (stringExpression != stringLiteral) {
+          issue.secondary(stringLiteral, null);
+        }
+      }
+    }
+  }
+
+  private static boolean isUnknownSymbol(PythonFile pythonFile, Map<String, Symbol> symbolsByName, StringLiteral stringLiteral) {
+    String name = stringLiteral.trimmedQuotesValue();
+    return stringLiteral.stringElements().stream().noneMatch(StringElement::isInterpolated) &&
```
gyula-sallai-sonarsource, line 102: _minor_ I would move the `&&` operator to the next line, as it is done for the other lines.

PR #756, opened 2020-06-02
SONARPY-718 Define harcoded external libraries symbols as .pyi files

dmitry-golovinov-sonarsource:


python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
```diff
@@ -213,7 +219,7 @@ private static ModuleDescription getResourceForModule(String moduleName, String
     String packageName = String.join(".", Arrays.copyOfRange(moduleNameHierarchy, 0, moduleNameHierarchy.length - 1));
     InputStream resource = TypeShed.class.getResourceAsStream(categoryPath + pathToModule + ".pyi");
     if (resource == null) {
-      resource = TypeShed.class.getResourceAsStream(categoryPath + moduleName + "/__init__.pyi");
+      resource = TypeShed.class.getResourceAsStream(categoryPath + pathToModule + "/__init__.pyi");
```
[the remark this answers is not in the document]
dmitry-golovinov-sonarsource: I've already been fixing it. Not quite sure why it was not like this before.

mod1/mod2/__init__.pyi being resolved as mod1.mod2/__init__.pyi making it impossible to declare subpackages.

python-checks/src/main/java/org/sonar/python/checks/ArgumentTypeCheck.java
```diff
@@ -177,8 +176,6 @@ public static String matchBuiltinCategory(Predicate<String> predicate) {
   We exclude types which can be associated to hard-coded symbols with missing members
```
andrea-guarino-sonarsource, line 176: This comment is outdated, can you please remove it?

python-frontend/src/main/resources/org/sonar/python/types/custom/cryptography/hazmat/primitives/asymmetric/rsa.pyi
```diff
@@ -0,0 +1,8 @@
+class RSAPublicKey:
```
andrea-guarino-sonarsource, line 1: I think we're missing the "from typing import Any"

dmitry-golovinov-sonarsource: Good point!

PR #761, opened 2020-06-04
SONARPY-688 Rule S5828: The "open" builtin should be called with a va…

guillaume-dequenne-sonarsource:
…lid mode


python-checks/src/main/java/org/sonar/python/checks/InvalidOpenModeCheck.java
```diff
@@ -0,0 +1,120 @@
...
+import org.sonar.plugins.python.api.tree.StringElement;
+import org.sonar.plugins.python.api.tree.StringLiteral;
+import org.sonar.plugins.python.api.tree.Tree;
+import org.sonar.python.tree.TreeUtils;
+
+@Rule(key = "S5828")
+public class InvalidOpenModeCheck extends PythonSubscriptionCheck {
+
+  private static final String VALID_MODES = "rwatb+Ux";
+  private static final Pattern INVALID_CHARACTERS = Pattern.compile("[^rwatb+Ux]");
```
andrea-guarino-sonarsource, line 44: Minor: I think you can use `VALID_MODES` here

PR #762, opened 2020-06-04
SONARPY-676 Rule S5806: Builtins should not be shadowed

dmitry-golovinov-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/BuiltinShadowingAssignmentCheck.java
```diff
@@ -0,0 +1,156 @@
...
+    context.registerSyntaxNodeConsumer(Tree.Kind.ASSIGNMENT_EXPRESSION, this::checkAssignmentExpression);
+    context.registerSyntaxNodeConsumer(Tree.Kind.FUNCDEF, this::checkFunctionDeclaration);
+    context.registerSyntaxNodeConsumer(Tree.Kind.CLASSDEF, this::checkClassDefinition);
+    context.registerSyntaxNodeConsumer(Tree.Kind.IMPORT_FROM, ctx ->
+      ((ImportFrom) ctx.syntaxNode()).importedNames().forEach(m -> checkImportedNameAlias(ctx, m)));
+  }
+
+  private void checkAssignmentExpression(SubscriptionContext ctx) {
+    AssignmentExpression assignmentExpression = (AssignmentExpression) ctx.syntaxNode();
+    if (shouldReportIssue(assignmentExpression.lhsName(), assignmentExpression.lhsName())) {
```
guillaume-dequenne-sonarsource, line 81: _minor_ I think you can define a variable for `assignmentExpression.lhsName()`.

python-checks/src/main/java/org/sonar/python/checks/BuiltinShadowingAssignmentCheck.java
```diff
@@ -0,0 +1,156 @@
...
+
+  private boolean shouldReportIssue(Tree tree, Tree parent) {
+    return tree.is(Tree.Kind.NAME) && isBuiltInName((Name) tree) && !skipIssueCondition(parent);
+  }
+
+  private boolean isBuiltInName(Name name) {
+    return reservedNames.contains(name.name());
+  }
+
+  private static boolean skipIssueCondition(Tree name) {
```
guillaume-dequenne-sonarsource, line 146: _minor_ I'd rename this method into something that better reflects the fact that it returns a boolean (like `isException`, `shouldSkipIssueCondition`, or something like that)

its/ruling/src/test/resources/expected/python-S5806.json
```diff
@@ -0,0 +1,2049 @@
+{
```
andrea-guarino-sonarsource, line 1: isn't it a bit too noisy? I think it may be worth checking with @nicolas-harraudeau-sonarsource if we can avoid some of those issues

guillaume-dequenne-sonarsource: I share the feeling. It appears that some builtin names are quite popular names: `max, id, dict, type`, etc...Not sure what should be done: those are indeed TP, developers seem to be comfortable with shadowing builtins.

nicolas-harraudeau-sonarsource: I would like to see on peach if all projects are like this or if only a few of them do this. Personally I wouldn't remove any builtin from the list. Shadowing builtins can be a source of bugs in case of refactoring. Ex: you rename the variable but forget a few places and they use the builtin instead. There is also a simple solution: add an underscore at the end of the name or find another name.

There is however one thing which came to my mind this morning. We should not raise issues on parameters of methods overriding a method from a parent class. Developers can't change these parameters or they would break the Liskov substitution principle. These issues should only appear on the top overriden method. If we don't have the full class hierarchy we would not raise at all just to avoid FPs. Does it make sense @dmitry-golovinov-sonarsource ?

nicolas-harraudeau-sonarsource: Another solution would be to have a parameter enabling the rule on method and function parameters, and it would be disabled by default. This would avoid raising on parameter names when developers have to follow an API (ex: for duck typing).

nicolas-harraudeau-sonarsource: As discusser IRL I'll modify the RSPEC as follows:
* I will add a parameter enabling the validation of function/method/lambda parameters. It will be disabled, i.e. set to False, by default.
* I will document in the RSPEC description the exception about function/method/lambda parameters.

We also agreed that when the parameter is enabled we won't raise on overriding methods.

python-checks/src/main/java/org/sonar/python/checks/CheckUtils.java
```diff
@@ -113,4 +120,25 @@ public static boolean isCalledInClassBody(Symbol symbol, ClassDef classDef) {
     });
   }
 
+  public static Optional<FunctionSymbol> getOverriddenMethod(FunctionSymbol functionSymbol) {
```
andrea-guarino-sonarsource, line 123: I'd suggest moving this method to `SymbolUtils` or `TreeUtils` in python-frontend so you don't need to declare Mockito dependency

sonar-python-plugin/pom.xml
```diff
@@ -136,7 +136,7 @@
             <configuration>
               <rules>
                 <requireFilesSize>
-                  <maxsize>4500000</maxsize>
+                  <maxsize>4800000</maxsize>
                   <minsize>4200000</minsize>
```
guillaume-dequenne-sonarsource, line 140: _Minor_: Could you update `minsize` as well, to keep the same range?

python-checks/src/main/java/org/sonar/python/checks/CheckUtils.java
guillaume-dequenne-sonarsource: Agreed, I think it would make sense to have it in `SymbolUtils`. Also you wouldn't have to redefine the `pythonFile` helper method that way.

python-checks/src/main/java/org/sonar/python/checks/BuiltinShadowingAssignmentCheck.java
```diff
@@ -0,0 +1,192 @@
...
+      for (Parameter parameter : parameterList.nonTuple()) {
+        Name parameterName = parameter.name();
+        if (parameterName != null && isBuiltInName(parameterName)) {
+          ctx.addIssue(parameterName, String.format(MESSAGE, "parameter"));
+        }
+      }
+    }
+  }
+
+  private static boolean checkIfMethodIsOverride(ParameterList parameterList) {
```
guillaume-dequenne-sonarsource, line 146: _Minor_: could you rename this into something like `isParentAMethodOverride` ? Instinctively I think a method named `checkSomething` is about raising an issue if some condition is met rather than return a boolean.

PR #763, opened 2020-06-05
SONARPY-717 Support for 'str.format()'-style string formatting

gyula-sallai-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/StringFormat.java
```diff
@@ -65,34 +141,251 @@ private StringFormat(List<ReplacementField> replacementFields) {
...
+      return true;
+    }
+
+    private int parseFieldName(char current, int pos) {
+      if (current == '{') {
+        state = ParseState.INIT;
+      } else {
+        state = ParseState.FIELD;
+        nesting++;
+        if (matcher.region(pos, value.length()).find()) {
```
andrea-guarino-sonarsource, line 347: can we find a more descripting name for `matcher`?

python-checks/src/main/java/org/sonar/python/checks/StringFormatMisuseCheck.java
```diff
@@ -44,32 +53,97 @@
...
+
+    Optional<StringFormat> formatOptional = StringFormat.createFromPrintfStyle(ctx, expression.leftOperand(), formatString);
+    if (!formatOptional.isPresent()) {
+      // The string format contains invalid syntax.
+      return;
+    }
+
+    StringFormat format = formatOptional.get();
+    Expression rhs = expression.rightOperand();
+    if (format.numExpectedArguments() == 0 && (isMapping(rhs) || rhs.type().canOnlyBe("list"))) {
```
andrea-guarino-sonarsource, line 137: minor: for "list" you can use `BuiltinTypes.LIST`

PR #764, opened 2020-06-05
SONARPY-678 Fix FPs for S5807

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/UndefinedNameAllPropertyCheck.java
```diff
@@ -149,14 +160,53 @@ private static StringLiteral retrieveStringLiteral(Tree tree) {
     return (StringLiteral) Expressions.singleAssignedValue((Name) tree);
   }
 
-  private static class UnresolvedWildcardImportVisitor extends BaseTreeVisitor {
+  private static boolean importsManipulatedAllProperty(FileInput fileInput) {
+    return fileInput.globalVariables().stream().anyMatch(s -> s.name().equals("__all__") && s.usages().stream().anyMatch(u -> u.kind().equals(Usage.Kind.IMPORT)));
```
gyula-sallai-sonarsource, line 164: _nitpick_ I would prefer to avoid nesting `anyMatch` and would rather have a loop for the global variables, but it's just a personal preference, feel free to disregard.

python-checks/src/main/java/org/sonar/python/checks/UndefinedNameAllPropertyCheck.java
```diff
@@ -149,14 +160,53 @@ private static StringLiteral retrieveStringLiteral(Tree tree) {
...
-      hasUnresolvedWildcardImport |= importFrom.hasUnresolvedWildcardImport();
+      hasWildcardImport |= importFrom.isWildcardImport();
+      shouldNotReportIssue |= importFrom.hasUnresolvedWildcardImport();
       super.visitImportFrom(importFrom);
     }
+
+    @Override
+    public void visitCallExpression(CallExpression callExpression) {
+      Symbol calleeSymbol = callExpression.calleeSymbol();
+      shouldNotReportIssue |= calleeSymbol != null && "globals".equals(calleeSymbol.fullyQualifiedName());
```
gyula-sallai-sonarsource, line 182: _nitpick_ I find statements like this a little bit hard to read, I would either:

- surround the right hand side in parentheses:
`shouldNotReportIssue |= (calleeSymbol != null && "globals".equals(calleeSymbol.fullyQualifiedName()));`
- or create a helper method to do this check and reuse it in line 190, e.g.:
`shouldNotReportIssue |= isSymbolWithName(calleeSymbol, "globals")`

PR #765, opened 2020-06-08
SONARPY-675 S5799: Avoid reporting issues on intentional concatenations

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/ImplicitStringConcatenationCheck.java
```diff
@@ -58,6 +64,13 @@ public void initialize(Context context) {
     });
   }
 
+  private static boolean isABoolean(StringElement first, StringElement second) {
```
gyula-sallai-sonarsource, line 67: I think this one should have a more descriptive name.

python-checks/src/main/java/org/sonar/python/checks/ImplicitStringConcatenationCheck.java
```diff
@@ -58,6 +64,13 @@ public void initialize(Context context) {
...
+    if (first.firstToken().column() + first.value().length() > MAX_COLUMN) {
+      return false;
+    }
+    return !END_LINE_PATTERN.matcher(first.trimmedQuotesValue()).matches() && !START_LINE_PATTERN.matcher(second.trimmedQuotesValue()).matches();
```
gyula-sallai-sonarsource, line 71: Feel free to disregard: I am not sure if we should use a regex for this one, I think it would be a little bit cleaner if we checked the first and last characters of these lines "explicitly".

guillaume-dequenne-sonarsource: As discussed, I'll leave it that way as matching punctuation characters is made much simpler through the regex than by doing it explicitly.

PR #767, opened 2020-06-08
SONARPY-514 Rule S2755: XML parsers should not be vulnerable to XXE attacks

dmitry-golovinov-sonarsource:


python-checks/src/test/resources/checks/xmlParserXxeCheck.py
```diff
@@ -0,0 +1,104 @@
+from lxml import etree
+
+def case1():
+    parser = etree.XMLParser() # Noncompliant {{Disable access to external entities in XML parsing.}}
+    #        ^^^^^^^^^^^^^^^^^
+    tree1 = etree.parse('ressources/xxe.xml', parser)
+    #       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^<
+
```
andrea-guarino-sonarsource, line 8: we should have a test with unpacking expressions:

```python
parser = etree.XMLParser()
args = ['ressources/xxe.xml', parser]
etree.parse(*args) # we can accept this as FN
```

PR #768, opened 2020-06-09
SONARPY-676 Rule S5806: Builtins should not be shadowed. Reducing the noise.

dmitry-golovinov-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/BuiltinShadowingAssignmentCheck.java
```diff
@@ -120,7 +124,18 @@ private void checkAnnotatedAssignment(SubscriptionContext ctx) {
...
       Token equalToken = assignment.equalToken();
       if (equalToken != null && shouldReportIssue(variable)) {
-        ctx.addIssue(variable, String.format(MESSAGE, VRBL_ISSUE_TYPE));
+        raiseIssueForVariable(ctx, (Name) variable);
+      }
+    }
+  }
+
+  private void raiseIssueForVariable(SubscriptionContext ctx, Name variable) {
+    if (variable.symbol() != null) {
```
guillaume-dequenne-sonarsource, line 133: Nitpick: I think you can assign `variableSymbol` before so you don't have to access it twice.

dmitry-golovinov-sonarsource: Actually, this was a way to overcome issue with QG. It's a defensive check, but if I will move an assignment a line above -I will have to merge if clauses...

dmitry-golovinov-sonarsource: Found a way to overcome this

PR #769, opened 2020-06-09
SONARPY-685 Rule S3457: Strings should be formatted correctly 

gyula-sallai-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/StringFormatCorrectnessCheck.java
```diff
@@ -0,0 +1,234 @@
...
+      } else {
+        this.checkStrFormatStyle(ctx);
+      }
+    });
+    context.registerSyntaxNodeConsumer(Tree.Kind.STRING_LITERAL, StringFormatCorrectnessCheck::checkFStringLiteral);
+  }
+
+  private static void checkFStringLiteral(SubscriptionContext ctx) {
+    StringLiteral literal = (StringLiteral) ctx.syntaxNode();
+    if (TreeUtils.firstAncestorOfKind(literal, Tree.Kind.PLUS) != null) {
```
guillaume-dequenne-sonarsource, line 73: Not sure if this is a common case, but you could probably be more restrictive in case you'd want to catch wrongly formatted f-strings within uses of `PLUS` not related to the concatenation of said f-strings, like `foo(f"hello") + bar(f"world")`.

python-checks/src/test/resources/checks/stringFormatCorrectness.py
```diff
@@ -103,9 +103,25 @@ def logger_format():
...
+l1.error("Foo %s", "Bar")
+l1.error("Foo %s", "Bar", "too many")  # Noncompliant
+
+l2 = logging.getLogger('l2')
+l2 = logging.getLogger('l3')
+l2.error("Foo %s", "Bar")
+l2.error("Foo %s", "Bar", "too many")  # FN
+
+l3 = 'hello'
+l3.error("Foo %s", "Bar", "too many")  # FN
```
guillaume-dequenne-sonarsource, line 116: _Minor_ I guess that's not a FN as we don't want to raise on this one.

PR #771, opened 2020-06-10
SONARPY-710 Add API to access symbols generated from stub files

dmitry-golovinov-sonarsource:


python-frontend/src/main/java/org/sonar/python/SubscriptionVisitor.java
```diff
@@ -140,6 +155,16 @@ public PythonFile pythonFile() {
       return pythonVisitorContext.pythonFile();
     }
 
+    @Override
+    public Map<String, Set<Symbol>> getTypeShedSymbols() {
+      return SubscriptionVisitor.getTypeShedSymbols();
```
andrea-guarino-sonarsource, line 160: Why can't we return `Typeshed.typeShedSymbols` and remove the static methods on SubscriptionVisitor?

python-frontend/src/main/java/org/sonar/python/SubscriptionVisitor.java
```diff
@@ -140,6 +155,16 @@ public PythonFile pythonFile() {
...
+    }
+
+    @Override
+    public Set<Symbol> getBuiltinGlobalSymbols() {
```
andrea-guarino-sonarsource, line 164: I would rename it to `builtinSymbols()`

python-frontend/src/main/java/org/sonar/python/SubscriptionVisitor.java
```diff
@@ -140,6 +155,16 @@ public PythonFile pythonFile() {
       return pythonVisitorContext.pythonFile();
     }
 
+    @Override
+    public Map<String, Set<Symbol>> getTypeShedSymbols() {
+      return SubscriptionVisitor.getTypeShedSymbols();
+    }
+
+    @Override
```
andrea-guarino-sonarsource, line 163: Please add `@Beta` annotation

python-frontend/src/main/java/org/sonar/python/SubscriptionVisitor.java
```diff
@@ -140,6 +155,16 @@ public PythonFile pythonFile() {
       return pythonVisitorContext.pythonFile();
     }
 
+    @Override
```
andrea-guarino-sonarsource, line 158: Please add `@Beta`

python-frontend/src/main/java/org/sonar/python/SubscriptionVisitor.java
```diff
@@ -140,6 +155,16 @@ public PythonFile pythonFile() {
...
+    public Map<String, Set<Symbol>> getTypeShedSymbols() {
```
andrea-guarino-sonarsource, line 159: Because they don't reference only typeshed symbols, maybe I'd rename it to `stubFilesSymbols()`

python-frontend/src/main/java/org/sonar/plugins/python/api/SubscriptionContext.java
```diff
@@ -42,6 +45,10 @@
 
   PythonFile pythonFile();
 
+  Map<String, Set<Symbol>> stubFilesSymbols();
```
andrea-guarino-sonarsource, line 48: `@Beta` should be placed in the interface

andrea-guarino-sonarsource, line 48: do we really need to return a Map? AFAICT, ExternalTypeHierarchies in sonar-sec doesn't need to know anything about modules.
I think we can just return `Collection<Symbol>` with the union of all external symbols.
Moreover I think we can drop the 'builtinSymbols' API and merge them in this returned collection

python-frontend/src/main/java/org/sonar/python/SubscriptionVisitor.java
```diff
@@ -140,6 +145,18 @@ public PythonFile pythonFile() {
       return pythonVisitorContext.pythonFile();
     }
 
+    @Beta
```
andrea-guarino-sonarsource, line 148: I think we can remove `@Beta` here

python-frontend/src/main/java/org/sonar/plugins/python/api/SubscriptionContext.java
andrea-guarino-sonarsource, line 48: I believe this API deserves a comment

python-frontend/src/main/java/org/sonar/plugins/python/api/SubscriptionContext.java
```diff
@@ -42,6 +45,12 @@
 
   PythonFile pythonFile();
 
+  /**
+   * Returns externally declared symbols available in this context.
```
andrea-guarino-sonarsource, line 49: I would something like:
`Returns symbols declared in stub files (e.g. typeshed) used in the analyzed project.`

PR #782, opened 2020-06-11
SONARPY-740 Rule S5756: Calls should not be made to non-callable values

guillaume-dequenne-sonarsource:
Current implementation is missing secondary locations: they will be introduced through a new API based on flow sensitive type inference.


python-checks/src/test/resources/checks/nonCallableCalled.py
```diff
@@ -0,0 +1,113 @@
...
+    list_var()  # Noncompliant
+
+    tuple_var = ()
+    tuple_var()  # Noncompliant
+
+    dict_var = {}
+    dict_var()  # Noncompliant
+
+    set_var = set()
+    set_var()  # FN (set has unresolved type hierarchy)
```
andrea-guarino-sonarsource, line 30: maybe we can hardcode it in the rule?

python-checks/src/main/java/org/sonar/python/checks/NonCallableCalledCheck.java
```diff
@@ -0,0 +1,55 @@
...
+@Rule(key = "S5756")
+public class NonCallableCalledCheck extends PythonSubscriptionCheck {
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.CALL_EXPR, ctx -> {
+      CallExpression callExpression = (CallExpression) ctx.syntaxNode();
+      Expression callee = callExpression.callee();
+      if (!callee.type().canHaveMember("__call__")) {
+        Optional<String> nameOptional = nameFromExpression(callee);
```
andrea-guarino-sonarsource, line 39: minor: I think I would prefer `nameFromExpression` to return `null` in this case

python-checks/src/test/resources/checks/nonCallableCalled.py
guillaume-dequenne-sonarsource: Done in aaaac0c3236b8c3d45c1c19f26873e584d0a457d. That said, I assume there will be many other types in that situation.

PR #783, opened 2020-06-12
Update sonar-python APIs for MMF-1912

gyula-sallai-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/FunctionSymbolImpl.java
```diff
@@ -85,7 +85,7 @@ public void setParametersWithType(ParameterList parametersList) {
     hasVariadicParameter = functionSymbol.hasVariadicParameter();
     parameters.addAll(functionSymbol.parameters());
     functionDefinitionLocation = functionSymbol.definitionLocation();
-    declaredReturnType = ((FunctionSymbolImpl) functionSymbol).declaredReturnType();
+    declaredReturnType = functionSymbol.underlyingType();
```
vilchik-elena, line 88: `declaredReturnType` does not sound the same as `underlyingType`

python-frontend/src/main/java/org/sonar/plugins/python/api/types/InferredType.java
```diff
@@ -44,4 +44,7 @@
   @Beta
   boolean isCompatibleWith(InferredType other);
 
+  @Beta
+  Optional<Symbol> declaringSymbol();
```
andrea-guarino-sonarsource, line 48: Within ticket SONARPY-728 (https://github.com/SonarSource/sonar-python/pull/789), we're adding 
```java
Collection<Symbol> declaringSymbols();
``` 
to handle cases where we have more than one (i.e. union types).
Could you please rebase as soon as it will be merged and use that API instead?

python-frontend/src/main/java/org/sonar/plugins/python/api/symbols/ClassSymbol.java
```diff
@@ -33,6 +35,8 @@
 
   Set<Symbol> declaredMembers();
 
+  Map<Symbol, InferredType> fieldsWithTypes();
```
andrea-guarino-sonarsource, line 38: I'm not so sure about this API `fieldsWithTypes`.
I think I would prefer having an API on Symbol `InferredType declaredType()`

dmitry-golovinov-sonarsource: Well. We don't need this on anything but fields. Exposing this on symbol level - we will expose this information for functions and classes (probably, later for module symbols as well) - where it makes no sense. At the same time, I think that field could be ambiguous - so that we will miss it in this map. But they are useless for sonar python

dmitry-golovinov-sonarsource:
returnTypeHints for a function

python-frontend/src/main/java/org/sonar/plugins/python/api/types/InferredType.java
andrea-guarino-sonarsource: After discussion, we decided not to expose the "declaring Symbol" of an Inferred Type. 
We will instead expose "type hints" from FunctionSymbol

python-frontend/src/main/java/org/sonar/plugins/python/api/symbols/ClassSymbol.java
andrea-guarino-sonarsource: After discussion, we decided we will expose "type hints" of members in ClassSymbol, without exposing `InferredType`

python-frontend/src/main/java/org/sonar/plugins/python/api/symbols/FunctionSymbol.java
```diff
@@ -48,6 +50,9 @@
   @CheckForNull
   LocationInFile definitionLocation();
 
+  @Beta
+  Optional<ClassSymbol> returnTypeHint();
```
andrea-guarino-sonarsource, line 54: I think it should return a `Optional<Symbol>`.
In order to be consistent with AST, let's call it `returnTypeAnnotation`.

python-frontend/src/main/java/org/sonar/plugins/python/api/symbols/ClassSymbol.java
```diff
@@ -33,6 +34,9 @@
 
   Set<Symbol> declaredMembers();
 
+  @Beta
+  Map<Symbol, ClassSymbol> memberTypeHints();
```
andrea-guarino-sonarsource, line 38: I think it should return a `Map<Symbol, Symbol>`.
In order to be consistent with AST let's call it `memberTypeAnnotations`

python-frontend/src/main/java/org/sonar/python/semantic/FunctionSymbolImpl.java
```diff
@@ -225,6 +227,15 @@ public InferredType declaredReturnType() {
     return declaredReturnType;
   }
 
+  @Override
+  public Optional<ClassSymbol> returnTypeHint() {
+    if (declaredReturnType instanceof RuntimeType) {
```
andrea-guarino-sonarsource, line 232: I think we shouldn't return the typeClass of RuntimeType. I think we should instead return the symbol of the "TypeAnnotation" tree

python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
```diff
@@ -276,6 +279,23 @@ public void visitFunctionDef(FunctionDef functionDef) {
...
+        Optional.ofNullable(variable.symbol()).ifPresent(symbol -> setAnnotatedType(symbol, annotatedAssignment));
+      }
+      super.visitAnnotatedAssignment(annotatedAssignment);
+    }
+
+    private static void setAnnotatedType(Symbol symbol, AnnotatedAssignment annotatedAssignment) {
+      TypeAnnotation typeAnnotation = annotatedAssignment.annotation();
+      if (symbol.is(Symbol.Kind.OTHER)) {
+        SymbolImpl other = (SymbolImpl) symbol;
+        other.setInferredType(InferredTypes.declaredType(typeAnnotation));
```
andrea-guarino-sonarsource, line 295: I don't like that we use `setInferredType` for something that it's not inferred.
I think `SymbolImpl` should have a `typeAnnotation` setter, and we should set the symbol of `annotatedAssignment` instead of calling `InferredTypes.declaredType(typeAnnotation)`

python-frontend/src/main/java/org/sonar/plugins/python/api/symbols/Symbol.java
```diff
@@ -35,6 +36,10 @@
 
   Kind kind();
 
+  @Beta
+  @CheckForNull
+  Symbol annotatedType();
```
andrea-guarino-sonarsource, line 41: I think we should be consistent between `returnTypeAnnotation()` and `annotatedType()`: 
- either we return `Optional<Symbol>` for both 
- or we return a nullable Symbol for both.

Nitpick: Also names don't sound consistent: 
- either 'returnTypeAnnotation()' and 'typeAnnotation()' 
- or 'annotatedReturnType()' and 'annotatedType()'

python-frontend/src/main/java/org/sonar/python/semantic/Scope.java
```diff
@@ -151,7 +151,7 @@ private Symbol copySymbol(String symbolName, Symbol symbol) {
         .collect(Collectors.toSet());
       return new AmbiguousSymbolImpl(symbolName, symbol.fullyQualifiedName(), alternativeSymbols);
     } else if (symbol.is(Symbol.Kind.OTHER)) {
-      SymbolImpl res = new SymbolImpl(symbolName, symbol.fullyQualifiedName());
+      SymbolImpl res = new SymbolImpl(symbolName, symbol.fullyQualifiedName(), ((SymbolImpl) symbol).annotatedType());
```
andrea-guarino-sonarsource, line 154: I think `SymbolImpl` cast is unnecessary

andrea-guarino-sonarsource, line 154: I think we should also use this constructor in ProjectLevelSymbolTable.
https://github.com/SonarSource/sonar-python/blob/e6dd6dd1bbde8e461040bf59250670e6408a4ffa/python-frontend/src/main/java/org/sonar/python/semantic/ProjectLevelSymbolTable.java#L76

python-frontend/src/test/java/org/sonar/python/types/TypeTestUtils.java
```diff
@@ -0,0 +1,28 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.python.types;
+
+import org.sonar.plugins.python.api.symbols.ClassSymbol;
+
+public class TypeTestUtils {
+  public static RuntimeType getRuntimeType(ClassSymbol s) {
```
andrea-guarino-sonarsource, line 25: Do we really need a utility for this?
It looks like it's not used anywhere, did you forget to remove it maybe?

python-frontend/src/main/java/org/sonar/python/semantic/SymbolImpl.java
```diff
@@ -132,4 +153,16 @@ public void removeUsages() {
   Map<String, Symbol> getChildrenSymbolByName() {
     return Collections.unmodifiableMap(childrenSymbolByName);
   }
+
+  @Nullable
+  public static Symbol getTypeSymbolFromExpression(Expression expression) {
```
andrea-guarino-sonarsource, line 158: do we need this to be public? What about having `setAnnotatedTypeName` taking a TypeAnnotation tree instead as parameter and calling this function?

python-frontend/src/main/java/org/sonar/plugins/python/api/symbols/Symbol.java
```diff
@@ -35,6 +36,13 @@
 
   Kind kind();
 
+  /**
+   * Returns fully qualified name of the type if any
+   */
+  @Beta
+  @CheckForNull
+  String annotatedTypeName();
```
andrea-guarino-sonarsource, line 44: As discussed, I think it would be nice to add a unit test (for both `annotatedTypeName` and `annotatedReturnTypeName`) for ProjectLevelSymbolTable to assert we keep information about annotated types.
Something similair to what we did for metaclasses

https://github.com/SonarSource/sonar-python/blob/d299c45234efe70567244369e9d813d7dcdc839e/python-frontend/src/test/java/org/sonar/python/semantic/ProjectLevelSymbolTableTest.java#L266

dmitry-golovinov-sonarsource:
We decided not run ReturnTypeVisitor on Project Level Symbols as of now

python-frontend/src/test/java/org/sonar/python/PythonVisitorCheckTest.java
```diff
@@ -127,7 +130,27 @@ public void initialize(SubscriptionCheck.Context context) {
...
+  @Test
+  public void stubFilesSymbols() throws IOException {
+    Collection<Symbol> symbols = TypeShed.stubFilesSymbols();
+    PythonVisitorContext context = TestPythonVisitorRunner.createContext(FILE);
+
+    SymbolsRecordingCheck check = new SymbolsRecordingCheck();
+    check.scanFile(context);
+    SubscriptionVisitor.analyze(Collections.singletonList(check), context);
+
+    assertThat(check.symbols).isEqualTo(TypeShed.stubFilesSymbols());
```
guillaume-dequenne-sonarsource, line 142: I guess you could use `symbols` assigned earlier. Or otherwise remove the variable altogether as it is not used.

dmitry-golovinov-sonarsource: Fixed (removed the variable completely), but rewrote history as was asked earlier to keep history clean.

dmitry-golovinov-sonarsource:
> This PR is becoming a bit messy. Can you please keep
> 
>     * 1 commit for `stubFilesSymbols` API
> 
>     * 1 commit for type annotation related API
> 
>     * 1 commit for new custom stubs

History should be cleaner now.

PR #787, opened 2020-06-17
SONARPY-734 Fix memory leak in BuiltinShadowingAssignmentCheck (S5806)

andrea-guarino-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/BuiltinShadowingAssignmentCheck.java
```diff
@@ -66,6 +66,7 @@
 
   @Override
   public void initialize(Context context) {
+    variableIssuesRaised.clear();
```
guillaume-dequenne-sonarsource, line 69: In my opinion a clear at each visit of a `FILE_INPUT` would be a bit more explicit with regard to which lifespan we want for the map's content, but I don't have that strong an opinion on this.

andrea-guarino-sonarsource: Agreed, I will do that

PR #793, opened 2020-06-22
SONARPY-504 Rule S5607: Operators should be used on compatible types

andrea-guarino-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/IncompatibleOperandsCheck.java
```diff
@@ -0,0 +1,163 @@
...
+    InferredType type = unaryExpression.expression().type();
+    Token operator = unaryExpression.operator();
+    String memberName = UNARY_SPECIAL_METHODS_BY_OPERATOR.get(operator.value());
+    if (memberName != null && !type.canHaveMember(memberName)) {
+      addIssue(operator, "Fix this invalid \"" + operator.value() + "\" operation on a type which doesn't support it.");
+    }
+    super.visitUnaryExpression(unaryExpression);
+  }
+
+  private void checkComparisonOperands(Token operator, InferredType left, InferredType right) {
```
guillaume-dequenne-sonarsource, line 112: This method is used for more than comparison operations. It might be better to have more generic names to reflect that.

python-checks/src/main/java/org/sonar/python/checks/IncompatibleOperandsCheck.java
```diff
@@ -0,0 +1,168 @@
...
+
+  private void checkOperands(Token operator, InferredType left, InferredType right) {
+    if (typeSymbols(left).stream().anyMatch(ClassSymbol::hasDecorators) || typeSymbols(right).stream().anyMatch(ClassSymbol::hasDecorators)) {
+      return;
+    }
+    SpecialMethodNames specialMethodNames = BINARY_SPECIAL_METHODS_BY_OPERATOR.get(operator.value());
+    if (specialMethodNames == null) {
+      return;
+    }
+    boolean canHaveLessThan = left.canHaveMember(specialMethodNames.left);
```
guillaume-dequenne-sonarsource, line 122: I'm sorry it wasn't explicit in my previous comment but I believe those names (and also `hasIncompatibleLessThan` and `hasIncompatibleGreaterThan ` on line 135 & 136 should be more generic as well.

PR #794, opened 2020-06-22
SONARPY-733 Exclude metaclasses for S3862 & S5644

guillaume-dequenne-sonarsource:


python-checks/src/test/resources/checks/itemOperationsTypeCheck/itemOperations_getitem.py
```diff
@@ -144,19 +144,29 @@ def __getitem__(cls, key):
...
 
 def type_annotations():
   """No issue as type annotations do no call item methods"""
   from typing import Awaitable
   def my_func() -> Awaitable[bool]: ... # OK
   def my_other_func(arg: Awaitable[bool]): ... # OK
   x: Awaitable[bool] # OK
+
+
+def annotated_classes():
```
andrea-guarino-sonarsource, line 165: do you mean decorated class?

python-frontend/src/main/java/org/sonar/python/semantic/SymbolUtils.java
```diff
@@ -95,14 +95,31 @@ static void resolveTypeHierarchy(ClassDef classDef, @Nullable Symbol symbol, Pyt
       return;
     }
     for (Argument argument : argList.arguments()) {
-      Symbol argumentSymbol = getSymbolFromArgument(argument);
-      if (argumentSymbol == null) {
+      if (!argument.is(Kind.REGULAR_ARGUMENT)) {
         classSymbol.setHasSuperClassWithoutSymbol();
+        continue;
```
andrea-guarino-sonarsource, line 100: I think I would prefer having an `else` branch instead of `continue`

python-frontend/src/main/java/org/sonar/python/semantic/ClassSymbolImpl.java
```diff
@@ -150,9 +151,14 @@ public boolean hasUnresolvedTypeHierarchy(boolean includeAmbiguousSymbols) {
     return Optional.empty();
   }
 
+  @Override
+  public boolean hasMetaClass() {
+    return hasMetaClass;
```
andrea-guarino-sonarsource, line 156: It looks like in Python2 you define metaclasses with an attribute: https://docs.python.org/2.7/reference/datamodel.html#__metaclass__
Maybe we can have
```java
return hasMetaClass || membersByName.get("__metaclass__") != null
```

python-frontend/src/main/java/org/sonar/python/semantic/SymbolUtils.java
```diff
@@ -95,14 +95,31 @@ static void resolveTypeHierarchy(ClassDef classDef, @Nullable Symbol symbol, Pyt
...
+    Name keyword = regularArgument.keywordArgument();
+    if (keyword != null) {
+      if (keyword.name().equals("metaclass")) {
+        classSymbol.setHasMetaClass();
       } else {
-        Symbol normalizedArgumentSymbol = normalizeSymbol(argumentSymbol, pythonFile, symbolsByName);
-        if (normalizedArgumentSymbol != null) {
-          classSymbol.addSuperClass(normalizedArgumentSymbol);
-        }
+        classSymbol.setHasSuperClassWithoutSymbol();
```
andrea-guarino-sonarsource, line 112: it looks like keyword arguments don't alter type hierarchy (at least not explictely).
They are instead passed to `__init_subclass__()` method

https://docs.python.org/3/reference/datamodel.html#object.__init_subclass__

Note that in python 2 they are a syntax error.

python-frontend/src/main/java/org/sonar/python/semantic/SymbolUtils.java
```diff
@@ -204,9 +219,9 @@ public static boolean isTypeShedFile(PythonFile pythonFile) {
 
   /**
    * @return the offset between parameter position and argument position:
-   *   0 if there is no implicit first parameter (self, cls, etc...)
-   *   1 if there is an implicit first parameter
-   *  -1 if unknown (intent is not clear from context)
+   * 0 if there is no implicit first parameter (self, cls, etc...)
```
andrea-guarino-sonarsource, line 222: unnecessary change

PR #796, opened 2020-06-23
SONARPY-728 Improve issue reporting for S3862, S5644, S5756 and S5607

andrea-guarino-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/ItemOperationsTypeCheck.java
```diff
@@ -106,16 +116,18 @@ private static boolean isValidSubscription(Expression subscriptionObject, String
       if (symbol == null) {
         return true;
       }
-      if (symbol.is(Symbol.Kind.FUNCTION, Symbol.Kind.CLASS)) {
+      if (symbol.is(FUNCTION, Symbol.Kind.CLASS)) {
```
guillaume-dequenne-sonarsource, line 119: _Nitpick:_ `Symbol.Kind.CLASS` can be shortened to `CLASS` as well.

PR #808, opened 2020-07-07
SONARPY-749 Add new style importing of Pylint reports

guillaume-dequenne-sonarsource:


sonar-python-plugin/src/main/java/org/sonar/plugins/python/pylint/PylintRulesDefinition.java
```diff
@@ -0,0 +1,37 @@
...
+import org.sonar.api.server.rule.RulesDefinition;
+import org.sonar.plugins.python.Python;
+import org.sonarsource.analyzer.commons.ExternalRuleLoader;
+
+public class PylintRulesDefinition implements RulesDefinition {
+
+  private static final String RULES_JSON = "org/sonar/plugins/python/pylint/rules.json";
+
+  private static final ExternalRuleLoader RULE_LOADER = new ExternalRuleLoader(PylintSensor.LINTER_KEY, PylintSensor.LINTER_NAME, RULES_JSON, Python.KEY);
+
```
gyula-sallai-sonarsource, line 31: _nitpick_ I would drop this extra empty line.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/pylint/PylintReportReader.java
```diff
@@ -0,0 +1,63 @@
...
+import org.sonar.plugins.python.TextReportReader;
+
+public class PylintReportReader extends TextReportReader {
+
+  private static final Logger LOG = Loggers.get(PylintReportReader.class);
+  private static final Pattern PYLINT_DEFAULT = Pattern.compile("(.+):(\\d+):(\\d+): (\\S+): (.*)");
+  private static final Pattern PATTERN_LEGACY = Pattern.compile("(.+):(\\d+): \\[(.*)\\] (.*)");
+
+  protected Issue parseLine(String line) {
+
```
gyula-sallai-sonarsource, line 36: _nitpick_ I would drop this empty line.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/pylint/PylintReportReader.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.regex.Pattern;
+import org.sonar.api.utils.log.Logger;
+import org.sonar.api.utils.log.Loggers;
+import org.sonar.plugins.python.TextReportReader;
+
+public class PylintReportReader extends TextReportReader {
+
+  private static final Logger LOG = Loggers.get(PylintReportReader.class);
+  private static final Pattern PYLINT_DEFAULT = Pattern.compile("(.+):(\\d+):(\\d+): (\\S+): (.*)");
+  private static final Pattern PATTERN_LEGACY = Pattern.compile("(.+):(\\d+): \\[(.*)\\] (.*)");
```
gyula-sallai-sonarsource, line 33: The naming is a little inconsistent for me, I would prefer something like `DEFAULT_PATTERN` and `LEGACY_PATTERN`, or `PYLINT_DEFAULT` and `PYLINT_LEGACY`.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/flake8/Flake8ReportReader.java
```diff
@@ -19,89 +19,42 @@
...
+        String filePath = m.group(1);
+        int lineNumber = Integer.parseInt(m.group(2));
+        int columnNumber = Integer.parseInt(m.group(3));
+        // Flake8 issues are off by 1 compared to SQ conventions
+        columnNumber -= 1;
+        String ruleKey = m.group(4);
+        String message = m.group(5);
+        return new Issue(filePath, ruleKey, message, lineNumber, columnNumber);
       }
+      m = PYLINT_PATTERN.matcher(line);
```
gyula-sallai-sonarsource, line 48: I would add a comment here that justifies why are we using a Pylint pattern here (maybe also worth explaining that even though we have the same pattern in `PylintReportReader`, the implementation is slightly different).

sonar-python-plugin/src/main/java/org/sonar/plugins/python/ExternalIssuesSensor.java
```diff
@@ -62,7 +78,40 @@ private void logUnresolvedInputFiles(Set<String> unresolvedInputFiles) {
     logger().warn("Failed to resolve {} file path(s) in " + linterName() + " report. No issues imported related to file(s): {}", unresolvedInputFiles.size(), fileList);
   }
 
-  protected abstract void importReport(File reportPath, SensorContext context, Set<String> unresolvedInputFiles);
+  private void logFileCantBeRead(Exception e, File reportPath) {
+    logger().error("No issues information will be saved as the report file '{}' can't be read. " +
+      e.getClass().getSimpleName() + ": " + e.getMessage(), reportPath, e);
```
gyula-sallai-sonarsource, line 83: It's a little weird to me that we are mixing string formatting and string concatenation here.

sonar-python-plugin/src/test/java/org/sonar/plugins/python/pylint/PylintSensorTest.java
```diff
@@ -0,0 +1,238 @@
...
+      String projectId = projectDir.getFileName().toString() + "-project";
+      context.fileSystem().add(TestInputFileBuilder.create(projectId, projectDir.toFile(), file.toFile())
+        .setCharset(UTF_8)
+        .setLanguage(language(file))
+        .setContents(new String(Files.readAllBytes(file), UTF_8))
+        .setType(InputFile.Type.MAIN)
+        .build());
+    } catch (IOException e) {
+      throw new IllegalStateException(e);
+    }
```
gyula-sallai-sonarsource, line 220: Not sure I understand why we are doing this, `executeSensorImporting` has already declared that it may throw `IOException`

sonar-python-plugin/src/main/java/org/sonar/plugins/python/flake8/Flake8ReportReader.java
guillaume-dequenne-sonarsource: After having a second look at this, the different formats are only subtly different and I decided to simply make a more generic version of the regexes and logic in `TextReportReader` rather than keep the 2 implementations.

sonar-python-plugin/src/test/java/org/sonar/plugins/python/pylint/PylintSensorTest.java
guillaume-dequenne-sonarsource: After double checking it seems that it's because this method is called in a `forEach` statement, and because of that the `IOException` it might throw would not be handled by the `executeSensorImporting` throw declaration.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/TextReportReader.java
```diff
@@ -31,11 +31,19 @@
 import org.sonar.api.utils.log.Logger;
 import org.sonar.api.utils.log.Loggers;
 
-public class Flake8ReportReader {
+/**
+ * Common implementation to parse Flake8 and Pylint reports
+ */
+public class TextReportReader {
+
+  private final int reportOffset;
```
gyula-sallai-sonarsource, line 39: _nitpick_ I would add an empty line here just to indicate that this is an instance field. Maybe also would move it below the static fields.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/TextReportReader.java
```diff
@@ -31,11 +31,19 @@
...
+  protected static final Pattern DEFAULT_PATTERN = Pattern.compile("(.+):(\\d+):(\\d+): (\\S+[^:]):? (.*)");
+  protected static final Pattern LEGACY_PATTERN = Pattern.compile("(.+):(\\d+): \\[(.*)\\] (.*)");
+  private static final Logger LOG = Loggers.get(TextReportReader.class);
 
-  private static final Logger LOG = Loggers.get(Flake8ReportReader.class);
-  private static final Pattern DEFAULT_PATTERN = Pattern.compile("(.+):(\\d+):(\\d+): (\\S+) (.*)");
-  private static final Pattern PYLINT_PATTERN = Pattern.compile("(.+):(\\d+): \\[(.*)\\] (.*)");
+  public TextReportReader(int columnStartIndex) {
```
gyula-sallai-sonarsource, line 44: _minor, subjective_ Having to call this constructor with some magic numbers seems a little off to me. I think it would look nicer if we introduced integer constants and used them something like this:
```
new TextReportReader(TextReportReader.COLUMN_ZERO_BASED)
new TextReportReader(TextReportReader.COLUMN_ONE_BASED)
```

Feel free to disregard if you feel that this is too much.

PR #811, opened 2020-07-13
SONARPY-753 Import Pylint rules metadata

guillaume-dequenne-sonarsource:


sonar-python-plugin/src/test/java/org/sonar/plugins/python/pylint/PylintRulesDefinitionTest.java
```diff
@@ -43,7 +43,7 @@ public void pylint_external_repository() {
     RulesDefinition.Rule rule = repository.rule("C0121");
     assertThat(rule).isNotNull();
     assertThat(rule.name()).isEqualTo("Singleton comparison");
-    assertThat(rule.htmlDescription()).isEqualTo("This is external rule <code>pylint:C0121</code>. No details are available.");
+    assertThat(rule.htmlDescription()).isEqualTo("Used when an expression is compared to singleton values like True, False or None.");
```
pynicolas, line 46: I'm not sure whether we want that: is that what we did for other analyzers? Should we keep those descriptions up-to-date with newer versions of pylint?

sonar-python-plugin/src/test/java/org/sonar/plugins/python/pylint/PylintSensorTest.java
```diff
@@ -175,6 +177,7 @@ public void issues_other_formats() throws IOException {
     assertThat(first.ruleKey()).hasToString("external_pylint:E1300");
     assertThat(first.type()).isEqualTo(RuleType.CODE_SMELL);
     assertThat(first.severity()).isEqualTo(Severity.MAJOR);
+    assertThat(first.remediationEffort()).isEqualTo(3);
```
pynicolas, line 180: I think we should also update integration tests to ensure that remediation costs are correctly saved.

guillaume-dequenne-sonarsource:
After further discussion, we don't want to keep maintaining those existing metadata and we'll drop them.
I'm therefore closing this PR.

PR #812, opened 2020-07-13
SONARPY-694 S3330 should raise a Security Hotpost when Flask is configured to create session cookie with HttpOnly=False

dmitry-golovinov-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/hotspots/HttpOnlyCookieCheck.java
```diff
@@ -51,4 +66,91 @@ String message() {
...
+      ctx.addIssue(assignmentStatement.assignedValue(), message());
+    }
+  }
+
+  private void dictionarySessionCookieHttponlyCheck(SubscriptionContext ctx) {
+    DictionaryLiteral dict = (DictionaryLiteral) ctx.syntaxNode();
+    Optional<Expression> falsySetting = searchForFalsySessionCookieHttponlyInDictionary(dict);
+    if (falsySetting.isPresent()) {
+      ctx.addIssue(falsySetting.get(), message());
+    }
```
guillaume-dequenne-sonarsource, line 106: You could replace this by:

```falsySetting.ifPresent(expression -> ctx.addIssue(expression, message()));```

File: python-checks/src/main/java/org/sonar/python/checks/hotspots/HttpOnlyCookieCheck.java [hunk 2]
```diff
@@ -51,4 +66,91 @@ String message() {
...
+  }
+
+  private static final String SESSION_COOKIE_HTTPONLY = "SESSION_COOKIE_HTTPONLY";
+
+  private void dictConstructorSessionCookieHttponlyCheck(SubscriptionContext ctx) {
+    CallExpression callExpression = (CallExpression) ctx.syntaxNode();
+    Optional<Expression> falsySetting = searchForFalsySessionCookieHttponlyInDictCons(callExpression);
+    if (falsySetting.isPresent()) {
+      ctx.addIssue(falsySetting.get(), message());
+    }
```
guillaume-dequenne-sonarsource, line 128: Same here

PR #813, opened 2020-07-15
SONARPY-752 Update documentation for Pylint report importing

guillaume-dequenne-sonarsource:


sonar-python-plugin/src/main/resources/static/documentation.md
```diff
@@ -19,34 +19,15 @@ Discover and update the Python-specific [properties](/analysis/analysis-paramete
 ## Pylint
 [Pylint](http://www.pylint.org/) is an external static source code analyzer, it can be used in conjunction with SonarSource Python analyzer.
 
-You can enable Pylint rules directly in your Python Quality Profile. Their rule keys start with "*Pylint:*".
+It is possible to import Pylint reports by passing a comma-delimited list of report paths to the `sonar.python.pylint.reportPaths` property.
```
nicolas-harraudeau-sonarsource, line 22: I would personally remove the whole "Pylint" section on this page. We have nothing for flake8 and we now support the default pylint format so we don't need to provide a command, or do we? If we only need to tell users to use a specific option it can be explained on the "Importing External Issues" page.

nicolas-harraudeau-sonarsource, line 22: I would personally remove the whole "Pylint" section on this page. We have nothing for flake8 and we now support the default pylint format so we don't need to provide a command, or do we? If we only need to tell users to use a specific option it can be explained on the "Importing External Issues" page.

guillaume-dequenne-sonarsource: From my understanding the default format can vary depending on the version of Pylint, and there can be many customized format that we don't support as well. 
So just to make sure users know which one to use, I felt like keeping the information somewhere (but I agree here might not be the best place).

PR #819, opened 2020-07-27
Log execution time of creation of project symbol table

guillaume-dequenne-sonarsource:


sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonScanner.java
```diff
@@ -87,8 +87,12 @@ public PythonScanner(
     this.parser = PythonParser.create();
 
     // computes "globalSymbolsByModuleName"
+    long startTime = System.currentTimeMillis();
+    LOG.debug("Starting to build project level symbol table");
```
andrea-guarino-sonarsource, line 91: Given that we already have a message saying `Starting global symbols computation` , do we really need this?

guillaume-dequenne-sonarsource: Indeed we don't!

PR #823, opened 2020-08-04
SONARPY-773 Add type names for Union types

guillaume-dequenne-sonarsource:


python-checks/src/test/resources/checks/incompatibleOperands/comparison.py
```diff
@@ -10,7 +10,7 @@ def builtin_noncompliant(p):
     x = 42
   else:
     x = complex(1)
-  x > "1" # Noncompliant {{Fix this invalid ">" operation between incompatible types.}}
+  x > "1" # Noncompliant {{Fix this invalid ">" operation between incompatible types (Union[int, complex] and str).}}
```
andrea-guarino-sonarsource, line 13: I'm not sure we want to do that, sometimes we infer a union type because of a limitation of our analyzer. In that case message might be somewhat confusing.
For example:
```python
def f():
  try:
   x = 42
   x > "1" # Noncompliant {{Fix this invalid ">" operation between incompatible types (Union[int, complex] and str).}}
   x = complex(1)
  except:
    ...
```

guillaume-dequenne-sonarsource: Indeed, I forgot about this kind of case. Would you suggest we don't display Union type names due to that?
Maybe we can find some heuristic to display them only when we are sure they are actual union types (i.e either when coming from TypeShed or from flow sensitive type inference)...Wdyt?

andrea-guarino-sonarsource: Maybe, however to me it's not clear how to easily apply those heuristic, do you have some ideas?
In my opinion we can avoid display Union type names for the time being also because in MMF-1975 we plan to have secondary locations explaining type inference steps.

guillaume-dequenne-sonarsource: After further discussion, another problematic case would be the following:
```
def call_something():
  import re
  x = re.match("hello", "hello")
  x() # Noncompliant {{Fix this call; "x" has type Union[Match, Match, NoneType] and it is not callable.}}
```

We also create union types when we have an ambiguity between a Python2 and Python3 symbol, which we don't want to display as such. Ensuring correct behavior here might require a bit more effort and is not so high priority, so I'll close this PR for now.

PR #825, opened 2020-08-06
SONARPY-751 Rule S5886: Function return types should be consistent with their type hint

guillaume-dequenne-sonarsource:


python-checks/src/test/resources/checks/functionReturnType.py
```diff
@@ -0,0 +1,228 @@
...
+  def my_int_returns_none() -> int:
+    return  # Noncompliant
+
+  def my_int_union_type_ok(cond) -> int:
+    if cond:
+      x = 42
+    else:
+      x = "hello"
+    return x  # OK
+
```
andrea-guarino-sonarsource, line 109: I think we could add a test case:
```python
  def foo(cond) -> int:
    if cond:
      return 42
    else:
      return "hello"
```

guillaume-dequenne-sonarsource: There is a test case similar to this one on line 94-98, which does flag the `str` return as `Noncompliant`.

andrea-guarino-sonarsource: ops sorry I missed it

python-frontend/src/main/java/org/sonar/python/types/InferredTypes.java
```diff
@@ -149,23 +150,32 @@ private static DeclaredType declaredTypeFromTypeAnnotation(Expression expression
...
+    }
     return null;
   }
 
   private static InferredType runtimeTypefromTypeAnnotation(Expression expression, Map<String, Symbol> builtinSymbols) {
     if (expression.is(Kind.NAME) && !((Name) expression).name().equals("Any")) {
       Symbol symbol = ((Name) expression).symbol();
-      if (symbol != null && "typing.Text".equals(symbol.fullyQualifiedName())) {
-        return InferredTypes.runtimeType(builtinSymbols.get("str"));
+      if (symbol != null ) {
```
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: We still branch depending on `symbol.fullyQualifiedName()` right after so I think we still need the null check.

PR #827, opened 2020-08-10
SONARPY-775 Fix FN on S5797: flow sensitive inference not enabled for nested try/except

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/types/TypeInference.java
```diff
@@ -98,10 +99,19 @@ private static void inferTypesAndMemberAccessSymbols(FunctionDef functionDef, Py
...
-    if (TreeUtils.hasDescendant(functionDef, tree -> tree.is(Tree.Kind.TRY_STMT))) {
+    TryStatementVisitor tryStatementVisitor = new TryStatementVisitor(functionDef);
+    functionDef.accept(tryStatementVisitor);
+    if (tryStatementVisitor.hasTryStatement) {
       // CFG doesn't model precisely try-except statements. Hence we fallback to AST based type inference
       visitor.processPropagations(trackedVars);
       functionDef.accept(new BaseTreeVisitor() {
+        @Override
+        public void visitFunctionDef(FunctionDef visited) {
+          if (functionDef == visited) {
```
andrea-guarino-sonarsource, line 110: I think that to avoid this condition we can call the visitor on the body like `functionDef.body().accept(new BaseTreeVisitor() {})` at line 107

python-frontend/src/main/java/org/sonar/python/types/TypeInference.java
```diff
@@ -118,6 +128,32 @@ public void visitName(Name name) {
...
+    }
+
+    @Override
+    public void visitClassDef(ClassDef classDef) {
+      // don't visit nested classes
+    }
+
+    @Override
+    public void visitFunctionDef(FunctionDef visited) {
+      if (functionDef == visited) {
```
andrea-guarino-sonarsource, line 146: Also here, I think we can call the visitor on the body of the function

PR #828, opened 2020-08-10
SONARPY-743 RSPEC-5864 Calling a non-callable type

andrea-guarino-sonarsource:


its/ruling/src/test/resources/expected/python-S5607.json
```diff
@@ -1,4 +1,11 @@
 {
+'project:biopython/Bio/Align/AlignInfo.py':[
+562,
+610,
+],
```
guillaume-dequenne-sonarsource, line 5: Nice issues 😃

python-checks/src/main/java/org/sonar/python/checks/ConfusingTypeCheckingCheck.java
```diff
@@ -0,0 +1,49 @@
...
+  @Override
+  public void initialize(Context context) {
+    new NonCallableCalledCheck().initialize(context);
+  }
+
+  private static class NonCallableCalledCheck extends NonCallableCalled {
+
+    @Override
+    public boolean isNonCallableType(InferredType type) {
+      return type.canHaveMember("__call__") && !type.declareMember("__call__");
```
guillaume-dequenne-sonarsource, line 38: I'm a bit confused by `type.canHaveMember("__call__")`. Is this a way of asserting that we're dealing with a `DeclaredType`? It feels a bit indirect and implementation dependent to me.

That said I'm not sure I have a suggestion for a more direct check (checking directly that `type` is an instance of `DeclaredType` should do the trick but it might be heavy-handed...but then again this rule is tailored for `DeclaredType` only so it might make sense)

python-checks/src/main/java/org/sonar/python/checks/NonCallableCalled.java
```diff
@@ -0,0 +1,77 @@
...
+import org.sonar.plugins.python.api.tree.Tree;
+import org.sonar.plugins.python.api.types.InferredType;
+import org.sonar.python.types.InferredTypes;
+
+import static org.sonar.python.types.InferredTypes.typeClassLocation;
+
+public abstract class NonCallableCalled extends PythonSubscriptionCheck {
+
+  // List of non callable types with unresolved type hierarchies
+  protected static final List<String> NON_CALLABLE_TYPES = Arrays.asList("set", "frozenset");
```
guillaume-dequenne-sonarsource, line 39: Following the fix on `copyWithoutUsages`, I think you can remove this.

python-checks/src/test/resources/checks/confusingTypeChecking/nonCallableCalled.py
```diff
@@ -0,0 +1,21 @@
...
+  param3() # Noncompliant
+  fs = frozenset()
+  fs() # OK, raised by S5756
+
+def derived(x: int):
+  x.conjugate()() # FN
+  y = x or 'str'
+  y() # Noncompliant
+  z = x + 'str'
+  z() # FN?
```
guillaume-dequenne-sonarsource, line 21: Not sure we should consider this a FN. We should raise a code smell for the impossible `+` operation above, though, then we shouldn't be able to resolve a type for `z` I assume.
I believe this one would be a FN though (we should infer a DeclaredType of `str` for `y`):
```
def foo(x: str):
  y = x + "hello"
  y() # FN
```

python-frontend/src/main/java/org/sonar/plugins/python/api/types/InferredType.java
```diff
@@ -32,6 +32,13 @@
   @Beta
   boolean canHaveMember(String memberName);
 
+  /**
+   * Used to handle uncertainty of declared types. It return false when a member is not present in a declared type,
+   * while canHaveMember returns False because one of its subtype may have it.
+   */
+  @Beta
+  boolean declareMember(String memberName);
```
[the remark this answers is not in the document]
andrea-guarino-sonarsource: I named it like this to be consistent with `resolveMember`. I think we're already inconsistent (we have `isCompatibleWith`), not sure which one we should prefer

python-frontend/src/main/java/org/sonar/plugins/python/api/types/InferredType.java
```diff
@@ -32,6 +32,13 @@
   @Beta
   boolean canHaveMember(String memberName);
 
+  /**
+   * Used to handle uncertainty of declared types. It return false when a member is not present in a declared type,
+   * while canHaveMember returns False because one of its subtype may have it.
```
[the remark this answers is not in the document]
andrea-guarino-sonarsource: Indeed :) Thanks!

python-checks/src/test/resources/checks/confusingTypeChecking/nonCallableCalled.py
andrea-guarino-sonarsource: yes indeed, I agree

python-checks/src/main/java/org/sonar/python/checks/ConfusingTypeCheckingCheck.java
andrea-guarino-sonarsource: Yes, the idea is to avoid raising twice the same issue (one for the bug rule and one for code smell).

I think we cannot check only for `type` being instance of `DeclaredType`. It could also be a `UnionType` containing `DeclaredType` and we still want to raise an issue.
I agree it's a bit indirect; if you agree I'll add a comment explaining this

PR #832, opened 2020-08-11
SONARPY-751 S5886: Fix FPs on iterators and unknown type aliases

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/types/DeclaredType.java
```diff
@@ -83,6 +83,9 @@ public boolean isIdentityComparableWith(InferredType other) {
   @Override
   public boolean canBeOrExtend(String typeName) {
     if (typeName.equals("typing.Tuple") || typeName.equals("tuple")) {
```
andrea-guarino-sonarsource, line 85: as discussed I think we shouldn't have this logic in the API. I think we can accept the FN on tuple for the time being and having a proper fix for return statement tree

PR #839, opened 2020-08-18
SONARPY-743 Avoid raising FP when 'isinstance' is used

andrea-guarino-sonarsource:


python-frontend/src/test/java/org/sonar/python/types/TypeInferenceTest.java
```diff
@@ -524,4 +524,60 @@ public void execution_order_assignment_statement() {
...
   }
+
+  @Test
+  public void isinstance_flow_sensitive() {
+    assertThat(lastExpression(
+      "def f(x: int):",
+      "  if isinstance(x, Foo):",
+      "    ...",
+      "  x"
+      ).type()).isEqualTo(anyType());
```
guillaume-dequenne-sonarsource, line 535: Is that correct? Here the last `x` is independent from the `isinstance` check.

I see why we'd want to stay conservative here and have `x` as `ANY` nevertheless, but I think a test case within the `if` body would still be useful. 

In the future when we have more confidence about `isinstance` checks, I could see the type of `x` here remain `DECL_INT`.

andrea-guarino-sonarsource: Makes sense, I will add a test to check inside the if branch body. I think it still makes sense to keep this test, so that it will break as soon as we'll implement `isinstance` properly

PR #842, opened 2020-08-18
SONARPY-745 S5864: Iteration operation on a non-iterable type

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/ConfusingTypeCheckingCheck.java
```diff
@@ -106,4 +109,38 @@ public String message(@Nullable String name, String missingMethod) {
...
+      String typeNameString = typeName != null ? String.format("has type \"%s\" and", typeName) : "";
+      return isForLoop && isAsyncIterable(expression) ?
+        String.format("Add \"async\" before \"for\"; Previous type checks suggest that %s %s is an async generator.", expressionNameString, typeNameString) :
+        String.format("Replace this expression; Previous type checks suggest that %s %s isn't iterable.", expressionNameString, typeNameString);
+    }
+
+    @Override
+    boolean isAsyncIterable(Expression expression) {
+      InferredType type = expression.type();
+      return type.declaresMember("__aiter__");
```
andrea-guarino-sonarsource, line 136: Here we don't need to check `containsDeclaredType` because it's called only in method `message` right? Can you add a comment for that?

PR #847, opened 2020-08-20
SONARPY-778 Rule S5890: Values assigned to variables should match their type annotations

guillaume-dequenne-sonarsource:


python-checks/src/test/resources/checks/inconsistentTypeHintCheck.py
```diff
@@ -0,0 +1,109 @@
...
+  """We should avoid raising FPs on type aliases"""
+  my_float: SupportsFloat = 42  # OK
+  my_iterable: Iterable = []  # OK
+  my_generator: Generator = get_generator()  # OK
+
+
+def collections():
+  my_list: List = {}  # Noncompliant
+
+  my_str_list_nok: List[str] = [1, 2, 3]  # Out of scope
```
andrea-guarino-sonarsource, line 66: I think we can mark it as FN, as we'll probably want to detect it when we'll implement MMF-1977

python-checks/src/main/java/org/sonar/python/checks/InconsistentTypeHintCheck.java
```diff
@@ -0,0 +1,77 @@
...
+      AnnotatedAssignment annotatedAssignment = (AnnotatedAssignment) ctx.syntaxNode();
+      Expression assignedExpression = annotatedAssignment.assignedValue();
+      if (assignedExpression == null) {
+        return;
+      }
+      checkAnnotatedAssignment(ctx, annotatedAssignment, assignedExpression);
+    });
+  }
+
+  private static void checkAnnotatedAssignment(org.sonar.plugins.python.api.SubscriptionContext ctx, AnnotatedAssignment annotatedAssignment, Expression assignedExpression) {
```
andrea-guarino-sonarsource, line 50: why did you use the fully qualified name for SubscriptionContext?

python-checks/src/test/resources/checks/inconsistentTypeHintCheck.py
```diff
@@ -0,0 +1,109 @@
+from typing import SupportsFloat, List, Iterable, Generator, Set, Union, Type
```
andrea-guarino-sonarsource, line 1: minor: usually we don't add the `Check` suffix in the python file name

andrea-guarino-sonarsource (approved):
LGTM, please check some minor comments

PR #849, opened 2020-08-24
SONARPY-777 Fix FP on S1764 when expressions contain function calls

guillaume-dequenne-sonarsource:


python-checks/src/test/resources/checks/identicalExpressionOnBinaryOperator.py
```diff
@@ -41,3 +41,21 @@ def work():
...
+  if foo().x == foo().x: ...
+  if MyClass() == MyClass(): ...
+  my_class = MyClass()
+  if my_class.bar() == my_class.bar(): ...
+
+def no_issues_within_try_except():
+  try:
+    foo(c)
+  except ValueError:
+    return c / c
```
andrea-guarino-sonarsource, line 61: why do we want to exclude this one?

guillaume-dequenne-sonarsource: I didn't keep the comment for this one and simplified a bit the example, but it's an actual case where they want to assess the value of parameter `c`: `0/0` which would raise an exception `inf/inf` which would return `nan` and any other number which would return `1`.
I'm not sure if there are other legitimate uses for this construct, but I prefer to avoid being noisy in try/except blocks for that reason.

andrea-guarino-sonarsource: ok..as discussed please create a separate ticket for the try/except FP

PR #853, opened 2020-08-25
SONARPY-786 RSPEC-3827: Extract detection of never defined variable i…

andrea-guarino-sonarsource:
…n a separate rule (RSPEC-5953)


python-checks/src/main/java/org/sonar/python/checks/UnresolvedSymbolsCheck.java
```diff
@@ -0,0 +1,94 @@
...
+import org.sonar.plugins.python.api.SubscriptionContext;
+import org.sonar.plugins.python.api.tree.BaseTreeVisitor;
+import org.sonar.plugins.python.api.tree.CallExpression;
+import org.sonar.plugins.python.api.tree.FileInput;
+import org.sonar.plugins.python.api.tree.ImportFrom;
+import org.sonar.plugins.python.api.tree.Name;
+import org.sonar.plugins.python.api.tree.Tree;
+
+@Rule(key = "S5953")
+public class UnresolvedSymbolsCheck extends PythonSubscriptionCheck {
```
guillaume-dequenne-sonarsource, line 37: I'd suggest `UndefinedSymbolsCheck` or `UnknownSymbolsCheck`. I feel like `UnresolvedSymbolsCheck` implies we actually want to raise on symbols we are unable to resolve due to technical limitations of our analyzer.

python-checks/src/main/java/org/sonar/python/checks/UnresolvedSymbolsCheck.java
```diff
@@ -0,0 +1,94 @@
...
+    @Override
+    public void visitName(Name name) {
+      if (name.isVariable() && name.symbol() == null) {
+        nameIssues.computeIfAbsent(name.name(), k -> new ArrayList<>()).add(name);
+      }
+    }
+
+    @Override
+    public void visitImportFrom(ImportFrom importFrom) {
+      hasUnresolvedWildcardImport |= importFrom.hasUnresolvedWildcardImport();
```
guillaume-dequenne-sonarsource, line 81: Maybe we can consider avoiding raising issues on files which use any wildcard imports. Even though the rule is going out of Sonar_way, I feel our resolution of such import is not strong enough to raise issues here (if `moduleA` wildcard imports `moduleB` and then `moduleC` wildcard imports `moduleA`, we will fail to resolve symbols from `moduleB` despite them being available to use (and `hasUnresolvedWildcardImport()` won't necessarily be `true` if I'm not mistaken).
 
Anyway, if you agree with this, I believe it can be done in a separate ticket.

guillaume-dequenne-sonarsource (approved):
lgtm, minor comments

python-checks/src/main/java/org/sonar/python/checks/UnresolvedSymbolsCheck.java
andrea-guarino-sonarsource: Agreed, Let's handle this in a separate ticket

python-checks/src/main/java/org/sonar/python/checks/UnresolvedSymbolsCheck.java
andrea-guarino-sonarsource: Ok, I'll rename it to `UndefinedSymbolsCheck`

PR #854, opened 2020-08-25
SONARPY-690 Update default regex for S101 to not raise on valid PEP-8…

guillaume-dequenne-sonarsource:
… class names


andrea-guarino-sonarsource (approved):
LGTM, don't forget to update RSPEC as mentioned in SONARPY-690

PR #856, opened 2020-08-26
SONARPY-788 Fix parsing error on annotated assignment with yield expr…

guillaume-dequenne-sonarsource:
…ession

### DIFF
--- a/python-frontend/src/main/java/org/sonar/python/api/PythonGrammar.java
+++ b/python-frontend/src/main/java/org/sonar/python/api/PythonGrammar.java
@@ -186,8 +186,8 @@ public enum PythonGrammar implements GrammarRuleKey {
         b.sequence(AUGASSIGN, b.firstOf(YIELD_EXPR, TESTLIST)),
         b.zeroOrMore("=", b.firstOf(YIELD_EXPR, TESTLIST_STAR_EXPR))));
 
-    // https://github.com/python/cpython/blob/v3.6.4/Grammar/Grammar#L43
-    b.rule(ANNASSIGN).is(":", TEST, b.optional( "=", b.firstOf(YIELD_EXPR, TEST)));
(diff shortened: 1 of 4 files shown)


python-frontend/src/main/java/org/sonar/python/api/PythonGrammar.java
```diff
@@ -187,7 +187,7 @@ public static void grammar(LexerfulGrammarBuilder b) {
         b.zeroOrMore("=", b.firstOf(YIELD_EXPR, TESTLIST_STAR_EXPR))));
 
     // https://github.com/python/cpython/blob/v3.6.4/Grammar/Grammar#L43
-    b.rule(ANNASSIGN).is(":", TEST, b.optional( "=", TEST));
+    b.rule(ANNASSIGN).is(":", TEST, b.optional( "=", b.firstOf(YIELD_EXPR, TEST)));
```
merlin-s, line 190: The comment should probably updated to
 https://github.com/python/cpython/blob/v3.8.0/Grammar/Grammar#L87

Also the rule should probably read:
b.rule(ANNASSIGN).is(":", TEST, b.optional( "=", b.firstOf(YIELD_EXPR, TESTLIST_STAR_EXPR)));

This is not directly related to this PR, but important nontheless.

guillaume-dequenne-sonarsource: Hello,

Thank you for the feedback! Do you know why it should be a `TESTLIST_STAR_EXPR` though? I saw the grammar you are referring to, but I couldn't find a valid Python example that couldn't be parsed using simply a `TEST`.

Using `TESTLIST_STAR_EXPR` here requires some additional changes in the `PythonTreeMaker`, which I was not willing to do if it's just a case of "the grammar permits it but further syntactic checks prevent it".

andrea-guarino-sonarsource (approved):
LGTM

PR #861, opened 2020-08-28
SONARPY-744 Avoid FP when subscription object is used inside 'in expr…

andrea-guarino-sonarsource:
…essions'

### DIFF
--- a/python-checks/src/main/java/org/sonar/python/checks/ConfusingTypeCheckingCheck.java
+++ b/python-checks/src/main/java/org/sonar/python/checks/ConfusingTypeCheckingCheck.java
@@ -30,11 +30,13 @@ import org.sonar.plugins.python.api.SubscriptionContext;
 import org.sonar.plugins.python.api.symbols.ClassSymbol;
 import org.sonar.plugins.python.api.symbols.Symbol;
 import org.sonar.plugins.python.api.tree.Expression;
+import org.sonar.plugins.python.api.tree.InExpression;
 import org.sonar.plugins.python.api.tree.IsExpression;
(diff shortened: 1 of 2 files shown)


python-checks/src/main/java/org/sonar/python/checks/ConfusingTypeCheckingCheck.java
```diff
@@ -110,7 +112,21 @@ public boolean isValidSubscription(Expression subscriptionObject, String require
...
+
+    private static boolean isUsedInsideInExpression(Expression subscriptionObject) {
+      return TreeUtils.getSymbolFromTree(subscriptionObject)
+        .filter(symbol -> symbol.usages().stream().anyMatch(usage -> isRightOperandInExpression(usage.tree())))
+        .isPresent();
+    }
+
+    private static boolean isRightOperandInExpression(Tree tree) {
+      Tree parent = tree.parent();
+      if (parent instanceof InExpression) {
```
guillaume-dequenne-sonarsource, line 126: _Minor_: I think you can use `Tree::is` here instead.

guillaume-dequenne-sonarsource (approved):
LGTM, 1 minor comment

python-checks/src/main/java/org/sonar/python/checks/ConfusingTypeCheckingCheck.java
andrea-guarino-sonarsource: If you don't mind, I'd prefer to use `instanceof` here so that it will also work in case of parent being `null`

PR #876, opened 2021-01-19
SONARPY-818 Parser should allow decorators to be any valid expression

guillaume-dequenne-sonarsource:


andrea-guarino-sonarsource (changes requested):
I believe we shouldn't have `nameTreesFromExpression` as a public utility method.
Also I think it's better to deprecate also arguments, rightPar and leftPar

python-frontend/src/main/java/org/sonar/python/semantic/FunctionSymbolImpl.java
```diff
@@ -107,18 +111,19 @@ FunctionSymbolImpl copyWithoutUsages() {
 
   private static boolean isInstanceMethod(FunctionDef functionDef) {
     return !functionDef.name().name().equals("__new__") && functionDef.isMethodDefinition() && functionDef.decorators().stream()
-      .map(decorator -> {
-        List<Name> names = decorator.name().names();
-        return names.get(names.size() - 1).name();
-      })
+      .map(decorator -> TreeUtils.decoratorNameFromExpression(decorator.expression()))
+      .filter(Objects::nonNull)
+      .flatMap(s -> Arrays.stream(s.split("\\.")))
```
andrea-guarino-sonarsource, line 116: I think we don't need this `flatMap` . We want to check that decorator name is equal to "staticmethod" or "classmethod"

python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -648,8 +648,6 @@ public void visitQualifiedExpression(QualifiedExpression qualifiedExpression) {
 
     @Override
     public void visitDecorator(Decorator decorator) {
```
andrea-guarino-sonarsource, line 650: I believe we can remove `visitDecorator` method

python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -648,8 +648,6 @@ public void visitQualifiedExpression(QualifiedExpression qualifiedExpression) {
...
-      Name nameTree = decorator.name().names().get(0);
```
andrea-guarino-sonarsource, line 651: I think that we can remove this lines because decorator name is now a variable. I think it would be nice to add a test case inside `NameImpl` to assert that

andrea-guarino-sonarsource (approved):
LGTM, some minor comment + a unit test we could add

python-frontend/src/main/java/org/sonar/python/tree/TreeUtils.java
```diff
@@ -273,6 +275,31 @@ public static String nameFromExpression(Expression expression) {
...
+    String nameOfQualifier = decoratorNameFromExpression(qualifier);
+    if (nameOfQualifier != null) {
+      exprName = nameOfQualifier + "." + exprName;
+    } else {
+      exprName = null;
+    }
+    return exprName;
+  }
+
+  public static String decoratorNameFromExpression(Expression expression) {
```
andrea-guarino-sonarsource, line 290: Please add `@CheckForNull`

python-frontend/src/main/java/org/sonar/python/semantic/FunctionSymbolImpl.java
guillaume-dequenne-sonarsource: Indeed, thanks! I thought you could use them with another qualified name like `abc.staticmethod` hence why I kept it. But after double checking it I was mistaken.

python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
guillaume-dequenne-sonarsource: As discussed, I added the test in `SymbolTableBuilderTest` instead.

PR #882, opened 2021-01-27
SONARPY-820 Type inference: standard collections as generics

andrea-guarino-sonarsource:


python-frontend/src/test/java/org/sonar/python/tree/CallExpressionImplTest.java
```diff
@@ -202,4 +207,20 @@ public void inner_class_constructor_declared_type() {
...
+    assertThat(lastExpressionInFunction("dict[str, int]()").type()).isEqualTo(DICT);
+    assertThat(lastExpressionInFunction("set[int]()").type()).isEqualTo(SET);
+    assertThat(lastExpressionInFunction("frozenset[int]()").type().canOnlyBe("frozenset")).isTrue();
+    assertThat(lastExpressionInFunction("from collections import deque; deque[int]()").type().canOnlyBe("collections.deque")).isTrue();
+    assertThat(lastExpressionInFunction("from collections import defaultdict; defaultdict[str, int]()").type().canOnlyBe("collections.defaultdict")).isTrue();
+    assertThat(lastExpressionInFunction("from collections import OrderedDict; OrderedDict[str, int]()").type().canOnlyBe("collections.OrderedDict")).isTrue();
+    assertThat(lastExpressionInFunction("from collections import Counter; Counter[int]()").type().canOnlyBe("collections.Counter")).isTrue();
+    assertThat(lastExpressionInFunction("from collections import ChainMap; ChainMap[str, int]()").type().canOnlyBe("collections.ChainMap")).isTrue();
+    assertThat(lastExpressionInFunction("from foo import bar; bar[str]()").type().canOnlyBe("bar.foo")).isFalse();
```
pynicolas, line 223: Shouldn't it be `.canOnlyBe("foo.bar")`?
In fact, it may be better to assert that it's equal to `ANY`.

python-frontend/src/test/java/org/sonar/python/tree/CallExpressionImplTest.java
```diff
@@ -202,4 +207,20 @@ public void inner_class_constructor_declared_type() {
...
     assertThat(typeName(type)).isEqualTo("B");
   }
+
+  @Test
+  public void test_generic_collections() {
+    assertThat(lastExpressionInFunction("list[int]()").type()).isEqualTo(LIST);
+    assertThat(lastExpressionInFunction("tuple[int]()").type()).isEqualTo(TUPLE);
+    assertThat(lastExpressionInFunction("dict[str, int]()").type()).isEqualTo(DICT);
+    assertThat(lastExpressionInFunction("set[int]()").type()).isEqualTo(SET);
+    assertThat(lastExpressionInFunction("frozenset[int]()").type().canOnlyBe("frozenset")).isTrue();
```
pynicolas, line 217: Maybe we could add something like:
```
assertThat(lastExpressionInFunction("int[str]()").type()).isEqualTo(anyType());
```

python-frontend/src/main/java/org/sonar/python/semantic/ClassSymbolImpl.java
```diff
@@ -73,22 +74,23 @@ public ClassSymbolImpl(ClassDef classDef, @Nullable String fullyQualifiedName, P
   }
 
   public ClassSymbolImpl(String name, @Nullable String fullyQualifiedName) {
-    this(name, fullyQualifiedName, null, false, false, null);
+    this(name, fullyQualifiedName, null, false, false, null, false);
   }
 
   public ClassSymbolImpl(String name, @Nullable String fullyQualifiedName, @Nullable LocationInFile definitionLocation,
-                         boolean hasDecorators, boolean hasMetaClass, @Nullable String metaclassFQN) {
+                         boolean hasDecorators, boolean hasMetaClass, @Nullable String metaclassFQN, boolean supportsGenerics) {
```
pynicolas, line 81: An alternative solution would have been to compute this value lazily. Instead of having a constructor argument, we would compute `supportsGenerics` only when `supportsGenerics()` is called by looking at the super classes.

pynicolas (approved):
I wonder if we should review some rules, or at least some of their test cases.
For example, we may have updated https://github.com/SonarSource/sonar-python/blob/master/python-checks/src/test/resources/checks/confusingTypeChecking/itemOperations.py#L10

python-frontend/src/main/java/org/sonar/python/semantic/SymbolUtils.java
```diff
@@ -119,6 +119,9 @@ private static void addParentClass(PythonFile pythonFile, Map<String, Symbol> sy
     if (argumentSymbol == null) {
       classSymbol.setHasSuperClassWithoutSymbol();
     } else {
+      if (argumentSymbol.name().equals("Generic")) {
```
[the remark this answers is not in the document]
andrea-guarino-sonarsource: very good point! thanks!

python-frontend/src/main/java/org/sonar/python/semantic/ClassSymbolImpl.java
andrea-guarino-sonarsource: This might be a bit problematic, because when constructing TypeShed symbols we don't add `Generic` superclass in order to avoid having incomplete hierarchies. 
https://github.com/SonarSource/sonar-python/blob/be0952ea43fd81fd934e26465ea16762a0a97d7d/python-frontend/src/main/java/org/sonar/python/semantic/SymbolUtils.java#L134

PR #886, opened 2021-01-29
SONARPY-726 Fix parsing errors on f-strings formatted expressions

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/api/PythonGrammar.java
```diff
@@ -207,7 +207,7 @@ public static void grammar(LexerfulGrammarBuilder b) {
     b.rule(F_STRING_CONTENT).is(b.zeroOrMore(b.firstOf(GenericTokenType.UNKNOWN_CHAR, FORMATTED_EXPR)));
     b.rule(FORMATTED_EXPR).is(
       PythonPunctuator.LCURLYBRACE,
-      EXPR,
+      TEST,
```
pynicolas, line 210: That's better.
However, we still don't handle [all the possibilities of the grammar](https://docs.python.org/3/reference/lexical_analysis.html#formatted-string-literals).
For example, we still have an FP on unused local var for:
```
def f1():
    x = 42
    print(f"{41,x}")
```
However, that's not explicitly in the scope of SONARPY-815, so I suggest to create another ticket to tackle it later.

python-frontend/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
```diff
@@ -1967,6 +1972,20 @@ public void string_interpolation() {
...
+    assertThat(elmt.isInterpolated()).isTrue();
+    assertThat(elmt.formattedExpressions()).hasSize(1);
+    assertThat(elmt.formattedExpressions().get(0).expression().is(Kind.CONDITIONAL_EXPR)).isTrue();
+
+    stringLiteral = (StringLiteral) parse("f\"missing {'previous C' if pCatom is None else 'N'} atom\"", treeMaker::expression);
+    assertThat(stringLiteral.stringElements()).hasSize(1);
+    elmt = stringLiteral.stringElements().get(0);
+    assertThat(elmt.isInterpolated()).isTrue();
+    assertThat(elmt.formattedExpressions()).hasSize(1);
+    assertThat(elmt.formattedExpressions().get(0).expression().is(Kind.CONDITIONAL_EXPR)).isTrue();
```
pynicolas, line 1988: I don't see what those test cases bring. Maybe I'm missing something.

pynicolas (approved):

python-frontend/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
guillaume-dequenne-sonarsource: I wanted to have a test ensuring the entire string gets parsed correctly, but indeed having 1 of them is debatable and 2 redundant, so I'll just keep one.

python-frontend/src/main/java/org/sonar/python/api/PythonGrammar.java
guillaume-dequenne-sonarsource: Indeed, thanks for pointing this out.
I'll reopen SONARPY-726 and mention this in the ticket, as this PR only partly fixes the underlying problem.

PR #890, opened 2021-02-22
SONARPY-826 Rule S2612: Setting loose POSIX file permissions is secur…

guillaume-dequenne-sonarsource:
…ity sensitive


python-checks/src/test/resources/checks/hotspots/filePermissions.py
```diff
@@ -0,0 +1,84 @@
...
+os.chmod("/tmp/fs", stat.S_IXGRP) # Compliant ------x---
+os.chmod("/tmp/fs", stat.S_IRWXG) # Compliant ----rwx---
+os.chmod("/tmp/fs", stat.S_IRWXU | stat.S_IRWXG)  # Compliant -rwxrwx---
+os.chmod("/tmp/fs", stat.S_IRGRP | stat.S_IRUSR)  # Compliant -r--r-----
+
+os.chmod("/tmp/fs", stat.S_IROTH)  # Noncompliant
+os.chmod("/tmp/fs", stat.S_IWOTH)  # Noncompliant
+os.chmod("/tmp/fs", stat.S_IXOTH)  # Noncompliant
+os.chmod("/tmp/fs", stat.S_IRWXO)  # Noncompliant
+os.chmod("/tmp/fs", stat.S_IRWXU | stat.S_IRWXG | stat.S_IRWXO)  # FN
```
andrea-guarino-sonarsource, line 83: why those one are FN? 
I think we should be able to implement it no?

guillaume-dequenne-sonarsource: We certainly could be able to implement it but I felt like we'd need to draw the line as to how complex of an expression we'd try to resolve and that such cases would be rare enough that a more basic approach would be sufficient.

python-frontend/src/main/java/org/sonar/python/tree/NumericLiteralImpl.java
```diff
@@ -57,6 +57,9 @@ public long valueAsLong() {
     if (literalValue.startsWith("0b") || literalValue.startsWith("0B")) {
       return Integer.valueOf(literalValue.substring(2), 2);
     }
+    if (literalValue.startsWith("0o") || literalValue.startsWith("0O")) {
```
andrea-guarino-sonarsource, line 60: as discussed, Python 2 octal literals are prefixed with `0`. 
This means we might have an FP with the following code:
```python
os.chmod("foo", 0750)
```

guillaume-dequenne-sonarsource: Good catch! Fixed in d4b8d61bd603df59b4190b95794446f1cc0a362e.

andrea-guarino-sonarsource (approved):
LGTM, I think ruling needs to be updated

PR #891, opened 2021-02-22
SONARPY-829 Rule S5659: JWT should be signed and verified with strong cipher algorithms

pynicolas:


guillaume-dequenne-sonarsource (approved):
LGTM

PR #893, opened 2021-02-24
SONARPY-828 Rule S3329: Cipher Block chaining IV's should be random a…

guillaume-dequenne-sonarsource:
…nd unique


python-checks/src/main/java/org/sonar/python/checks/CipherBlockChainingCheck.java
```diff
@@ -0,0 +1,162 @@
...
+import org.sonar.plugins.python.api.tree.RegularArgument;
+import org.sonar.plugins.python.api.tree.Tree;
+import org.sonar.python.tree.TreeUtils;
+
+import static java.util.Arrays.asList;
+
+@Rule(key = "S3329")
+public class CipherBlockChainingCheck extends PythonSubscriptionCheck {
+
+  private static final HashSet<String> pyCryptoSensitiveFQNs = new HashSet<>();
```
andrea-guarino-sonarsource, line 43: _minor_: why not `PYCRIPTO_SENSITIVE_FQNS`?

python-checks/src/main/java/org/sonar/python/checks/CipherBlockChainingCheck.java
```diff
@@ -0,0 +1,162 @@
...
+  private static boolean isStaticInitializationVector(Expression expression) {
+    if (expression.is(Tree.Kind.CALL_EXPR) || TreeUtils.hasDescendant(expression, tree -> tree.is(Tree.Kind.CALL_EXPR))) {
+      return false;
+    }
+    if (expression.is(Tree.Kind.NAME)) {
+      Expression singleAssignedValue = Expressions.singleAssignedValue((Name) expression);
+      if (singleAssignedValue == null) {
+        return false;
+      }
+      return isStaticInitializationVector(singleAssignedValue);
```
andrea-guarino-sonarsource, line 158: I think we can have a StackOverflowException in the following case:
```python
def soe():
    static_vector = other_static
    other_static = static_vector
    enc2 = AES.new(key, AES.MODE_CBC, static_vector)  # 2nd location
    cipher_text = enc2.encrypt(pad(data, AES.block_size))  # Noncompliant
```

andrea-guarino-sonarsource (changes requested):
I think there is a risk of StackOverflowException in the recursive function call `isStaticInitializationVector`

python-checks/src/main/java/org/sonar/python/checks/CipherBlockChainingCheck.java
guillaume-dequenne-sonarsource: Indeed, thanks!

PR #894, opened 2021-02-24
SONARPY-826 S2612 to handle binary OR expressions

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/FilePermissionsCheck.java
```diff
@@ -75,21 +76,29 @@ private static void checkSensitiveArgument(List<Argument> arguments , int sensit
...
+  private static boolean isUnsafeExpression(Expression expression, int safeModulo) {
     if (expression instanceof HasSymbol) {
       Symbol symbol = ((HasSymbol) expression).symbol();
       if (symbol != null && SENSITIVE_CONSTANTS.contains(symbol.fullyQualifiedName())) {
-        ctx.addIssue(modeArgument, MESSAGE);
-        return;
+        return true;
       }
     }
+    if (expression.is(Tree.Kind.BITWISE_OR)) {
```
pynicolas, line 91: Can I nitpick? :)
We have a FN for the following case:
```
x = stat.S_IRWXU | stat.S_IRWXG | stat.S_IRWXO
os.chmod("/tmp/fs", x)
```
In fact, we also have a FN for a simpler case:
```
x = stat.S_IRWXO
os.chmod("/tmp/fs", x)
```

PR #895, opened 2021-02-25
SONARPY-831 Rule S3752: Allowing both safe and unsafe HTTP methods is security-sensitive

andrea-guarino-sonarsource:


python-frontend/src/main/java/org/sonar/python/types/FlowSensitiveTypeInference.java
```diff
@@ -133,6 +133,7 @@ public void visitName(Name name) {
 
       @Override
       public void visitFunctionDef(FunctionDef pyFunctionDefTree) {
+        scan(pyFunctionDefTree.decorators());
```
pynicolas, line 136: I'm not sure why we do this change. If we want to avoid FNs or FPs, I think that could be reflected with at least one test case in a rule test.

python-checks/src/main/java/org/sonar/python/checks/hotspots/UnsafeHttpMethodsCheck.java
```diff
@@ -121,4 +135,31 @@ private static boolean isDjangoView(FunctionDef functionDef) {
...
+      .findFirst();
+  }
+
+  private boolean isFlaskRouteDecorator(CallExpression callExpression) {
+    Symbol calleeSymbol = callExpression.calleeSymbol();
+    if (calleeSymbol == null) {
+      return false;
+    }
+    if (calleeSymbol.fullyQualifiedName() == null) {
+      return isFlaskImported && calleeSymbol.name().equals("route");
```
pynicolas, line 154: I think I understand why we do that and I don't really like it.
If that's possible, I'd prefer a cleaner solution: maybe we could adjust how we handle module variables.
We can talk about it.

andrea-guarino-sonarsource: After discussion, I'll change the implementation to avoid keeping a state, however keeping this "hack" to infer if a decorator defines a flask route.
I also created [SONARPY-834](https://jira.sonarsource.com/browse/SONARPY-834) which enables type inference at module level and would allow us to have a cleaner implementation of this.

python-frontend/src/main/java/org/sonar/python/types/FlowSensitiveTypeInference.java
andrea-guarino-sonarsource: After discussion, I will drop this commit as we already use the heuristic to look if `flask` is imported in a file.

python-checks/src/main/java/org/sonar/python/checks/hotspots/UnsafeHttpMethodsCheck.java
```diff
@@ -0,0 +1,124 @@
...
+      if (decorator.expression().is(CALL_EXPR)) {
+        CallExpression callExpression = (CallExpression) decorator.expression();
+        Symbol symbol = callExpression.calleeSymbol();
+        if (symbol != null && "django.views.decorators.http.require_http_methods".equals(symbol.fullyQualifiedName())) {
+          checkRequireHttpMethodsDecorator(ctx, callExpression);
+          return;
+        }
+      }
+    }
+    ctx.addIssue(functionDef.name(), MESSAGE);
```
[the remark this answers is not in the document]
andrea-guarino-sonarsource: I create ticket [SONARPY-835](https://jira.sonarsource.com/browse/SONARPY-835) to keep track of it

pynicolas (approved):
LGTM. I let you choose how you want to merge it.

PR #915, opened 2021-05-04
Update SonarQube API to 8.9

guillaume-dequenne-sonarsource:


pom.xml
```diff
@@ -89,7 +89,7 @@
     <logback.version>1.2.3</logback.version>
     <maven.project.version>2.2.1</maven.project.version>
     <mockito.version>3.9.0</mockito.version>
-    <sonar.version>7.9</sonar.version>
+    <sonar.version>8.8.0.42792</sonar.version>
```
andrea-guarino-sonarsource, line 92: I think we could use 8.9 version, i.e. 8.9.0.43852. Also, can you also update `sonarQubeMinVersion`?

its/plugin/python-custom-rules-plugin/pom.xml
```diff
@@ -49,6 +54,7 @@
         <configuration>
           <pluginClass>org.sonar.samples.python.CustomPythonRulesPlugin</pluginClass>
           <requirePlugins>python:${project.version}</requirePlugins>
+          <sonarQubeMinVersion>${sonarQubeMinVersion}</sonarQubeMinVersion>
```
andrea-guarino-sonarsource, line 57: why did you need to change this?

andrea-guarino-sonarsource (approved):
LGTM, I think we can keep https://jira.sonarsource.com/browse/SONARPY-849 open to remember to update `sonarQubeMinVersion` as soon as LTS release if official

its/plugin/python-custom-rules-plugin/pom.xml
guillaume-dequenne-sonarsource: Without this, the custom rules IT would crash because min SQ version seems to default to `sonar.version` when this property is missing, but the custom rules plugin IT seems to use an hardcoded 7.9 version (and I believed it was better to break backward compatibility with 7.9 explicitly once 8.9 is officially released as the LTS version).

PR #919, opened 2021-05-06
SONARPY-852 Declare PythonIndexer as SonarLint component with module scope

guillaume-dequenne-sonarsource:


sonar-python-plugin/src/test/java/org/sonar/plugins/python/SonarLintPythonIndexerTest.java
```diff
@@ -0,0 +1,185 @@
...
+    assertThat(logTester.logs(LoggerLevel.DEBUG)).contains("Failed to load file \"nonexistent.py\" to the project symbol table");
+  }
+
+  @Test
+  public void test_indexer_modified_file() throws IOException {
+    InputFile file1 = inputFile("main.py");
+    InputFile file2 = inputFile("mod.py");
+    SonarLintPythonIndexer pythonIndexer = new SonarLintPythonIndexer(new TestModuleFileSystem(Arrays.asList(file1, file2)));
+    pythonIndexer.buildOnce(context);
+    ProjectLevelSymbolTable projectLevelSymbolTable = pythonIndexer.projectLevelSymbolTable();
```
andrea-guarino-sonarsource, line 158: I would suggest declaring these vars as fields:
```java
  InputFile file1;
  InputFile file2;
  SonarLintPythonIndexer pythonIndexer;
  ProjectLevelSymbolTable projectLevelSymbolTable;
```

and move this initialization in the `@Before init()` method

andrea-guarino-sonarsource (changes requested):
Overall looks good, nice job!
I left some comments. 
I would suggest to maybe move those new "Indexer" related classes in a new package "index". 
Wdyt?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/SonarQubePythonIndexer.java
```diff
@@ -0,0 +1,48 @@
...
+
+  private final List<InputFile> files;
+  private static final Logger LOG = Loggers.get(SonarQubePythonIndexer.class);
+
+  public SonarQubePythonIndexer(List<InputFile> files) {
+    this.files = files;
+  }
+
+  @Override
+  public void buildOnce(SensorContext context) {
```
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: If we want to avoid constructing a project level symbol table for big projects in SonarLint, I thought the SonarLint-specific could go there and that the two methods would eventually be different from each others. Do you think it make sense to keep them separate for that reason, or should we just refactor when we actually get to it?

andrea-guarino-sonarsource (changes requested):

sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonPlugin.java
```diff
@@ -79,6 +80,9 @@ public void define(Context context) {
       addBanditExtensions(context);
       addFlake8Extensions(context);
     }
+    if (sonarRuntime.getProduct() == SonarProduct.SONARLINT) {
```
andrea-guarino-sonarsource, line 83: we also need to check sonarlint version to avoid `ClassNotFound` exceptions in older versions of SonarLint

andrea-guarino-sonarsource: we'll do it via a separate PR

PR #923, opened 2021-05-11
Avoid loading SonarlintPythonIndexer on older versions of SonarLint

guillaume-dequenne-sonarsource:


sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonPlugin.java
```diff
@@ -81,7 +84,12 @@ public void define(Context context) {
       addFlake8Extensions(context);
     }
     if (sonarRuntime.getProduct() == SonarProduct.SONARLINT) {
-      context.addExtension(SonarLintPythonIndexer.class);
+      try {
```
andrea-guarino-sonarsource, line 87: Can we wrap this logic in a function like `isSonarLintPluginAPIProvided()` (or something similar)?

andrea-guarino-sonarsource (changes requested):
As discussed, can we try adding a UT or IT?

andrea-guarino-sonarsource (approved):

PR #925, opened 2021-05-17
Add performance measurement for project level symbol table

guillaume-dequenne-sonarsource:


sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonSensor.java
```diff
@@ -110,6 +119,20 @@ public void execute(SensorContext context) {
     return Collections.unmodifiableList(list);
   }
 
+  private static PerformanceMeasure.Duration createPerformanceMeasureReport(SensorContext context) {
+    return PerformanceMeasure.reportBuilder()
+      .activate(context.config().getBoolean(PERFORMANCE_MEASURE_PROPERTY).orElse(Boolean.FALSE))
+      .allowSingleThreadMode()
```
andrea-guarino-sonarsource, line 125: looks like this has been removed finally: https://github.com/SonarSource/sonar-analyzer-commons/pull/143/commits/fcf2dbd2009c5f68e85c4cdadc48b444bbedd716

sonar-python-plugin/src/main/java/org/sonar/plugins/python/indexer/SonarQubePythonIndexer.java
```diff
@@ -44,5 +46,6 @@ public void buildOnce(SensorContext context) {
     globalSymbolsStep.execute(files, context);
     long stopTime = System.currentTimeMillis() - startTime;
     LOG.debug("Time to build the project level symbol table: " + stopTime + "ms");
```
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: I was a bit hesitant because I found I'm still using this for debugging, to avoid the overhead of the performance monitoring, but indeed it doesn't make much sense to keep this on master anyway.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonSensor.java
guillaume-dequenne-sonarsource: I updated the dependency accordingly. Anyway we'll still need to update it to its released version when available.

sonar-python-plugin/pom.xml
```diff
@@ -146,8 +150,8 @@
             <configuration>
               <rules>
                 <requireFilesSize>
-                  <maxsize>4810000</maxsize>
-                  <minsize>4310000</minsize>
+                  <maxsize>4910000</maxsize>
```
andrea-guarino-sonarsource, line 153: when I build locally I have `4917432` . Can you please increase to "4500000 - 5000000"?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/indexer/SonarQubePythonIndexer.java
```diff
@@ -36,13 +37,12 @@ public SonarQubePythonIndexer(List<InputFile> files) {
 
   @Override
   public void buildOnce(SensorContext context) {
+    PerformanceMeasure.Duration duration = PerformanceMeasure.start("ProjectLevelSymbolTable");
```
andrea-guarino-sonarsource, line 40: do we do it only for SQ / SC? I think we could add this statistics also for SonarLint

guillaume-dequenne-sonarsource: From my understanding the performance monitoring makes mostly sense for monitoring on Peach, thus in SQ context.
Although there could be more usages to this and it might make sense to have it in SL context, I felt it's not necessary for now (optimizing symbol table construction for SQ is essentially the same as doing it for SL as of today).
If you think it's worth having it nevertheless, I can still add it though.

andrea-guarino-sonarsource:
I think it would be nice to create a `doc/properties.md` file in the repo, to document this kind of "internal" properties we're adding. Wdyt?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/indexer/SonarQubePythonIndexer.java
andrea-guarino-sonarsource: ok makes sense. Maybe we can add a small comment to explain why we only measure SQ / SC

andrea-guarino-sonarsource (approved):
LGTM

PR #927, opened 2021-05-17
Avoid Persisting BaseDir file, but use its absolute path instead

andrea-guarino-sonarsource:


guillaume-dequenne-sonarsource (approved):
LGTM

PR #929, opened 2021-05-19
SONARPY-857 Add cross-file secondary location for SonarLint

guillaume-dequenne-sonarsource:


andrea-guarino-sonarsource (approved):
LGTM, we manually tested that the impact on memory is low
- less than 1MB for `black`

PR #931, opened 2021-05-21
Update rules metadata

guillaume-dequenne-sonarsource:


python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S5042.html
```diff
@@ -23,9 +23,95 @@ <h2>Sensitive Code Example</h2>
 <pre>
 import tarfile
 
-tar = tarfile.open("attachment.tar.gz")  # Sensitive
-tar.extractall()
-tar.close()
+tfile = tarfile.open("TarBomb.tar")
```
andrea-guarino-sonarsource, line 26: I think we didn't implement yet this new change: https://jira.sonarsource.com/browse/SONARPY-792

Can you please roll it back?

PR #938, opened 2021-07-05
SONARPY-869 Deserialize protobuf files containing symbol information …

andrea-guarino-sonarsource:
…for "annoy" library

### DIFF
--- a/python-frontend/pom.xml
+++ b/python-frontend/pom.xml
@@ -37,6 +37,68 @@
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
+      <version>${protobuf.version}</version>
+    </dependency>
   </dependencies>
-
+  <build>
+    <extensions>
+      <extension>
+        <groupId>kr.motd.maven</groupId>
+        <artifactId>os-maven-plugin</artifactId>
+        <version>1.6.2</version>
+      </extension>
+    </extensions>
+    <plugins>
+      <plugin>
+        <groupId>org.xolstice.maven.plugins</groupId>
+        <artifactId>protobuf-maven-plugin</artifactId>
+        <version>0.6.1</version>
+        <executions>
+          <execution>
+            <id>generate-protobuf-java-sources</id>
+            <goals>
+              <goal>compile</goal>
+            </goals>
+            <configuration>
+              <protocArtifact>com.google.protobuf:protoc:${protobuf.version}:exe:${os.detected.classifier}
+              </protocArtifact>
+              <protoSourceRoot>${project.basedir}/src/main/protobuf</protoSourceRoot>
+              <outputDirectory>${project.build.directory}/generated-sources/protobuf</outputDirectory>
+            </configuration>
+          </execution>
+        </executions>
+        <dependencies>
+          <dependency>
+            <groupId>com.google.protobuf</groupId>
+            <artifactId>protoc</artifactId>
+            <version>${protobuf.version}</version>
+            <type>exe</type>
+            <classifier>${os.detected.classifier}</classifier>
+          </dependency>
+        </dependencies>
+      </plugin>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>build-helper-maven-plugin</artifactId>
+        <version>3.2.0</version>
+        <executions>
+          <execution>
+            <id>add-protobuf-generated-sources</id>
+            <phase>generate-sources</phase>
+            <goals>
+              <goal>add-source</goal>
+            </goals>
+            <configuration>
+              <sources>
+                <source>${project.build.directory}/generated-sources/protobuf</source>
+              </sources>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
 </project>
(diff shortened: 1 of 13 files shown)


python-frontend/pom.xml
```diff
@@ -37,6 +37,68 @@
...
       <artifactId>guava</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
+      <version>${protobuf.version}</version>
+    </dependency>
   </dependencies>
-
+  <build>
```
guillaume-dequenne-sonarsource, line 46: Just to make sure, this is the same thing we do in sonar-security to compile the Java Protobuf classes automatically, right?
Will it need adaptations to build the Python classes as well? (Should we even do it here as well?)

python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
```diff
@@ -340,4 +354,77 @@ private static void setParameterTypes(Symbol symbol, FunctionDef functionDef) {
...
+    if (fqnSplittedByDot.length == 2 && fqnSplittedByDot[0].equals("builtins") && builtins.containsKey(localName)) {
+      return builtins.get(localName);
+    }
+    for (Map<String, Symbol> symbolsByFqn : typeShedSymbols.values()) {
+      Symbol symbol = symbolsByFqn.get(fullyQualifiedName);
+      if (symbol != null) {
+        return symbol;
+      }
+    }
+    return builtins.getOrDefault(fullyQualifiedName, null);
```
guillaume-dequenne-sonarsource, line 427: Can it happen that we actually retrieve a builtin here if the name didn't start with `builtins`? (for serialized Typeshed at least)

python-frontend/src/main/java/org/sonar/python/types/InferredTypes.java
```diff
@@ -154,6 +155,30 @@ public static InferredType fromTypeshedTypeAnnotation(TypeAnnotation typeAnnotat
...
   }
 
+  public static InferredType fromTypeshedProtobuf(@Nullable SymbolsProtos.Type type) {
+    if (type == null) {
+      return anyType();
+    }
+    switch (type.getKind()) {
+      case INSTANCE:
+        String typeName = type.getSimpleName();
+        return typeName != null ? runtimeType(TypeShed.symbolWithFQN(typeName)) : anyType();
```
guillaume-dequenne-sonarsource, line 165: Here as well it seems `typeName` won't be null.

guillaume-dequenne-sonarsource (changes requested):
Overall looks good. Left a few comments and clarifying questions. I played a bit the devil's advocate and I'm not sure everything needs or can actually be addressed (especially the part that relates to the handling of super classes).

python-frontend/src/test/java/org/sonar/python/types/TypeShedTest.java
```diff
@@ -234,4 +244,140 @@ public void stub_files_symbols() {
...
+  @Test
+  public void deserialize_annoy_protobuf() {
+    Map<String, Symbol> deserializedAnnoySymbols = TypeShed.symbolsForModule("annoy").stream()
+      .collect(Collectors.toMap(Symbol::fullyQualifiedName, s -> s));
+    assertThat(deserializedAnnoySymbols.values()).extracting(Symbol::kind, Symbol::fullyQualifiedName)
+      .containsExactlyInAnyOrder(tuple(Kind.CLASS, "annoy._Vector"), tuple(Kind.CLASS, "annoy.AnnoyIndex"));
+
+    ClassSymbol vector = (ClassSymbol) deserializedAnnoySymbols.get("annoy._Vector");
+    assertThat(vector.superClasses()).extracting(Symbol::kind, Symbol::fullyQualifiedName)
+      .containsExactlyInAnyOrder(tuple(Kind.AMBIGUOUS, "typing.Sized"));
```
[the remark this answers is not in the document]
andrea-guarino-sonarsource: Yes that's correct

sonar-python-plugin/pom.xml
```diff
@@ -150,8 +150,8 @@
             <configuration>
               <rules>
                 <requireFilesSize>
-                  <maxsize>5000000</maxsize>
-                  <minsize>4500000</minsize>
+                  <maxsize>6500000</maxsize>
+                  <minsize>6000000</minsize>
```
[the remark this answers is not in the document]
andrea-guarino-sonarsource: Yes, this is because of protobuf depedency which takes around 1.6MB

python-frontend/pom.xml
andrea-guarino-sonarsource: Yes, that's the same as sonar-security. I believe it makes sense to also build the Python classes here, either on this pom or in a dedicated pom if we decide to have a separate module

python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
andrea-guarino-sonarsource: I think you're right, it should never happen for serialized Typeshed

python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
```diff
@@ -340,4 +354,77 @@ private static void setParameterTypes(Symbol symbol, FunctionDef functionDef) {
...
+    }
+    for (SymbolsProtos.OverloadedFunctionSymbol overloadedFunctionSymbol : moduleSymbol.getOverloadedFunctionsList()) {
+      deserializedSymbols.put(overloadedFunctionSymbol.getFullname(),
+        AmbiguousSymbolImpl.create(overloadedFunctionSymbol.getDefinitionsList().stream().map(FunctionSymbolImpl::new).collect(Collectors.toSet())));
+    }
+    return deserializedSymbols;
+  }
+
+
+  private static void addSuperClasses(Map<String, Symbol> deserializedSymbols, SymbolsProtos.ClassSymbol classSymbolProto, ClassSymbolImpl classSymbol) {
```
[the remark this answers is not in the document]
andrea-guarino-sonarsource: I think it makes sense! Thanks for the nice suggestion!

guillaume-dequenne-sonarsource (approved):
LGTM

PR #945, opened 2021-07-21
SONARPY-631 Add a python version parameter and raise a warning when it is not set

andrea-guarino-sonarsource:


python-frontend/src/main/java/org/sonar/python/types/DeclaredType.java
```diff
@@ -139,7 +145,7 @@ public String toString() {
   }
 
   public String typeName() {
```
guillaume-dequenne-sonarsource, line 147: Not sure if this could break anything, but we should probably consider changing this to use the `pretty_printed_name` of Typeshed types here when available instead of recomputing them (not in this PR though...)

python-frontend/src/main/java/org/sonar/python/types/DeclaredType.java
```diff
@@ -149,10 +155,16 @@ public String typeName() {
...
   Symbol getTypeClass() {
+    if (typeClass == null) {
+      return TypeShed.typeShedClass(builtinFullyQualifiedName);
+    }
     return typeClass;
   }
 
   public Set<Symbol> alternativeTypeSymbols() {
+    if (alternativeTypeSymbols == null) {
+      return resolveAlternativeSymbols(getTypeClass(), typeArgs);
```
guillaume-dequenne-sonarsource, line 166: I guess here we could set `alternativeTypeSymbols = resolveAlternativeSymbols(getTypeClass(), typeArgs)` to avoid redoing the computation?
Though I think we discussed it and concluded it would be better to just recompute so that it would be a bit easier to manage in SonarLint?

python-frontend/src/main/java/org/sonar/plugins/python/api/ProjectPythonVersion.java
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  private static PythonVersion currentVersion = PythonVersion.allVersions();
+
+  public static PythonVersion currentVersion() {
+    return currentVersion;
+  }
+
+  public static void setCurrentVersion(PythonVersion currentVersion) {
+    ProjectPythonVersion.currentVersion = currentVersion;
```
guillaume-dequenne-sonarsource, line 34: Related to my other comment, it's kind of weird to me that what actually holds the project python version is the attribute `supportedVersions` of a field of type `PythonVersion` nested in `ProjectPythonVersion`.

I feel it could be a bit simpler, e.g by having the `PythonVersion` class as a fully util class whose `fromString` method would return the `List<Version>` which would be what's actually stored in `ProjectPythonVersion`, wdyt?

python-frontend/src/test/java/org/sonar/python/types/TypeShedTest.java
```diff
@@ -67,6 +74,19 @@ public void str() {
...
+    assertThat(strClass.superClasses()).extracting(Symbol::kind, Symbol::name).containsExactlyInAnyOrder(tuple(Kind.CLASS, "Sequence"));
+    assertThat(strClass.resolveMember("removeprefix")).isEmpty();
+    assertThat(strClass.resolveMember("removesuffix")).isEmpty();
+
+    setPythonVersion(PythonVersion.fromString("2.7"));
+    strClass = TypeShed.typeShedClass("str");
+    assertThat(strClass.superClasses()).extracting(Symbol::kind, Symbol::name)
+      .containsExactlyInAnyOrder(tuple(Kind.CLASS, "Sequence"), tuple(Kind.CLASS, "basestring"));
+
+    setPythonVersion(PythonVersion.allVersions());
```
guillaume-dequenne-sonarsource, line 89: I guess that's needed to avoid polluting other tests?
Is there an alternative to this (like using `BeforeEach` or something like that)? I guess not without performance impact, but I don't like that the test somehow needs to do the cleanup itself.

guillaume-dequenne-sonarsource (approved):
LGTM. I think the way we handle `PythonVersion` could be a bit more straightforward but that's mostly personal preference (and perhaps the current approach makes more sense if we want to extend a bit the accepted inputs with things like ranges, etc...).

python-frontend/src/test/java/org/sonar/python/types/TypeShedTest.java
andrea-guarino-sonarsource: Yeah I don't like that either.
However I initially tried with `BeforeEach` and it had quite a significant performance impact and IMO it's not really worth it.
Wdyt?

python-frontend/src/main/java/org/sonar/python/types/DeclaredType.java
andrea-guarino-sonarsource: Yes, I did this to ease the "cache" invalidation. I think that the declared types we're now computing lazily don't have `type args ` so we will never execute line 166 for the time being.
However I believe it's cleaner to have it already done properly to avoid any inconsistency problem later. wdyt?

python-frontend/src/main/java/org/sonar/plugins/python/api/ProjectPythonVersion.java
andrea-guarino-sonarsource: Agreed, we don't need to abstract Python version in a class for the time being!
I think we can indeed have `PythonVersionUtils.fromString()` returning a `Set<Version>` and `ProjectPythonVersion` storing just this set

python-frontend/src/main/java/org/sonar/python/types/DeclaredType.java
guillaume-dequenne-sonarsource: Makes sense to me!

python-frontend/src/test/java/org/sonar/python/types/TypeShedTest.java
guillaume-dequenne-sonarsource: Indeed I assumed that would be the issue. I think it's fine as it is though, not ideal, but we know why we do it that way.

andrea-guarino-sonarsource: I will add a comment so that we won't forget

python-frontend/src/main/java/org/sonar/python/types/DeclaredType.java
andrea-guarino-sonarsource: Agreed, we can have a separate ticket for it

andrea-guarino-sonarsource: Ticket created: https://jira.sonarsource.com/browse/SONARPY-880

PR #957, opened 2021-08-09
Use moved `DirectoryScanner` to sonar-analyzer-commons 1.16.0.719

yassin-kammoun-sonarsource:
We needed to support wildcard patterns for coverage reports in SonarJS. Since sonar-python already supported that feature, we extracted the `DirectoryScanner` logic and moved it to sonar-analyzer-commons (see [#149](https://github.com/SonarSource/sonar-analyzer-commons/pull/149)). The implementation is basically the same from the original. There are a few renaming here and there as well as a slightly changed unit test, though.


guillaume-dequenne-sonarsource (approved):
LGTM

PR #970, opened 2021-11-02
SONARPY-885 Rule S5996 Regex boundaries should not be used in a way that can never be matched

nils-werner-sonarsource:


karim-ouerghemmi-sonarsource (approved):
LGTM!

PR #971, opened 2021-11-02
SONARPY-891 Rule S5857 Character classes should be preferred over reluctant quantifiers in regular expressions

nils-werner-sonarsource:


karim-ouerghemmi-sonarsource (approved):
LGTM!

PR #972, opened 2021-11-02
SONARPY-889 Rule S5869 Character classes in regular expressions should not contain the same character twice

nils-werner-sonarsource:


karim-ouerghemmi-sonarsource (approved):
LGTM!

PR #973, opened 2021-11-02
SONARPY-893 Rule S5843 Regular expressions should not be too complicated

nils-werner-sonarsource:


karim-ouerghemmi-sonarsource (approved):
LGTM

PR #974, opened 2021-11-02
SONARPY-921 Rule S5856 Regular expressions should be syntactically valid

nils-werner-sonarsource:


karim-ouerghemmi-sonarsource (approved):
LGTM. Let's rebase before merging.

PR #975, opened 2021-11-02
SONARPY-860 Support `with` statements with parenthesized context mana…

andrea-guarino-sonarsource:
…gers


guillaume-dequenne-sonarsource (approved):
LGTM, you could consider merging on master directly as this is not strictly related to `MMF-2541` and already causes parse errors on real world projects.
Although I have no strong opinion on that.

PR #978, opened 2021-11-03
SONARPY-923 Analyze regex in variables whose values we can infer

karim-ouerghemmi-sonarsource:
I had to make the regex cache's key also depend on the flags. We didn't have to do that in PHP because the global flags are part of the string literal.

One thing I'm not 100% convinced about yet is that, if we do raise on the value of a variable, we might have raised because of some flags. However, in the location we raise, those flags are not visible. I'm not certain if this will lead to unclear issues. We can go with it now, and adapt if that will be the case.


nils-werner-sonarsource:
> One thing I'm not 100% convinced about yet is that, if we do raise on the value of a variable, we might have raised because of some flags. However, in the location we raise, those flags are not visible. I'm not certain if this will lead to unclear issues. We can go with it now, and adapt if that will be the case.

I agree this is a fair point which I also realize during reviewing the results. However, we should be consistent with all languages. Java also does not highlight the flags if this could be relevant for the issue. Let's create a common issue for this.

nils-werner-sonarsource (approved):
LGTM!

PR #979, opened 2021-11-03
SONARPY-918 Match / case statement: as patterns

andrea-guarino-sonarsource:


python-frontend/src/main/java/org/sonar/python/tree/PythonTreeMaker.java
```diff
@@ -822,9 +823,31 @@ public Guard guard(AstNode guardNode) {
...
-    AstNode literalPattern = pattern.getFirstChild(PythonGrammar.LITERAL_PATTERN);
+  public static Pattern pattern(AstNode pattern) {
+    if (pattern.is(PythonGrammar.CLOSED_PATTERN)) {
+      return closedPattern(pattern);
+    }
+    return asPattern(pattern);
+  }
+
+  private static AsPattern asPattern(AstNode asPattern) {
+    Pattern pattern = literalPattern(asPattern.getFirstChild(PythonGrammar.CLOSED_PATTERN).getFirstChild());
```
pynicolas, line 834: Does that work for `case x as y:`? (I'm not sure that's valid)

python-frontend/src/main/java/org/sonar/plugins/python/api/tree/Name.java
```diff
@@ -22,7 +22,7 @@
 /**
  * See https://docs.python.org/3/reference/expressions.html#atom-identifiers
  */
-public interface Name extends Expression, HasSymbol {
+public interface Name extends Expression, HasSymbol, Pattern {
```
pynicolas, line 25: I'm not sure that's a good change. That would mean that:
* there's no easy way to visit all capture patterns.
* we have something which implements `Pattern` everywhere we have an identifier (I know that's what we already do with `Name` implementing `Expression`). I don't know if it's a real problem.

An alternative would be to introduce a `CapturePattern` class which would wrap an instance of `Name`. 
Maybe we can discuss that.

andrea-guarino-sonarsource: Following our discussion, I will create `CapturePattern` wrapper

python-frontend/src/main/java/org/sonar/python/tree/PythonTreeMaker.java
andrea-guarino-sonarsource: Yes , I think it's valid and it works after commit 782042ffdfa6371194397ac50b955dfb6f7d625d. I'll add a test

pynicolas: Did you add a test? As we call `literalPattern`, I don't get how it works...

andrea-guarino-sonarsource: Sorry, I misinterpreted your comment. I added a test in [`CasePatternsTest`](https://github.com/SonarSource/sonar-python/blob/22336e3e8b5fccd5491964c6dbf0597095a4dc92/python-frontend/src/test/java/org/sonar/python/parser/compound/statements/CasePatternsTest.java#L75) for the low level. grammar. However the problem you mentioned happens during the strongly typed AST translation.

I will treat this in a separate PR

PR #983, opened 2021-11-04
SONARPY-914 Match / case statement: support sequence patterns

andrea-guarino-sonarsource:


python-frontend/src/main/java/org/sonar/python/tree/PythonTreeMaker.java
```diff
@@ -843,8 +846,60 @@ private static Pattern closedPattern(AstNode closedPattern) {
...
+    return new SequencePatternImpl(toPyToken(leftDelimiter.getToken()), patterns, commas, toPyToken(rightDelimiter.getToken()));
+  }
+
+  private static void addPatternsAndCommasFromMaybeSequencePattern(@Nullable AstNode maybeSequencePattern,List<Pattern> patterns, List<Token> commas) {
+    if (maybeSequencePattern == null) {
+      return;
     }
-    throw new IllegalStateException(String.format("Pattern %s not recognized.", closedPattern.getName()));
+    patterns.addAll(maybeSequencePattern.getChildren(PythonGrammar.MAYBE_STAR_PATTERN).stream().map(PythonTreeMaker::maybeStarPattern).collect(Collectors.toList()));
+    commas.addAll(maybeSequencePattern.getChildren(PythonPunctuator.COMMA).stream().map(node -> toPyToken(node.getToken())).collect(Collectors.toList()));
```
pynicolas, line 888: I think that the `punctuators` method could be useful here.

python-frontend/src/test/java/org/sonar/python/parser/compound/statements/CasePatternsTest.java
```diff
@@ -81,4 +81,29 @@ public void capture() {
...
+      .matches("(head, *tail)");
+
+    setRootRule(PythonGrammar.SEQUENCE_PATTERN);
+    assertThat(p).notMatches("(x)"); // this is a group pattern instead
+
+    setRootRule(PythonGrammar.CASE_BLOCK);
+    assertThat(p)
+      .matches("case 'foo', x, z: ...");
+
+  }
```
pynicolas, line 108: I'm not sure it's a good idea to add tests here because we don't test that `PythonTreeMaker` against those pieces of code. We could add those cases to `PythonTreeMakerMatchStatementTest` to verify in a very concise way that:
* `PythonTreeMaker` doesn't crash on them
* The `Tree` instance has all tokens.

python-frontend/src/test/java/org/sonar/python/tree/PythonTreeMakerMatchStatementTest.java
```diff
@@ -171,4 +174,93 @@ public void capture_pattern() {
     assertThat(capturePattern.name().name()).isEqualTo("x");
     assertThat(capturePattern.children()).containsExactly(capturePattern.name());
   }
+
+  @Test
+  public void sequence_pattern() {
```
pynicolas, line 179: Can we find a way to make those tests shorter? For example, do we get more confidence from our tests by checking the result of the `children` method for each of the following cases?

pynicolas (approved):

python-frontend/src/main/java/org/sonar/python/tree/PythonTreeMaker.java
```diff
@@ -805,7 +807,8 @@ public MatchStatement matchStatement(AstNode matchStmt) {
 
   public CaseBlock caseBlock(AstNode caseBlock) {
     Token caseKeyword = toPyToken(caseBlock.getTokens().get(0));
-    Pattern pattern = pattern(caseBlock.getFirstChild(PythonGrammar.PATTERN).getFirstChild());
+    AstNode patternOrSequence = caseBlock.getFirstChild(PythonGrammar.PATTERNS).getFirstChild();
```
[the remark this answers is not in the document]
andrea-guarino-sonarsource: Nice, I forgot about `.skip()` method. Thanks! :)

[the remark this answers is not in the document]
andrea-guarino-sonarsource: In this specific case I'd prefer keeping the code as it is, as I find it easier to have this 1-to-1 relation between grammar and code.

python-frontend/src/test/java/org/sonar/python/parser/compound/statements/CasePatternsTest.java
andrea-guarino-sonarsource: That's a fair point, I will move them to `PythonTreeMakerMatchStatementTest ` class

python-frontend/src/test/java/org/sonar/python/tree/PythonTreeMakerMatchStatementTest.java
andrea-guarino-sonarsource: Indeed, I will simplify those tests and I will take the method you wrote inside "or patterns" PR that factorizes pattern creation  :) 
I also agree that we don't really need to check children as we already check that all tokens are there.

PR #986, opened 2021-11-05
SONARPY-919 Match / case statement: OR patterns

pynicolas:


python-frontend/src/main/java/org/sonar/plugins/python/api/tree/OrPattern.java
```diff
@@ -0,0 +1,39 @@
...
+ *   case "foo" | "bar": ...
+ * </pre>
+ *
+ * See https://docs.python.org/3/reference/compound_stmts.html#or-patterns
+ */
+public interface OrPattern extends Pattern {
+
+  List<Pattern> patterns();
+
+  List<Token> separators();
```
andrea-guarino-sonarsource, line 37: do we need to have `separators` tokens in the interface?
I think they might be potentially useful for reporting issue locations,  however I don't know if we would ever need to access a specific separator. Do you have some other use case in mind?

andrea-guarino-sonarsource (approved):
LGTM, Just one question about `OrPattern#separators()`

python-frontend/src/main/java/org/sonar/plugins/python/api/tree/OrPattern.java
pynicolas: I don't know, I saw that [`ExpressionList`](https://github.com/SonarSource/sonar-python/blob/master/python-frontend/src/main/java/org/sonar/plugins/python/api/tree/ExpressionList.java) has a list of commas.
It could be useful if we want to raise an issue on the following code:
```
case 'foo' | 'foo': ...
```
What do you think?

andrea-guarino-sonarsource: I think that regarding tokens we've not been super consisting in the past. (sometimes we did keep them in the API, sometimes we didn't).

Anyway I'm fine with keeping `separators()`

PR #990, opened 2021-11-08
SONARPY-916 Match / case statement: support class patterns

andrea-guarino-sonarsource:


python-frontend/src/main/java/org/sonar/python/api/PythonGrammar.java
```diff
@@ -482,6 +487,12 @@ public static void compoundStatements(LexerfulGrammarBuilder b) {
     // TODO: ADD wildcard pattern for "*_"
     b.rule(STAR_PATTERN).is("*", CAPTURE_PATTERN);
 
+    b.rule(CLASS_PATTERN).is(NAME_OR_ATTR, "(", b.optional(PATTERN_ARGS), ")");
+    b.rule(NAME_OR_ATTR).is(NAME, b.zeroOrMore(".", NAME));
+    b.rule(PATTERN_ARGS).is(PATTERN_ARG, b.zeroOrMore(b.sequence(",", PATTERN_ARG)), b.optional(","));
+    b.rule(PATTERN_ARG).is(b.firstOf(KEYWORD_PATTERN, PATTERN));
+    b.rule(KEYWORD_PATTERN).is(NAME, "=", PATTERN);
```
guillaume-dequenne-sonarsource, line 494: We probably could follow the [official grammar](https://docs.python.org/3/reference/compound_stmts.html#class-patterns) and disallow keyword patterns if they come before positional ones. That way, we wouldn't need to check this in the `PythonTreeMaker`. Wdyt?

guillaume-dequenne-sonarsource (approved):
LGTM, though I wonder whether we could simplify things a bit by having the grammar itself ensure no positional pattern can come before keyword ones.

python-frontend/src/main/java/org/sonar/python/api/PythonGrammar.java
andrea-guarino-sonarsource: I initially followed the official grammar and I had a problem matching the following code:
```python
Foo(x, prop=42)
```
The problem being that the production rules:
```
pattern_arguments = positional_patterns ["," keyword_patterns]
positional_patterns ::=  ",".pattern+
```
will never match `keyword_patterns`, because `positional_patterns` will parse also `prop` as being a `CAPTURE_PATTERN`  because in my understanding PEG grammars (like SSLR is) won't backtrack (as lookahead is always 1).

I believe to have the "positional-before-keyword" constraint encoded in the grammar itself would be a bit more complex, so I decided to keep the grammar simple and have a postcondition check instead.

Alternatively we could also relax this constraint and be more flexible then the official grammar, as this is not very important for us.

Wdyt?

guillaume-dequenne-sonarsource: Right, that makes sense. I was puzzled that the official grammar was presented that way as they are PEG as well, but I guess they are doing other stuff under the hood to make it work.

PR #996, opened 2021-11-10
SONARPY-909 Add name bindings for capture patterns

andrea-guarino-sonarsource:


python-frontend/src/test/java/org/sonar/python/semantic/MatchStatementSymbolsTest.java
```diff
@@ -0,0 +1,177 @@
...
+    Symbol x = asPattern.alias().name().symbol();
+    assertThat(z.name()).isEqualTo("z");
+    assertThat(x.name()).isEqualTo("x");
+    assertThat(z.usages()).extracting(Usage::kind).containsExactly(PATTERN_DECLARATION, OTHER);
+    assertThat(x.usages()).extracting(Usage::kind).containsExactly(PATTERN_DECLARATION, OTHER);
+  }
+
+  @Test
+  public void or_pattern() {
+    OrPattern orPattern = patternFromCase("case A(x) | y: print(x + y)");
```
guillaume-dequenne-sonarsource, line 126: I have a syntax error with this construct:
```
SyntaxError: alternative patterns bind different names
```
I assume some similar case could never happen, as one case would always either not match or get short circuited.
Perhaps it would be more realistic to just test non-binding usages for or patterns?

python-frontend/src/test/java/org/sonar/python/semantic/MatchStatementSymbolsTest.java
```diff
@@ -0,0 +1,177 @@
+/*
```
guillaume-dequenne-sonarsource, line 1: For the sake of completeness, maybe we could have one test case with using guards as well?

guillaume-dequenne-sonarsource (changes requested):
Looks good. I think it would be worth improving one of the tests which contains a syntax error.

python-frontend/src/test/java/org/sonar/python/semantic/MatchStatementSymbolsTest.java
andrea-guarino-sonarsource: Good point, thanks!

guillaume-dequenne-sonarsource (approved):
LGTM

PR #999, opened 2021-11-11
SONARPY-924 Add Python 3.10 to supported versions

andrea-guarino-sonarsource:


python-frontend/src/main/java/org/sonar/plugins/python/api/PythonVersionUtils.java
```diff
@@ -75,6 +76,7 @@ public String serializedValue() {
     STRING_VERSION_MAP.put("3.7", V_37);
     STRING_VERSION_MAP.put("3.8", V_38);
     STRING_VERSION_MAP.put("3.9", V_39);
+    STRING_VERSION_MAP.put("3.10", V_310);
   }
   private static final Version MIN_SUPPORTED_VERSION = V_27;
   private static final Version MAX_SUPPORTED_VERSION = V_39;
```
guillaume-dequenne-sonarsource, line 82: Shouldn't we set `MAX_SUPPORTED_VERSION` to V_310 now?

guillaume-dequenne-sonarsource (changes requested):
Looks good, just a small comment.

By the way, this should probably be a separate ticket, but I believe we'd want to serialize the builtins from Typeshed for Python 3.10 as well, so the appropriate symbols can be loaded when that version is set (and so that it's part of the symbols for all versions as well).

I'm not sure how important this is to do as I don't know if there are any changes to the builtins with 3.10, but we should probably at least consider the idea. Wdyt?

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1030, opened 2021-12-10
SONARPY-949 S5756 (NonCallableCalled): avoid reporting on typeshed sy…

andrea-guarino-sonarsource:
…mbols having type 'Callable[..., T]'


python-frontend/src/main/java/org/sonar/python/types/InferredTypes.java
```diff
@@ -156,8 +156,10 @@ public static InferredType fromTypeshedProtobuf(SymbolsProtos.Type type) {
         String typeName = type.getFullyQualifiedName();
         return typeName.isEmpty() ? anyType() : runtimeType(TypeShed.symbolWithFQN(typeName));
       case TYPE_ALIAS:
-      case CALLABLE:
         return fromTypeshedProtobuf(type.getArgs(0));
+      case CALLABLE:
+        // this should be handled as a function type
```
guillaume-dequenne-sonarsource, line 161: Do we want a ticket for that? Are there FNs we would be able to avoid if we tracked such type correctly?

guillaume-dequenne-sonarsource (approved):
LGTM

python-frontend/src/main/java/org/sonar/python/types/InferredTypes.java
andrea-guarino-sonarsource: Good point, I'll create a ticket for this

PR #1031, opened 2021-12-10
SONARPY-950 S5708 (CaughtExceptionCheck) should not report on Ambiguo…

andrea-guarino-sonarsource:
…us Symbols that might inherit from BaseException


guillaume-dequenne-sonarsource (approved):
LGTM

PR #1032, opened 2021-12-10
SONARPY-942 Serialize unanalyzed overloaded items when regular ones are missing

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
```diff
@@ -452,6 +452,10 @@ public static boolean isValidForProjectPythonVersion(List<String> validForPython
         symbols.add(new FunctionSymbolImpl(((SymbolsProtos.FunctionSymbol) descriptor), isInsideClass));
       }
       if (descriptor instanceof OverloadedFunctionSymbol) {
+        if (((OverloadedFunctionSymbol) descriptor).getDefinitionsList().size() < 2) {
+          LOG.error("Overloaded function symbols should have at least two definitions.");
```
andrea-guarino-sonarsource, line 456: I think we shouldn't silently continue (as we do now). Instead I believe we should throw an Exception

andrea-guarino-sonarsource (approved):
LGTM, I think we should throw an exception when overloaded symbols < 2

PR #1034, opened 2021-12-14
SONARPY-954 Typeshed symbols retrieved by FQN should be consistent

andrea-guarino-sonarsource:


python-frontend/src/test/java/org/sonar/python/types/TypeShedTest.java
```diff
@@ -448,6 +448,16 @@ public void pythonVersions() {
...
   }
 
+  @Test
+  public void symbolWithFQN_should_be_consistent() {
+    // smtplib imports typing.Sequence only in Python3, hence typing.Sequence has kind CLASS
+    TypeShed.symbolsForModule("smtplib");
+    Symbol sequence = TypeShed.symbolWithFQN("typing.Sequence");
+    assertThat(sequence.kind()).isEqualTo(Kind.AMBIGUOUS);
+    Map<String, Symbol> typing = TypeShed.symbolsForModule("typing").stream().collect(Collectors.toMap(Symbol::name, Function.identity()));
+    assertThat(sequence).isEqualTo(typing.get("Sequence"));
```
guillaume-dequenne-sonarsource, line 458: _Nitpick:_ You could probably even assert `isSameAs` here.

python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
```diff
@@ -477,12 +477,6 @@ public static Symbol symbolWithFQN(String fullyQualifiedName) {
...
     }
-    for (Map<String, Symbol> symbolsByFqn : typeShedSymbols.values()) {
-      Symbol symbol = symbolsByFqn.get(fullyQualifiedName);
-      if (symbol != null) {
-        return symbol;
-      }
-    }
     String[] fqnSplittedByDot = fullyQualifiedName.split("\\.");
     String moduleName = Arrays.stream(fqnSplittedByDot, 0, fqnSplittedByDot.length - 1).collect(Collectors.joining("."));
     Set<Symbol> symbols = symbolsForModule(moduleName);
```
guillaume-dequenne-sonarsource, line 482: On line 484 (which I can't highlight as it wasn't changed), shouldn't we try to find a symbol with a given `name` instead of `fullyQualifiedName`?
So that we can retrieve symbols from "façade" modules that export imported symbols whose actual FQN might be different?
Something like that is done on line 190-193 on the other `symbolWithFQN` method.

guillaume-dequenne-sonarsource (changes requested):
Looks good, I left a small suggestion, but I'm not sure it actually changes anything as of now.
Also there is a typo in the PR name: retrivied -> retrieved

python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
andrea-guarino-sonarsource: I think it makes sense indeed. The only concern might be on performance, as we'd have to iterate on all the symbols of a module. We should probably optimize map lookup by name instead of having fqn as key

guillaume-dequenne-sonarsource: Indeed. I guess it can also be done later and just have some TODO + referencing an appropriate ticket for that, given this PR is really about fixing a inconsistency bug.

PR #1036, opened 2021-12-15
SONARPY-941 Make sure there are no conflicting symbols having the sam…

andrea-guarino-sonarsource:
…e name across Python versions

### DIFF
--- a/python-frontend/src/test/java/org/sonar/python/types/TypeShedTest.java
+++ b/python-frontend/src/test/java/org/sonar/python/types/TypeShedTest.java
@@ -28,7 +28,6 @@ import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.sonar.api.utils.log.LogTester;
 import org.sonar.api.utils.log.LoggerLevel;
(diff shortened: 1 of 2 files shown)


python-frontend/src/test/java/org/sonar/python/types/TypeShedTest.java
```diff
@@ -46,6 +46,7 @@
 import org.sonar.python.types.protobuf.SymbolsProtos;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
```
guillaume-dequenne-sonarsource, line 49: It seems this import (as well as `org.junit.Ignore`on line 31) is unused.

guillaume-dequenne-sonarsource (approved):
LGTM.
To be clear, does that mean this issue was actually fixed in a prior PR and this one is just here to update the tests?

python-frontend/src/test/java/org/sonar/python/types/TypeShedTest.java
andrea-guarino-sonarsource: thanks, I'll remove it!

andrea-guarino-sonarsource:
> 

Yes exactly, I'll add a comment in ticket to clarify the actual fix has been done in another PR

andrea-guarino-sonarsource:
Thinking a bit more about the solution we took in `TypeShed. disambiguateSymbolsWithSameName`, I think it makes sense to have a debug log when we disambiguate with a Python 3 symbol. This might be helpful during an investigation. I'll add this log in a new  commit, let me know if you think it's not a good idea

PR #1038, opened 2021-12-15
SONARPY-951 Translate starred parameter types to descriptors

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/FunctionSymbolImpl.java
```diff
@@ -218,9 +227,28 @@ private void addParameter(org.sonar.plugins.python.api.tree.Parameter parameter,
...
+          // if the form “*identifier” is present, it is initialized to a tuple receiving any excess positional parameters
+          isPositionalVariadic = true;
+          declaredType = InferredTypes.TUPLE;
+        }
+        if ("**".equals(starToken.value())) {
+          //  If the form “**identifier” is present, it is initialized to a new ordered mapping receiving any excess keyword arguments
+          isKeywordVariadic = true;
+          declaredType = InferredTypes.DICT;
+        }
+      }
```
andrea-guarino-sonarsource, line 246: can we keep this inside `getParameterType` method? I think `addParameter` is already quite long and it has Cognitive complexity of 18

python-frontend/src/main/java/org/sonar/plugins/python/api/symbols/FunctionSymbol.java
```diff
@@ -69,6 +69,8 @@
     boolean isVariadic();
     boolean isKeywordOnly();
     boolean isPositionalOnly();
+    boolean isKeywordVariadic();
```
andrea-guarino-sonarsource, line 72: I think it makes sense to add a short JavaDoc to say this returns true when you have `**param` (and similarly for `isPositionalVariadic`

andrea-guarino-sonarsource (approved):
Looks good, I left two minors suggestions

python-frontend/src/main/java/org/sonar/python/semantic/FunctionSymbolImpl.java
guillaume-dequenne-sonarsource: The issue is that this piece of code now produces the declared type as well as the boolean `isPositionalVariadic` and `isKeywordVariadic`.

While it's a bit complex due to the nesting and the multiple if/else, I felt it still was more understandable that way than if I wrapped this info in a different structure for example.

PR #1039, opened 2021-12-15
Fix remaining StackOverflowError condition

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/ProjectLevelSymbolTable.java
```diff
@@ -138,24 +137,13 @@ public Symbol getSymbol(@Nullable String fullyQualifiedName) {
...
     if (fullyQualifiedName == null) return null;
-    if (queriedSymbolNames.contains(fullyQualifiedName)) {
-      // cyclic dependencies
-      queriedSymbolNames = new HashSet<>();
-      String[] fqnSplitByDot = fullyQualifiedName.split("\\.");
-      localSymbolName =  localSymbolName != null ? localSymbolName : fqnSplitByDot[fqnSplitByDot.length - 1];
-      return new SymbolImpl(localSymbolName, fullyQualifiedName);
-    }
     Descriptor descriptor = globalDescriptorsByFQN().get(fullyQualifiedName);
-    if (descriptor == null) {
```
andrea-guarino-sonarsource, line 150: that's great that `queriedSymbolsNames` got removed!

andrea-guarino-sonarsource (commented):
I see no unit test for the remaining StackOverflowError, can you add one?

python-frontend/src/main/java/org/sonar/python/semantic/ProjectLevelSymbolTable.java
guillaume-dequenne-sonarsource: Indeed! As you suspected, it really wasn't a good solution in the end.

guillaume-dequenne-sonarsource:
> I see no unit test for the remaining StackOverflowError, can you add one?

I added a test ensuring we reuse the same symbols for super types and type hints when fetching the symbols from a given module. It's a bit difficult to craft a test actually producing a StackOverflow before and not after because I suspect it depends on the order of iteration of multiple sets (members of class, descriptors of module, etc...).

andrea-guarino-sonarsource (approved):

PR #1040, opened 2021-12-17
Avoid copying symbols that come from the project table and fix bug for ambiguous descriptors with null FQN

guillaume-dequenne-sonarsource:
Thanks to descriptors, symbols from the project symbol table are always new and shouldn't require any copy logic (`copyWithoutUsage` or `copySymbol`).

This PR also fix a bug when a descriptor within an ambiguous descriptor has a `null` FQN and cannot be retrieved properly by `createdSymbols`.

### DIFF
--- a/python-frontend/src/main/java/org/sonar/python/index/DescriptorUtils.java
+++ b/python-frontend/src/main/java/org/sonar/python/index/DescriptorUtils.java
@@ -117,6 +117,7 @@ public class DescriptorUtils {
     )).collect(Collectors.toList());
   }
 
+  // TODO SONARPY-958: Cleanup the symbol construction from descriptors by extracting this logic in a builder class
   public static Symbol symbolFromDescriptor(Descriptor descriptor, ProjectLevelSymbolTable projectLevelSymbolTable,
                                             @Nullable String localSymbolName, Map<Descriptor, Symbol> createdSymbolsByDescriptor, Map<String, Symbol> createdSymbolsByFqn) {
     if (createdSymbolsByDescriptor.containsKey(descriptor)) {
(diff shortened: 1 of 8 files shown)


python-frontend/src/main/java/org/sonar/python/index/DescriptorUtils.java
```diff
@@ -161,34 +170,36 @@ private static void addMembers(ClassSymbolImpl classSymbol, ClassDescriptor clas
...
-          if (createdSymbols.containsKey(superClassFqn)) {
-            return createdSymbols.get(superClassFqn);
+          if (createdSymbolsByFqn.containsKey(superClassFqn)) {
+            return createdSymbolsByFqn.get(superClassFqn);
           }
-          Symbol symbol = projectLevelSymbolTable.getSymbol(superClassFqn, null, createdSymbols);
-          symbol = symbol != null ? symbol : typeshedSymbolWithFQN(superClassFqn);
-          createdSymbols.put(superClassFqn, symbol);
+          Symbol symbol = projectLevelSymbolTable.getSymbol(superClassFqn, null, createdSymbolsByDescriptor, createdSymbolsByFqn);
+          symbol = symbol != null ? symbol : ((SymbolImpl) typeshedSymbolWithFQN(superClassFqn)).copyWithoutUsages();
```
andrea-guarino-sonarsource, line 183: Maybe we could call the `copyWithoutUsages` inside `typeshedSymbolWithFQN`?

python-frontend/src/main/java/org/sonar/python/index/DescriptorUtils.java
```diff
@@ -115,42 +116,50 @@ public static AmbiguousDescriptor ambiguousDescriptor(AmbiguousSymbol ambiguousS
   }
 
   public static Symbol symbolFromDescriptor(Descriptor descriptor, ProjectLevelSymbolTable projectLevelSymbolTable,
```
andrea-guarino-sonarsource, line 120: I think we should clean-up a bit those static methods that create symbols: this could become a Builder Class where we keep projectLevelSymbolTable, createdSymbolsByDescriptor and createdSymbolsByFqn as internal states. I don't think we need to do it in this PR, however  believe it's important to do it before releasing new version to avoid adding more technical dept. 

Let's add a TODO comment and maybe a task ticket in the Kanbar.

andrea-guarino-sonarsource (approved):
LGTM, I think it's ok to merge this PR to fix the problems we have on Peach. However, I think we should clean-up `DescriptorUtils` before shipping it to production

PR #1042, opened 2022-01-03
SONARPY-938 Protobuf typeshed symbols should contain information about imported modules

andrea-guarino-sonarsource:


python-frontend/src/test/java/org/sonar/python/types/TypeShedTest.java
```diff
@@ -193,9 +193,10 @@ public void package_inner_submodules_symbols() {
   @Test
   public void package_relative_import() {
     Map<String, Symbol> osSymbols = symbolsForModule("os");
-    // TODO: Add imported symbols SONARPY-938
-    // Symbol sysSymbol = osSymbols.get("sys");
-    // assertThat(sysSymbol.kind()).isEqualTo(Kind.AMBIGUOUS);
+    Symbol sysSymbol = osSymbols.get("sys");
+    assertThat(sysSymbol.kind()).isEqualTo(Kind.OTHER);
+    Set<String> sysExportedSymbols = symbolsForModule("sys").keySet();
+    assertThat(((SymbolImpl) sysSymbol).getChildrenSymbolByName().values()).extracting(Symbol::name).containsAll(sysExportedSymbols);
```
guillaume-dequenne-sonarsource, line 199: Do you think we should eventually move to actually use a `ModuleSymbol` type for this?
Certainly not in this PR, but for longer term I'm not sure.

python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
```diff
@@ -470,7 +470,13 @@ public static boolean isValidForProjectPythonVersion(List<String> validForPython
         symbols.add(fromOverloadedFunction(((OverloadedFunctionSymbol) descriptor), isInsideClass));
       }
       if (descriptor instanceof SymbolsProtos.VarSymbol) {
-        symbols.add(new SymbolImpl((SymbolsProtos.VarSymbol) descriptor));
+        SymbolsProtos.VarSymbol varSymbol = (SymbolsProtos.VarSymbol) descriptor;
+        SymbolImpl symbol = new SymbolImpl(varSymbol);
+        if (varSymbol.getIsImportedModule()) {
+          Set<Symbol> moduleExportedSymbols = searchTypeShedForModule(varSymbol.getFullyQualifiedName());
+          moduleExportedSymbols.forEach(symbol::addChildSymbol);
```
guillaume-dequenne-sonarsource, line 477: I don't think we should copy the symbols here as it would create a lot of them, but are we sure we don't risk anything by creating so many links between existing instances?
Probably not and it certainly won't matter once we move to descriptors, but just wanted to point it out to make sure.

guillaume-dequenne-sonarsource:
LGTM, left minor questions but I don't think they need addressing.

guillaume-dequenne-sonarsource (approved):
LGTM, left minor questions but I don't think they need addressing.

python-frontend/src/test/java/org/sonar/python/types/TypeShedTest.java
andrea-guarino-sonarsource: Indeed, I think it would be cleaner to have `ModuleSymbol` in the long run! Although I already tried once and I found some unexpected problems: https://jira.sonarsource.com/browse/SONARPY-658

python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
andrea-guarino-sonarsource: I think they are already copied when we actually import module symbols inside the local symbol table (inside Scope#copySymbol), so I'd say we don't need to copy them here. Does it make sense?

PR #1044, opened 2022-01-05
SONARPY-963 Custom stubs should have priority over protobuf generated symbols. Am…

andrea-guarino-sonarsource:
…bigous Symbols of all classes should be disambiguated

### DIFF
--- a/python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
+++ b/python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
@@ -198,15 +198,15 @@ public class TypeShed {
       return new HashSet<>();
     }
     modulesInProgress.add(moduleName);
-    Collection<Symbol> symbolsFromProtobuf = getSymbolsFromProtobufModule(moduleName).values();
-    if (!symbolsFromProtobuf.isEmpty()) {
-      modulesInProgress.remove(moduleName);
-      return new HashSet<>(symbolsFromProtobuf);
-    }
     Set<Symbol> customSymbols = new HashSet<>(getModuleSymbols(moduleName, CUSTOM_THIRD_PARTY, builtinGlobalSymbols).values());
     if (!customSymbols.isEmpty()) {
       modulesInProgress.remove(moduleName);
       return customSymbols;
+    }
+    Collection<Symbol> symbolsFromProtobuf = getSymbolsFromProtobufModule(moduleName).values();
+    if (!symbolsFromProtobuf.isEmpty()) {
+      modulesInProgress.remove(moduleName);
+      return new HashSet<>(symbolsFromProtobuf);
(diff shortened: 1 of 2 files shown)


python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
```diff
@@ -283,12 +283,28 @@ private static Symbol disambiguateWithLatestPythonSymbol(Set<Symbol> alternative
     return latestPythonSymbol;
   }
 
+
+  /**
+   * Returns stub symbols to be used by SonarSecurity.
+   * Ambiguous symbols that contain all class symbols are disambiguated with latest Python version.
```
guillaume-dequenne-sonarsource, line 289: The phrasing is a bit weird, I assume you mean "...that only contain class symbols"?

python-frontend/src/test/java/org/sonar/python/types/TypeShedTest.java
```diff
@@ -488,4 +488,10 @@ public void variables_from_protobuf() throws TextFormat.ParseException {
     assertThat(symbols.values()).extracting(Symbol::kind, Symbol::fullyQualifiedName, Symbol::annotatedTypeName)
       .containsExactlyInAnyOrder(tuple(Kind.OTHER, "mod.foo", "str"), tuple(Kind.OTHER, "mod.bar", null));
   }
+
+  @Test
+  public void stubFilesSymbols_should_not_contain_ambiguous_symbols_of_classes() {
```
guillaume-dequenne-sonarsource, line 493: I'm not sure I understand, I have this test pass even when I revert the implementation of `stubFilesSymbols` to its previous one. Yet, as this change has an actual impact on sonar-security ITs, I wonder how could this test pass without the change?

guillaume-dequenne-sonarsource (approved):
The change looks good, I'm a bit puzzled about the test for `stubFilesSymbols`, are we fixing an existing problem or being defensive (in which case I guess it's normal the test still pass without the change)?
In any case, as this has an impact on sonar-security ITs, I think this change deserves its own ticket.

python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
andrea-guarino-sonarsource: Indeed thanks, I will change it!

python-frontend/src/test/java/org/sonar/python/types/TypeShedTest.java
andrea-guarino-sonarsource: Indeed, my bad. I forgot to populate the cache of typeshed symbols. The tests passes because `Typeshed.typeShedSymbols` is empty in this test

PR #1046, opened 2022-01-06
SONARPY-935 Ensure there are no deprecated rules in the default quality profile

guillaume-dequenne-sonarsource:


andrea-guarino-sonarsource (commented):
Am I right that the unit test `test_no_deprecated_rule_in_default_profile ` won't fail if we update rule-api and readd by mistake S1523 and S4721? Didn't we say we wanted to hardcode them?

guillaume-dequenne-sonarsource:
> Am I right that the unit test `test_no_deprecated_rule_in_default_profile ` won't fail if we update rule-api and readd by mistake S1523 and S4721? Didn't we say we wanted to hardcode them?

Yes, and then I realized what actually happened was that the json for those rules had been correctly reverted but not the `Sonar_way_profile.json` file so I figured this was a good test to have. I didn't like too much the idea of having to maintain a hard-coded list of deprecated rules. At the same time it's true that we are still at risk of changing those status by mistakes if we do not pay attention during the rule-api execution, so I might just add this check as well.

PR #1047, opened 2022-01-07
SONARPY-962 Cleanup Typeshed class and fix FQN of methods of class symbols inheriting from private typeshed symbols

andrea-guarino-sonarsource:


python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
```diff
@@ -193,6 +149,87 @@ public static Symbol symbolWithFQN(String stdLibModuleName, String fullyQualifie
...
+  public static boolean isValidForProjectPythonVersion(List<String> validForPythonVersions) {
+    if (validForPythonVersions.isEmpty()) {
+      return true;
+    }
+    HashSet<String> intersection = new HashSet<>(validForPythonVersions);
+    intersection.retainAll(supportedPythonVersions);
+    return !intersection.isEmpty();
+  }
+
+  public static Set<Symbol> symbolsFromDescriptor(Set<Object> descriptors, boolean isInsideClass) {
```
guillaume-dequenne-sonarsource: Now that we have actual `Descriptor` classes, perhaps it would be better to rename this concept `ProtobufDescriptor` or something similar to clearly distinguish them.

guillaume-dequenne-sonarsource (approved):
LGTM, thanks for the refactoring of the `TypeShed` class, it makes things much clearer.
I left a minor renaming suggestion, feel free to actually implement it or not.

PR #1049, opened 2022-01-10
Typeshed#stubFilesSymbols should not contain null symbols for ambigou…

andrea-guarino-sonarsource:
…s third party symbols


guillaume-dequenne-sonarsource (approved):
LGTM

PR #1051, opened 2022-01-10
SONARPY-966 Update rules metadata

guillaume-dequenne-sonarsource:


python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S5843.html
```diff
@@ -3,43 +3,6 @@
 <p>The complexity of a regular expression is determined as follows:</p>
 <p>Each of the following operators increases the complexity by an amount equal to the current nesting level and also increases the current nesting
 level by one for its arguments:</p>
-<ul>
```
andrea-guarino-sonarsource, line 6: is this change expected?

guillaume-dequenne-sonarsource: This piece of metadata seems related to the Java implementation of the rule and duplicate of the Python one starting line 43 .

andrea-guarino-sonarsource (approved):

PR #1060, opened 2022-01-25
SONARPY-896 NOSONAR annotation should silence issues on multiline str…

guillaume-dequenne-sonarsource:
…ings


andrea-guarino-sonarsource (approved):
LGTM. The ticket says
>Out of consistency with other linters, we should also silence issues if the "NOSONAR" comment is placed after the opening quotes.

Could you please update the description or put a comment explaining we decided not to do that?

guillaume-dequenne-sonarsource:
I updated the comment to say that despite other linters doing it, we won't, as an alternative that doesn't impact the resulting string is possible now.

PR #1062, opened 2022-01-26
SONARPY-970 Serialize stub class members to Protobuf

guillaume-dequenne-sonarsource:


python-frontend/src/main/protobuf/symbols.proto
```diff
@@ -74,13 +74,14 @@ message ClassSymbol {
...
   repeated FunctionSymbol methods = 5;
   repeated OverloadedFunctionSymbol overloaded_methods = 6;
-  bool has_decorators = 7;
-  bool has_metaclass = 8;
-  bool is_enum = 9;
-  bool is_generic = 10;
-  bool is_protocol = 11;
-  optional string metaclass_name = 12;
-  repeated string valid_for = 13;
+  repeated VarSymbol vars = 7;
```
andrea-guarino-sonarsource, line 77: I think it's a bit weird to call this `vars`. What about `attributes`?

python-frontend/src/main/protobuf/symbols.proto
```diff
@@ -74,13 +74,14 @@ message ClassSymbol {
...
+  bool has_decorators = 8;
+  bool has_metaclass = 9;
+  bool is_enum = 10;
+  bool is_generic = 11;
+  bool is_protocol = 12;
+  optional string metaclass_name = 13;
+  repeated string valid_for = 14;
```
andrea-guarino-sonarsource, line 84: I think usually in protobuf we shouldn't change field numbers to guarantee backward compatibility. I would suggest to add the new field after `valid_for` with field number `14`.
It's true that we own both serializer and deserializer, so there shouldn't be risk in changing field numbers, however I think it's a good practice not to change those.
In addition, I think that many of changes we see in protobuf binaries might be bacause of this.

andrea-guarino-sonarsource (changes requested):
Overall it's good, I would suggest to avoid changing existing field numbers in `symbols.proto`

python-frontend/src/main/protobuf/symbols.proto
guillaume-dequenne-sonarsource: Indeed... I used `var` in `typeshed_serializer` because that allows to reuse the existing merge functions for both classes and modules through duck typing. I'd go for keeping that and then serialize to `attributes` on the protobuf side then, wdyt? Or would you prefer we have class attributes everywhere for consistency?

andrea-guarino-sonarsource: ok for me, we can keep `var` in `typeshed_serializer` and just change it the protobuf

PR #1064, opened 2022-01-28
 SONARPY-945 Use precomputed Typeshed symbols for custom stub files 

guillaume-dequenne-sonarsource:


python-frontend/src/test/java/org/sonar/python/types/CustomStubSanityTest.java
```diff
@@ -39,6 +40,8 @@
 public class CustomStubSanityTest {
 
   @Test
+  @Ignore
+  //TODO: remove or replace?
```
andrea-guarino-sonarsource, line 44: Why is this test ignored? I think this test currently fails because `Typeshed.supportedPythonVersions` is null.
I think we could call `Typeshed.resetBuiltinSymbols()` at the beginning of test and that should make it pass.

python-frontend/src/test/java/org/sonar/python/types/TypeShedTest.java
```diff
@@ -264,6 +264,15 @@ public void package_django_class_property_type() {
     assertThat(((ClassSymbol) requestSymbol).declaredMembers().iterator().next().annotatedTypeName()).isEqualTo("django.http.request.QueryDict");
   }
 
+  @Test
+  public void package_lxml_reexported_symbol_fqn() {
+    Map<String, Symbol> lxmlEtreeSymbols = symbolsForModule("lxml.etree");
+    Symbol requestSymbol = lxmlEtreeSymbols.get("ElementTree");
```
andrea-guarino-sonarsource, line 270: _Minor_ : is `requestSymbol` name intended? I would call it `elementTree`

python-frontend/typeshed_serializer/serializer/symbols.py
```diff
@@ -337,7 +338,8 @@ def __init__(self, name: str, fullname: str, is_imported_module=False, type_desc
 
     @classmethod
     def from_var(cls, var: mpn.Var, name: str = None):
-        return cls(var.name if name is None else name, var.fullname, type_descriptor=TypeDescriptor(var.type) if var.type else None)
+        return cls(var.name if name is None else name, var.fullname,
```
andrea-guarino-sonarsource, line 341: is this change intended?

python-frontend/typeshed_serializer/serializer/typeshed_serializer.py
```diff
@@ -129,14 +151,16 @@ def serialize_typeshed_stdlib_multiple_python_version():
...
     merged_modules = symbols_merger.merge_multiple_python_versions(is_third_parties)
     for mod in merged_modules:
-        symbols.save_module(merged_modules[mod], is_debug=is_debug, debug_dir="output_merge", is_stdlib=not is_third_parties)
+        symbols.save_module(merged_modules[mod], dir_name, is_debug=is_debug, debug_dir="output_merge")
 
 
 def main():
-    save_merged_symbols()
-    save_merged_symbols(is_third_parties=True)
+    #save_merged_symbols()
```
andrea-guarino-sonarsource: Unnecessary change?

sonar-python-plugin/pom.xml
```diff
@@ -150,8 +150,8 @@
             <configuration>
               <rules>
                 <requireFilesSize>
-                  <maxsize>9300000</maxsize>
-                  <minsize>8300000</minsize>
+                  <maxsize>13000000</maxsize>
```
andrea-guarino-sonarsource, line 153: I think you forgot to remove the old protobuf files from the previous location. After deleting those, we won't probably need to change jar size constraints.

When we serialize protobuf from `typeshed_serializer`, I think we should add code to clean the existing protobuf files. However , no need to do it in this PR, let's do that in a separate ticket/PR.

andrea-guarino-sonarsource (changes requested):
I left a few comments in the PR. Also, unit tests are missing for python code.

python-frontend/src/test/java/org/sonar/python/types/CustomStubSanityTest.java
guillaume-dequenne-sonarsource: In fact I think I made my mind to remove, but forgot. This test parses the custom stubs and verifies they're valid. I guess instead it should be removed and something else should do the same in the `Python` code (I thought maybe even just raise some error in the production code if Mypy returns with errors in fact), wdyt?

python-frontend/typeshed_serializer/serializer/symbols.py
guillaume-dequenne-sonarsource: It's just a formatting change because the line was too long I assume. I don't remember I did it to be honest though, and it could have been separated from actual changes, but now that it's here I guess it can stay?

python-frontend/typeshed_serializer/serializer/typeshed_serializer.py
guillaume-dequenne-sonarsource: Sorry, my mistake when reordering the commit, somehow this changed gets reverted in the .protobuf bulk update instead of not existing at all...

andrea-guarino-sonarsource (approved):
LGTM

PR #1065, opened 2022-01-31
SONARPY-902 S930 (ArgumentNumberCheck) should not report on instance …

andrea-guarino-sonarsource:
…methods called from class methods

### DIFF
--- a/python-checks/src/main/java/org/sonar/python/checks/ArgumentNumberCheck.java
+++ b/python-checks/src/main/java/org/sonar/python/checks/ArgumentNumberCheck.java
@@ -22,6 +22,7 @@ package org.sonar.python.checks;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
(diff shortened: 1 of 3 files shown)


python-checks/src/main/java/org/sonar/python/checks/ArgumentNumberCheck.java
```diff
@@ -101,6 +106,27 @@ private static void checkPositionalParameters(SubscriptionContext ctx, CallExpre
...
   }
 
+  private static boolean isCalledAsClassMethod(QualifiedExpression callee) {
+    return TreeUtils.getSymbolFromTree(callee.qualifier())
+      .filter(ArgumentNumberCheck::isFirstParameterOfClassMethod)
+      .isPresent();
+  }
+
+  private static boolean isFirstParameterOfClassMethod(Symbol symbol) {
+    return symbol instanceof SelfSymbolImpl && isParamOfClassMethod(symbol);
```
pynicolas, line 116: About `symbol instanceof SelfSymbolImpl`:
* Tests are still green if I remove it.
* The meaning of this expression looks unclear to me. Maybe we could consider adding a method `isSelf()` on `Symbol` (and I'm not sure what it would mean, my Python knowledge is rusty). Maybe that would require a change in the public API which we're not completely comfortable with.

andrea-guarino-sonarsource: `symbol instanceof SelfSymbolImpl` is indeed a shortcut to express "the symbol represents the first parameter of a method". I also thought about adding `isSelf()` to `Symbol` however I'm not sure it's really worth it adding this to the public API. I think I will go with a utility method on `SymbolUtils`

> Tests are still green if I remove it.
Good point, I will add a test using another parameter! Thanks!

andrea-guarino-sonarsource: While trying to write a test for `symbol instanceof SelfSymbolImpl`, I realized this condition is not needed. The only way to call another method from a parameter is via the first parameter (being `self` or `cls`).  I will remove it and I will simplify the code. Thanks!

PR #1066, opened 2022-01-31
SONARPY-904 S1066 (CollapsibleIfStatements): Reduce noise when breaki…

andrea-guarino-sonarsource:
…ng line length limit, when using walrus operator and when a comment is present


python-checks/src/main/java/org/sonar/python/checks/CollapsibleIfStatementsCheck.java
```diff
@@ -58,11 +63,36 @@ public void visitIfStatement(IfStatement ifStatement) {
...
+      || enclosingIfStatement.condition().is(ASSIGNMENT_EXPRESSION)
+      || hasCommentsBetweenEnclosingAndChildIf(singleIfChild, enclosingIfStatement);
+  }
+
+  private static boolean hasCommentsBetweenEnclosingAndChildIf(IfStatement singleIfChild, IfStatement enclosingIfStatement) {
+    return TreeUtils.tokens(enclosingIfStatement).stream()
+      .anyMatch(token -> !token.trivia().isEmpty() && token.trivia().get(0).token().line() < singleIfChild.firstToken().line());
+  }
+
+  private static boolean wouldCauseLongLineLength(IfStatement singleIfChild, IfStatement enclosingIf) {
```
guillaume-dequenne-sonarsource, line 89: It's a bit nitpicky but I think we should also account for the fact that, on a single line, an `and` as well as 2 additional spaces surrounding it would be required as well.
Currently, rewriting the following snippet with 2 nested ifs raises an issue, despite the line being 84 characters long:
```
if blaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa and bluuuuuu:
  ...

```

guillaume-dequenne-sonarsource (changes requested):
Looks good, just a tiny comment about the way we estimate the resulting line length.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1068, opened 2022-02-01
SONARPY-967 Handle Typeshed Python2 modules whose name differ from th…

andrea-guarino-sonarsource:
…eir Python 3 counterpart by capitalization only

### DIFF
--- a/python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
+++ b/python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
@@ -75,6 +75,16 @@ public class TypeShed {
   private static final Set<String> BUILTINS_TO_DISAMBIGUATE = new HashSet<>(
     Arrays.asList(INT, FLOAT, COMPLEX, STR, BuiltinTypes.SET, DICT, LIST, TUPLE, NONE_TYPE, BOOL, "type", "super", "frozenset", "memoryview"));
 
+  private static final Map<String, String> MODULES_TO_DISAMBIGUATE = new HashMap<>();
+  static {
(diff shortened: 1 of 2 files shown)


python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
```diff
@@ -79,6 +79,16 @@
...
     Arrays.asList(INT, FLOAT, COMPLEX, STR, BuiltinTypes.SET, DICT, LIST, TUPLE, NONE_TYPE, BOOL, "type", "super", "frozenset", "memoryview"));
 
+  private static final Map<String, String> MODULES_TO_DISAMBIGUATE = new HashMap<>();
+  static {
+    MODULES_TO_DISAMBIGUATE.put("ConfigParser", "2@ConfigParser");
+    MODULES_TO_DISAMBIGUATE.put("Queue", "2@Queue");
+    MODULES_TO_DISAMBIGUATE.put("SocketServer", "2@SocketServer");
+    MODULES_TO_DISAMBIGUATE.put("configparser", "configparser");
+    MODULES_TO_DISAMBIGUATE.put("queue", "queue");
+    MODULES_TO_DISAMBIGUATE.put("socketserver", "socketserver");
```
guillaume-dequenne-sonarsource, line 89: I guess those aren't even needed since we'll default to the `moduleName` anyway when the entry is not found, right?
Also we probably could have a short comment explaining those modules differ in Python 2 and Python 3 by capitalization only.

guillaume-dequenne-sonarsource (approved):
LGTM.

PR #1070, opened 2022-02-02
 SONARPY-972 Remove Typeshed parsing logic 

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -273,11 +269,9 @@ public void visitFileInput(FileInput tree) {
       createScope(tree, null);
       enterScope(tree);
       moduleScope = currentScope();
-      if (!SymbolUtils.isTypeShedFile(pythonFile) || !BASE_MODULES.contains(pythonFile.fileName())) {
-        Map<String, Symbol> typeShedSymbols = TypeShed.builtinSymbols();
-        for (String name : BuiltinSymbols.all()) {
-          currentScope().createBuiltinSymbol(name, typeShedSymbols);
-        }
+      Map<String, Symbol> typeShedSymbols = TypeShed.builtinSymbols();
+      for (String name : BuiltinSymbols.all()) {
```
andrea-guarino-sonarsource, line 273: probably we don't need anymore to go through `BuiltinSymbols.all()` , however let's do that in a separate PR :)

andrea-guarino-sonarsource (approved):
LGTM! 
I think there two additional things to consider:
- I believe we could (should?) keep the `LICENSE` file from typeshed
- Can we reduce the size constraints of plugin jar?

guillaume-dequenne-sonarsource:
> LGTM! I think there two additional things to consider:
> 
>     * I believe we could (should?) keep the `LICENSE` file from typeshed
> 
>     * Can we reduce the size constraints of plugin jar?

With regards to the license file, we still have it through our submodule of Typeshed in the serializer, so I think that should be enough?
Also unfortunately, plugin size still sits nicely within the constraints at 8.7MB (8.3 - 9.3 being the current range).

PR #1074, opened 2022-02-03
SONARPY-975 Generate typeshed protobuf messages in cirrus-ci only when needed

andrea-guarino-sonarsource:


.cirrus.yml
```diff
@@ -33,7 +33,31 @@ win_vm_definition: &WINDOWS_VM_DEFINITION
     type: n1-standard-8
     use_ssd: true
 
+generate_typeshed_task:
```
guillaume-dequenne-sonarsource, line 36: With this approach, aren't we building `python-frontend` twice when there are changes to the serializer? Once with the `generateTypeshed` flag, and once without?

I guess we don't risk overriding anything but maybe we could also have the build script setting `generateTypeshed` conditionally depending on the changes, so that there is no duplication? One counter argument could be that then it's less visible whether Typeshed has been generated or not (though maybe we could simply have a log for that?)

guillaume-dequenne-sonarsource (approved):
LGTM. I left a suggestion for a different approach, not sure whether it's feasible though.

.cirrus.yml
andrea-guarino-sonarsource: As discussed, it looks like we could only call `changesInclude` inside values of `skip` and `only_if` properties. Going to keep it as a separate task for the time being

PR #1076, opened 2022-02-04
Execute typeshed serializer unit tests during CI

andrea-guarino-sonarsource:


.cirrus.yml
```diff
@@ -82,6 +82,8 @@ build_task:
   build_script:
     - git submodule update --init
     - source cirrus-env BUILD
+    - mvn -f python-frontend/pom.xml protobuf:compile-python@generate-protobuf-python-sources
+    - tox -c python-frontend/typeshed_serializer/tox.ini
```
guillaume-dequenne-sonarsource, line 86: If you run tox like that, doesn't it run both the tests and the serialization? I'm not sure what is the exact command to only run the tests, though...

guillaume-dequenne-sonarsource (commented):
I'm not sure about `tox` behaviour here

.cirrus.yml
andrea-guarino-sonarsource: Indeed thanks!
After some manual test, I believe that running directly `python -m pytest --cov=serializer --cov-report xml:cov.xml --ignore=resources` will only execute the tests

PR #1080, opened 2022-02-07
 SONARPY-974 Add extended ruling test for bug rules 

guillaume-dequenne-sonarsource:


.cirrus.yml
```diff
@@ -139,9 +139,33 @@ ruling_task:
...
+    SONARSOURCE_QA: true
+  maven_cache:
+    folder: ${CIRRUS_WORKING_DIR}/.m2/repository
+  submodules_script:
+    - git submodule update --init
+  ruling_script:
+    - source cirrus-env QA
+    - source set_maven_build_version $BUILD_NUMBER
+    - cd its/ruling
+    - mvn verify -Dsonar.runtimeVersion=LATEST_RELEASE -Dmaven.test.redirectTestOutputToFile=false -B -e -V -Dtest=PythonExtendedRulingTest
```
andrea-guarino-sonarsource, line 165: I think we can run methods of `PythonExtendedRulingTest` in parallel.
I tried locally by adding the following to `its/ruling/pom.xml`:
```xml
  <build>
    <plugins>
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-surefire-plugin</artifactId>
        <configuration>
          <parallel>methods</parallel>
          <threadCount>10</threadCount>
        </configuration>
      </plugin>
    </plugins>
  </build> 
```

and I had an improvement from 6:16 min 4:31 min . I think it might be worth trying if we can get some improvement on Cirrus-CI. Probably the threadCount should be the same of the configured CPU.

its/ruling/src/test/java/org/sonar/python/it/PythonExtendedRulingTest.java
```diff
@@ -0,0 +1,189 @@
...
+  public void test_autokeras() throws IOException {
+    executeBuild(buildWithCommonProperties("autokeras"));
+  }
+
+  @Test
+  public void test_black() throws IOException {
+    SonarScanner build = buildWithCommonProperties("black");
+    build.setProperty("sonar.sources", "src");
+    build.setProperty("sonar.tests", "tests");
+    build.setProperty("sonar.test.exclusions", "tests/data/async_as_identifier.py");
```
andrea-guarino-sonarsource, line 70: I think we should have this level of configuration for all other projects as well.  We can probably copy it from `peachee-languages`

- This should help us reduce the analysis time, avoiding to run rules on tests files
- I'd expect we have high quality issues from this `extendedRulingTest` , so probably it's better to avoid analyzing test sources and only focus on production code

its/ruling/src/test/java/org/sonar/python/it/PythonExtendedRulingTest.java
```diff
@@ -0,0 +1,189 @@
...
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.sonar.python.it.RulingHelper.bugRuleKeys;
+import static org.sonar.python.it.RulingHelper.getOrchestrator;
+
+public class PythonExtendedRulingTest {
```
andrea-guarino-sonarsource, line 38: I'd put a comment explaining why we decided to have two ruling and this is meant to run only bug rules

its/ruling/src/test/java/org/sonar/python/it/RulingHelper.java
```diff
@@ -0,0 +1,98 @@
...
+
+  static List<String> bugRuleKeys() {
+    return Arrays.asList(
+      "PreIncrementDecrement",
+      "S1045",
+      "S1143",
+      "S1226",
+      "S1656",
+      "S1716",
+      "S1717",
```
andrea-guarino-sonarsource, line 50: looks like this rule is deprecated, so probably no need to include it

its/ruling/src/test/java/org/sonar/python/it/RulingHelper.java
```diff
@@ -0,0 +1,98 @@
...
+      "S1763",
+      "S1764",
+      "S1862",
+      "S2159",
+      "S2190",
+      "S2201",
+      "S2275",
+      "S2711",
+      "S2712",
+      "S2733",
```
andrea-guarino-sonarsource, line 61: This one is deprecated as well

its/ruling/src/test/java/org/sonar/python/it/RulingHelper.java
```diff
@@ -0,0 +1,98 @@
...
+
+  static Orchestrator getOrchestrator() {
+    return Orchestrator.builderEnv()
+      .setSonarVersion(System.getProperty(SQ_VERSION_PROPERTY, DEFAULT_SQ_VERSION))
+      .addPlugin(FileLocation.byWildcardMavenFilename(new File("../../sonar-python-plugin/target"), "sonar-python-plugin-*.jar"))
+      .addPlugin(MavenLocation.of("org.sonarsource.sonar-lits-plugin", "sonar-lits-plugin", "0.10.0.2181"))
+      .build();
+  }
+
+  static List<String> bugRuleKeys() {
```
andrea-guarino-sonarsource, line 42: I think it might be annoying to manually maintain this list when we'll add new bug rules or deprecate / remove existing ones. 
What about going through all json descriptors and filter rules having `"type": "BUG"` and `"status": "READY"`?

andrea-guarino-sonarsource (changes requested):

its/ruling/src/test/java/org/sonar/python/it/RulingHelper.java
```diff
@@ -0,0 +1,98 @@
...
+      "S5756",
+      "S5796",
+      "S5807",
+      "S5828",
+      "S5842",
+      "S5850",
+      "S5855",
+      "S5856",
+      "S5868",
+      "S5953",
```
andrea-guarino-sonarsource, line 91: This one is the rule about undefined symbols and I'm afraid could be quite noisy. What about excluding this rule?

its/ruling/src/test/java/org/sonar/python/it/RulingHelper.java
guillaume-dequenne-sonarsource: Agreed, it would be much cleaner. Surprinsingly, I didn't manage to get this to work in a clean way, I created SONARPY-984 to fix this in a later sprint if that's okay for you.

PR #1082, opened 2022-02-08
Update wss-unified-agent.config

andrea-guarino-sonarsource:


.cirrus.yml
```diff
@@ -77,7 +77,12 @@ ws_scan_task:
   depends_on:
     - build
   gke_container:
-    <<: *CONTAINER_DEFINITION
+    dockerfile: .cirrus/python-env.Dockerfile
+    builder_image_project: language-team
+    builder_image_name: docker-builder-lt-v1
+    cluster_name: cirrus-ci-lt-cluster
+    zone: us-central1-a
+    namespace: default
```
nils-werner-sonarsource, line 85: Since this is the same container definition as for the build task, it may be worth storing it in something like BUILD_CONTAINER_DEFINITION to be more maintainable.

nils-werner-sonarsource (approved):
LGTM! Please see my comment.

.cirrus.yml
andrea-guarino-sonarsource: Indeed, makes sense! thanks :)

PR #1088, opened 2022-03-02
SONARPY-264 : Rule S1135 : Track uses of 'TODO' tags

marco-bearzi-sonarsource:


python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S1135.html
```diff
@@ -0,0 +1,13 @@
+<p><code>FIXME</code> tags are commonly used to mark places where a bug is suspected, but which the developer wants to deal with later.</p>
```
pynicolas, line 1: TODO: Make this rule description target TODO instead of FIXME.

pynicolas (approved):

PR #1089, opened 2022-03-07
SONARPY-234: Rule 1291: Track uses of 'NOSONAR' comments

marco-bearzi-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/NoSonarCommentCheck.java
```diff
@@ -0,0 +1,49 @@
...
+import org.sonar.plugins.python.api.tree.Trivia;
+
+import static org.sonar.python.metrics.FileLinesVisitor.containsNoSonarComment;
+
+
+@Rule(key = "NoSonar")
+
+public class NoSonarCommentCheck extends PythonSubscriptionCheck {
+
+  private static final String MESSAGE = "Is //NOSONAR used to exclude false-positive or to hide real quality flaw ?";
```
pynicolas, line 35:
```suggestion
  private static final String MESSAGE = "Is //NOSONAR used to exclude false-positive or to hide real quality flaw?";
```

python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S1291.html
```diff
@@ -0,0 +1,4 @@
+<p>Any issue to quality rule can be deactivated with the <code>NOSONAR</code> marker. This marker is pretty useful to exclude false-positive results
```
pynicolas, line 1: We should not have both `NoSonar.html` and `S1291.html`.

python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S1291.json
```diff
@@ -0,0 +1,17 @@
+{
```
pynicolas, line 1: We should not have both `NoSonar.json` and `S1291.json`.

pynicolas:
I believe we should have an integration test for the NoSonar rule.
[Example](https://github.com/SonarSource/sonar-php/blob/master/its/plugin/tests/src/test/java/com/sonar/it/php/NoSonarTest.java#L61-L62)

python-checks/src/main/java/org/sonar/python/checks/NoSonarCommentCheck.java
```diff
@@ -0,0 +1,49 @@
...
+import org.sonar.plugins.python.api.tree.Trivia;
+
+import static org.sonar.python.metrics.FileLinesVisitor.containsNoSonarComment;
+
+
+@Rule(key = "NoSonar")
+
+public class NoSonarCommentCheck extends PythonSubscriptionCheck {
+
+  private static final String MESSAGE = "Is //NOSONAR used to exclude false-positive or to hide real quality flaw?";
```
nils-werner-sonarsource, line 35:
```suggestion
  private static final String MESSAGE = "Is #NOSONAR used to exclude false-positive or to hide real quality flaw?";
```
`//` would not be the correct comment opener in Python

python-checks/src/test/resources/checks/nosonarComment.py
```diff
@@ -0,0 +1,27 @@
+def divide(numerator, denominator):
+# Noncompliant@+1
```
nils-werner-sonarsource, line 2:
```suggestion
# Noncompliant@+1 {{Is #NOSONAR used to exclude false-positive or to hide real quality flaw?}}
```
Besides checking the location, you can also check for the message of an issue.

python-frontend/src/main/java/org/sonar/python/metrics/FileLinesVisitor.java
```diff
@@ -138,14 +138,19 @@ private void visitToken(Token token) {
...
       addNoSonarLines(trivia, parentToken);
     } else if (!isBlank(commentLine)) {
       linesOfComments.add(line);
     }
   }
 
+  public static boolean containsNoSonarComment(Trivia trivia){
+    String commentLine = getContents(trivia.token().value());
+    return commentLine.contains("NOSONAR");
+  }
```
nils-werner-sonarsource, line 152: I'm not sure if it's really necessary to create a public method here. Especially when your check accesses the helper method from a non-check-related class like this visitor. I would leave the FileLineVisitor as it is and just duplicate the one-line logic in the check.

nils-werner-sonarsource (approved):
LGTM! Please see my comment. I'm happy to discuss or clarify.

python-frontend/src/main/java/org/sonar/python/metrics/FileLinesVisitor.java
marco-bearzi-sonarsource: It has been discussed and, while the location of this method is indeed not the best, we've chosen to keep it there for the moment to have a consistency when detecting the NOSONAR marker as it is also used in visitComment() right above.

python-checks/src/test/resources/checks/nosonarComment.py
marco-bearzi-sonarsource: Added this message to the test file as well as another Noncompliant@+1 above this line since, due to the current way to detect NOSONAR marker, it will also raise a issue

PR #1090, opened 2022-03-08
SONARPY-282: Rule 1940: Boolean checks should not be inverted

marco-bearzi-sonarsource:


python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S1940.html
```diff
@@ -0,0 +1,12 @@
...
+<h2>Noncompliant Code Example</h2>
+<pre>
+if ( !(a == 2)) { ...}  // Noncompliant
+boolean b = !(i &lt; 10);  // Noncompliant
+</pre>
+<h2>Compliant Solution</h2>
+<pre>
+if (a != 2) { ...}
+boolean b = (i &gt;= 10);
+</pre>
```
pynicolas, line 11: That doesn't look like python code.

python-checks/src/main/java/org/sonar/python/checks/BooleanCheckNotInvertedCheck.java
```diff
@@ -0,0 +1,101 @@
...
+public class BooleanCheckNotInvertedCheck extends PythonSubscriptionCheck {
+
+  private static final String MESSAGE = "Use the opposite operator (\"%s\") instead.";
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.NOT, ctx -> checkNotOutsideParentheses(ctx, (UnaryExpression) ctx.syntaxNode()));
+  }
+
+  private static void checkNotOutsideParentheses(SubscriptionContext ctx, UnaryExpression original) {
```
pynicolas, line 44: The method name doesn't seem very appropriate. I suggest something like `checkUnaryNot` or `checkNotExpression`.

python-checks/src/main/java/org/sonar/python/checks/BooleanCheckNotInvertedCheck.java
```diff
@@ -0,0 +1,101 @@
...
+    while (negatedExpr.is(Tree.Kind.PARENTHESIZED)) {
+      negatedExpr = ((ParenthesizedExpression) negatedExpr).expression();
+    }
+    if(negatedExpr.is(Tree.Kind.COMPARISON)) {
+      BinaryExpression binaryExp = (BinaryExpression) negatedExpr;
+      // Don't raise warning with "not a == b == c" because a == b != c is not equivalent
+      if(!binaryExp.leftOperand().is(Tree.Kind.COMPARISON)) {
+        ctx.addIssue(original, String.format(MESSAGE, oppositeOperator(binaryExp.operator())));
+      }
+    } else if(negatedExpr.is(Tree.Kind.IN) || negatedExpr.is(Tree.Kind.IS) ) {
```
pynicolas, line 55:
```suggestion
    } else if (negatedExpr.is(Tree.Kind.IN, Tree.Kind.IS)) {
```

python-checks/src/main/java/org/sonar/python/checks/BooleanCheckNotInvertedCheck.java
```diff
@@ -0,0 +1,101 @@
...
+import org.sonar.plugins.python.api.tree.UnaryExpression;
+
+@Rule(key = "S1940")
+public class BooleanCheckNotInvertedCheck extends PythonSubscriptionCheck {
+
+  private static final String MESSAGE = "Use the opposite operator (\"%s\") instead.";
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.NOT, ctx -> checkNotOutsideParentheses(ctx, (UnaryExpression) ctx.syntaxNode()));
```
pynicolas, line 41: Adding an import for `Tree.Kind` may slightly improve readability.

python-checks/src/test/resources/checks/booleanCheckNotInverted.py
```diff
@@ -0,0 +1,51 @@
...
+
+l = not ((((((a > b))))))  # Noncompliant
+
+o = not a == b  # Noncompliant
+
+m = not a == b == 1
+
+n = not (not (a == b))  # Noncompliant
+
+p = not (1 == a == b) == 2 # Noncompliant
```
pynicolas, line 31: * I'm not sure what this test is about
* It's not clear on which expression we expect an issue

python-checks/src/test/resources/checks/booleanCheckNotInverted.py
```diff
@@ -0,0 +1,51 @@
...
+
+q = not (a == 1) == b == c == (d and e)
+
+r = not (a and 1) == b == c == (d and e)
+s = a is not b
+
+t = not (a is b)  # Noncompliant {{Use the opposite operator ("is not") instead.}}
+#   ^^^^^^^^^^^^
+u = not (a is not b)  # Noncompliant
```
pynicolas, line 40: I believe that adding an assertion on the issue message could be valuable, at least to check it's not "is not not".

pynicolas (approved):

PR #1091, opened 2022-03-09
SONARPY-267 Rule S1172: Unused function parameters should be removed

nils-werner-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/UnusedFunctionParametersCheck.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  private static void checkFunctionParameter(SubscriptionContext ctx, FunctionDef functionDef, Set<Symbol> symbols) {
+    if (CheckUtils.isCallingLocalsFunction(functionDef) || maybeOverridingMethod(functionDef) || isInterfaceMethod(functionDef)) {
+      return;
+    }
+
+    symbols.stream()
+      .map(UnusedFunctionParametersCheck::getUnusedParameter)
+      .filter(Objects::nonNull)
```
pynicolas, line 55: Minor: I don't really like this way of filtering.
Vague idea, I'm not sure its' better:
```
symbols.stream()
  .filter(symbol -> !"self".equals(symbol.name()))
  .map(symbol -> symbol.usages())
  .filter(usages -> usages.size() == 1 && usages.get(0).tree().parent().is(Kind.PARAMETER))
  .map(usages -> (Parameter) usages.get(0).tree().parent())
  .foreach(...)
```

python-checks/src/main/java/org/sonar/python/checks/UnusedFunctionParametersCheck.java
```diff
@@ -0,0 +1,88 @@
...
+    return functionDef.body().statements().stream()
+      .allMatch(statement -> statement.is(Kind.PASS_STMT, Kind.RAISE_STMT)
+        || (statement.is(Kind.EXPRESSION_STMT) && isStringExpression((ExpressionStatement) statement)));
+  }
+
+  private static boolean isStringExpression(ExpressionStatement stmt) {
+    return stmt.expressions().stream().allMatch(expr -> expr.is(Kind.STRING_LITERAL));
+  }
+
+  private static boolean maybeOverridingMethod(FunctionDef functionDef) {
```
pynicolas, line 81: That's really the kind of feature I would expect from the symbol table. Let's keep it like that for the moment.

python-checks/src/test/resources/checks/unusedFunctionParameters.py
```diff
@@ -0,0 +1,54 @@
...
+
+def g(read_param):
+    print(read_param)
+
+
+def f(unread_param):
+    return locals()
+
+
+def g(read_param):  # Noncompliant
```
pynicolas, line 14: `read_param` is not a good name here

pynicolas (approved):

PR #1092, opened 2022-03-10
SONARPY-259 Rule S138: Functions should not have too many lines of code

nils-werner-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/TooManyLinesInFunctionCheck.java
```diff
@@ -0,0 +1,95 @@
...
+
+    private void scan(Tree element) {
+      Deque<Tree> stack = new ArrayDeque<>();
+      stack.push(element);
+      while (!stack.isEmpty()) {
+        Tree currentElement = stack.pop();
+        if (currentElement.is(Tree.Kind.TOKEN)) {
+          linesOfCode.addAll(countTokenLines((Token) currentElement));
+        }
+        for (int i = currentElement.children().size() - 1; i >= 0; i--) {
```
pynicolas, line 87: Why do we iterate in reverse order?

python-frontend/src/main/java/org/sonar/python/metrics/FileLinesVisitor.java
```diff
@@ -101,15 +101,21 @@ private void visitNode(SubscriptionContext ctx) {
     }
   }
 
-  private void handleDocString(@Nullable StringLiteral docstring) {
+  protected void handleDocString(@Nullable StringLiteral docstring) {
+    linesOfDocstring.addAll(countDocstringLines(docstring));
+  }
+
+  public static Set<Integer> countDocstringLines(@Nullable StringLiteral docstring) {
```
pynicolas, line 108: The name of this method seems misleading to me. I would remove "count" from it.
Maybe something like `docStringLineNumbers` would be clearer.

python-frontend/src/main/java/org/sonar/python/metrics/FileLinesVisitor.java
```diff
@@ -121,18 +127,24 @@ private void visitToken(Token token) {
...
     }
 
+    linesOfCode.addAll(countTokenLines(token));
+
+    for (Trivia trivia : token.trivia()) {
+      visitComment(trivia, token);
+    }
+  }
+
+  public static Set<Integer> countTokenLines(Token token) {
```
pynicolas, line 137: Same thing: `tokenLineNumbers`? `getTokenLineNumbers`?

python-frontend/src/main/java/org/sonar/python/metrics/FileLinesVisitor.java
```diff
@@ -121,18 +127,24 @@ private void visitToken(Token token) {
...
+    Set<Integer> lines = new HashSet<>();
```
pynicolas, line 138: Most tokens wouldn't need a `HashSet`. I wonder if this change has a visible impact on performance (e.g. our ruling). Probably not.

python-checks/src/test/resources/checks/tooManyLinesInFunctionCustom.py
```diff
@@ -0,0 +1,22 @@
...
+    print(1)
+    print(1)
+    print(1)
+    print(1)
+    pass
+
+def function_6_and_empty_lines():
+    print(1)
+
+    print(1)
```
pynicolas, line 14: What's the point of this test?

python-checks/src/test/resources/checks/tooManyLinesInFunctionCustom.py
```diff
@@ -0,0 +1,22 @@
+# Noncompliant@+1 {{This function "function_6" has 6 lines of code, which is greater than the 5 authorized. Split it into smaller functions.}}
```
pynicolas, line 1: I feel that our unit tests are a bit weak. I would like to see more cases covered where each case differs from another with exactly one difference. We may discuss about it.

pynicolas (commented):
You finally found a rather simple way to implement this rule: nice job!

PR #1096, opened 2022-03-14
SONARPY-253: Rule S2761: Double prefix operators not and ~ should not be used 

marco-bearzi-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/DoublePrefixOperatorCheck.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  private static void doubleInversionCheck(SubscriptionContext ctx, UnaryExpression original) {
+    Expression invertedExpr = original.expression();
+    while (invertedExpr.is(Tree.Kind.PARENTHESIZED)) {
+      invertedExpr = ((ParenthesizedExpression) invertedExpr).expression();
+    }
+    if(invertedExpr.is(Tree.Kind.NOT, Tree.Kind.BITWISE_COMPLEMENT) && original.is(invertedExpr.getKind())) {
+      ctx.addIssue(original, String.format(MESSAGE, original.operator().value()));
+    }
```
andrea-guarino-sonarsource, line 50: Regarding `BITWISE_COMPLEMENT` : I think we should check that the type of the operand is `integer` to avoid FPs when `~` operator is overloaded.
For example:

```python
class Foo:
  def __invert__(self):
     return self.bar == 42

f  = Foo()
~~f # here ~~ it might have a different semantic 
```

andrea-guarino-sonarsource (changes requested):
Nice work, I think it's important to acknowledge `not not` has a valid use case, and that we want to raise an issue for its poor readability.
Also, we should make sure `~~` has not been overloaded to avoid FP.

python-checks/src/main/java/org/sonar/python/checks/DoublePrefixOperatorCheck.java
```diff
@@ -0,0 +1,69 @@
...
+
+  private static void doubleInversionCheck(SubscriptionContext ctx, UnaryExpression original) {
+    Expression invertedExpr = original.expression();
+    boolean doubleInversionFollowed = true;
+    while (invertedExpr.is(Tree.Kind.PARENTHESIZED)) {
+      doubleInversionFollowed = false;
+      invertedExpr = ((ParenthesizedExpression) invertedExpr).expression();
+    }
+
+    if(invertedExpr.is(Tree.Kind.NOT, Tree.Kind.BITWISE_COMPLEMENT) && original.is(invertedExpr.getKind())) {
```
andrea-guarino-sonarsource, line 53: I believe the code is not correctly formatted. You could call "format" action on IntelliJ to have the correct whitespaces

python-checks/src/main/java/org/sonar/python/checks/DoublePrefixOperatorCheck.java
```diff
@@ -0,0 +1,69 @@
...
+import org.sonar.plugins.python.api.tree.Tree;
+import org.sonar.plugins.python.api.tree.UnaryExpression;
+import org.sonar.python.types.InferredTypes;
+
+
+@Rule(key = "S2761")
+
+public class DoublePrefixOperatorCheck extends PythonSubscriptionCheck {
+
+  private static final String MESSAGE = "Use the (\"%s\") operator just once or not at all.?";
```
andrea-guarino-sonarsource, line 36: why the question mark at the end of the message? Also I would remove the the parenthesis around the operator.

python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S2761.html
```diff
@@ -0,0 +1,24 @@
...
+
+c = not not a # Noncompliant
+d = ~~b # Noncompliant
+</pre>
+<h2>Compliant Solution</h2>
+<pre>
+a = 0
+b = False
+
+c = not a # Compliant
```
andrea-guarino-sonarsource, line 18: To be consistent, I think the corresponding compliant code should be:
```python
a = 0
b = False

c = bool(a)
d = b
```

python-checks/src/main/java/org/sonar/python/checks/DoublePrefixOperatorCheck.java
```diff
@@ -0,0 +1,69 @@
...
+      doubleInversionFollowed = false;
+      invertedExpr = ((ParenthesizedExpression) invertedExpr).expression();
+    }
+
+    if(invertedExpr.is(Tree.Kind.NOT, Tree.Kind.BITWISE_COMPLEMENT) && original.is(invertedExpr.getKind())) {
+      if(doubleInversionFollowed){
+        if(invertedExpr.is(Tree.Kind.NOT)){
+          ctx.addIssue(original, MESSAGE_NOT);
+        }else{
+          // Inferred BITWISE_COMPLEMENT; type() allows to get the type inside parentheses
```
andrea-guarino-sonarsource, line 58: In my opinion, I don't think we really need this comment

python-checks/src/main/java/org/sonar/python/checks/DoublePrefixOperatorCheck.java
```diff
@@ -0,0 +1,69 @@
...
+          if(((UnaryExpression) invertedExpr).expression().type() == InferredTypes.INT){
+            ctx.addIssue(original, String.format(MESSAGE, original.operator().value()));
+          }
+          // Overloaded __invert__ should not raise any warning
```
andrea-guarino-sonarsource, line 62: I would prefer put this comment outside at the beginning of `doubleInversionCheck` method.

andrea-guarino-sonarsource (approved):
Good work! I left a few minor comments

PR #1097, opened 2022-03-15
SONARPY-212 Rule S3801: Functions should use "return" consistently

pynicolas:


python-checks/src/main/java/org/sonar/python/checks/ConsistentReturnCheck.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  private static void addIssue(SubscriptionContext ctx, FunctionDef functionDef, List<Tree> endStatements) {
+    PreciseIssue issue = ctx.addIssue(functionDef.name(), MESSAGE);
+    for (Tree statement : endStatements) {
+      if (statement.is(Kind.RETURN_STMT)) {
+        ReturnStatement returnStatement = (ReturnStatement) statement;
+        boolean hasValue = hasValue(returnStatement);
+        issue.secondary(statement, String.format("Return %s value", hasValue ? "with" : "without"));
+      } else if (statement.is(Kind.IF_STMT, Kind.FOR_STMT, Kind.WHILE_STMT)) {
```
andrea-guarino-sonarsource, line 80: I think we should highlight only the first token for all compound statements, i.e. also for match/case statements and function / class definitions (although it's unlike to end a function with those).

python-checks/src/test/resources/checks/consistentReturn.py
```diff
@@ -0,0 +1,114 @@
...
+  while True:
+    x = get_random_string(32)
+    if valid(x):
+      return x
+
+def while_with_unknown_condition(x): # Noncompliant
+  while condition(x):
+    x = get_random_string(32)
+    if valid(x):
+      return x
```
andrea-guarino-sonarsource, line 114: We have FP with unreachable code like the following:

```python
def unreachable(x):
  if x:
    return 42
  else:
    return 0
  return

```

I think we should probably exclude unreachable cfg blocks

andrea-guarino-sonarsource (changes requested):
Overall looks good: I think we should probably avoid the FPs with dead code. We can have a short discussion if you think otherwise

python-checks/src/main/java/org/sonar/python/checks/ConsistentReturnCheck.java
```diff
@@ -0,0 +1,98 @@
...
+public class ConsistentReturnCheck extends PythonSubscriptionCheck {
+
+  public static final String MESSAGE = "Refactor this function to use \"return\" consistently.";
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Kind.FUNCDEF, ctx -> {
+      FunctionDef functionDef = (FunctionDef) ctx.syntaxNode();
+      ControlFlowGraph cfg = ControlFlowGraph.build(functionDef, ctx.pythonFile());
+      if (cfg == null || hasExceptOrFinally(cfg)) {
```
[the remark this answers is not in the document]
pynicolas: I actually suspect that iterating over all descendant trees would be more costly than building a CFG.

python-checks/src/test/resources/checks/consistentReturn.py
pynicolas: I didn't do it, we may do it later.

andrea-guarino-sonarsource (approved):
LGTM!

PR #1100, opened 2022-03-17
SONARPY-215 Rule S3699: The output of functions that don't return any…

andrea-guarino-sonarsource:
…thing should not be used


marco-bearzi-sonarsource (approved):
LGTM!

PR #1103, opened 2022-03-23
SONARPY-985 Show UI warning when errors occur in coverage report parsing

andrea-guarino-sonarsource:


sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonReportSensor.java
```diff
@@ -77,7 +77,9 @@ public void execute(SensorContext context) {
         // try absolute path
         File file = new File(reportPath);
         if (!file.exists()) {
-          LOG.warn("No report was found for {} using pattern {}", reportPathPropertyKey, reportPath);
+          String formattedMessage = String.format("No report was found for %s using pattern %s.", reportPathPropertyKey, reportPath);
```
marco-bearzi-sonarsource, line 80: From a consistency point of view, maybe it is better to either drop the '.' or make the other log messages consistent by adding a '.', but this is picky.

marco-bearzi-sonarsource (approved):
LGTM!

PR #1104, opened 2022-03-23
SONARPY-977 Rule S6323: Alternation in regular expressions should not…

andrea-guarino-sonarsource:
… contain empty alternatives


python-checks/src/test/resources/checks/regex/emptyAlternativeCheck.py
```diff
@@ -0,0 +1,33 @@
...
+    # Noncompliant@+3
+    re.match('''(
+    \'[^\']*(\'|$)|     # - a string that starts with a quote, up until the next quote or the end of the string
+    |                   # or
+    \S                  # - a non-whitespace character
+    )''', input, re.X)
+
+    re.match("|mandatory|-optional", input)              # Noncompliant
+    re.match("(mandatory|(|O|o|)ptional|)", input)       # Noncompliant
+    re.match("(|mandatory|optional)?", input)            # Noncompliant
```
marco-bearzi-sonarsource, line 21: Adding a check for the message would help make sure the correct quantifiers have been detected.

marco-bearzi-sonarsource (approved):
LGTM! Just a comment to check that the error message contains the correct quantifiers

python-checks/src/test/resources/checks/regex/emptyAlternativeCheck.py
andrea-guarino-sonarsource: After discussion, it looks like the message keeps the same as for other issues. However I added the assertion to make it more explicit :)

PR #1108, opened 2022-03-24
SONARPY-980 Rule S6353: Regular expression quantifiers and character classes should be used concisely

nils-werner-sonarsource:


andrea-guarino-sonarsource (approved):
LGTM!

PR #1109, opened 2022-03-24
SONARPY-983 Rule S6395: Non-capturing groups without quantifier should not be used

nils-werner-sonarsource:


its/ruling/src/test/resources/expected/python-S6395.json
```diff
@@ -0,0 +1,43 @@
+{
+'project:biopython/Bio/SearchIO/ExonerateIO/exonerate_text.py':[
+32,
+32,
+36,
+36,
```
andrea-guarino-sonarsource, line 6: To be honest, I'm not sure about those issue. I have the feeling that having those parenthesis make the regex more readable. Maybe we shouldn't raise an issue if the non-capturing group is part of an alternation?

Example:
```
(?:[<>]+ \w+ Intron \d+ [<>]+)|(?:\.+))
```

nils-werner-sonarsource: This problem affects all analyzers that use the Finder. This change must be made on the analyzer-commons side. For this reason, we will create a ticket for analyzer-commons and implement it in a timely manner.

andrea-guarino-sonarsource (approved):

PR #1112, opened 2022-03-25
SONARPY-981 Rule S6328: Replacement strings should reference existing regular expression groups

marco-bearzi-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/regex/GroupReplacementCheck.java
```diff
@@ -0,0 +1,103 @@
...
+    GroupFinder groupFinder = new GroupFinder();
+    groupFinder.visit(regexParseResult);
+    checkReplacement(regexFunctionCall, groupFinder.groups);
+  }
+
+  private void checkReplacement(CallExpression tree, Set<CapturingGroupTree> groups) {
+    RegularArgument regArg = nthArgumentOrKeyword(1, "replacement", tree.arguments());
+    if(regArg == null){
+      return;
+    }
```
andrea-guarino-sonarsource, line 68:
```suggestion
    if(regArg == null) return;
```
Very minor (my personal preference)

python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S6328.json
```diff
@@ -0,0 +1,17 @@
+{
+  "title": "Replacement strings should reference existing regular expression groups",
+  "type": "BUG",
```
andrea-guarino-sonarsource, line 3: Because this rule is a bug, we should also add it in the extended_ruling test . 
We can do that together!

python-checks/src/test/resources/checks/regex/groupReplacementCheck.py
```diff
@@ -0,0 +1,36 @@
+import re
```
andrea-guarino-sonarsource, line 1: I think there is an empty python file that has been commited by mistake (i.e. python-checks/src/test/resources/checks/regex/groupReplacementEmptyCheck.py)

python-checks/src/main/java/org/sonar/python/checks/CheckList.java
```diff
@@ -150,6 +151,7 @@ private CheckList() {
       FunctionReturnTypeCheck.class,
       FunctionUsingLoopVariableCheck.class,
       GenericExceptionRaisedCheck.class,
+      GroupReplacementCheck.class,
```
andrea-guarino-sonarsource, line 154: I think it should go after `GraphemeClustersInClassesCheck` to be alphabetically sorted

python-checks/src/main/java/org/sonar/python/checks/regex/GroupReplacementCheck.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  private static List<Integer> collectReferences(String replacement) {
+    Matcher match = REFERENCE_PATTERN.matcher(replacement);
+    List<Integer> references = new ArrayList<>();
+    while (match.find()) {
+      // extract reference number out of one of the possible 2 groups of the regex
+      for (int i = 1; i <= 2; i++) {
+        Optional.ofNullable(match.group(i)).map(Integer::valueOf).filter(ref -> ref != 0).ifPresent(references::add);
+      }
```
andrea-guarino-sonarsource, line 83: Given it's always two iterations, do we really need a for loop?

andrea-guarino-sonarsource (approved):
LGTM! Just left two minor comments!

PR #1123, opened 2022-04-29
SONARPY-1013 Rule S6252: Disabling versioning of S3 buckets is security-sensitive

marco-bearzi-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/cdk/S3BucketVersioningCheck.java
```diff
@@ -0,0 +1,81 @@
...
+    " versioning. Make sure it is safe here.";
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.CALL_EXPR, this::visitNode);
+  }
+
+  public void visitNode(SubscriptionContext ctx) {
+    CallExpression node = (CallExpression) ctx.syntaxNode();
+    Optional.ofNullable(node.calleeSymbol()).ifPresent(nodeSymbol -> {
```
nils-werner-sonarsource, line 48:
```suggestion
    Optional.ofNullable(node.calleeSymbol())
      .filter(nodeSymbol -> "aws_cdk.aws_s3.Bucket".equals(nodeSymbol.fullyQualifiedName()))
      .ifPresent(nodeSymbol -> {
```
You can use the `Optional::filter(Predicate)` method to check this condition if there is no else branch of it.

python-checks/src/main/java/org/sonar/python/checks/cdk/S3BucketVersioningCheck.java
```diff
@@ -0,0 +1,81 @@
...
+      if ("aws_cdk.aws_s3.Bucket".equals(nodeSymbol.fullyQualifiedName())) {
+        Optional<RegularArgument> version = getVersionArgument(node.arguments());
+        if (version.isPresent()) {
+          version.filter(a -> isExpressionFalse(a.expression())).ifPresent(v -> ctx.addIssue(v, MESSAGE));
+        } else {
+          ctx.addIssue(node.callee(), MESSAGE);
+        }
```
nils-werner-sonarsource, line 55: Let's discuss with the team if we move to Java 11 to make use of the `Optional::ifPresentOrElse` method. Nevertheless, this does not block the PR.

python-checks/src/test/resources/checks/cdk/s3BucketVersioning.py
```diff
@@ -0,0 +1,42 @@
+from aws_cdk import aws_s3 as s3
+
+a = int(10)
```
nils-werner-sonarsource, line 3: I would put this call to the end of the file and choose a better function name like `function_call_for_coverage` to show the intention of this call

python-checks/src/test/resources/checks/cdk/s3BucketVersioning.py
```diff
@@ -0,0 +1,42 @@
...
+third_versioning = True
+bucket = s3.Bucket(self, "MyUnversionedBucket",
+                   versioned=third_versioning # Compliant
+                   )
+
+bucket = s3.Bucket(self, "MyUnversionedBucket",
+                   versioned=True # Compliant
+                   )
+bucket = s3.Bucket(self, "MyUnversionedBucket",
+                   versioned=s3.Bucket(self, "InsideBucket") # Noncompliant
```
nils-werner-sonarsource, line 40: Is this a real test case? I could imagine this leads to an exception due to the `versioned` argument expects a boolean.
Maybe wrap this structure in something more realistic like `processBucket(self, bucket=s3.Bucket(...`

python-checks/src/test/resources/checks/cdk/s3BucketVersioning.py
```diff
@@ -0,0 +1,42 @@
...
+                                              # ^^^^^^^^^^^^^^^
+
+# Noncompliant@+1 {{Make sure using unversioned S3 bucket is safe here. Omitting 'versioned=True' disables S3 bucket versioning. Make sure it is safe here.}}
+bucket = s3.Bucket(self, "MyUnversionedBucket")
+#        ^^^^^^^^^
+
+bucket_versioning = False
+bucket = s3.Bucket(self, "MyUnversionedBucket",
+                   versioned=bucket_versioning # Noncompliant
+                   # highlight `versioned=bucket_versioning` as 2st location
```
nils-werner-sonarsource, line 15: You can remove this comment. It was used for implementation purposes only. In test cases, the highlighting is shown with the `^` comments.

python-checks/src/test/resources/checks/cdk/s3BucketVersioning.py
```diff
@@ -0,0 +1,42 @@
...
+a = int(10)
+
+bucket = s3.Bucket(self, "MyUnversionedBucket", versioned=False) # Noncompliant
+                                              # ^^^^^^^^^^^^^^^
+
+# Noncompliant@+1 {{Make sure using unversioned S3 bucket is safe here. Omitting 'versioned=True' disables S3 bucket versioning. Make sure it is safe here.}}
+bucket = s3.Bucket(self, "MyUnversionedBucket")
+#        ^^^^^^^^^
+
+bucket_versioning = False
```
nils-werner-sonarsource, line 12: We should add a secondary location for the assignment. So it's more straightforward for users why we raise an issue on line 14.

nils-werner-sonarsource (changes requested):
So far it looks good to me. Please address my suggestions and add a secondary location for single assigned values.

python-checks/src/main/java/org/sonar/python/checks/cdk/S3BucketVersioningCheck.java
```diff
@@ -0,0 +1,81 @@
...
+
+  public void visitNode(SubscriptionContext ctx) {
+    CallExpression node = (CallExpression) ctx.syntaxNode();
+    Optional.ofNullable(node.calleeSymbol()).ifPresent(nodeSymbol -> {
+      if ("aws_cdk.aws_s3.Bucket".equals(nodeSymbol.fullyQualifiedName())) {
+        Optional<RegularArgument> version = getVersionArgument(node.arguments());
+        if (version.isPresent()) {
+          version.filter(a -> isExpressionFalse(a.expression())).ifPresent(v -> ctx.addIssue(v, MESSAGE));
+        } else {
+          ctx.addIssue(node.callee(), MESSAGE);
```
nils-werner-sonarsource, line 54: We should also have a different message for omitting the argument. See updated RSPEC PR for more context

python-checks/src/main/java/org/sonar/python/checks/cdk/S3BucketVersioningCheck.java
```diff
@@ -0,0 +1,98 @@
...
+
+  private static Optional<RegularArgument> getVersionArgument(List<Argument> args) {
+    return args.stream()
+      .map(RegularArgument.class::cast)
+      .filter(a -> a.keywordArgument() != null)
+      .filter(a -> "versioned".equals(a.keywordArgument().name()))
+      .findAny();
+  }
+
+  private static List<Tree> treesOfFalseExpression(Expression expression) {
```
nils-werner-sonarsource, line 77:
```suggestion
  private static List<Tree> propagatedSenitiveExpressions(Expression expression) {
```

nils-werner-sonarsource (approved):
LGTM! See function name suggestion.

python-checks/src/main/java/org/sonar/python/checks/cdk/S3BucketVersioningCheck.java
marco-bearzi-sonarsource: Yes that's a better name, I've changed it

PR #1128, opened 2022-05-04
 SONARPY-1014 Rule S6245: Disabling server-side encryption of S3 bucket is security-sensitive

marco-bearzi-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/cdk/S3BucketServerEncryptionCheck.java
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  public void visitNode(SubscriptionContext ctx) {
+    CallExpression node = (CallExpression) ctx.syntaxNode();
+    Optional.ofNullable(node.calleeSymbol())
+      .filter(nodeSymbol -> "aws_cdk.aws_s3.Bucket".equals(nodeSymbol.fullyQualifiedName()))
+      .ifPresent(nodeSymbol -> {
+        List<RegularArgument> args = getEncryptionArguments(node.arguments());
+        Optional<RegularArgument> optEncryptionType = args.stream().filter(a -> "encryption".equals(getArgumentName(a))).findFirst();
+        Optional<RegularArgument> optEncryptionKey = args.stream().filter(a -> "encryption_key".equals(getArgumentName(a))).findFirst();
```
nils-werner-sonarsource, line 59: Instead of having one method which extracts both arguments and splits the arguments later on again, you can simply modify the method and parameterize the argument name.

python-checks/src/main/java/org/sonar/python/checks/cdk/S3BucketServerEncryptionCheck.java
```diff
@@ -0,0 +1,119 @@
...
+            });
+        } else {
+          if (!optEncryptionKey.isPresent()) {
+            ctx.addIssue(node.callee(), MESSAGE_OMITTING);
+          }
+        }
+      });
+  }
+
+  private static List<RegularArgument> getEncryptionArguments(List<Argument> args) {
```
nils-werner-sonarsource, line 75:
```suggestion
  private static RegularArgument getEncryptionArguments(List<Argument> args, String argName) {
```
This way you simplify also the logic of `visitNode`

python-checks/src/main/java/org/sonar/python/checks/cdk/S3BucketServerEncryptionCheck.java
```diff
@@ -0,0 +1,119 @@
...
+          optEncryptionType.map(regArg -> propagatedSensitiveExpressions(regArg.expression()))
+            .filter(trees -> !trees.isEmpty())
+            .ifPresent(trees -> {
+              PreciseIssue issue = ctx.addIssue(trees.get(0).parent(), MESSAGE);
+              trees.stream().skip(1).forEach(expression -> issue.secondary(expression.parent(), MESSAGE_SECONDARY));
+            });
+        } else {
+          if (!optEncryptionKey.isPresent()) {
+            ctx.addIssue(node.callee(), MESSAGE_OMITTING);
+          }
```
nils-werner-sonarsource, line 70: I would extract the `encryption_key` only if necessary. This means to call `getEncryptionArgument` only if `encryption` is missing.

python-checks/src/main/java/org/sonar/python/checks/cdk/S3BucketServerEncryptionCheck.java
```diff
@@ -0,0 +1,119 @@
...
+      trees.add(expression);
+    }
+    return trees;
+  }
+
+  private static boolean isUnencrypted(Expression expression) {
+    Symbol symbol = ((QualifiedExpression) expression).symbol();
+    return Optional.ofNullable(symbol)
+      .map(s -> "aws_cdk.aws_s3.BucketEncryption.UNENCRYPTED".equals(s.fullyQualifiedName()))
+      .orElse(false);
```
nils-werner-sonarsource, line 117:
```suggestion
    return Optional.ofNullable(symbol)
      .map(Symbol::fullyQualifiedName)
      .filter("aws_cdk.aws_s3.BucketEncryption.UNENCRYPTED"::equals)
      .isPresent();
```

python-checks/src/main/java/org/sonar/python/checks/cdk/S3BucketServerEncryptionCheck.java
```diff
@@ -0,0 +1,119 @@
...
+    } else {
+      trees.add(expression);
+    }
+    return trees;
+  }
+
+  private static boolean isUnencrypted(Expression expression) {
+    Symbol symbol = ((QualifiedExpression) expression).symbol();
+    return Optional.ofNullable(symbol)
+      .map(s -> "aws_cdk.aws_s3.BucketEncryption.UNENCRYPTED".equals(s.fullyQualifiedName()))
```
nils-werner-sonarsource, line 116: I would locate the FQN in a constant

python-checks/src/test/resources/checks/cdk/s3BucketServerEncryption.py
```diff
@@ -0,0 +1,49 @@
...
+    type_unencrypted = s3.BucketEncryption.UNENCRYPTED
+#   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^> {{Propagated setting.}}
+    bucket = s3.Bucket(self,"MyEncryptedBucket",
+                       encryption=type_unencrypted,     # NonCompliant {{Objects in the bucket are not encrypted. Make sure it is safe here.}}
+#                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+                       encryption_key=my_encryption_key2)
+
+bucket = s3.Bucket(self, "MyEncryptedBucket", encryption="UNENCRYPTED")  # NonCompliant
+
+coverage = s3.Bucket(self, "bucket", public_read_acces=False, enforce_ssl=None) # NonCompliant
```
nils-werner-sonarsource, line 48: What happens if `enforce_ssl` is True?

python-checks/src/main/java/org/sonar/python/checks/cdk/S3BucketServerEncryptionCheck.java
```diff
@@ -0,0 +1,65 @@
...
+  public static final String MESSAGE = "Objects in the bucket are not encrypted. Make sure it is safe here.";
+  public static final String OMITTING_MESSAGE = "Omitting 'encryption' disables server-side encryption. Make sure it is safe here.";
+
+  @Override
+  void visitBucketConstructor(SubscriptionContext ctx, CallExpression bucket) {
+    Optional<ArgumentTrace> optEncryptionType = getArgument(ctx, bucket, "encryption");
+    if (optEncryptionType.isPresent()) {
+      optEncryptionType.ifPresent(argumentTrace -> {
+        List<Expression> list = argumentTrace.trace();
+        argumentTrace.addIssueIf(isUnencrypted(list.get(list.size() - 1)), MESSAGE);
```
nils-werner-sonarsource, line 46: You don't have to pick the last item on the list. The predicate which is tested inside `addIssueIf` will actually try to get any match. Out of the types which can be part of the trace list, only the last one can match this expression. The list only contains expressions of type NAME which are a reference to another expression or which is a value expression. The predicate can only match the value expression. Tipp: compare this to `isFalse`

python-checks/src/test/resources/checks/cdk/s3BucketServerEncryption.py
```diff
@@ -0,0 +1,47 @@
...
+def create_bucket():
+    type_unencrypted = s3.BucketEncryption.UNENCRYPTED
+#   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^> {{Propagated setting.}}
+    bucket = s3.Bucket(self,"MyEncryptedBucket",
+                       encryption=type_unencrypted,     # NonCompliant {{Objects in the bucket are not encrypted. Make sure it is safe here.}}
+#                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+                       encryption_key=my_encryption_key2)
+
+
+coverage = s3.Bucket(self, "bucket", encryption=b)  # Compliant
```
nils-werner-sonarsource, line 47:
```suggestion
coverage = s3.Bucket(self, "bucket", encryption=unresolved_reference)  # Compliant
```

nils-werner-sonarsource (approved):
LGTM! Small improvement suggestion to reduce the complexity.

python-checks/src/main/java/org/sonar/python/checks/cdk/S3BucketServerEncryptionCheck.java
marco-bearzi-sonarsource: I was thinking that it may slightly reduce the overhead if there is a long trace, but in practice there are probably not too long traces. I'll remove this modification

PR #1129, opened 2022-05-04
SONARPY-1011 Rule S6265: Granting access to S3 buckets to all or authenticated users is security-sensitive

marco-bearzi-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/cdk/S3BucketGrantedAccessCheck.java
```diff
@@ -0,0 +1,95 @@
...
+      .ifPresent(argument -> argument.addIssueIf(this::isNotPrivate, MESSAGE));
+  }
+
+  protected boolean isNotPrivate(Expression expression) {
+    return Optional.ofNullable(expression)
+      .filter(QualifiedExpression.class::isInstance)
+      .map(QualifiedExpression.class::cast)
+      .map(QualifiedExpression::symbol)
+      .map(Symbol::fullyQualifiedName)
+      .map(s -> !S3_BUCKET_PRIVATE_ACCESS_POLICY.equals(s))
```
nils-werner-sonarsource, line 91: You should flip the logic to raise an issue if the argument is set to the public access policy. Otherwise, you create noise if the value is something like `s3.bar.WHAT_EVER`.

python-checks/src/main/java/org/sonar/python/checks/cdk/S3BucketGrantedAccessCheck.java
```diff
@@ -0,0 +1,95 @@
...
+      .orElse(false);
```
nils-werner-sonarsource, line 92:
```suggestion
      .filter(s -> !S3_BUCKET_PRIVATE_ACCESS_POLICY.equals(s))
      .isPresent();
```
Regardless of the logic problem: `.map(...).orElse(false)` should always be `.filter(...).isPresent()`

nils-werner-sonarsource (approved):
LGTM. You have to adjust the message and you have to flip the logic to not raise FPs.

PR #1136, opened 2022-05-25
BUILD-1577 migrate to new infra

julien-carsique-sonarsource:
- BUILD-1551 update repox credentials
- BUILD-1577 migrate to new infra


drautureau-sonarsource (changes requested):
We have an issue with the new image. This is why I am blocking this PR

PR #1137, opened 2022-05-25
SONARPY-1034 : Add verifier support for testing quick-fixes

marco-bearzi-sonarsource:
[JIRA ticket](https://jira.sonarsource.com/browse/SONARPY-1034)

### DIFF
--- a/python-checks-testkit/src/main/java/org/sonar/python/checks/utils/PythonQuickFixVerifier.java
+++ b/python-checks-testkit/src/main/java/org/sonar/python/checks/utils/PythonQuickFixVerifier.java
@@ -1,0 +1,105 @@
+/*
+ * SonarQube Python Plugin
+ * Copyright (C) 2011-2022 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.python.checks.utils;
+
+import com.sonar.sslr.api.AstNode;
+import java.net.URI;
+import java.util.Collections;
+import java.util.List;
+import org.sonar.plugins.python.api.IssueLocation;
+import org.sonar.plugins.python.api.PythonCheck;
+import org.sonar.plugins.python.api.PythonCheck.PreciseIssue;
+import org.sonar.plugins.python.api.PythonFile;
+import org.sonar.plugins.python.api.PythonSubscriptionCheck;
+import org.sonar.plugins.python.api.PythonVisitorContext;
+import org.sonar.plugins.python.api.tree.FileInput;
+import org.sonar.python.SubscriptionVisitor;
+import org.sonar.python.parser.PythonParser;
+import org.sonar.python.quickfix.IssueWithQuickFix;
+import org.sonar.python.semantic.ProjectLevelSymbolTable;
+import org.sonar.python.tree.PythonTreeMaker;
+
+public class PythonQuickFixVerifier {
+  private PythonQuickFixVerifier() {
+  }
+
+  private static List<PreciseIssue> scanFileForIssues(PythonCheck check, PythonVisitorContext context) {
+    check.scanFile(context);
+    if (check instanceof PythonSubscriptionCheck) {
+      SubscriptionVisitor.analyze(Collections.singletonList((PythonSubscriptionCheck) check), context);
+    }
+    return context.getIssues();
+  }
+
+  public static List<PreciseIssue> getIssuesWithQuickFix(String codeWithIssue, PythonCheck check) {
+    PythonParser parser = PythonParser.create();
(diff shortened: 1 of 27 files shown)


python-checks-testkit/src/main/java/org/sonar/python/checks/utils/PythonQuickFixVerifier.java
```diff
@@ -0,0 +1,105 @@
...
+
+    return scanFileForIssues(check, visitorContext);
+  }
+
+  public static String applyQuickFix(String codeWithIssue, IssueWithQuickFix issueWithQuickFix) {
+    IssueLocation loc = issueWithQuickFix.getQuickFixes().get(0).getTextEdits().get(0).issueLocation;
+    String replacement = loc.message();
+    int start = convertPositionToIndex(codeWithIssue, loc.startLine(), loc.startLineOffset());
+    int end = convertPositionToIndex(codeWithIssue, loc.endLine(), loc.endLineOffset());
+    return codeWithIssue.substring(0, start + 1) + replacement + codeWithIssue.substring(end + 1);
```
pynicolas, line 69: I would prefer the `+ 1` to be handled by `convertPositionToIndex` because I don't like it to return an index such as `-1`.

python-checks/src/test/java/org/sonar/python/checks/ClassMethodFirstArgumentNameCheckTest.java
```diff
@@ -19,13 +19,39 @@
...
+    List<PythonCheck.PreciseIssue> issues = PythonQuickFixVerifier
+      .getIssuesWithQuickFix(codeWithIssue, new ClassMethodFirstArgumentNameCheck());
+
+    assertThat(issues).hasSize(1);
+    IssueWithQuickFix issue = (IssueWithQuickFix) issues.get(0);
+
+    assertThat(issue.getQuickFixes()).hasSize(1);
+
+    String codeQFApplied = PythonQuickFixVerifier.applyQuickFix(codeWithIssue, issue);
+    assertThat(codeQFApplied).isEqualTo(codeFixed);
```
pynicolas, line 55: I would expect all of that to be handled in a single method call to PythonQuickFixVerifier, for example:
```
PythonQuickFixVerifier.verify(new ClassMethodFirstArgumentNameCheck(), codeWithIssue, codeFixed);
```

python-checks/src/test/java/org/sonar/python/checks/ClassMethodFirstArgumentNameCheckTest.java
```diff
@@ -19,13 +19,39 @@
...
 public class ClassMethodFirstArgumentNameCheckTest {
 
   @Test
   public void test() {
     PythonCheckVerifier.verify("src/test/resources/checks/classMethodFirstArgumentNameCheck.py", new ClassMethodFirstArgumentNameCheck());
+
+    String codeWithIssue = "class A():\n\n" +
+      "    @classmethod\n" +
+      "    def area(bob, length, width):\n" +
+      "        return length*width\n";
```
pynicolas, line 40: I think the code example could be simplified: fewer parameters, trivial body (`pass`)...

python-checks/src/test/java/org/sonar/python/checks/quickfix/PythonQuickFixVerifier.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  private static int convertPositionToIndex(String fileContent, int line, int lineOffset) {
+    int currentLine = 1;
+    int currentIndex = -1;
+    while (currentLine < line) {
+      currentIndex = fileContent.indexOf("\n", currentIndex + 1);
+      currentLine++;
+    }
+    return currentIndex + lineOffset + 1;
```
pynicolas, line 99: To me, it would be more understandable like that:
```suggestion
    int currentLine = 1;
    int currentIndex = 0;
    while (currentLine < line) {
      currentIndex = fileContent.indexOf("\n", currentIndex) + 1;
      currentLine++;
    }
    return currentIndex + lineOffset;
```
However, I'm not sure that it would be correct...

python-checks/src/test/java/org/sonar/python/checks/quickfix/PythonQuickFixVerifier.java
```diff
@@ -0,0 +1,125 @@
...
+
+    PythonVisitorContext visitorContext = new PythonVisitorContext(parse,
+      pythonFile, null, "",
+      ProjectLevelSymbolTable.empty());
+
+    return scanFileForIssues(check, visitorContext);
+  }
+
+  public static String applyQuickFix(String codeWithIssue, IssueWithQuickFix issueWithQuickFix) {
+    IssueLocation loc = issueWithQuickFix.getQuickFixes().get(0).getTextEdits().get(0).issueLocation;
```
pynicolas, line 85: I would prefer to have the assertion about the number of quick fixes just above this line so that we can be sure that `get(0)` cannot fail.

python-checks/src/test/java/org/sonar/python/checks/quickfix/PythonQuickFixVerifierTest.java
```diff
@@ -0,0 +1,73 @@
...
+import org.sonar.plugins.python.api.PythonSubscriptionCheck;
+import org.sonar.python.checks.ClassMethodFirstArgumentNameCheck;
+import org.sonar.python.quickfix.IssueWithQuickFix;
+import org.sonar.python.quickfix.PythonQuickFix;
+import org.sonar.python.quickfix.PythonTextEdit;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+
+public class PythonQuickFixVerifierTest {
```
pynicolas, line 35: I would prefer to test only the `verify` method (I don't think the other methods need to be public), create a mock implementation of `PythonCheck` and test cases such as:
* No issue is raised (we expect an `AssertionError`)
* More than 1 issue is raised
* Exactly 1 issue is raised, but it doesn't have a quickfix
* Exactly 1 issue with 1 quickfix but applying it doesn't give the expected result
* Verification is successful

pynicolas:
@marco-bearzi-sonarsource Please rebase the PR against master.

PR #1141, opened 2022-06-01
BUILD-1577 fix win build

julien-carsique-sonarsource:
win task wait for linux build tasks to get a BUILD number (see cirrus-env)


julien-carsique-sonarsource:
Also unset `SONARSOURCE_QA` to skip its (circumvent BUILD-1625)

PR #1144, opened 2022-06-07
DBD-200 Add Dataflow Bug Detection rules to the default quality profile

guillaume-dequenne-sonarsource:


andrea-guarino-sonarsource (approved):
LGTM, for the merge commit message I would use something more explicit like we did on Java, e.g. "Add Dataflow Bug Detection rules to the default quality profile"

PR #1147, opened 2022-06-09
SONARPY-1032: Add quick fix for S3923 (AllBranchesAreIdenticalCheck)

marco-bearzi-sonarsource:


nils-werner-sonarsource (changes requested):
Conditionals are not handled correctly and elif branches are not handled at all.

nils-werner-sonarsource (approved):
When the `conditional` case is solved it's ready to be merged.

PR #1152, opened 2022-06-14
SONARPY-1041: Fix the quick-fix for S1940 (BooleanCheckNotInvertedCheck)

marco-bearzi-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/BooleanCheckNotInvertedCheck.java
```diff
@@ -120,8 +122,36 @@ private static void createQuickFix(IssueWithQuickFix issue, String oppositeOpera
...
+    if (tree.children().isEmpty()) {
+      if (tree.is(Kind.EXPRESSION_LIST)) {
+        return "";
+      }
+      return ((Token) tree).value();
+    }
+    // Regular argument and tuple may have children
+    if (tree.children().size() == 1 && !tree.children().get(0).is(Kind.REGULAR_ARGUMENT, Kind.TUPLE, Kind.PARENTHESIZED)) {
+      return tree.firstToken().value();
+    }
```
andrea-guarino-sonarsource, line 148: If I remove those lines, tests are still green: is this condition really needed?

python-checks/src/main/java/org/sonar/python/checks/BooleanCheckNotInvertedCheck.java
```diff
@@ -120,8 +122,36 @@ private static void createQuickFix(IssueWithQuickFix issue, String oppositeOpera
...
+    return tree.children().stream()
+      .map(BooleanCheckNotInvertedCheck::getText)
+      .reduce("", (acc, b) -> isSpaceNotNeeded(acc, b) ? (acc + b) : (acc + " " + b));
```
andrea-guarino-sonarsource, line 151: why `b`? I would maybe name it `currentChar` or `toAdd`

python-checks/src/main/java/org/sonar/python/checks/BooleanCheckNotInvertedCheck.java
```diff
@@ -120,8 +122,36 @@ private static void createQuickFix(IssueWithQuickFix issue, String oppositeOpera
...
+  }
+
+  private static boolean isSpaceNotNeeded(String acc, String toAdd) {
+    return acc.isBlank() || acc.endsWith("(") || acc.endsWith("[")  || "]".equals(toAdd) || "(".equals(toAdd) || ")".equals(toAdd) || ",".equals(toAdd);
```
andrea-guarino-sonarsource, line 155: how did you determine this? I think a comment would be necessary. 
Wouldn't be better to decide when a space is needed iterating on tokens, instead that iterating over characters?

andrea-guarino-sonarsource (changes requested):
Please have a look at my suggestions / clarification questions

python-checks/src/main/java/org/sonar/python/checks/BooleanCheckNotInvertedCheck.java
```diff
@@ -120,8 +122,36 @@ private static void createQuickFix(IssueWithQuickFix issue, String oppositeOpera
...
+    return getText(toUse.leftOperand()) + " " + oppositeOperator + " " + getText(toUse.rightOperand());
+  }
+
+  private static String getText(Tree tree) {
+    if (tree.children().isEmpty()) {
+      if (tree.is(Kind.EXPRESSION_LIST)) {
+        return "";
+      }
+      return ((Token) tree).value();
+    }
```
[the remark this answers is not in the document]
marco-bearzi-sonarsource: You're right! And your solution also takes care of kinds of trees which are not token, other than expression_list

python-checks/src/main/java/org/sonar/python/checks/BooleanCheckNotInvertedCheck.java
marco-bearzi-sonarsource: Indeed, it was a remnant of another logic I used: I was returning the value of the tree directly if there was only one child. But if we only return the value when there is no child, as you suggested, those lines are useless

python-checks/src/main/java/org/sonar/python/checks/BooleanCheckNotInvertedCheck.java
marco-bearzi-sonarsource: Yep, it would be clearer

python-checks/src/main/java/org/sonar/python/checks/BooleanCheckNotInvertedCheck.java
marco-bearzi-sonarsource: Yes, I'll write a comment about this. For the time being, it is easier to work directly on the text/characters as we completely rewrite the boolean expression. But indeed, in the future we may want to use tokens

python-checks/src/test/java/org/sonar/python/checks/BooleanCheckNotInvertedCheckTest.java
```diff
@@ -54,30 +57,135 @@ public void operatorStringTest() {
...
+    codeWithIssue = "a = not (() in c)";
+    codeFixed = "a = () not in c";
+    PythonQuickFixVerifier.verify(check, codeWithIssue, codeFixed);
+
+    codeWithIssue = "a = not ((1,) in c)";
+    codeFixed = "a = (1,) not in c";
+    PythonQuickFixVerifier.verify(check, codeWithIssue, codeFixed);
+
+    codeWithIssue = "a = not ((1,2,3,4) in c)";
+    codeFixed = "a = (1, 2, 3, 4) not in c";
```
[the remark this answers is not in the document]
marco-bearzi-sonarsource: Yes, we've decided to modify the formatting to follow PEP-8 conventions, if possible and if not too complex for each case.
For this check, it has been decided to rewrite the above lines so that we don't add spacing after the commas

python-checks/src/main/java/org/sonar/python/checks/BooleanCheckNotInvertedCheck.java
```diff
@@ -120,8 +122,33 @@ private static void createQuickFix(IssueWithQuickFix issue, String oppositeOpera
...
+  }
+
+  private static String getText(Tree tree) {
+    if (tree.is(Kind.TOKEN)) {
+      return ((Token) tree).value();
+    }
+
+    return tree.children().stream()
+      .map(BooleanCheckNotInvertedCheck::getText)
+      .reduce("", (acc, currentChar) -> isSpaceNotNeeded(acc, currentChar) ? (acc + currentChar) : (acc + " " + currentChar));
```
andrea-guarino-sonarsource, line 145: I think there is still a problem with "f strings" like the following:
```java
    codeWithIssue = "x = not (a == f'foo${b}')";
    codeFixed = "x = a != f'foo${b}'";
    PythonQuickFixVerifier.verify(check, codeWithIssue, codeFixed);
```

I would suggest to use `TreeUtils.tokens(tree)` which handles already f-strings:
```suggestion
    return TreeUtils.tokens(tree).stream()
      .map(Token::value)
      .reduce("", (acc, currentChar) -> isSpaceNotNeeded(acc, currentChar) ? (acc + currentChar) : (acc + " " + currentChar));
```

andrea-guarino-sonarsource (changes requested):
I left a few other comments: it seems there is a problem with f-strings

python-checks/src/main/java/org/sonar/python/checks/BooleanCheckNotInvertedCheck.java
```diff
@@ -120,8 +122,33 @@ private static void createQuickFix(IssueWithQuickFix issue, String oppositeOpera
   }
 
   private static PythonTextEdit getReplaceEdit(BinaryExpression toUse, Expression toReplace, String oppositeOperator) {
-    List<Tree> children = toReplace.parent().parent().children();
+    // To be able to remove the 'not' sign and the possible space afterwards
+    Tree notAncestor = Optional.ofNullable(TreeUtils.firstAncestorOfKind(toReplace, Kind.NOT))
+      .orElseThrow(() -> new IllegalArgumentException("A NOT tree must be found. It was not the case."));
```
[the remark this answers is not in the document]
marco-bearzi-sonarsource: Good point!

python-checks/src/main/java/org/sonar/python/checks/BooleanCheckNotInvertedCheck.java
marco-bearzi-sonarsource: Yes, I didn't cover this. It should be the case now

andrea-guarino-sonarsource (approved):
LGTM, can you please reorganize the commits as follows before merging?
1) Change version to 3.15.1
2) SONARPY-1041: Fix the quick-fix for S1940 (containing all the rest of the changes)

PR #1159, opened 2022-08-03
SONARPY-760 Rule S5905: Assert should not be called on a tuple literal

rudy-regazzoni-sonarsource:


guillaume-dequenne-sonarsource (approved):
LGTM.

PR #1165, opened 2022-08-05
 SONARPY-763 Rule S1607: A reason should be provided when skipping a test

rudy-regazzoni-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/tests/SkippedTestNoReasonCheck.java
```diff
@@ -0,0 +1,119 @@
...
+@Rule(key = "S1607")
+public class SkippedTestNoReasonCheck extends PythonSubscriptionCheck {
+
+  private static final String MESSAGE = "Provide a reason for skipping this test.";
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.DECORATOR, ctx -> {
+      Decorator decorator = (Decorator) ctx.syntaxNode();
+
```
guillaume-dequenne-sonarsource, line 45: _Nitpick:_  There are many extra empty lines which don't look necessary given the code for the check is quite readable. They could probably be removed.

python-checks/src/main/java/org/sonar/python/checks/tests/SkippedTestNoReasonCheck.java
```diff
@@ -0,0 +1,119 @@
...
+    context.registerSyntaxNodeConsumer(Tree.Kind.CALL_EXPR, ctx -> {
+      CallExpression callExpression = (CallExpression) ctx.syntaxNode();
+
+      checkCallExpressionSkipWithNoOrEmptyReason(ctx, callExpression, "pytest.skip");
+    });
+  }
+
+  private static void checkDecoratorSkipWithoutReason(SubscriptionContext ctx, Decorator decorator, String decoratorSkipName) {
+    Expression expression = decorator.expression();
+    Symbol symbol = getSymbolFromExpression(expression);
```
guillaume-dequenne-sonarsource, line 59: As of now, it seems we wouldn't do anything if the decorator is a `NAME` (because `getSymbolFromExpression` doesn't handle that case).
I'd suggest to get rid of the `getSymbolFromExpression` method and to replace it here with something like:
```suggestion
    Symbol symbol = null;
    if (expression instanceof HasSymbol) {
      symbol = ((HasSymbol) expression).symbol();
    }
```
That way you can handle any decorator expression which has a symbol with the FQN you seek, without having to care about the underlying tree.

python-checks/src/test/resources/checks/tests/skippedTestNoReasonWithImport.py
```diff
@@ -0,0 +1,8 @@
+from pytest.mark import skip
+
+# Pytest
+class UsingPytest():
+    @skip()  # Noncompliant {{Provide a reason for skipping this test.}}
+#   ^^^^^^^
+    def test_pytest_mark_skip_no_reason():
```
guillaume-dequenne-sonarsource, line 7: You could add a test for which the decorator is a name (`@skip`). Currently, it wouldn't be reported due to my other comment, but it probably should.

python-checks/src/test/resources/checks/tests/skippedTestNoReason.py
```diff
@@ -0,0 +1,72 @@
...
+    @"my string"
+    def test_decorator_without_symbol_ok():
+        assert 1 == 2
+
+    @pytest.mark.skip()  # Noncompliant {{Provide a reason for skipping this test.}}
+#   ^^^^^^^^^^^^^^^^^^^
+    def test_skip_without_args_ko():
+        assert 1 == 2
+
+    @pytest.mark.skip("")  # Noncompliant {{Provide a reason for skipping this test.}}
```
guillaume-dequenne-sonarsource, line 49: For the sake of completeness, I think we could have a test for which the parameter is a string literal (something like `@pytest.mark.skip("a valid reason")`) as we currently don't have one.

python-checks/src/main/java/org/sonar/python/checks/tests/SkippedTestNoReasonCheck.java
```diff
@@ -0,0 +1,119 @@
...
+
+  private static final String MESSAGE = "Provide a reason for skipping this test.";
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.DECORATOR, ctx -> {
+      Decorator decorator = (Decorator) ctx.syntaxNode();
+
+      checkDecoratorSkipWithoutReason(ctx, decorator, "unittest.case.skip");
+      checkDecoratorSkipWithoutReason(ctx, decorator, "pytest.mark.skip");
```
guillaume-dequenne-sonarsource, line 47: To simplify things a bit, I'd suggest putting the fully qualified names of the skip decorators in a static set for the rule, only call the `checkDecoratorSkipWithoutReason` method once, and when checking the FQN, just check that it belongs to said set before checking the arguments.
That way, you avoid executing the logic to retrieve the symbol twice as it's not needed.

its/ruling/src/test/resources/expected/python-S1607.json
```diff
@@ -0,0 +1,26 @@
+{
+  'project:mypy-0.782/mypy/test/data.py':[
+    217,
+  ],
+  'project:mypy-0.782/mypy/test/testfinegrained.py':[
+    77,
+  ],
+  'project:mypy-0.782/mypy/test/testipc.py':[
+    59,
```
guillaume-dequenne-sonarsource, line 9: This case (as well as a few others from ruling) puzzles me a bit.
The developer is not providing a reason in the arguments for the call, but there is still a reason provided through comments above the decorator.
We may want to avoid reporting issues for this case (to be verified, as an explicit skip through pytest/unittest call might provide some additional info when running all the test suite).

Personally, I think it's fine if we report an issue in such case, but I'd probably add a test case in our UTs to document the behavior.

python-checks/src/main/java/org/sonar/python/checks/tests/SkippedTestNoReasonCheck.java
```diff
@@ -0,0 +1,119 @@
...
+    });
+
+    context.registerSyntaxNodeConsumer(Tree.Kind.CALL_EXPR, ctx -> {
+      CallExpression callExpression = (CallExpression) ctx.syntaxNode();
+
+      checkCallExpressionSkipWithNoOrEmptyReason(ctx, callExpression, "pytest.skip");
```
guillaume-dequenne-sonarsource, line 53: Related to the above comment, for consistency I'd probably put the `pytest.skip` in some kind of singleton set to have a similar implementation as `checkDecoratorSkipWithoutReason` if you end up changing that one.

guillaume-dequenne-sonarsource (changes requested):
Looking good. I left a few suggestions to make the rule a bit more robust and to have some additional test cases.

python-checks/src/main/java/org/sonar/python/checks/tests/SkippedTestNoReasonCheck.java
rudy-regazzoni-sonarsource: I tend to abuse of bit of empty lines to somewhat aerate my code, indeed I agree it's not necessary in this case.

python-checks/src/main/java/org/sonar/python/checks/tests/SkippedTestNoReasonCheck.java
rudy-regazzoni-sonarsource: That's a good idea, will implement this.

python-checks/src/main/java/org/sonar/python/checks/tests/SkippedTestNoReasonCheck.java
rudy-regazzoni-sonarsource: Same as above, will implement this.

python-checks/src/main/java/org/sonar/python/checks/tests/SkippedTestNoReasonCheck.java
rudy-regazzoni-sonarsource: This doesn't seems to handle the case of `CallExpression`, which does not implement `HasSymbol` interface. For those, we do not call the method `symbol()` but `calleeSymbol()`.
I still implemented your suggestion in the method `getSymbolFromExpression()` instead of the `QualifiedExpression` check, since as you mentioned it allow to not take care of the tree for this one.

python-checks/src/test/resources/checks/tests/skippedTestNoReason.py
rudy-regazzoni-sonarsource: Correct, will add this !

python-checks/src/test/resources/checks/tests/skippedTestNoReasonWithImport.py
rudy-regazzoni-sonarsource: I added this test case.
If for "my other comment" you refer to the HasSymbol part, then since I slightly adapted your proposition, this test case does report properly an error, which seems fine.

its/ruling/src/test/resources/expected/python-S1607.json
rudy-regazzoni-sonarsource: That's a good point, I added the test case with some comment.
I personally agree with you, for me we should indeed raise an issue as I believe the reason should really be in the method/decorator rather than in the comment, but probably both points have valid arguments.

python-checks/src/main/java/org/sonar/python/checks/tests/SkippedTestNoReasonCheck.java
```diff
@@ -38,19 +41,23 @@ public class SkippedTestNoReasonCheck extends PythonSubscriptionCheck {
...
+  private static final Set<String> skipCallExpressionsFQN = new HashSet<>(List.of("pytest.skip"));
+
   @Override
   public void initialize(Context context) {
     context.registerSyntaxNodeConsumer(Tree.Kind.DECORATOR, ctx -> {
       Decorator decorator = (Decorator) ctx.syntaxNode();
-
-      checkDecoratorSkipWithoutReason(ctx, decorator, "unittest.case.skip");
-      checkDecoratorSkipWithoutReason(ctx, decorator, "pytest.mark.skip");
+      for (String skipDecoratorFQN : skipDecoratorsFQN) {
```
guillaume-dequenne-sonarsource, line 51: This is not what I had in mind: I think the `checkDecoratorSkipWithoutReason` should be called only once.

When you're comparing the actual decorator name with the one you provide as argument (on line 72), just try to see if you can find the name in the set instead. That way you wouldn't have to perform the same operation for each case (granted, there are only 2 and the code doesn't have to do much, so there is not really a problem performance-wise, but I think it's still a bit cleaner).

python-checks/src/main/java/org/sonar/python/checks/tests/SkippedTestNoReasonCheck.java
```diff
@@ -38,19 +41,23 @@ public class SkippedTestNoReasonCheck extends PythonSubscriptionCheck {
...
+        checkDecoratorSkipWithoutReason(ctx, decorator, skipDecoratorFQN);
+      }
     });
 
     context.registerSyntaxNodeConsumer(Tree.Kind.CALL_EXPR, ctx -> {
       CallExpression callExpression = (CallExpression) ctx.syntaxNode();
-
-      checkCallExpressionSkipWithNoOrEmptyReason(ctx, callExpression, "pytest.skip");
+      for (String skipCallExpressionFQN : skipCallExpressionsFQN) {
```
guillaume-dequenne-sonarsource, line 58: Same here, we shouldn't have to iterate over those names (even if there is only one 😄 ).

guillaume-dequenne-sonarsource (changes requested):
Looking good, just one last tiny change requested.

python-checks/src/main/java/org/sonar/python/checks/tests/SkippedTestNoReasonCheck.java
rudy-regazzoni-sonarsource: I see ! That's indeed better, that's also what I implemented for SONARPY-766.
This has far more sense, will implement this.

guillaume-dequenne-sonarsource (approved):
LGTM.

PR #1170, opened 2022-08-10
 SONARPY-766 Rule S5918: Tests should be skipped explicitly 

rudy-regazzoni-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/tests/TestShouldBeSkippedExplicitlyCheck.java
```diff
@@ -0,0 +1,128 @@
...
+      }
+      super.visitQualifiedExpression(pyQualifiedExpressionTree);
+    }
+
+    private static String tryGetName(Expression expression) {
+      if (expression.is(Tree.Kind.NAME)) {
+        return ((Name) expression).name();
+      }
+      return "";
+    }
```
marco-bearzi-sonarsource, line 125: In order not to return "", you could use `Optional.of()` over `expression` and then filter it over the name and check whether it is in the set with `isPresent()`. But that's merely a suggestion

python-checks/src/test/resources/checks/tests/testShouldBeSkippedExplicitly.py
```diff
@@ -0,0 +1,162 @@
...
+#     ^^^^^^
+
+    directory = self._prepare_directory(num_classes=3, count=5)
+
+    with self.assertRaisesRegex(ValueError, '`labels` argument should be'):
+      _ = image_dataset.image_dataset_from_directory(
+          directory, labels=None)
+
+
+return
```
marco-bearzi-sonarsource, line 162: Is it necessary here ?

marco-bearzi-sonarsource (changes requested):
Good PR ! See the few comments to address

python-checks/src/main/java/org/sonar/python/checks/tests/TestShouldBeSkippedExplicitlyCheck.java
```diff
@@ -0,0 +1,128 @@
...
+    "assertNotIsInstance", "assertRaises", "assertRaisesRegexp", "assertRaisesRegex", "assertAlmostEqual", "assertNotAlmostEqual",
+    "assertGreater", "assertGreaterEqual", "assertLess", "assertLessEqual", "assertRegexpMatches", "assertNotRegexpMatches",
+    "assertItemsEqual", "assertDictContainsSubset"));
+  private static final Tree.Kind[] literalsKind = {Tree.Kind.STRING_LITERAL, Tree.Kind.NUMERIC_LITERAL, Tree.Kind.LIST_LITERAL,
+    Tree.Kind.BOOLEAN_LITERAL_PATTERN, Tree.Kind.NUMERIC_LITERAL_PATTERN, Tree.Kind.NONE_LITERAL_PATTERN, Tree.Kind.STRING_LITERAL_PATTERN,
+    Tree.Kind.SET_LITERAL, Tree.Kind.DICTIONARY_LITERAL};
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.FUNCDEF, ctx -> {
```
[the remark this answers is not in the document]
rudy-regazzoni-sonarsource: Very good point ! I believe we discussed this with @guillaume-dequenne-sonarsource 
What do you think ? Should I go forward and add this check ?

python-checks/src/main/java/org/sonar/python/checks/tests/TestShouldBeSkippedExplicitlyCheck.java
rudy-regazzoni-sonarsource: I like the approach, will implement this.

python-checks/src/test/resources/checks/tests/testShouldBeSkippedExplicitly.py
rudy-regazzoni-sonarsource: I believe this is not required anymore.
This was used to cover a specific usecase in the first version of my code, which was using a very different approach : subscribe over RETURN statement instead of FUNCTION DEF.
Having a return out of a function definition was throwing an error.
In any case, I should at least add a comment here to explain why we have this, but I will check and probably remove it as with the new approach this doesn't cover any test cases.

marco-bearzi-sonarsource (approved):
LGTM! Excepted the little change about the `Optional` part, it looks OK for me. Maybe @guillaume-dequenne-sonarsource can also have a look ?

python-checks/src/main/java/org/sonar/python/checks/tests/TestShouldBeSkippedExplicitlyCheck.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @Override
+    public void visitAssertStatement(AssertStatement pyAssertStatementTree) {
+      hasANativeAssert = true;
+      super.visitAssertStatement(pyAssertStatementTree);
+    }
+
+    @Override
+    public void visitQualifiedExpression(QualifiedExpression pyQualifiedExpressionTree) {
+      Optional<String> name = tryGetName(pyQualifiedExpressionTree.qualifier());
```
guillaume-dequenne-sonarsource, line 114: This could probably be made even simpler by checking if `pyQualifiedExpressionTree.qualifier().is(Tree.Kind.NAME)` and then just casting it and accessing its `name()` directly. That way, you don't need a different method or optional, etc (this is a very common pattern we use).
Maybe irrelevant if you end up using the callee symbol instead.

guillaume-dequenne-sonarsource (changes requested):
Looking good. There are a few edge cases that may need addressing, and I left a couple of nitpick comments, but still good overall!

python-checks/src/main/java/org/sonar/python/checks/tests/TestShouldBeSkippedExplicitlyCheck.java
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: I think this makes sense when checking `unittest` cases, because you always have to import unittest (and be in a class that inherits from `unittest.TestCase`) for a test to be discoverable (from the top of my head, specifics to be verified).
For `pytest`, I believe that you don't need to import it. When `pytest` will run, it will discover anything that starts with `test` as a test, without necessarily being imported. So I'd definitely not check this one.

For `unittest`, I'd be fine not checking it either (though it may be more correct to do so), but in that case I'd probably document the potential FP (calling `self.assertX` in a class that is not a `unittest.TestCase` for example).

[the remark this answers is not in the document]
rudy-regazzoni-sonarsource: I ended up adding the check of unittest presence.
It took more time than expected, because :
- sometimes unittest framework is wrapped around another framework, so the FQN is not necessary "unittest.case.TestCase" but can be something like "twisted.trial.unittest.TestCase". After a discussion with @andrea-guarino-sonarsource, we switched to a different approach : ensure that the FQN both contains "unittest" and "TestCase"
- they can be several level of superclass, so we have to add in the scope parent classes recursively

It is interesting to note that we get far fewer issue from PythonRulingTest : this is because the project tensoerflow have apparently they own test classes definitions (although it look a lot like unittest).

python-checks/src/main/java/org/sonar/python/checks/tests/TestShouldBeSkippedExplicitlyCheck.java
```diff
@@ -0,0 +1,128 @@
...
+      if (!functionDef.name().name().startsWith("test")) {
+        return;
+      }
+
+      ReturnStatement returnStatement = getReturnStatementInIfFirstStatement(functionDef);
+      if (returnStatement == null) {
+        return;
+      }
+
+      // check there is an assert like statement in the function
```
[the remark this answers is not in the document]
rudy-regazzoni-sonarsource: Haha yes that's correct !
As stated during our earlier conversation, this was probably a comment a accidentally left, because in my programming process is sometimes write few comment about my intent before replacing them with real code.
As you said, the method name itself is explicit and so this comment just add some noise, i removed it.
I replaced it with a comment to explain the use case instead.

python-checks/src/main/java/org/sonar/python/checks/tests/TestShouldBeSkippedExplicitlyCheck.java
```diff
@@ -0,0 +1,129 @@
...
+@Rule(key = "S5918")
+public class TestShouldBeSkippedExplicitlyCheck extends PythonSubscriptionCheck {
+
+  private static final String MESSAGE = "Skip this test explicitly.";
+  private static final Set<String> supportedAssertMethods = new HashSet<>(Arrays.asList("assertEqual", "assertNotEqual", "assertTrue",
+    "assertFalse", "assertIs", "assertIsNot", "assertIsNone", "assertIsNotNone", "assertIn", "assertNotIn", "assertIsInstance",
+    "assertNotIsInstance", "assertRaises", "assertRaisesRegexp", "assertRaisesRegex", "assertAlmostEqual", "assertNotAlmostEqual",
+    "assertGreater", "assertGreaterEqual", "assertLess", "assertLessEqual", "assertRegexpMatches", "assertNotRegexpMatches",
+    "assertItemsEqual", "assertDictContainsSubset"));
+  private static final Tree.Kind[] literalsKind = {Tree.Kind.STRING_LITERAL, Tree.Kind.NUMERIC_LITERAL, Tree.Kind.LIST_LITERAL,
```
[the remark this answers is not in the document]
rudy-regazzoni-sonarsource: I removed both SET and DICTIONARY from the literals list in order to avoid those potential FP, even if as you said it's probably unlikely to happen ;)

python-checks/src/main/java/org/sonar/python/checks/tests/TestShouldBeSkippedExplicitlyCheck.java
```diff
@@ -0,0 +1,129 @@
...
+    boolean hasACallToAssert = false;
+
+    @Override
+    public void visitAssertStatement(AssertStatement pyAssertStatementTree) {
+      hasANativeAssert = true;
+      super.visitAssertStatement(pyAssertStatementTree);
+    }
+
+    @Override
+    public void visitQualifiedExpression(QualifiedExpression pyQualifiedExpressionTree) {
```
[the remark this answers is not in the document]
rudy-regazzoni-sonarsource: Thanks for checking, keeping it as this in that case !

python-checks/src/main/java/org/sonar/python/checks/tests/TestShouldBeSkippedExplicitlyCheck.java
rudy-regazzoni-sonarsource: Thanks for confirming, I was afraid that it would be a bit to bulky to put directly in the IF condition.
I adapted.

python-checks/src/main/java/org/sonar/python/checks/tests/ImplicitlySkippedTestCheck.java
```diff
@@ -0,0 +1,187 @@
...
+      }
+
+      return null;
+    }
+    return null;
+  }
+
+  private static boolean containsAssertion(FunctionDef functionDef) {
+    List<String> parentClasses = getParentClassFromFunctionDef(functionDef);
+    AssertionVisitor assertVisitor = new AssertionVisitor(parentClasses);
```
guillaume-dequenne-sonarsource, line 125: _Minor suggestion:_ I feel we can simplify this a bit by making the assumption that only one test framework is used at once (and that we probably won't add support for many test frameworks in the near future).

We could check before invoking the visitor that the parent classes contain `unittest` and `TestCase`. And if yes, we can provide the list of assertion methods to the visitor (if we are not in a `unittest.TestCase` or derivative, then we provide an empty list).

I'm okay with keeping the current implementation though, this is just a way I think this could be simplified.

python-checks/src/test/resources/checks/tests/implicitlySkippedTestWithWrapper.py
```diff
@@ -0,0 +1,24 @@
+import os
```
guillaume-dequenne-sonarsource, line 1: It looks like this directly comes from cases found on ruling.

In general, it is better to try and reduce those cases a bit to have more minimal test cases.
Some of the imports could probably be removed (for example `runtime` don't feel strictly necessary for the test - any random `cond()` could replace the use of `runtime.platformType` for instance).

Despite `twisted.trial.unittest` surely being a thing, I guess we don't actually resolve this and we could replace it by some random and obviously fake import just to highlight that what we do is look for `unittest` and `TestCase` in the name without doing a full resolution.

We could probably also remove the mixin, as I guess it doesn't impact the result either (also multiple inheritance is already tested somewhere else).

python-checks/src/test/resources/checks/tests/implicitlySkippedTest.py
```diff
@@ -0,0 +1,237 @@
...
+        self.patch(utils, 'rmdirRecursive', fail)
+        self.make_command(fs.RemoveDirectory, dict(
+            dir='workdir',
+        ), True)
+        d = self.run_command()
+
+        def check(_):
+            self.assertIn({'rc': -1}, self.get_updates(), self.builder.show())
+        d.addCallback(check)
+        return d
```
guillaume-dequenne-sonarsource, line 203: I guess this part could be simplified with just a simple assertion (this seems copied from actual code, which removes a bit of clarity from the test).
Perhaps if you want a test where the assertion is actually called in a nested function, this could be done in a separate test.

guillaume-dequenne-sonarsource (changes requested):
This is looking good, though I still have a few comments.
About unit tests, although it's a good idea to take actual code to get some inspiration and verifying real-world behavior, it's important to ensure we can reduce the real-world code into minimalist examples, to make sure we minimize noise and ensure clarity about what is being tested.

python-checks/src/main/java/org/sonar/python/checks/tests/ImplicitlySkippedTestCheck.java
rudy-regazzoni-sonarsource: I adapted a bit, I kept the part where we don't test specifically for unittest framework as I don't like to have this framework hardcoded in the logic, but I modified the part you mentioned in order to simplify and only pass the set of assert methods to the visitor.

python-checks/src/test/resources/checks/tests/implicitlySkippedTest.py
rudy-regazzoni-sonarsource: Good point, i cleaned the test files

python-checks/src/main/java/org/sonar/python/checks/tests/ImplicitlySkippedTestCheck.java
```diff
@@ -121,20 +124,20 @@ private static ReturnStatement getReturnStatementFromFirstIfStatement(FunctionDe
...
-  private static List<String> getParentClassFromFunctionDef(FunctionDef functionDef) {
+  private static Optional<TestFramework> getParentClassTestFrameworkFromFunctionDef(FunctionDef functionDef) {
     List<String> libraries = new ArrayList<>();
     ClassDef classDef = (ClassDef) TreeUtils.firstAncestorOfKind(functionDef, Tree.Kind.CLASSDEF);
     if (classDef != null) {
       libraries.addAll(getInheritedClassesFQN(classDef));
     }
 
-    return libraries;
+    return testFrameworks.stream().filter(testFramework -> testFramework.matchAnyProvidedClasses(libraries)).findFirst();
```
guillaume-dequenne-sonarsource, line 140: I'd suggest replacing this with:

```
    return testFrameworks.stream()
      .filter(testFramework -> testFramework.matchAnyProvidedClasses(libraries))
      .findFirst()
      .map(t -> t.supportedAssertMethods)
      .orElseGet(Collections::emptySet);
```

and make the method return a set of assertion methods instead of an optional of test framework.
That way, you can get rid of the `emptyFramework` you had to define.

guillaume-dequenne-sonarsource (approved):
LGTM.

PR #1172, opened 2022-08-11
SONARPY-758 Rule S5899: Test methods should be discoverable

marco-bearzi-sonarsource:
Link to [RSPEC PR](https://github.com/SonarSource/rspec/pull/1165)


python-checks/src/main/java/org/sonar/python/checks/tests/NotDiscoverableTestMethodCheck.java
```diff
@@ -0,0 +1,130 @@
...
+    .collect(Collectors.toList());
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.CLASSDEF, ctx -> {
+      ClassDef classDefinition = (ClassDef) ctx.syntaxNode();
+
+      if (inheritsOnlyFromUnitTest(classDefinition)) {
+        HashMap<String, List<Tree>> functionOccurrences = new HashMap<>();
+        List<FunctionDef> allDefinitions = new ArrayList<>();
```
guillaume-dequenne-sonarsource, line 65: I suggest to define these as:

```
Map<FunctionSymbol, FunctionDef> suspiciousFunctionsAndDefinitions = new HashMap<>();
Set<Tree> allDefinitions = new HashSet<>();
```

python-checks/src/main/java/org/sonar/python/checks/tests/NotDiscoverableTestMethodCheck.java
```diff
@@ -0,0 +1,130 @@
...
+        // Assuming there is no definition of function in another one
+        for (Statement statement : classDefinition.body().statements()) {
+          if (statement.is(Tree.Kind.FUNCDEF)) {
+            FunctionDef functionDef = ((FunctionDef) statement);
+            String nameFunction = functionDef.name().name();
+            // If it doesn't override existing methods and doesn't start with test, we add it to the map
+            if (!overrideExistingMethod(nameFunction) && !nameFunction.startsWith("test")) {
+              addValueToMap(functionOccurrences, functionDef.name(), nameFunction);
```
guillaume-dequenne-sonarsource, line 73: I'd suggest to do something like this instead:
```
Symbol symbol = functionDef.name().symbol();
// If it doesn't override existing methods and doesn't start with test, we add it to the map
if (!overrideExistingMethod(nameFunction) && !nameFunction.startsWith("test") && symbol != null && symbol.is(Symbol.Kind.FUNCTION)) {
    suspiciousFunctionsAndDefinitions.put(((FunctionSymbol) symbol), functionDef);
}
```
(Maybe the checks on the symbol can be done separately)

python-checks/src/main/java/org/sonar/python/checks/tests/NotDiscoverableTestMethodCheck.java
```diff
@@ -0,0 +1,130 @@
...
+        // Add tree everytime a token contains the name of function (call + reference functions are then taken care of)
+        allDefinitions.forEach(def -> retrieveReferenceInBody(def, functionOccurrences));
+
+        // More than 1 occurrence of the method means it is used inside the class -> do not raise
+        // Only raise when one (non-test) method is not used
+        functionOccurrences.entrySet()
+          .stream()
+          .filter((nameListEntry -> nameListEntry.getValue().size() == 1))
+          .forEach(nameListEntry -> ctx.addIssue(nameListEntry.getValue().get(0), MESSAGE));
+      }
```
guillaume-dequenne-sonarsource, line 88: I think this part is a bit complex. What you're doing is going through each method definition's body to try to find usages of the suspicious methods, to check if they are actually helpers.

You can do this more simply by using the semantic API of sonar-python. In particular, when you have a `FunctionDef#name`, its associated symbol should most likely be a `FunctionSymbol` (if it's not the case, it may be because the symbol is defined multiple times, or other edge cases like that, for which we don't want to report issues anyway). 

When you have a `Symbol`, you can access its usages through the `Symbol#usages` API. By default, there is always at least one usage (the definition of the symbol). If you have more, you can check that the associated tree has indeed one of the method definitions as its parent, by doing something like:


```
suspiciousFunctionsAndDefinitions.forEach((s, d) -> {
  List<Usage> usages = s.usages();
  if (usages.size() == 1 || usages.stream().noneMatch(u -> TreeUtils.firstAncestor(u.tree(), allDefinitions::contains) != null)) {
    ctx.addIssue(d.name(), MESSAGE);
 }
});
```

Then you can remove `retrieveReferenceInBody` and `addValueToMap` which are now unused.

guillaume-dequenne-sonarsource (changes requested):
Looks good!
I suggest a few small changes to make the implementation a bit simpler by using the semantic API.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1173, opened 2022-08-17
SONARPY-758 Set S5899 scope to ALL

guillaume-dequenne-sonarsource:


nils-werner-sonarsource (approved):
LGTM! Do we want to have a default check to validate the default scope for all the other rules?

guillaume-dequenne-sonarsource:
> LGTM! Do we want to have a default check to validate the default scope for all the other rules?

It's done on the `PythonCheckTest` for the default implementation. I believe it's okay to only have additional tests when a check overrides the default.

PR #1175, opened 2022-08-24
 SONARPY-761 Rule S5845: Assertions of dissimilar types should not be made

rudy-regazzoni-sonarsource:
As discussed, two commits :
- the original code, with reference to the BUILTINS_TYPE_CATEGORY map directly from SillyEquality class
- the refactoring where BUILTINS_TYPE_CATEGORY is moved into InferredTypes utils class, also moved the method to use it + added a method to retrieve the unmodifiable version of the Map (used by ConfusingTypeCheckingCheck class to do a custom check)


python-checks/src/main/java/org/sonar/python/checks/tests/AssertOnDissimilarTypesCheck.java
```diff
@@ -0,0 +1,166 @@
...
+      return Optional.empty();
+    }
+
+    Symbol symbol = ((Name) expr).symbol();
+    if (symbol == null) {
+      return Optional.empty();
+    }
+
+    Usage lastAssignment = symbol.usages().stream()
+      .takeWhile(usage -> usage != ((Name) expr).usage())
```
guillaume-dequenne-sonarsource, line 128: I'm not sure I understand the `takeWhile`. Why not have the condition `usage != ((Name) expr).usage()` as part of the filter as well?

guillaume-dequenne-sonarsource (changes requested):
Looking good! I left a few comments. There is one potential for FP when keyword arguments are used (sorry about that, I completely forgot about it...), and we might want to have a helper method to check whether we are in a `unittest.TestCase` since many rules could want to check that (let's discuss this, as I'm implementing something similar in a rule as well). 
I shared other small observations but nothing major.

python-checks/src/main/java/org/sonar/python/checks/tests/AssertOnDissimilarTypesCheck.java
rudy-regazzoni-sonarsource: As discussed takeWhile is different from filter in the sense that it stop at the first evaluation which return false instead of filtering it out and pursuing.
I added a test case where a variable is assigned after the self.assert statement, in this case we ensure that the last assignment before the self.assert statement is put in secondary and not the one after.

rudy-regazzoni-sonarsource: So again, as discussed adding the use case clearly showed that this approach was not working : usages() does not list usage by order.
I added a sorted() based on the line number of the token associated to the usage, seems to do the trick.

python-checks/src/main/java/org/sonar/python/checks/tests/AssertOnDissimilarTypesCheck.java
```diff
@@ -0,0 +1,166 @@
...
+      .filter(usage -> usage.kind() == Usage.Kind.ASSIGNMENT_LHS)
+      .reduce((first, second) -> second)
+      .orElse(null);
+
+    return Optional.ofNullable(lastAssignment)
+      .map(assignment -> (AssignmentStatement) TreeUtils.firstAncestorOfKind(assignment.tree(), Tree.Kind.ASSIGNMENT_STMT));
+  }
+
+  private static boolean checkArgumentsCantBeIdentical(Expression left, Expression right) {
```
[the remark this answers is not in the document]
rudy-regazzoni-sonarsource: I got the same thought, I noticed that we are calling this method all the time with inverting the boolean result, so I was tempted to invert the method name/result directly, but then I noticed its "brother" method checkArgumentsCantBeEqual does not have the same case, so it would have mean to either break consistency between them or to choose one which result will be always inverted.

From my perspective, I would put both of them as "CanBeEqual" instead of their current behavior/name "CantBeEqual", to avoid encountering the double negation "!CantBeEqual" which seems less readable for me. I replaced to this way.

Agree with your last paragraph, I switched to "canArgumentsBeXXX" since "checkXXX" method is usually for method raising issues.

rudy-regazzoni-sonarsource:
To be reworked once we have the UnittestUtils ;)

python-checks/src/main/java/org/sonar/python/checks/tests/AssertOnDissimilarTypesCheck.java
```diff
@@ -81,18 +84,20 @@ private static void checkArguments(SubscriptionContext ctx, CallExpression callE
...
+    Expression right = secondArg.expression();
+    if (canArgumentsBeIdentical(left, right)) {
       return;
     } else if (isAnAssertIdentity) {
       shouldRaise = true;
     }
-    if (isAnAssertEquality && checkArgumentsCantBeEqual(left, right)) {
+    if (isAnAssertEquality && !canArgumentsBeEqual(left, right)) {
       shouldRaise = true;
     }
```
guillaume-dequenne-sonarsource, line 102: I just realize we can further simplify this by using a single:
```
if (isAnAssertIdentity || !canArgumentsBeEqual(left, right)) {
    shouldRaise = true;
}
```
Since at this point, if we're in an identity assertion, we know we must raise anyway, and if we're not, that means we are in an assert equality.

python-checks/src/test/resources/checks/tests/assertOnDissimilarTypes.py
```diff
@@ -22,16 +22,20 @@ def __ne__(self, other):
...
+    aInd = A()
+#   ^^^^^^^^^^> {{Last assignment of "aInd"}}
+    bInd = B()
+#   ^^^^^^^^^^> {{Last assignment of "bInd"}}
+
+    self.assertIs(aInd, bInd) # Noncompliant {{Change this assertion to not compare dissimilar types (A and B).}}
+#                 ^^^^^^^^^^
+    aInd = A()
+    self.assertEqual(first="string", second=True) # Noncompliant
+    self.assertIs(msg="my message", first="myString", second=a) # Noncompliant
```
guillaume-dequenne-sonarsource, line 38: For the sake of completeness, I'd probably add a compliant test where the arguments are in an unexpected order where we don't raise an FP because we correctly understand the keyword arguments.

guillaume-dequenne-sonarsource (commented):
I left 2 minors comments that you may choose to address when rebasing with the `UnittestUtils` changes. It's really minor though.

guillaume-dequenne-sonarsource (approved):
LGTM!

PR #1176, opened 2022-08-24
 SONARPY-762 Rule S5906: The most specific assertion should be preferred

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/tests/DedicatedAssertionCheck.java
```diff
@@ -0,0 +1,239 @@
...
+  private static final String IS_INSTANCE = "isinstance";
+
+  private static final Map<String, String> ASSERT_TRUE_SUGGESTIONS = Map.of(
+    "==", ASSERT_EQUAL,
+    "!=", ASSERT_NOT_EQUAL,
+    ">", "assertGreater",
+    ">=", "assertGreaterEqual",
+    "<", "assertLess",
+    "<=", "assertLessEqual",
+    IN, "assertIn",
```
rudy-regazzoni-sonarsource, line 72: Nitpick, sound a bit strange to have sometimes constant defined values and sometimes a String literal.
I believe it is because the constant defined values are used later in the switch ?
Not sure if my remark is relevant, just about uniformity.

python-checks/src/main/java/org/sonar/python/checks/tests/DedicatedAssertionCheck.java
```diff
@@ -0,0 +1,239 @@
...
+    if (dedicatedAssertion != null) {
+      ctx.addIssue(callExpression, String.format(MESSAGE, dedicatedAssertion));
+    }
+  }
+
+  private static void checkAssertEqualOrNotEqual(CallExpression callExpression, SubscriptionContext ctx, boolean isAssertEqual) {
+    List<Argument> arguments = callExpression.arguments();
+    if (arguments.size() < 2) {
+      return;
+    }
```
rudy-regazzoni-sonarsource, line 169: This check does not seems to be required since you are testing just after that both firstArg and secondArg are not null.

python-checks/src/main/java/org/sonar/python/checks/tests/DedicatedAssertionCheck.java
```diff
@@ -0,0 +1,239 @@
...
+    if (secondDedicatedAssertion != null) {
+      ctx.addIssue(callExpression, String.format(MESSAGE, secondDedicatedAssertion));
+    }
+  }
+
+  private static void checkAssertAlmostEqualOrNot(CallExpression callExpression, SubscriptionContext ctx, String methodName) {
+    List<Argument> arguments = callExpression.arguments();
+    if (arguments.size() < 2) {
+      return;
+    }
```
rudy-regazzoni-sonarsource, line 194: Same as above, maybe this check can be removed as this use case is handled by the next check.

python-checks/src/main/java/org/sonar/python/checks/tests/DedicatedAssertionCheck.java
rudy-regazzoni-sonarsource: That's a good question, indeed for operator it feel more readable to keep them as it is.
Also I'm afraid that if we replace everything by constant, the class may start with 20+ lines of such definitions, which may not be desirable.
We can either let as it is, replace only method name or replace everything, I let you see what is the more readable.

guillaume-dequenne-sonarsource: Then I think I'll let it as it is. It's not ideal but no the solution is, really 😄.

python-checks/src/main/java/org/sonar/python/checks/tests/DedicatedAssertionCheck.java
guillaume-dequenne-sonarsource: Good point! Thanks!

rudy-regazzoni-sonarsource (approved):
LGTM !

PR #1178, opened 2022-08-25
Introduce utils class for unittest checks

guillaume-dequenne-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM !

python-frontend/src/main/java/org/sonar/python/tests/UnittestUtils.java
```diff
@@ -0,0 +1,75 @@
...
+
+  private UnittestUtils() {
+
+  }
+
+  // All methods of unittest https://docs.python.org/3/library/unittest.html#unittest.TestCase
+  public static final Set<String> RUN_METHODS = Set.of("setUp", "tearDown", "setUpClass", "tearDownClass", "run", "skiptTest",
+    "subTest", "debug");
+
+  public static final Set<String> ASSERTIONS_METHODS = Set.of("assertEqual",
```
rudy-regazzoni-sonarsource, line 41: I see that you merged both Python 2 and Python 3 methods, maybe we should also add the line to Python 2 methods and/or add a comment to mention this ?

rudy-regazzoni-sonarsource (commented):
LGTM, just a small comment about Python 2/3 unittest methods.

PR #1180, opened 2022-08-29
 SONARPY-765 Rule S5915: Assertions should not be made at the end of blocks expecting an exception

rudy-regazzoni-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/tests/AssertAfterRaiseCheck.java
```diff
@@ -0,0 +1,101 @@
...
+import org.sonar.plugins.python.api.tree.Tree;
+import org.sonar.plugins.python.api.tree.WithStatement;
+import org.sonar.python.tests.PytestUtils;
+import org.sonar.python.tests.UnittestUtils;
+
+@Rule(key = "S5915")
+public class AssertAfterRaiseCheck extends PythonSubscriptionCheck {
+  private static final String MESSAGE_MULTIPLE_STATEMENT = "Don’t perform an assertion here; An exception is expected to be raised before its execution.";
+  private static final String MESSAGE_SINGLE_STATEMENT = "Refactor this test; if this assertion’s argument raises an exception, the assertion will never get executed.";
+  private static final String MESSAGE_SECONDARY = null;
```
guillaume-dequenne-sonarsource, line 41: I'm not sure it's a good practice to have secondary locations with null message. If none has been defined in the RSPEC, maybe we can create one.
What about `An exception is expected to be raised in this block.` ?

python-frontend/src/main/java/org/sonar/python/tests/PytestUtils.java
```diff
@@ -0,0 +1,42 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.python.tests;
+
+import java.util.Set;
+import org.sonar.plugins.python.api.symbols.Symbol;
+import org.sonar.plugins.python.api.tree.QualifiedExpression;
+import org.sonar.python.tree.TreeUtils;
+
+public class PytestUtils {
```
guillaume-dequenne-sonarsource, line 27: I tend to think this PytestUtils is not really needed, given that pytest is generally much more simple than unittest. We can probably check for the patterns/fully qualified names in the rules themselves.

python-checks/src/main/java/org/sonar/python/checks/tests/AssertAfterRaiseCheck.java
```diff
@@ -0,0 +1,101 @@
...
+    return withStatement.withItems().stream()
+      .filter(withItem -> withItem.test().is(Tree.Kind.CALL_EXPR))
+      .map(withItem -> ((CallExpression) withItem.test()).callee())
+      .filter(callee -> callee.is(Tree.Kind.QUALIFIED_EXPR))
+      .map(QualifiedExpression.class::cast)
+      .anyMatch(callee -> isPytestRaise(callee) || isUnittestRaise(callee));
+  }
+
+  public boolean isPytestRaise(QualifiedExpression callee) {
+    return PytestUtils.isPytest(callee) && PytestUtils.RAISE_METHODS.contains(callee.name().name());
```
guillaume-dequenne-sonarsource, line 70: I think just checking that the callee has `pytest.raises` as fully qualified name is enough. We can also just test that `pytest.raises` is in the FQN without being strictly equal, to account for potential wrappers.
In any case, we probably don't expect a lot of false positive here: this pattern seems to be a mistake for any framework.

python-checks/src/main/java/org/sonar/python/checks/tests/AssertAfterRaiseCheck.java
```diff
@@ -0,0 +1,101 @@
...
+          .secondary(IssueLocation.preciseLocation(withStatement.firstToken(), withStatement.colon(), MESSAGE_SECONDARY));
+      }
+    });
+  }
+
+  public boolean isWithStatementItemARaise(WithStatement withStatement) {
+    return withStatement.withItems().stream()
+      .filter(withItem -> withItem.test().is(Tree.Kind.CALL_EXPR))
+      .map(withItem -> ((CallExpression) withItem.test()).callee())
+      .filter(callee -> callee.is(Tree.Kind.QUALIFIED_EXPR))
```
guillaume-dequenne-sonarsource, line 64: You probably don't need to filter by qualified expression, one could do:
```
from pytest import raises
```
Or even:
```
from pytest import raises as my_raise
```
And then the callee would not be a qualified expression. Though it's probably quite uncommon...

its/ruling/src/test/resources/expected/python-S5915.json
```diff
@@ -3,19 +3,4 @@
     641,
     1482,
   ],
-  'project:tensorflow/python/data/experimental/kernel_tests/take_while_test.py':[
```
guillaume-dequenne-sonarsource: We probably wouldn't mind raising on these, even if we can't be sure it is `unittest`. Even in another framework, the pattern is probably a mistake.

guillaume-dequenne-sonarsource (changes requested):
Looking good, I left a few comments.

python-checks/src/main/java/org/sonar/python/checks/tests/AssertAfterRaiseCheck.java
rudy-regazzoni-sonarsource: That's what I was thinking, I saw that all methods about adding secondary had specifically "Nullable" but I was not sure that was a good Idea.
Indeed there is nothing in RSPEC, the message you provided sounds good to me !

python-frontend/src/main/java/org/sonar/python/tests/PytestUtils.java
rudy-regazzoni-sonarsource: I agree with you, especially with the fact that I already have a dedicated method `isPytestRaise` for this.
I removed this class.

its/ruling/src/test/resources/expected/python-S5915.json
rudy-regazzoni-sonarsource: I removed the call to `UnittestUtils.isWithinUnittestTestCase` in order to have those usecase raised again.
Indeed check on the name itself seems quite enough to highlight that this is very probably a mistake, the risk of FP seems unlikely to me. Not sure if we should do the same elsewhere in the end, but this is maybe defeating the purpose of the `isWithinUnittestTestCase` itself ^^'

Not sure how the tensorflow project is working : it seems to have it's own unittest framework implementation with specific name, but methods and usage are the same.

guillaume-dequenne-sonarsource (approved):
LGTM.

PR #1184, opened 2022-08-31
SONARPY-797 FP on S4426: mismatch between key length and key exponent

nils-werner-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM !

PR #1187, opened 2022-09-01
SONARPY-1106 Fix wrong issue message on S5906 for NOT IN statement

rudy-regazzoni-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/tests/DedicatedAssertionCheck.java
```diff
@@ -55,30 +56,33 @@ public class DedicatedAssertionCheck extends PythonSubscriptionCheck {
...
+    Map.entry(">", "assertGreater"),
+    Map.entry(">=", "assertGreaterEqual"),
+    Map.entry("<", "assertLess"),
+    Map.entry("<=", "assertLessEqual"),
+    Map.entry(IN, "assertIn"),
+    Map.entry(NOT_IN, "assertNotIn"),
+    Map.entry(IS, "assertIs"),
+    Map.entry(IS_NOT, "assertIsNot"),
+    Map.entry(IS_INSTANCE, "assertIsInstance")
   );
```
nils-werner-sonarsource, line 79: Why did you replace `Map.of` with `Map.ofEntries`? IMO it reduces the readability of the code.

nils-werner-sonarsource (approved):
LGTM! Please see comment

python-checks/src/main/java/org/sonar/python/checks/tests/DedicatedAssertionCheck.java
rudy-regazzoni-sonarsource: I fully agree with you, but we don't have the choice :D

I was thinking that `Map.of` was using a **varargs** style parameter which would allow an unlimited number of arguments, but that's not the case.
Instead, this methods is overridden from 0 to 10 different map values, meaning that if you want more that 10 values in your `Map`, you cannot use it and must switch to an alternative.

image

In that case we are unlucky : with the **NOT_IN** map value we add, we now have 11 values.
I searched for alternatives and apparently the only one available is the usage of `Map.ofEntries`, which is using **varargs**.

nils-werner-sonarsource: Ah, you are right. I totally forgot that this method is limited. So everything is fine.

PR #1188, opened 2022-09-01
SONARPY-1021 Fix FP on S1721 when using the walrus operator

rudy-regazzoni-sonarsource:


nils-werner-sonarsource (approved):
LGTM! I'm not happy with `isParenthesisWithoutAssignmentInside` but I don't have an alternative,

PR #1190, opened 2022-09-01
SONARPY-876 S5795 (IdentityComparisonWithCachedTypesCheck) shouldn't raise on 'is not None'

nils-werner-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/ComparisonToNoneCheck.java
```diff
@@ -67,10 +69,6 @@ private static void addIssue(SubscriptionContext ctx, Tree tree, String comparis
     ctx.addIssue(tree, String.format("Remove this %s; it will always be %s.", comparisonKind, resultAsString));
   }
 
-  private static boolean isNone(InferredType type) {
-    return type.canOnlyBe(BuiltinTypes.NONE_TYPE);
-  }
-
   private static boolean cannotBeNone(InferredType type) {
     return !type.canBeOrExtend(BuiltinTypes.NONE_TYPE);
   }
```
rudy-regazzoni-sonarsource, line 74: I see you have moved the method `isNone()` to `CheckUtils`, which sounds right to me.
Maybe we should also move the method `cannotBeNone()` the same way, WDYT ?

rudy-regazzoni-sonarsource (approved):
LGTM !
Just a quick remark about potentially moving another method to CheckUtils.

python-checks/src/main/java/org/sonar/python/checks/ComparisonToNoneCheck.java
nils-werner-sonarsource: I don't like to move methods to util classes that are only used in a single check. There are only a few cases where this is applicable. Eg a Utils class is only for a specific type of helpers like a dedicated InferredType utils class.

PR #1192, opened 2022-09-01
SONARPY-1037 S1481 should not report on dummy variables

rudy-regazzoni-sonarsource:


nils-werner-sonarsource (approved):
LGTM! You should create a second unit test to test a custom pattern.

PR #1193, opened 2022-09-01
SONARPY-1086 Fix FP on S1764 for left bitwise shift

rudy-regazzoni-sonarsource:


rudy-regazzoni-sonarsource:
Not totally sure if I'm correct, from my perspective we should not check on shift operators, so I removed them from the list of Kinds to check + the method that was dedicated to this operator.

python-checks/src/test/resources/checks/identicalExpressionOnBinaryOperator.py
```diff
@@ -23,7 +23,8 @@ def work():
 k = 5 - 5 # Noncompliant
 l = 5 + 5
 m = 5 * 5
-n = 3 << 3 # Noncompliant
+n = 3 << 3
+n = 3 >> 3
```
nils-werner-sonarsource, line 27: IMO this is still predictable. We shouldn't raise on variables like `a << a`

nils-werner-sonarsource (changes requested):
We should not ignore predictable operations only variable-based ones.

python-checks/src/test/resources/checks/identicalExpressionOnBinaryOperator.py
rudy-regazzoni-sonarsource: Well at first I was thinking the same, but then I started to dig in the code/unit test and apparently, from my understanding, the current implementation of this rule does not aim to detect such usecase.
It does not care about operand being variables or literal, it is checking if they are equivalent.

Some example of non compliant code as per current implementation :
```
j = 5 / 5 # Noncompliant
k = 5 - 5 # Noncompliant
o = 3 & 3 # Noncompliant
p = 3 ^ 3 # Noncompliant
```

And some example of compliant code :
```
l = 5 + 5
m = 5 * 5
```

For me those last examples are also predictable, but it's more about having always the same result independently from the left and right operand identical values.
x - x will always give 0, while x + x is not predictable (2x)

After thinking about it, x << x is also not predictable (x * 2**x), but x >> x is (always give 0).

So instead of skipping both << and >> operators, we should only skip << and still check for >>.

WDYT ?

nils-werner-sonarsource: Greate investigation. Yeah, let's only skip <<.

nils-werner-sonarsource (approved):
LGTM!

PR #1194, opened 2022-09-02
SONARPY-990 S1481 should not raise when variable is assigned inside decorator

nils-werner-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM !

PR #1195, opened 2022-09-02
SONARPY-1107 Adapt S5918 message when returning a call to a skip test method

rudy-regazzoni-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/tests/ImplicitlySkippedTestCheck.java
```diff
@@ -92,6 +101,30 @@ private static ReturnStatement getReturnStatementFromFirstIfStatement(FunctionDe
...
+    return expressions.stream()
+      .filter(expression -> expression.is(Tree.Kind.CALL_EXPR))
+      .map(CallExpression.class::cast)
+      .anyMatch(callExpr -> isPytestSkip(callExpr) || isUnittestSkip(callExpr));
+  }
+
+  private static boolean isPytestSkip(CallExpression callExpression) {
+    return Optional.of(callExpression).stream()
+      .filter(callExpr -> callExpr.callee().is(Tree.Kind.QUALIFIED_EXPR))
+      .map(callExpr -> (QualifiedExpression) callExpr.callee())
```
nils-werner-sonarsource, line 114: This part is identical in `isPytestSkip` and `isUnittestSkip`. Maybe you can move it to `hasAnySkipStatement`.

python-checks/src/main/java/org/sonar/python/checks/tests/ImplicitlySkippedTestCheck.java
```diff
@@ -92,6 +101,30 @@ private static ReturnStatement getReturnStatementFromFirstIfStatement(FunctionDe
...
+      .filter(Objects::nonNull)
+      .anyMatch(symbol -> "pytest.skip".equals(symbol.fullyQualifiedName()));
+  }
+
+  private static boolean isUnittestSkip(CallExpression callExpression) {
+    return Optional.of(callExpression).stream()
+      .filter(callExpr -> callExpr.callee().is(Tree.Kind.QUALIFIED_EXPR))
+      .map(callExpr -> (QualifiedExpression) callExpr.callee())
+      .anyMatch(qualifExpr -> qualifExpr.qualifier().is(Tree.Kind.NAME) && "self".equals(((Name) qualifExpr.qualifier()).name())
+        && UnittestUtils.SKIP_METHOD.equals(qualifExpr.name().name()));
```
nils-werner-sonarsource, line 125: Why do you have a constant for this string and not for "pytest.skip". I wouldn't create a constant in `UnittestUtils` if it's only used in this check.

nils-werner-sonarsource (approved):
LGTM! See comments which could be addressed.

PR #1197, opened 2022-09-05
SONARPY-1040 S1854 (DeadStoreCheck) should not raise FP with assignment expressions

nils-werner-sonarsource:


python-checks/src/test/resources/checks/deadStore.py
```diff
@@ -327,3 +327,28 @@ def no_separator_after_deadstore():
...
+
+
+def assignment_expression_in_for_loop_definition_3():
+    (var := 2)
+    print (var)
+    for _ in (var := 2):  # Noncompliant
+    #        ^^^^^^^^^^
+        ...
+
+def complex_loop():
```
andreaguarino, line 350: what do you mean by `complex_loop`?

andreaguarino (approved):
LGTM, can you please build sonar-security locally with snapshot version of sonar-python to check there is no impact?

PR #1200, opened 2022-09-07
SONARPY-1102 Fix FP on S5914 for nonlocal variables

rudy-regazzoni-sonarsource:


rudy-regazzoni-sonarsource:
Not totally satisfied with this : it eliminate FP with 'nonlocal' variable but without using the CFG to determine the use/potential values.
From my understanding, it seems our current Dataflow Analyzer does not support properly those 'nonlocal' variable, but I couldn't find a way to modify/adapt this, so my current solution is more like a quick fix to "don't raise an issue in case of nonlocal : we can't determine if it's a real issue or a FP"

python-checks/src/main/java/org/sonar/python/checks/tests/UnconditionalAssertionCheck.java
```diff
@@ -143,8 +143,13 @@ private boolean isUnconditional(RegularArgument argument) {
 
     if (expression.is(NAME) || expression.is(QUALIFIED_EXPR)) {
       Symbol symbol = ((HasSymbol) expression).symbol();
-      if (symbol != null && isClassOrFunction(symbol)) {
-        return true;
+      if (symbol != null) {
+        if (isClassOrFunction(symbol)) {
+          return true;
+        }
+        if (symbol.usages().stream().anyMatch(usage -> usage.tree().parent().is(Tree.Kind.NONLOCAL_STMT))) {
```
nils-werner-sonarsource, line 150: Maybe create a method with a more descriptive name for this stream condition.

nils-werner-sonarsource (approved):
LGTM! See suggestion for readability.

PR #1207, opened 2022-09-20
SONARPY-1113 Rule S6330: Using unencrypted SQS queues is security-sensitive

rudy-regazzoni-sonarsource:


python-checks/src/test/java/org/sonar/python/checks/cdk/UnencryptedSqsQueueCheckTest.java
```diff
@@ -0,0 +1,31 @@
...
+package org.sonar.python.checks.cdk;
+
+import org.junit.Test;
+import org.sonar.python.checks.utils.PythonCheckVerifier;
+
+public class UnencryptedSqsQueueCheckTest {
+
+  @Test
+  public void test() {
+    PythonCheckVerifier.verify("src/test/resources/checks/cdk/unencryptedSqsQueueCheckTest.py", new UnencryptedSqsQueueCheck());
```
nils-werner-sonarsource, line 29:
```suggestion
    PythonCheckVerifier.verify("src/test/resources/checks/cdk/unencryptedSqsQueueCheck.py", new UnencryptedSqsQueueCheck());
```
So far I know, we do not add the test suffix in the python analyzer context

nils-werner-sonarsource (approved):
LGTM! It's a new concept you use in UnencryptedSqsQueueCheck to have two kinds of sub-checks. Normally, we have consumers registered to the FQNs. But for now, it's fine. Maybe I will implement another concept and we can decide which to use.

PR #1209, opened 2022-09-21
SONARPY-1117 Rule S6303: Using unencrypted RDS databases is security-sensitive

rudy-regazzoni-sonarsource:


nils-werner-sonarsource (changes requested):
Please clean up the PR and remove commits that are already on master.

python-checks/src/main/java/org/sonar/python/checks/cdk/DisabledRDSEncryptionCheck.java
```diff
@@ -0,0 +1,91 @@
...
+        checkCfnDatabaseArguments(subscriptionContext, callExpression);
+      }
+    });
+  }
+
+  protected void checkDatabaseArguments(SubscriptionContext ctx, CallExpression resourceConstructor) {
+    Map<String, ArgumentTrace> map = getArguments(ctx, resourceConstructor,
+      List.of(ARG_ENCRYPTED, ARG_ENCRYPTION_KEY));
+    ArgumentTrace argEncrypted = map.get(ARG_ENCRYPTED);
+    ArgumentTrace argEncryptionKey = map.get(ARG_ENCRYPTION_KEY);
```
nils-werner-sonarsource, line 56: I don't like too much having a method that wraps another to create a List. Which is to explode to get the single items again instead of having two expressions directly. But I see that you want to get rid of the `Optionals`. However, I don't see a smarter solution right now. Maybe you can move the `getArguments` method to this check because it is exclusively used here.

python-checks/src/main/java/org/sonar/python/checks/cdk/DisabledRDSEncryptionCheck.java
```diff
@@ -0,0 +1,91 @@
...
+      .filter(argumentTrace ->
+        argumentTrace.hasExpression(expression ->
+          Optional.of(expression)
+            .filter(expr -> expr.is(Tree.Kind.STRING_LITERAL)).map(StringLiteral.class::cast)
+            .filter(stringLiteral -> stringLiteral.trimmedQuotesValue().toLowerCase(Locale.ROOT).startsWith("aurora"))
+            .isPresent()
+        )
+      ).isPresent();
+  }
+}
```
nils-werner-sonarsource, line 90:
```suggestion
  protected boolean isEngineAurora(SubscriptionContext ctx, CallExpression resourceConstructor) {
    return getArgument(ctx, resourceConstructor, "engine")
      .filter(argumentTrace -> argumentTrace.hasExpression(startsWith("aurora"))).isPresent();
  }

  protected Predicate<Expression> startsWith(String expected) {
    return e -> getStringValue(e).filter(expected::startsWith).isPresent();
  }

  protected Optional<String> getStringValue(Expression expression) {
    try {
      return Optional.of(((StringLiteral) expression).trimmedQuotesValue());
    } catch (ClassCastException e) {
      return Optional.empty();
    }
  }
```
An idea to make it more readable.

nils-werner-sonarsource (approved):
LGTM! Please check see my comments.

PR #1212, opened 2022-09-21
SONARPY-1123 Rule S5332: Using clear-text protocols is security-sensitive for AWS::ElasticLoadBalancingV

nils-werner-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM ! Nothing much to say, code is very clear. Eventually I was surprised to see this `getProtocolArgument` method declaration to not repeat `"protocol"` thrice, I would probably have used a constant instead, but this way seems totally fine to me.
Also I like the replacement of isFqnValue -> isFqn, I will try to make this Predicate approach my own.

nils-werner-sonarsource:
> LGTM ! Nothing much to say, code is very clear. Eventually I was surprised to see this `getProtocolArgument` method declaration to not repeat `"protocol"` thrice, I would probably have used a constant instead, but this way seems totally fine to me. Also I like the replacement of isFqnValue -> isFqn, I will try to make this Predicate approach my own.

PR #1213, opened 2022-09-22
SONARPY-1116 Rule S6308: Using unencrypted OpenSearch domains is security-sensitive

rudy-regazzoni-sonarsource:


rudy-regazzoni-sonarsource:
Not sure what to do with the issue message part : `...this {OpenSearch|Elasticsearch} domain...`
I supposed that this was the message to use, but maybe we should determine if we are in OpenSearch or Elasticsearch ? In that case I would need to dig a bit more about how to do it.

python-checks/src/main/java/org/sonar/python/checks/cdk/DisabledESDomainEncryptionCheck.java
```diff
@@ -0,0 +1,82 @@
...
+import org.sonar.plugins.python.api.tree.CallExpression;
+import org.sonar.plugins.python.api.tree.DictionaryLiteral;
+import org.sonar.plugins.python.api.tree.Expression;
+import org.sonar.plugins.python.api.tree.KeyValuePair;
+import org.sonar.plugins.python.api.tree.StringLiteral;
+import org.sonar.plugins.python.api.tree.Tree;
+
+@Rule(key = "S6308")
+public class DisabledESDomainEncryptionCheck extends AbstractCdkResourceCheck {
+  private static final String OMITTING_MESSAGE = "Omitting %s causes encryption of data at rest to be disabled for this {OpenSearch|Elasticsearch} domain." +
```
nils-werner-sonarsource, line 33: As you have already guessed, according to the specification, you must have a sensitive message. I.e. depending on the COntext use "OpenSearch" or "Elasticsearch". Work best with a format string.

python-checks/src/main/java/org/sonar/python/checks/cdk/DisabledESDomainEncryptionCheck.java
```diff
@@ -0,0 +1,82 @@
...
+    " Make sure it is safe here.";
+  private static final String UNENCRYPTED_MESSAGE = "Make sure that using unencrypted {OpenSearch|Elasticsearch} domains is safe here.";
```
nils-werner-sonarsource, line 35: Same as above.

python-checks/src/main/java/org/sonar/python/checks/cdk/DisabledESDomainEncryptionCheck.java
```diff
@@ -0,0 +1,82 @@
...
+  private static void checkCfnDomain(SubscriptionContext ctx, CallExpression callExpression) {
+    checkDomain(ctx, callExpression, "encryption_at_rest_options", "aws_cdk.aws_opensearchservice.CfnDomain.EncryptionAtRestOptionsProperty");
+  }
+
+  private static void checkDomain(SubscriptionContext ctx, CallExpression callExpression, String encryptionArgName, String argFqnMethod) {
+    getArgument(ctx, callExpression, encryptionArgName).ifPresentOrElse(
+      argEncryptedTrace ->
+        argEncryptedTrace.addIssueIf(expr -> isMethodCallWithoutEnabledOrSetToFalse(ctx, expr, argFqnMethod)
+          || (expr.is(Tree.Kind.DICTIONARY_LITERAL) && isDictionaryEnabledValueFalseOrAbsent(ctx, (DictionaryLiteral) expr))
+          , UNENCRYPTED_MESSAGE)
```
nils-werner-sonarsource, line 57: Hard to read. You should move it to a Predicate or combine two predicates with `Predicate.or(Predicate)`

python-checks/src/main/java/org/sonar/python/checks/cdk/DisabledESDomainEncryptionCheck.java
```diff
@@ -0,0 +1,82 @@
...
+      , () -> ctx.addIssue(callExpression.callee(), String.format(OMITTING_MESSAGE, encryptionArgName))
+    );
+  }
+
+  private static boolean isMethodCallWithoutEnabledOrSetToFalse(SubscriptionContext ctx, Expression expr, String argFqnMethod) {
```
nils-werner-sonarsource, line 62:
```suggestion
  private static boolean isSensitiveOptionObj(SubscriptionContext ctx, Expression expr, String argFqnMethod) {
```

python-checks/src/main/java/org/sonar/python/checks/cdk/DisabledESDomainEncryptionCheck.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+    if (!expr.is(Tree.Kind.CALL_EXPR)) {
+      return true;
+    }
+
+    return getArgument(ctx, (CallExpression) expr, ENABLED)
+      .filter(argEnabledTrace -> !argEnabledTrace.hasExpression(AbstractCdkResourceCheck::isFalse)).isEmpty();
+  }
+
+  private static boolean isDictionaryEnabledValueFalseOrAbsent(SubscriptionContext ctx, DictionaryLiteral dictionaryLiteral) {
```
nils-werner-sonarsource, line 74:
```suggestion
  private static boolean isSensitiveOptionDict(SubscriptionContext ctx, DictionaryLiteral dictionaryLiteral) {
```

python-checks/src/main/java/org/sonar/python/checks/cdk/DisabledESDomainEncryptionCheck.java
```diff
@@ -0,0 +1,82 @@
...
+    return dictionaryLiteral.elements().stream()
+      .filter(element -> element.is(Tree.Kind.KEY_VALUE_PAIR)).map(KeyValuePair.class::cast)
+      .filter(keyValuePair -> keyValuePair.key().is(Tree.Kind.STRING_LITERAL)
+        && ((StringLiteral) keyValuePair.key()).trimmedQuotesValue().equals(ENABLED))
+      .map(keyValuePair -> ArgumentTrace.build(ctx, keyValuePair.value()))
+      .allMatch(argTrace -> argTrace.hasExpression(AbstractCdkResourceCheck::isFalse));
+  }
```
nils-werner-sonarsource, line 81: Such long streams over maps are quite hard to read. Sometimes it's easier to create a small and simple subclass like `Option` which already trims keys, and resolves values when constructed. So you can focus in the stream on filtering the content, not on the preparation of it.

nils-werner-sonarsource (changes requested):
Please see my comments. As you have already guessed you have to change the message and I would like to ask you to clean up the code so that it is easier to review.

python-checks/src/main/java/org/sonar/python/checks/cdk/DisabledESDomainEncryptionCheck.java
```diff
@@ -0,0 +1,118 @@
...
+  private static final String UNENCRYPTED_MESSAGE = "Make sure that using unencrypted %s domains is safe here.";
+  private static final String ENABLED = "enabled";
+  private static final String OPENSEARCH = "OpenSearch";
+  private static final String ELASTICSEARCH = "Elasticsearch";
+
+  @Override
+  protected void registerFqnConsumer() {
+    checkFqn("aws_cdk.aws_opensearchservice.Domain",
+      checkDomain("encryption_at_rest", "aws_cdk.aws_opensearchservice.EncryptionAtRestOptions", OPENSEARCH));
+    checkFqn("aws_cdk.aws_opensearchservice.CfnDomain",
```
nils-werner-sonarsource, line 50: Maybe add a space between each call of `checkFqn` to increase readability

python-checks/src/main/java/org/sonar/python/checks/cdk/DisabledESDomainEncryptionCheck.java
```diff
@@ -0,0 +1,118 @@
...
+      if (!isFqn(argFqnMethod).test(expr)) {
+        return false;
+      }
+      if (!expr.is(Tree.Kind.CALL_EXPR)) {
+        return true;
+      }
+
+      return getArgument(ctx, (CallExpression) expr, ENABLED)
+        .filter(argEnabledTrace -> !argEnabledTrace.hasExpression(isFalse())).isEmpty();
+    };
```
nils-werner-sonarsource, line 85: Works, but can be improved to have more findings.
What if the function call, which is actually an object initiation, takes place outside the call and only a variable is used here? I would honestly only check the initiation of the option and not check if a function call is in a function call.
Meaning any call to `aws_cdk.aws_elasticsearch.EncryptionAtRestOptions` as an example. It may be that an option object is not used. But by definition, it is already insecure and should not be used in a domain. WDYT?

python-checks/src/main/java/org/sonar/python/checks/cdk/DisabledESDomainEncryptionCheck.java
```diff
@@ -0,0 +1,118 @@
...
+      .orElse(Collections.emptyList()).stream();
+  }
+
+  private static Predicate<KeyValuePair> isKey(String keyName) {
+    return keyValuePair -> Optional.of(keyValuePair)
+      .map(KeyValuePair::key)
+      .filter(key -> key.is(Tree.Kind.STRING_LITERAL)).map(StringLiteral.class::cast)
+      .filter(string -> string.trimmedQuotesValue().equals(keyName))
+      .isPresent();
+  }
```
nils-werner-sonarsource, line 113: Much better!
You miss creating a trace also for the keys and values. They also can be an expression -> variables -> and so their value can be defined somewhere else. Keep it for now. We can discuss an improvement later.

nils-werner-sonarsource (approved):
LGTM! Please see my comments for further improvements.

python-checks/src/main/java/org/sonar/python/checks/cdk/DisabledESDomainEncryptionCheck.java
rudy-regazzoni-sonarsource: Mmhh not totally sure what you mean here.
If I understand correctly, you are talking about detecting such usecase :
```
encryptionOption = aws_os.EncryptionAtRestOptions(enabled=False)
aws_os.Domain(encryption_at_rest=encryptionOption) # NonCompliant
```

This usecase is already detected because this method `isSensitiveOptionObj` is called on each expressions of the argumentTrace of the argument `encryption_at_rest`.

Your approach seems easier and possibly more reliable, but I see two relative disadvantage :
- as you say we will raise an issue even if the option object is not used (not a big issue from my perspective)
- we would not be able to easily highlight the call to `Domain` which would use this variable, although this does not seems to be specified in the RSPEC (no highlight.doc ?)

Please let me know if you meant something else.

python-checks/src/test/resources/checks/cdk/disabledESDomainEncryptionCheck.py
```diff
@@ -0,0 +1,57 @@
...
+    def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
+        super().__init__(scope, construct_id, **kwargs)
+        encryptionRestOptionMethodGood = aws_os.EncryptionAtRestOptions(enabled=True)
+        encryptionRestOptionMethodBad = aws_os.EncryptionAtRestOptions(enabled=False)
+        encryptionRestOptionDictionaryGood = {"enabled": True}
+        encryptionRestOptionDictionaryBad = {"enabled": False}
+        encrypted = True
+        not_encrypted = False
+
+        aws_os.Domain(self, "unencrypted-domain-1", version=aws_os.EngineVersion.OPENSEARCH_1_3) # NonCompliant{{Omitting encryption_at_rest causes encryption of data at rest to be disabled for this OpenSearch domain. Make sure it is safe here.}}
```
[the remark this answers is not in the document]
rudy-regazzoni-sonarsource: I very much agree, the tests as currently written are a nightmare to read.
Didn't know we could go this way, I will apply your suggestion ! I will just add few lines to explain what a normal call looks like for CDK.

python-checks/src/main/java/org/sonar/python/checks/cdk/DisabledESDomainEncryptionCheck.java
nils-werner-sonarsource: You're right. Missed the point, that every expression is checked!

PR #1215, opened 2022-09-22
SONARPY-1118 Rule S6332: Using unencrypted EFS file systems is security-sensitive

nils-werner-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM ! Nice change for `isFalse` and `isNone` into Predicate, will just need to adapt our other commits ;)

PR #1224, opened 2022-09-29
SONARPY-1152 Rule S5332: Adding precision on the issue by highlighting the root cause

rudy-regazzoni-sonarsource:


nils-werner-sonarsource (approved):
LGTM!

PR #1228, opened 2022-10-04
SONARPY-1159 Allowing unrestricted outbound communications is securit…

mstachniuk:
…y-sensitive


python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedOutboundCommunicationsCheck.java
```diff
@@ -0,0 +1,57 @@
...
+public class UnrestrictedOutboundCommunicationsCheck extends AbstractCdkResourceCheck {
+  @Override
+  protected void registerFqnConsumer() {
+    checkFqns(List.of("aws_cdk.aws_ec2.SecurityGroup", "aws_cdk.aws_ec2.SecurityGroup.from_security_group_id"), (subscriptionContext, callExpression) -> {
+      RegularArgument argument = TreeUtils.nthArgumentOrKeyword(3, "allow_all_outbound", callExpression.arguments());
+      if (argument == null) {
+        subscriptionContext.addIssue(callExpression, "Omitting \"allow_all_outbound\" enables unrestricted outbound communications. Make sure it is safe here.");
+      } else if (isArgumentTrue(argument.expression())) {
+        subscriptionContext.addIssue(argument, "Make sure that allowing unrestricted outbound communications is safe here.");
+      }
```
nils-werner-sonarsource, line 41: You can make use of a helper method as well as a structure that represents an argument flow
```suggestion
      CdkUtils.getArgument(subscriptionContext, callExpression, "allow_all_outbound").ifPresentOrElse(
        argument -> argument.addIssueIf(isTrue(), "Make sure that allowing unrestricted outbound communications is safe here."),
        () -> subscriptionContext.addIssue(callExpression.callee(), "Omitting \"allow_all_outbound\" enables unrestricted outbound communications. Make sure it is safe here."));
```
Also, it is enough to highlight only the callee and not the whole call expression. This way you can keep the focus on the constructor.

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedOutboundCommunicationsCheck.java
```diff
@@ -0,0 +1,57 @@
...
+      Name name = (Name) expression;
+      if (name.name().equals("True")) {
+        return true;
+      } else {
+        Expression assignedValue = Expressions.singleAssignedValue(name);
+        return isArgumentTrue(assignedValue);
+      }
+    }
+    return false;
+  }
```
nils-werner-sonarsource, line 56:
```suggestion
  public static Predicate<Expression> isTrue() {
    return expression -> Optional.ofNullable(expression)
      .filter(Name.class::isInstance).map(Name.class::cast)
      .map(Name::name)
      .filter("True"::equals).isPresent();
  }
```
In the other CDK-related checks we used predicate to make the checks more readable. Maybe it's an option for you.-.

python-checks/src/test/resources/checks/cdk/unrestrictedOutboundCommunicationsCheck.py
```diff
@@ -0,0 +1,49 @@
...
+    allow_all_outbound=True  # NonCompliant{{Make sure that allowing unrestricted outbound communications is safe here.}}
+)
+
+ec2.SecurityGroup.from_security_group_id(  # NonCompliant{{Omitting "allow_all_outbound" enables unrestricted outbound communications. Make sure it is safe here.}}
+    self,
+    "SensitiveDefault",
+    vpc=vpc
+)
+
+outbound_restriction=True
```
nils-werner-sonarsource, line 29: Using ExpressionFlow for arguments you could also highlight the assignment of the sensitive value as a secondary location.

nils-werner-sonarsource (approved):
LGTM!  Please take a look at my comments. You may want to use helper methods and structures that we have created for the other CDK checks. They will help you to keep the implementation leaner.

PR #1229, opened 2022-10-04
SONARPY-1137 Rule S6321: Administration services access should be restricted to specific IP addresses

rudy-regazzoni-sonarsource:


rudy-regazzoni-sonarsource:
I probably went to far with this one.
I tried to push a bit on our predicate logic so that we can define/write clear rules, but ended up writing lot of classes to reach this.
Also not totally sure if the rule written as it is is readable.
Please let me know your feedback, even if we don't push this code to production I hope it's a nice opportunity to discuss about the approach :)

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheck.java
```diff
@@ -0,0 +1,230 @@
...
+    this.ctx = ctx;
+
+    getArgument(ctx, callExpression, "security_group_ingress")
+      .flatMap(flow -> flow.getExpression(isListLiteral()).map(ListLiteral.class::cast))
+      .ifPresent(list -> list.elements().expressions()
+        // Process each expression in the Python list
+        .forEach(expression -> {
+          // Process each location of those expressions
+          CdkUtils.ExpressionFlow flow = CdkUtils.ExpressionFlow.build(ctx, expression);
+          for (Expression expr : flow.locations()) {
```
nils-werner-sonarsource, line 83: I think we can make use of the fact, that the last location of the flow is actually containing the value. All steps between will only be left-hand Name nodes of assignment expressions. WDYT about having a method in the ExpressionFlow like `value()` which returns `flow.getLast()`?

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheck.java
```diff
@@ -0,0 +1,230 @@
...
+    getCall(expression, "aws_cdk.aws_ec2.CfnSecurityGroup.IngressProperty")
+      .filter(callExpression -> isBadProtocolWithEmptyIpAddressAndAdminPort(callExpression) || isInvalidProtocolWithEmptyIpAddress(callExpression))
+      .ifPresent(callExpression -> {
+        getArgument(ctx, callExpression, CIDR_IP).ifPresent(flow -> flow.addIssue(MESSAGE));
+        getArgument(ctx, callExpression, CIDR_IPV6).ifPresent(flow -> flow.addIssue(MESSAGE));
+      });
+  }
+
+  private boolean isBadProtocolWithEmptyIpAddressAndAdminPort(CallExpression callExpression) {
+    return is(callExpression, IP_PROTOCOL, isString(BAD_PROTOCOL))
```
nils-werner-sonarsource, line 111: The name of the method ìs` should be more explicit about what actually is checked. Maybe something like ìsCallWithArgument`. Or you could think about a small class like `Call` which wrapped the `CallExpression` and has a method `boolean hasArgument(String arg, Predicate<Expression> predicate)`. This class could also contain the ctx, which is needed for the resolution of the argument.

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheck.java
```diff
@@ -0,0 +1,230 @@
...
+        .filter(Optional::isPresent).map(Optional::get).collect(Collectors.toList());
+
+      for (CdkUtils.ResolvedKeyValuePair pair : pairs) {
+        pair.key.getExpression(isStringLiteral()).map(StringLiteral.class::cast)
+          .ifPresent(key -> dict.map.put(key.trimmedQuotesValue(), pair));
+      }
+      return dict;
+    }
+
+    public boolean is(String key, Predicate<Expression> valuePredicate) {
```
nils-werner-sonarsource, line 213: This method name should also be a bit more explicit

nils-werner-sonarsource (commented):
I've already had a first look. I think you did a good job to reuse a lot of the tools. I added some comments you may work on.

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheck.java
rudy-regazzoni-sonarsource: Good idea for the Call class, I will add it !

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheck.java
```diff
@@ -0,0 +1,225 @@
...
+    if (isCallWithArgumentBadProtocolEmptyIpAddressAdminPort(call) || isCallWithArgumentInvalidProtocolEmptyIpAddress(call)) {
+      getArgument(call.ctx, call.callExpression, CIDR_IP).ifPresent(flow -> flow.addIssue(MESSAGE));
+      getArgument(call.ctx, call.callExpression, CIDR_IPV6).ifPresent(flow -> flow.addIssue(MESSAGE));
+    }
+  }
+
+  private static boolean isCallWithArgumentBadProtocolEmptyIpAddressAdminPort(Call call) {
+    return call.hasArgument(IP_PROTOCOL, isString(SENSITIVE_PROTOCOL))
+      && (call.hasArgument(CIDR_IP, isString(EMPTY_IPV4)) || call.hasArgument(CIDR_IPV6, isString(EMPTY_IPV6)))
+      && call.inInterval("from_port", "to_port", ADMIN_PORTS);
```
nils-werner-sonarsource, line 105: This method name does not feel intuitive. I would suggest something like `hasSensitvePortRange`

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheck.java
```diff
@@ -0,0 +1,225 @@
...
+    public Call(SubscriptionContext ctx, CallExpression callExpression) {
+      this.ctx = ctx;
+      this.callExpression = callExpression;
+    }
+
+    boolean hasArgument(String name, Predicate<Expression> predicate) {
+      return getArgument(ctx, callExpression, name).filter(flow -> flow.hasExpression(predicate)).isPresent();
+    }
+
+    public Optional<Long> asLong(String name) {
```
nils-werner-sonarsource, line 193: This name also is not intuitive. How a call should be represented as long? Maybe `getArgumentAsLong(String argumentName)`

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheck.java
```diff
@@ -0,0 +1,225 @@
...
+        .filter(Optional::isPresent).map(Optional::get).collect(Collectors.toList());
+
+      for (CdkUtils.ResolvedKeyValuePair pair : pairs) {
+        pair.key.getExpression(isStringLiteral()).map(StringLiteral.class::cast)
+          .ifPresent(key -> dict.map.put(key.trimmedQuotesValue(), pair));
+      }
+      return dict;
+    }
+
+    public boolean is(String key, Predicate<Expression> valuePredicate) {
```
nils-werner-sonarsource, line 156: Maybe `hasKeyValuePair` is a better name

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheck.java
```diff
@@ -0,0 +1,225 @@
...
+    default boolean inInterval(String minName, String maxName, long[] numbers) {
+      Optional<Long> min = asLong(minName);
+      Optional<Long> max = asLong(maxName);
+
+      if (min.isEmpty() || max.isEmpty()) {
+        return false;
+      }
+
+      return isInInterval(min.get(), max.get(), numbers);
+    }
```
nils-werner-sonarsource, line 214: I'm not sure if it's a nice way to have such a generic method inside an interface like this.

python-checks/src/test/resources/checks/cdk/unrestrictedAdministration_CfnSecurityGroup.py
```diff
@@ -0,0 +1,370 @@
+from aws_cdk import aws_ec2 as ec2
+
+### CfnSecurityGroup
+# A typical call to aws_cdk.aws_ec2.CfnSecurityGroup would look like this. For simplicity/readability sake, we will omit arguments non-related to the check.
+ec2.CfnSecurityGroup(self, "cfn-based-security-group", group_description="cfn based security group", group_name="cfn-based-security-group", vpc_id=vpc.vpc_id,
+    security_group_ingress=[
+        ec2.CfnSecurityGroup.IngressProperty(
+            ip_protocol="6",
+            cidr_ip="0.0.0.0/0", # Noncompliant
```
nils-werner-sonarsource, line 9: Maybe check the message at least once.

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheck.java
```diff
@@ -0,0 +1,225 @@
...
+          CdkUtils.ExpressionFlow flow = CdkUtils.ExpressionFlow.build(ctx, expression);
+          raiseIssueIfIngressPropertyCallWithSensitiveArgument(ctx, flow.getLast());
+          raiseIssueIfDictionaryWithSensitiveArgument(ctx, flow.getLast());
+        })
+      );
+  }
+
+  private static void checkCfnSecurityGroupIngress(SubscriptionContext ctx, CallExpression callExpression) {
+    checkCallCfnSecuritySensitive(new Call(ctx, callExpression));
+  }
```
nils-werner-sonarsource, line 86: IMO, you can have this logic inside the `registerFqnConsumer`

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheck.java
```diff
@@ -0,0 +1,225 @@
...
+  @Override
+  protected void registerFqnConsumer() {
+    checkFqn("aws_cdk.aws_ec2.CfnSecurityGroup", UnrestrictedAdministrationCheck::checkCfnSecurityGroup);
+    checkFqn("aws_cdk.aws_ec2.CfnSecurityGroupIngress", UnrestrictedAdministrationCheck::checkCfnSecurityGroupIngress);
+  }
+
+  // Checks methods
+  private static void checkCfnSecurityGroup(SubscriptionContext ctx, CallExpression callExpression) {
+    getArgument(ctx, callExpression, "security_group_ingress")
+      .flatMap(flow -> flow.getExpression(isListLiteral()).map(ListLiteral.class::cast))
```
nils-werner-sonarsource, line 73: Maybe move this logic to a method like `Optional<ListLiteral> getListExpression(ExpressionFlow expression)`

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheck.java
```diff
@@ -0,0 +1,225 @@
...
+      .ifPresent(list -> list.elements().expressions()
+        // Process each expression in the Python list
+        .forEach(expression -> {
+          CdkUtils.ExpressionFlow flow = CdkUtils.ExpressionFlow.build(ctx, expression);
+          raiseIssueIfIngressPropertyCallWithSensitiveArgument(ctx, flow.getLast());
+          raiseIssueIfDictionaryWithSensitiveArgument(ctx, flow.getLast());
+        })
+      );
```
nils-werner-sonarsource, line 81:
```
.map(list -> list.elements().expressions())
.orElse(Collections.emptyList())
.stream()
.map(expression -> CdkUtils.ExpressionFlow.build(ctx, expression).getLast())
.forEach(expression -> {
  raiseIssueIfIngressPropertyCallWithSensitiveArgument(ctx, expression);
  raiseIssueIfDictionaryWithSensitiveArgument(ctx, expression);
});
```
A possible way to get rid of the nesting.

nils-werner-sonarsource (approved):
LGTM! I've added some comments about refactoring. Decide on your own which one to address.

PR #1230, opened 2022-10-05
SONARPY-1063   Rules support PCI DSS Security Standard

nils-werner-sonarsource:


python-checks/src/test/resources/checks/regex/impossibleBoundariesCheck.py
```diff
@@ -26,9 +26,7 @@ def non_compliant(input):
     re.match(r'a\n\Ab', input, re.MULTILINE)  # Noncompliant
 
     # False positives because the end delimiter does not capture the newlines (SONARPHP-1238)
-    re.match(r'a$(\n)', input)  # Noncompliant
     re.match(r'a$./s', input)  # Noncompliant
-    re.match(r'a\Z(\n)', input)  # Noncompliant
```
rudy-regazzoni-sonarsource, line 30: If those FP are not present anymore, I would consider moving them in the compliant method if they are not raised anymore.

rudy-regazzoni-sonarsource (approved):
LGTM !

python-checks/src/test/resources/checks/regex/impossibleBoundariesCheck.py
nils-werner-sonarsource: Good point!

PR #1237, opened 2022-10-10
Rule S6302: Policies granting all privileges are security-sensitive for PolicyStatements

nils-werner-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/cdk/PrivilegePolicyCheck.java
```diff
@@ -0,0 +1,117 @@
...
+    if (action.hasExpression(isStringLiteral())) {
+      action.addIssueIf(isString("*"), MESSAGE, secondaryLocation(effect));
+    } else {
+      CdkUtils.getList(action).ifPresent(actions -> checkWildcardAction(ctx, actions, effect));
+    }
+  }
+
+  private static void checkWildcardAction(SubscriptionContext ctx, ListLiteral list, ExpressionFlow effect) {
+    CdkUtils.getListElements(ctx, list)
+      .stream()
```
nils-werner-sonarsource, line 96:
```suggestion
    CdkUtils.getListElements(ctx, list).stream()
```

python-checks/src/main/java/org/sonar/python/checks/cdk/CdkUtils.java
```diff
@@ -206,5 +255,9 @@ private ResolvedKeyValuePair(ExpressionFlow key, ExpressionFlow value) {
     static ResolvedKeyValuePair build(SubscriptionContext ctx, KeyValuePair pair) {
       return new ResolvedKeyValuePair(ExpressionFlow.build(ctx, pair.key()), ExpressionFlow.build(ctx, pair.value()));
     }
+
+    public ExpressionFlow value() {
```
mstachniuk, line 259: I understand the reason behind this method (to use `.map(ResolvedKeyValuePair::value)`) but `key` and `value` are package private and `.value` is used in many places. I think there is a god practice to use always either field access or getter/method access (for consistency in the code), but never both at the same time. WDYT?

python-checks/src/test/java/org/sonar/python/checks/cdk/PrivilegePolicyCheckTest.java
```diff
@@ -0,0 +1,45 @@
...
+import org.sonar.plugins.python.api.PythonCheck;
+import org.sonar.python.checks.utils.PythonCheckVerifier;
+
+public class PrivilegePolicyCheckTest {
+
+  PythonCheck check = new PrivilegePolicyCheck();
+
+  @Test
+  public void policyStatement() {
+    PythonCheckVerifier.verify("src/test/resources/checks/cdk/PrivilegePolicyCheck/policyStatement.py", check);
```
mstachniuk, line 32: PrivilegePolicyCheck - it should be snake case

mstachniuk (approved):
LGTM. I put some comments, most of them are newbie questions

python-checks/src/main/java/org/sonar/python/checks/cdk/CdkUtils.java
```diff
@@ -101,6 +105,50 @@ protected static Optional<ExpressionFlow> getArgument(SubscriptionContext ctx, C
...
+      .flatMap(Optional::stream)
+      .collect(Collectors.toList());
+  }
+
+  public static List<ResolvedKeyValuePair> resolveDictionary(SubscriptionContext ctx, DictionaryLiteral dict) {
+    return dict.elements().stream()
+      .map(e -> CdkUtils.getKeyValuePair(ctx, e))
+      .flatMap(Optional::stream)
+      .collect(Collectors.toUnmodifiableList());
+  }
```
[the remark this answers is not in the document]
nils-werner-sonarsource: I had the same thought, but was honestly too lazy. I will make up for it.

python-checks/src/test/resources/checks/cdk/PrivilegePolicyCheck/fromJson.py
```diff
@@ -0,0 +1,32 @@
+from aws_cdk import aws_iam as iam
+
+iam.PolicyStatement.from_json({
+    "Effect": "Allow",
+ #  ^^^^^^^^^^^^^^^^^> {{Related effect}}
+    "Action": ["*"],  # Noncompliant {{Make sure granting all privileges is safe here.}}
```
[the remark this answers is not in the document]
nils-werner-sonarsource: Should already work. I will add a testcase.

python-checks/src/main/java/org/sonar/python/checks/cdk/PrivilegePolicyCheck.java
```diff
@@ -0,0 +1,117 @@
...
+public class PrivilegePolicyCheck extends AbstractCdkResourceCheck {
+
+  private static final String MESSAGE = "Make sure granting all privileges is safe here.";
+  private static final String SECONDARY_MESSAGE = "Related effect";
+
+  @Override
+  protected void registerFqnConsumer() {
+    checkFqn("aws_cdk.aws_iam.PolicyStatement", (ctx, call) -> {
+      @Nullable ExpressionFlow effect = CdkUtils.getArgument(ctx, call, "effect").orElse(null);
+      CdkUtils.getArgumentAsList(ctx, call, "actions").ifPresent(actions -> {
```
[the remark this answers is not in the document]
nils-werner-sonarsource: Unfortunately, we do not have a convention for this. We should address times.

python-checks/src/main/java/org/sonar/python/checks/cdk/CdkUtils.java
nils-werner-sonarsource: I agree. Let's keep the field access for now

PR #1238, opened 2022-10-10
SONARPY-1154 Implement for Database migration resources

mstachniuk:


python-checks/src/main/java/org/sonar/python/checks/cdk/CdkPredicate.java
```diff
@@ -42,6 +42,13 @@ public static Predicate<Expression> isFalse() {
     return expression -> Optional.ofNullable(expression.firstToken()).map(Token::value).filter("False"::equals).isPresent();
   }
 
+  /**
+   * @return Predicate which tests if expression is boolean literal and is set to `true`
+   */
+  public static Predicate<Expression> isTrue() {
+    return expression -> Optional.ofNullable(expression.firstToken()).map(Token::value).filter("True"::equals).isPresent();
+  }
```
nils-werner-sonarsource, line 50: I think you can drop the method `org.sonar.python.checks.cdk.UnrestrictedOutboundCommunicationsCheck#isTrue` and replace it's call with the new predicate.

nils-werner-sonarsource (approved):
LGTM! Please check out my comment.

PR #1239, opened 2022-10-10
SONARPY-1135 Rule S6321: Administration services access should be res…

rudy-regazzoni-sonarsource:
…tricted to specific IP addresses -> level 2 construct


python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheck.java
```diff
@@ -43,146 +44,14 @@
...
-
-      if (min.isEmpty() || max.isEmpty()) {
-        return false;
-      }
-
-      return isInInterval(min.get(), max.get(), ADMIN_PORTS);
-    }
+  public void initialize(Context context) {
+    new UnrestrictedAdministrationCheckPartCfnSecurity().initialize(context);
+    new UnrestrictedAdministrationCheckPartConnections().initialize(context);
```
andreaguarino, line 51: I find this very strange: I guess it means we have two different AST visits for the same check? Does it mean we probably need to split this rule?

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheck.java
```diff
@@ -43,146 +44,14 @@
...
-    boolean hasSensitivePortRange(String minName, String maxName) {
-      Optional<Long> min = getArgumentAsLong(minName);
-      Optional<Long> max = getArgumentAsLong(maxName);
-
-      if (min.isEmpty() || max.isEmpty()) {
-        return false;
-      }
-
-      return isInInterval(min.get(), max.get(), ADMIN_PORTS);
-    }
```
andreaguarino, line 183: It would have been nice to split this refactoring in a separate commit, to ease review

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheck.java
```diff
@@ -43,146 +44,14 @@
...
+  public void initialize(Context context) {
+    new UnrestrictedAdministrationCheckPartCfnSecurity().initialize(context);
+    new UnrestrictedAdministrationCheckPartConnections().initialize(context);
   }
 
+  // Utils elements used by each classes implementing the rule
   static class Call {
```
andreaguarino, line 55: Why a wrapper on CallExpression? As far as I can see it's only proxying `hasArgument` and `getArgument`?

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheck.java
```diff
@@ -193,7 +62,11 @@ public Call(SubscriptionContext ctx, CallExpression callExpression) {
     }
 
     boolean hasArgument(String name, Predicate<Expression> predicate) {
-      return getArgument(ctx, callExpression, name).filter(flow -> flow.hasExpression(predicate)).isPresent();
+      return hasArgument(name, -1, predicate);
```
andreaguarino, line 65: why `-1`?

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheckPartCfnSecurity.java
```diff
@@ -0,0 +1,185 @@
+/*
```
andreaguarino, line 1: I'm not sure what I should review in this class: was it just copy-pasted from `UnrestrictedAdministrationCheck`?

python-checks/src/test/resources/checks/cdk/unrestrictedAdministration_Connections.py
```diff
@@ -0,0 +1,145 @@
...
+        aws_cdk.aws_ec2.Instance().connections.allow_from(ec2.Peer.any_ipv4(), ec2.Port.all_tcp()) # Noncompliant
+        aws_cdk.aws_ec2.LaunchTemplate().connections.allow_from(ec2.Peer.any_ipv4(), ec2.Port.all_tcp()) # Noncompliant
+        aws_cdk.aws_ec2.SecurityGroup().connections.allow_from(ec2.Peer.any_ipv4(), ec2.Port.all_tcp()) # Noncompliant
+        aws_cdk.aws_kinesisfirehose_alpha.DeliveryStream().connections.allow_from(ec2.Peer.any_ipv4(), ec2.Port.all_tcp()) # Noncompliant
+        aws_cdk.aws_stepfunctions_tasks.SageMakerCreateTrainingJob().connections.allow_from(ec2.Peer.any_ipv4(), ec2.Port.all_tcp()) # Noncompliant
+        aws_cdk.aws_stepfunctions_tasks.SageMakerCreateModel().connections.allow_from(ec2.Peer.any_ipv4(), ec2.Port.all_tcp()) # Noncompliant
+        aws_cdk.aws_stepfunctions_tasks.EcsRunTask().connections.allow_from(ec2.Peer.any_ipv4(), ec2.Port.all_tcp()) # Noncompliant
+        aws_cdk.aws_redshift_alpha.Cluster().connections.allow_from(ec2.Peer.any_ipv4(), ec2.Port.all_tcp()) # Noncompliant
+        aws_cdk.aws_opensearchservice.Domain().connections.allow_from(ec2.Peer.any_ipv4(), ec2.Port.all_tcp()) # Noncompliant
```
andreaguarino, line 134: do we need to test them all?

andreaguarino (changes requested):
Nice work, please have a look at my comments.
Some generated code should be removed.

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheck.java
rudy-regazzoni-sonarsource: This could be a possibility indeed.

The **CfnSecurity** part is doing checks on :
- aws_cdk.aws_ec2.CfnSecurityGroup
- aws_cdk.aws_ec2.CfnSecurityGroupIngress

While the **Connections** part is doing check on :
- aws_cdk.aws_ec2.Connections
- aws_cdk.aws_ec2.SecurityGroup
- aws_cdk.aws_ec2.*constructs*

The rule is quite big and on our side we splitted the ticket in two part that could be coded independently from each other.
They still do check similar things (ip protocol, ip address, ports...), but on different objects.

So in the end I choose to split the code in two different class for readability, as they don't directly share logic or elements.
WDYT should be done here ?

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheck.java
rudy-regazzoni-sonarsource: Indeed, in the end I'm not sure it's bringing much to the code and just obfuscate a bit more, I will try to remove it !

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheck.java
rudy-regazzoni-sonarsource: I use -1 as an invalid/impossible position, in case the concerned CallExpression is about a python method and an argument which can only be passed by name. In that case there is no valid position, so it should only retrieve the argument if it can find a valid name.

With a bit of insight now it does not look great :D Maybe should I replace `int` with `Integer`, so I can pass `null` value instead, but I would also need to modify `getArgument()` to support the `null` value and not call `TreeUtils.nthArgumentOrKeyword()` in that case, as it is expecting a `int` as `argPos`. Or I can also adapt `TreeUtils.nthArgumentOrKeyword()` to use `Integer` and support null value.

WDYT ?

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheckPartCfnSecurity.java
rudy-regazzoni-sonarsource: I confirm, no review required as it is indeed a copy/paste, I put this in the original commit message :

> Not necessary to review file [UnrestrictedAdministrationCheckPartCfnSecurity.java](https://github.com/SonarSource/sonar-python/pull/1239/files#diff-7bffa4cdc6308635fe43979de897f7756575feff19fd5ead9d2621958bbd3586), the content has been moved so it appears as new, but in fact It has been already reviewed in the https://github.com/SonarSource/sonar-python/pull/1229.

python-checks/src/test/resources/checks/cdk/unrestrictedAdministration_Connections.py
rudy-regazzoni-sonarsource: To be honest it helped me to make sure there was no error/missing/mistake in the writing of the Typeshed stubs.
I put them at the end of the Python test file in a dedicated method to avoid having them disturb much the readability of this test file.
Do you think they should be removed ? As the list of constructs was specifically provided in the original Jira ticket and in the RSPEC, I was thinking it would be best to make sure and ensure with our test that they were supported. Please let me know otherwise.

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheck.java
andreaguarino: I see, thanks for clarification! I would be a bit more explicit and have a dedicated `hasKeywordArgument` and probably use `TreeUtils.argumentByKeyword` utility

python-frontend/src/main/java/org/sonar/python/tree/TreeUtils.java
```diff
@@ -338,8 +339,12 @@ public static String fullyQualifiedNameFromQualifiedExpression(QualifiedExpressi
   @CheckForNull
   public static String fullyQualifiedNameFromExpression(Expression expression) {
     if (expression.is(Kind.NAME)) {
-      Symbol symbol = ((Name) expression).symbol();
-      return symbol != null ? symbol.fullyQualifiedName() : ((Name) expression).name();
+      if (expression.type() instanceof RuntimeType) {
+        return ((RuntimeType) expression.type()).getTypeClass().fullyQualifiedName();
```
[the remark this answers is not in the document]
rudy-regazzoni-sonarsource: Mmhh indeed.
I wanted to change the way `AbstractCdkResourceCheck` work, to allow our CDK rules to work on resolved inferred types with our recent type resolution progress.
But indeed this is changing this method that is used in many place.
I restored `fullyQualifiedNameFromExpression` and created a new method `resolveFullyQualifiedNameWithInferredType` which is doing that, It is now used in `AbstractCdkResourceCheck`.

rudy-regazzoni-sonarsource:
@andreaguarino I was able to address all your comment.
The new commit "Fix after review" contain those changed, including the restoration of third parties protobuf and the removing of the output/debug content.
Please let me know if you find anything else :)

python-checks/src/main/java/org/sonar/python/checks/cdk/UnrestrictedAdministrationCheckPartConnections.java
```diff
@@ -0,0 +1,127 @@
...
+    return (ctx, callExpression) ->
+      getArgument(ctx, callExpression, portName, 0).ifPresent(
+        flow -> {
+          if(flow.hasExpression(IS_SENSITIVE_PORT)) {
+            ctx.addIssue(getMethodExpression(callExpression), MESSAGE_BAD_METHOD);
+          }
+        });
+  }
+
+  private static Expression getMethodExpression(CallExpression callExpression) {
```
andreaguarino, line 120: If I understood correctly, this is to get the tree to use as primary location.
I would rename this method to something like "getMethodPrimaryLocation"

python-checks/src/main/java/org/sonar/python/checks/cdk/AbstractCdkResourceCheck.java
```diff
@@ -49,8 +50,8 @@ public void initialize(SubscriptionCheck.Context context) {
 
   protected void visitNode(SubscriptionContext ctx) {
     CallExpression node = (CallExpression) ctx.syntaxNode();
-    Optional.ofNullable(node.calleeSymbol())
-      .map(Symbol::fullyQualifiedName)
+    Optional.of(node)
+      .map(TreeUtils::fullyQualifiedNameFromExpression)
```
andreaguarino, line 54: it looks like the previous implementation still works, i.e.
 ```java
Optional.ofNullable(node.calleeSymbol())
```
So I would suggest to use the old code, and to revert the changes done inside `TreeUtils::fullyQualifiedNameFromExpression` , sorry! :D

File: python-frontend/src/main/java/org/sonar/python/types/RuntimeType.java [hunk 10]
```diff
@@ -28,7 +28,7 @@
 import org.sonar.plugins.python.api.types.InferredType;
 import org.sonar.python.semantic.ClassSymbolImpl;
 
-class RuntimeType implements InferredType {
+public class RuntimeType implements InferredType {
```
andreaguarino, line 31: I think this is not needed anymore

andreaguarino (changes requested):
Overall looks good,  I suggested some additional changes and the PR will be ready to be merged :)

python-checks/src/main/java/org/sonar/python/checks/cdk/AbstractCdkResourceCheck.java
rudy-regazzoni-sonarsource: My bad, this one was lost in the many change that happened regarding FQN, I didn't saw that I did not restore the code properly.

python-frontend/src/main/java/org/sonar/python/tree/TreeUtils.java
```diff
@@ -337,6 +337,12 @@ public static String fullyQualifiedNameFromQualifiedExpression(QualifiedExpressi
 
   @CheckForNull
   public static String fullyQualifiedNameFromExpression(Expression expression) {
+    if (expression instanceof HasSymbol) {
+      Symbol symbol = ((HasSymbol) expression).symbol();
+      if (symbol != null) {
+        return symbol.fullyQualifiedName();
+      }
+    }
```
andreaguarino, line 345: Because this is not needed anymore I would suggest to remove it. 
Alternatively, we can keep it and add a dedicated unit test for it

andreaguarino (approved):
LGTM! please check my last comment before merging to master :)

PR #1245, opened 2022-10-12
SONARPY-1155 Rule S6317: AWS IAM policies should not allow privilege escalation

nils-werner-sonarsource:
I introduced a data class that represents a PolicyStatement object. On the hand created by the constructor and on the other hand, created by the `from_json` method. They may look a bit overengineered but IMO we can improve PrivilegePolicyCheck(S6302) and ResourceAccessPolicyCheck(S6304) to reduce their complexity.


python-checks/src/main/java/org/sonar/python/checks/cdk/CdkPredicate.java
```diff
@@ -93,6 +94,13 @@ public static Predicate<Expression> isString(Set<String> expectedValues) {
     return expression -> CdkUtils.getString(expression).filter(expectedValues::contains).isPresent();
   }
 
+  /**
+   * @return Predicate which tests if expression is a string matches the pattern
+   */
+  public static Predicate<Expression> isString(Pattern pattern) {
```
mstachniuk, line 100: I suggest to rename the method to `matches` or `match`, then consumers code is more like English sentence and can't be confused with `isString(String expectedValue)`, see in other place:

```
    ExpressionFlow sensitiveAction = getSensitiveExpression(actions, isString(SENSITIVE_ACTIONS));
    ExpressionFlow sensitiveResource = getSensitiveExpression(resources, matches(SENSITIVE_RESOURCE_PATTERN));
```

python-checks/src/test/resources/checks/cdk/iamPrivilegeEscalation/fromJson.py
```diff
@@ -0,0 +1,62 @@
...
+        'Effect': 'Allow',
+# Noncompliant@+1 {{This policy is vulnerable to the "Update Login Profile " privilege escalation vector. Remove permissions or restrict the set of resources they apply to.}}
+        'Resource': '*'
+    #   ^^^^^^^^^^^^^^^
+    })
+
+iam.PolicyStatement.from_json({
+        'Action': 'iam:UpdateLoginProfile',
+        'Effect': 'Allow',
+        'Resource': ['*']  # Noncompliant
```
mstachniuk, line 15: Why there is no message after `# Noncompliant`?

mstachniuk (commented):
Please see my comments

python-checks/src/test/resources/checks/cdk/iamPrivilegeEscalation/fromJson.py
nils-werner-sonarsource: I'm usually not repeating the message if it's identical to the previous one. WDYT?

mstachniuk: ahhh cool I didn't know it can be omitted in that case

python-checks/src/main/java/org/sonar/python/checks/cdk/IamPrivilegeEscalationCheck.java
```diff
@@ -41,7 +41,7 @@ public class IamPrivilegeEscalationCheck extends AbstractCdkResourceCheck {
   private static final String ISSUE_MESSAGE_FORMAT = "This policy is vulnerable to the \"%s\" privilege escalation vector. " +
     "Remove permissions or restrict the set of resources they apply to.";
   private static final String SECONDARY_MESSAGE = "Permissions are granted on all resources.";
-  private static final Pattern SENSITIVE_RESOURCE_PATTERN = Pattern.compile("(\\*)|(arn:[^:]*:[^:]*:[^:]*:[^:]*:(role|user|group)/\\*)");
+  private static final Pattern SENSITIVE_RESOURCE_PATTERN = Pattern.compile("(\\*)|(arn:.*(role|user|group)/\\*)");
```
mstachniuk, line 44: I would add a `:` before `(...)` to be more precisely 

```suggestion
  private static final Pattern SENSITIVE_RESOURCE_PATTERN = Pattern.compile("(\\*)|(arn:.*:(role|user|group)/\\*)");
```

python-checks/src/main/java/org/sonar/python/checks/cdk/IamPrivilegeEscalationCheck.java
```diff
@@ -112,7 +112,7 @@ private static void checkPolicyStatement(PolicyStatement policyStatement) {
     }
 
     ExpressionFlow sensitiveAction = getSensitiveExpression(actions, isString(SENSITIVE_ACTIONS));
-    ExpressionFlow sensitiveResource = getSensitiveExpression(resources, isString(SENSITIVE_RESOURCE_PATTERN));
+    ExpressionFlow sensitiveResource = getSensitiveExpression(resources, CdkPredicate.matches(SENSITIVE_RESOURCE_PATTERN));
```
mstachniuk, line 115: static import :-)

mstachniuk (approved):

PR #1251, opened 2022-10-13
SONARPY-1161 Implement for database instance related resources

gyula-sallai-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/cdk/PublicNetworkAccessToCloudResourcesCheck.java
```diff
@@ -19,19 +19,68 @@
...
     );
   }
+
+  private static void checkDatabaseInstance(SubscriptionContext ctx, CallExpression call) {
+    Optional<CdkUtils.ExpressionFlow> vpcSubnets = CdkUtils.getArgument(ctx, call, "vpc_subnets");
+
+    Optional<CdkUtils.ExpressionFlow> subnetType = vpcSubnets.flatMap(flow ->
+      CdkUtils.getCall(flow.getLast(), "aws_cdk.aws_ec2.SubnetSelection")
+        .flatMap(subnetSelection -> CdkUtils.getArgument(flow.ctx(), subnetSelection, "subnet_type")));
+
```
rudy-regazzoni-sonarsource, line 60: Not sure if we can call `flow.getLast()` and expect to have the **CallExpression** in every use case, I would recommend to use `flow.getExpression(Predicate<Expression>)` with proper predicate to ensure you are getting the `aws_cdk.aws_ec2.SubnetSelection(...)` **CallExpression**.

Something like :
`CdkUtils.getCall(flow.getExpression(isFqn("aws_cdk.aws_ec2.SubnetSelection").and(expr -> expr.is(Tree.Kind.CALL_EXPR))))`

python-checks/src/main/java/org/sonar/python/checks/cdk/PublicNetworkAccessToCloudResourcesCheck.java
```diff
@@ -19,19 +19,68 @@
...
+      return;
+    }
+
+    // Raise issue if
+    //  - vpcSubnets is public and publicly_accessible is true
+    //  - vpcSubnets is unknown and publicly_accessible is true
+    //  - vpcSubnets is public and publicly_accessible is not set
+    Optional<CdkUtils.ExpressionFlow> publiclyAccessible = CdkUtils.getArgument(ctx, call, PUBLICLY_ACCESSIBLE_ARG_NAME);
+    publiclyAccessible.ifPresentOrElse(access -> access.addIssueIf(CdkPredicate.isTrue(), MESSAGE),
+      () -> subnetType.filter(isPublicSubnetSelection()).ifPresent(subnets -> subnets.addIssue(MESSAGE)));
```
rudy-regazzoni-sonarsource, line 71: I guess you can write this without storing the optional, purely opinion oriented ;)
```suggestion
  CdkUtils.getArgument(ctx, call, "publicly_accessible")
        .ifPresentOrElse(access -> access.addIssueIf(CdkPredicate.isTrue(), MESSAGE),
        () -> subnetType.filter(isPublicSubnetSelection()).ifPresent(subnets -> subnets.addIssue(MESSAGE)));
```

rudy-regazzoni-sonarsource (commented):
LGTM ! Just see my comment, I believe the `getLast()` usage should be replaced, although I'm not 100% sure if/how we can find an example where it wouldn't work ;)

PR #1255, opened 2022-10-14
SONARPY-1172 S6329 should not raise on CfnDbInstance if the subnet is known to be private

gyula-sallai-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/cdk/PublicNetworkAccessToCloudResourcesCheck.java
```diff
@@ -51,10 +54,58 @@ protected void registerFqnConsumer() {
...
     checkFqn("aws_cdk.aws_rds.CfnDBInstance", (subscriptionContext, callExpression) ->
-      getArgument(subscriptionContext, callExpression, PUBLICLY_ACCESSIBLE_ARG_NAME).ifPresent(
-        argument -> argument.addIssueIf(isTrue(), MESSAGE)
-      )
-    );
+      getArgument(subscriptionContext, callExpression, PUBLICLY_ACCESSIBLE_ARG_NAME).ifPresent(argument -> {
+        if (argument.hasExpression(isTrue()) && !isSafeNetwork(subscriptionContext, callExpression)) {
+          argument.addIssue(MESSAGE);
+        }
+      }));
```
nils-werner-sonarsource, line 61:
```suggestion
      getArgument(subscriptionContext, callExpression, PUBLICLY_ACCESSIBLE_ARG_NAME)
        .filter(argument -> argument.hasExpression(isTrue()) && !isSafeNetwork(subscriptionContext, callExpression))
        .ifPresent(argument -> argument.addIssue(MESSAGE)));
```
Instead of having a condition inside the ifPresent block you can have it as a filter.

python-checks/src/main/java/org/sonar/python/checks/cdk/PublicNetworkAccessToCloudResourcesCheck.java
```diff
@@ -51,10 +54,58 @@ protected void registerFqnConsumer() {
...
+      .map(QualifiedExpression.class::cast);
+
+    if (subnetGroupName.isEmpty()) {
+      return false;
+    }
+
+    QualifiedExpression qualifiedExpression = subnetGroupName.get();
+    if (!"ref".equals(qualifiedExpression.name().name())) {
+      return false;
+    }
```
nils-werner-sonarsource, line 76: This only covers the case the SubnetGroup object is referenced. What about
```
rds.CfnDBInstance(
    publicly_accessible=True,
    db_subnet_group_name=rds.CfnDBSubnetGroup(
        subnet_ids=ec2.SubnetType.PUBLIC
    )
)
```
Currently, the implementation is not detecting this simple case.

python-checks/src/main/java/org/sonar/python/checks/cdk/PublicNetworkAccessToCloudResourcesCheck.java
```diff
@@ -51,10 +54,58 @@ protected void registerFqnConsumer() {
...
+
+    ExpressionFlow subnetIdVpcQualifier = ExpressionFlow.build(subscriptionContext, subnetIds.get().qualifier());
+    Optional<CallExpression> vpcSelectSubnetCall = subnetIdVpcQualifier
+      .getExpression(isCallExpression())
+      .map(CallExpression.class::cast);
+
+    if (vpcSelectSubnetCall.isPresent()) {
+      return getArgument(subscriptionContext, vpcSelectSubnetCall.get(), "subnet_type")
+        .map(subnetType -> subnetType.hasExpression(isFqn("aws_cdk.aws_ec2.SubnetType.PUBLIC").negate()))
+        .orElse(false);
```
nils-werner-sonarsource, line 101: As I understand the ticket we should not raise an issue on this case:
```
rds_subnet_group_unknown_with_subnet_ids = rds.CfnDBSubnetGroup(
    subnet_ids=ec2.SubnetType.PRIVATE_WITH_NAT
)

rds.CfnDBInstance(
    publicly_accessible=True,
    db_subnet_group_name=rds_subnet_group_unknown_with_subnet_ids.ref
)
```
But we do. So I think is logic is incorrect but due to the Optional magic, we can not see it on the coverage.

python-checks/src/test/resources/checks/cdk/publicNetworkAccessDatabaseInstance.py
```diff
@@ -68,4 +68,34 @@
...
+)
+
+rds_subnet_group_unknown_with_subnet_ids_from_variable = rds.CfnDBSubnetGroup(
+    subnet_ids=get_subnets_from_somewhere.subnet_ids
+)
+
+rds.CfnDBInstance(
+    publicly_accessible=True, # Noncompliant
+    db_subnet_group_name=rds_subnet_group_unknown_with_subnet_ids_from_variable.ref
+)
```
nils-werner-sonarsource, line 101: You are missing the test cases I've used above to show that the logic is not complete.

python-checks/src/main/java/org/sonar/python/checks/cdk/PublicNetworkAccessToCloudResourcesCheck.java
```diff
@@ -51,10 +54,58 @@ protected void registerFqnConsumer() {
...
-      )
-    );
+      getArgument(subscriptionContext, callExpression, PUBLICLY_ACCESSIBLE_ARG_NAME).ifPresent(argument -> {
+        if (argument.hasExpression(isTrue()) && !isSafeNetwork(subscriptionContext, callExpression)) {
+          argument.addIssue(MESSAGE);
+        }
+      }));
+  }
+
+  private static boolean isSafeNetwork(SubscriptionContext subscriptionContext, CallExpression callExpression) {
```
nils-werner-sonarsource, line 64: This method is much too big and hard to follow. I would split it into smaller methods that can describe the subtask by method name.

nils-werner-sonarsource (changes requested):
It seems the implemented logic is not following the detection logic. You missed testing the case when `subnet_ids` is the safe like `aws_cdk.aws_ec2.SubnetType.PRIVATE_WITH_NAT`

gyula-sallai-sonarsource:
After discussing with the bubble, I am leaving this ticket for now: I am already working on another sprint, and this ticket is a good candidate for a later hardening, it’s not vital to have it merged ASAP. I am closing this PR.

PR #1257, opened 2022-10-18
SONARPY-1186 S6317 should not on any wildcard identifier in resources

nils-werner-sonarsource:


mstachniuk (approved):
LGTM. This is why i suggested in previous review https://github.com/SonarSource/sonar-python/pull/1245#discussion_r994328001 to implement it as non Regex, because using Regex is always one more problem to solve ;-)

PR #1259, opened 2022-10-18
SONARPY-1088 Add quick fixes for S2316 (BackticksUsageCheck)

mstachniuk:


python-checks/src/test/java/org/sonar/python/checks/BackticksUsageCheckTest.java
```diff
@@ -20,13 +20,24 @@
...
+  @Test
+  public void testQuickFix() {
+    String codeWithIssue = "def foo():\n" +
+            "    `num`\n" +
+            "    foo()";
+    String codeFixed = "def foo():\n" +
+            "    repr(num)\n" +
+            "    foo()";
+    PythonQuickFixVerifier.verify(new BackticksUsageCheck(), codeWithIssue, codeFixed);
+  }
```
nils-werner-sonarsource, line 42: Due to multiline problems we had in the past you could add a test like
```
print(`1
+2`)
```
which should become
```
print(repr(1
+2))
```

nils-werner-sonarsource (approved):
LGTM! Please see my comment.

PR #1261, opened 2022-10-19
SONARPY-1176 Make use of the PolicyStatement data class for applicable CDK-related checks

nils-werner-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/cdk/IamPrivilegeEscalationCheck.java
```diff
@@ -86,28 +82,14 @@ public class IamPrivilegeEscalationCheck extends AbstractCdkResourceCheck {
...
+  protected void checkAllowingPolicyStatement(PolicyStatement policyStatement) {
     ExpressionFlow actions = policyStatement.actions();
     ExpressionFlow resources = policyStatement.resources();
 
     if (actions == null
       || resources == null
       || policyStatement.principals() != null
       || policyStatement.conditions() != null
-      || CdkIamUtils.hasNotAllowEffect(policyStatement.effect())
     ) {
```
mstachniuk, line 93: I'm not sure, but probably there should be `!hasAllowEffect()` in if condition (replacement for `CdkIamUtils.hasNotAllowEffect(policyStatement.effect()`). WDYT?

python-checks/src/test/resources/checks/cdk/iamPolicyPublicAccessCheck.py
```diff
@@ -5,7 +5,7 @@
 iam.PolicyStatement(
   sid="AllowAnyPrincipal",
   effect=iam.Effect.ALLOW, # {{Access is set to "ALLOW" here.}}
-        #^^^^^^^^^^^^^^^^>
+# ^^^^^^^^^^^^^^^^^^^^^^^>
```
mstachniuk, line 8: Is it expected that refactoring change this behavior?

python-checks/src/main/java/org/sonar/python/checks/cdk/AbstractIamPolicyStatementCheck.java
```diff
@@ -28,54 +28,45 @@
 import org.sonar.plugins.python.api.tree.CallExpression;
 import org.sonar.plugins.python.api.tree.DictionaryLiteral;
 import org.sonar.plugins.python.api.tree.Expression;
-import org.sonar.plugins.python.api.tree.ListLiteral;
 
 import static org.sonar.python.checks.cdk.CdkPredicate.isFqn;
-import static org.sonar.python.checks.cdk.CdkPredicate.isString;
 
-public class CdkIamUtils {
+public abstract class AbstractIamPolicyStatementCheck extends AbstractCdkResourceCheck {
```
mstachniuk, line 34: The direction of changes and using abstract class looks nice, especially the rule implementation - small clean code. However:
1. Have you considered using the Util class instead of the abstract? (Delegation over inheritance)
2. Currently this class mix protected and public static methods, so I'm a little bit confused. Is it still Util like class or the class I should extend?

It's not a pushback from my side, I'm just challenging the current shape and usage of the class.

mstachniuk (commented):
Please look on my comments

python-checks/src/main/java/org/sonar/python/checks/cdk/IamPrivilegeEscalationCheck.java
nils-werner-sonarsource: It's already been checked by the parent class.

python-checks/src/test/resources/checks/cdk/iamPolicyPublicAccessCheck.py
nils-werner-sonarsource: This is expected. Normally, we highlight a complete assignment, or in this case a named argument. Eg. highlighting `true` may not be enough. However, in cases we have lists or directories, highlighting the complete argument with name could be too much. Unfortunately, we are not consistent here and we have to revisit the rules from time to time. 
As a resume from this comment, I will create a standard for highlighting which we can discuss within the bubble.

python-checks/src/main/java/org/sonar/python/checks/cdk/ResourceAccessPolicyCheck.java
```diff
@@ -19,75 +19,45 @@
...
+  private static boolean hasOnlyKmsActions(ExpressionFlow actions) {
+    return getSensitiveExpression(actions, notStartsWith("kms:")) == null;
   }
 
-  private static boolean hasOnlyKmsActions(SubscriptionContext ctx, ListLiteral actions) {
-    return CdkUtils.getListElements(ctx, actions)
-      .stream()
-      .allMatch(flow -> flow.hasExpression(CdkPredicate.startsWith("kms:")));
+  public static Predicate<Expression> notStartsWith(String expected) {
+    return expression -> CdkUtils.getString(expression).filter(str -> !str.toLowerCase(Locale.ROOT).startsWith(expected)).isPresent();
```
mstachniuk, line 56: long line

mstachniuk (approved):
LGTM, please check the comments

PR #1262, opened 2022-10-19
SONARPY-1188 S5717 should not provide a quick fix if the default value is not trivial

nils-werner-sonarsource:


mstachniuk (approved):
LGTM. Could you provide missing test coverage?

PR #1263, opened 2022-10-20
SONARPY-1076 S2710: Suggest alternative quick fix to rename the first parameter

mstachniuk:


python-checks/src/test/java/org/sonar/python/checks/utils/CodeTestUtils.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package org.sonar.python.checks.utils;
+
+import org.sonar.api.internal.apachecommons.lang.StringUtils;
+
+public class CodeTestUtils {
+
+  public static String code(String... lines) {
+    return StringUtils.join(lines, '\n');
+  }
```
nils-werner-sonarsource, line 28: :smile:

mstachniuk: Still not perfect, but I was inspired by `PythonTestUtils.parse(String... lines)`

PR #1270, opened 2022-10-26
SONARPY-1082 Add quick fixes for S139 (TrailingCommentCheck)

mstachniuk:


python-checks/src/main/java/org/sonar/python/checks/TrailingCommentCheck.java
```diff
@@ -41,23 +46,56 @@ public class TrailingCommentCheck extends PythonSubscriptionCheck {
...
       }
-      previousTokenLine = pyToken.line();
+      previousTokenLine = token.line();
     });
   }
+
+  private void addQuickFix(IssueWithQuickFix issue, Token commentToken) {
+    String line = lines.get(commentToken.line() - 1);
+    String indent = calculateIndent(line);
+    PythonTextEdit insert = new PythonTextEdit(
```
nils-werner-sonarsource, line 77: Maybe you can choose a more expressive name like `insertComment` as well as `removeTrailingComment` as PythonTextEdit names.

python-checks/src/main/java/org/sonar/python/checks/TrailingCommentCheck.java
```diff
@@ -41,23 +46,56 @@ public class TrailingCommentCheck extends PythonSubscriptionCheck {
...
+      indent + commentToken.value() + "\n",
+      commentToken.line(), 0,
+      commentToken.line(), 0);
```
nils-werner-sonarsource, line 80:
```suggestion
    PythonTextEdit insert = PythonTextEdit.insertAtPosition(commentToken.line(), 0, indent + commentToken.value() + "\n");
```
`PythonTextEdit.insertAtPosition` is private right now, but could reduce the repetition here if made public.

nils-werner-sonarsource (approved):
All in all, it looks good. I have added a few comments that on the one hand make the check faster again. Also, methods can be used in two places, which reduces the overhead.
Besides, I would like to ask again to choose a little bit more expressive commit messages. These can help with the review because they show the development process of the implementation in a better way.

PR #1275, opened 2022-10-27
SONARPY-1091 Add quick fix for S4144 (DuplicatedMethodImplementationC…

mstachniuk:
…heck)

### DIFF
--- a/python-checks/src/main/java/org/sonar/python/checks/DuplicatedMethodImplementationCheck.java
+++ b/python-checks/src/main/java/org/sonar/python/checks/DuplicatedMethodImplementationCheck.java
@@ -30,11 +30,15 @@ import org.sonar.plugins.python.api.tree.FunctionDef;
 import org.sonar.plugins.python.api.tree.Statement;
 import org.sonar.plugins.python.api.tree.StatementList;
 import org.sonar.plugins.python.api.tree.Tree;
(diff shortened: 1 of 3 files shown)


python-checks/src/main/java/org/sonar/python/checks/DuplicatedMethodImplementationCheck.java
```diff
@@ -83,6 +93,34 @@ private static boolean isOnASingleLine(StatementList statementList, boolean hasD
...
+        return;
+      }
+    }
+    boolean containsReturnStatement = originalMethod.body().statements().stream()
+      .anyMatch(s -> s.is(Tree.Kind.RETURN_STMT));
+    String replacementText = "";
+    if (containsReturnStatement) {
+      replacementText = replacementText + "return ";
+    }
+    replacementText = replacementText + originalMethod.name().name() + "()";
```
nils-werner-sonarsource, line 113: You can not call a method like this. You have to call the method on a receiver. In case of a non-static method, you should use the self parameter `self.foo()`. In case of a class method you can do a static call like `Bar.foo()`.

nils-werner-sonarsource (changes requested):
Your quick fix is producing invalid code. Method calls in python expecting a receiver. Please see my comments.

nils-werner-sonarsource (approved):
LGTM!

PR #1279, opened 2022-11-03
SONARPY-1191 The quick fix of S1854 (DeadStoreCheck) should remove indent before removed line

nils-werner-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/DeadStoreCheck.java
```diff
@@ -157,60 +166,81 @@ private static boolean isFunctionDeclarationSymbol(Symbol symbol) {
...
+      if (separator != null) {
+        return separator;
+      }
     }
+    return tree.lastToken();
+  }
 
-    Token currentFirstToken = currentTree.firstToken();
-    int indexOfCurrentTree = childrenOfParent.indexOf(currentTree);
+  private static boolean hasPreviousElementOnLine(Token firstTokenOfElement, @Nullable Tree previousTree) {
```
guillaume-dequenne-sonarsource, line 214: This is purely personal preference, but I have a feeling the `removeStatement` method is a bit easier to understand when I inline the `hasPreviousElementOnLine` and `hasNextElementOnLine` methods, as they are fairly simple. Feel free to disregard, as I understand that others may feel the exact opposite way 😄.

guillaume-dequenne-sonarsource (approved):
Looks good! I left a few comments but they are mostly stylistic suggestions.

As said in one of those, I think it would be great to have a utility class that handle stuff like `removeStatement` separately, so that it's easier to implement quick fixes in the long run, and check classes don't get bloated with quick fix logic.

PR #1286, opened 2022-11-14
SONARPY-1194 Serialize and cache the project level symbol table

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/plugins/python/api/LocationInFile.java
```diff
@@ -53,4 +64,28 @@ public int endLine() {
...
+
+  public DescriptorsProtos.LocationInFile toProtobuf() {
+    return DescriptorsProtos.LocationInFile.newBuilder()
+      .setFileId(fileId)
+      .setStartLine(startLine)
+      .setStartLineOffset(startLineOffset)
+      .setEndLine(endLine)
+      .setEndLineOffset(endLineOffset)
+      .build();
+  }
```
gyula-sallai-sonarsource, line 77: _subjective_ In the other analyzers, we prefer protobuf-related stuff to be in a separate class, I think it would make sense do it here as well, to avoid having as much boilerplate in the classes themselves.

python-frontend/src/main/java/org/sonar/plugins/python/api/caching/PythonReadCache.java
```diff
@@ -0,0 +1,32 @@
...
+import java.io.InputStream;
+import javax.annotation.CheckForNull;
+
+public interface PythonReadCache {
+  InputStream read(String var1);
+
+  @CheckForNull
+  byte[] readBytes(String var1);
+
+  boolean contains(String var1);
```
gyula-sallai-sonarsource, line 31: It would be nice to have some documentation on these, just as we have them in SonarJava (for example notes on whose responsibility it is to close the stream).

python-frontend/src/main/java/org/sonar/plugins/python/api/caching/PythonReadCache.java
```diff
@@ -0,0 +1,32 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.plugins.python.api.caching;
+
+import java.io.InputStream;
+import javax.annotation.CheckForNull;
+
+public interface PythonReadCache {
+  InputStream read(String var1);
```
gyula-sallai-sonarsource, line 26:
```suggestion
  InputStream read(String key);
```

Same below.

python-frontend/src/main/java/org/sonar/plugins/python/api/caching/PythonWriteCache.java
```diff
@@ -0,0 +1,30 @@
...
+package org.sonar.plugins.python.api.caching;
+
+import java.io.InputStream;
+
+public interface PythonWriteCache {
+  void write(String var1, InputStream var2);
+
+  void write(String var1, byte[] var2);
+
+  void copyFromPrevious(String var1);
```
gyula-sallai-sonarsource, line 29: Same as above, both with the comments (when does each method declared here throw) and the variable names.

python-frontend/src/main/java/org/sonar/plugins/python/api/LocationInFile.java
```diff
@@ -53,4 +64,28 @@ public int endLine() {
...
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+    LocationInFile that = (LocationInFile) o;
+    return startLine == that.startLine && startLineOffset == that.startLineOffset && endLine == that.endLine && endLineOffset == that.endLineOffset && fileId.equals(that.fileId);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(fileId, startLine, startLineOffset, endLine, endLineOffset);
+  }
```
gyula-sallai-sonarsource, line 90: As discussed: I think it would be better to remove these `equals` implementations that are currently only used for testing.

python-frontend/src/main/java/org/sonar/python/caching/CacheContextImpl.java
```diff
@@ -0,0 +1,61 @@
...
+  private PythonWriteCache writeCache;
+  private PythonReadCache readCache;
+
+
+  public CacheContextImpl() {
+    this.isCacheEnabled = false;
+    this.writeCache = null;
+    this.readCache = null;
+  }
+
```
gyula-sallai-sonarsource, line 38: _nitpick_ Extra newline

python-frontend/src/main/java/org/sonar/python/caching/Caching.java
```diff
@@ -0,0 +1,109 @@
...
+    String cacheKey = INPORT_MAP_CACHE_KEY_PREFIX + moduleFqn;
+    cacheContext.getWriteCache().write(cacheKey, importData);
+  }
+
+  public void writeProjectLevelSymbolTableEntry(String moduleFqn, Set<Descriptor> descriptors) {
+    String cacheKey = PROJECT_SYMBOL_TABLE_CACHE_KEY_PREFIX + moduleFqn;
+    cacheContext.getWriteCache().write(cacheKey, moduleDescriptor(descriptors).toByteArray());
+   }
+
+  public Optional<Set<Descriptor>> readProjectLevelSymbolTableEntry(String moduleFqn) {
```
gyula-sallai-sonarsource, line 61: Is there an advantage to `Optional` here instead of returning null?

python-frontend/src/main/java/org/sonar/python/caching/Caching.java
```diff
@@ -0,0 +1,109 @@
...
+    cacheContext.getReadCache().contains(moduleFqn);
+    byte[] bytes = cacheContext.getReadCache().readBytes(moduleFqn);
+    if (bytes == null) {
+      return Optional.empty();
+    }
+    try {
+      return Optional.of(DescriptorUtils.deserializeProtobufDescriptors(bytes));
+    } catch (InvalidProtocolBufferException e) {
+      e.printStackTrace();
```
gyula-sallai-sonarsource, line 70: I think we may want to re-throw an exception here.

python-frontend/src/main/java/org/sonar/python/caching/Caching.java
```diff
@@ -0,0 +1,109 @@
...
+    byte[] bytes = cacheContext.getReadCache().readBytes(cacheKey);
+    if (bytes != null) {
+      cacheContext.getWriteCache().copyFromPrevious(cacheKey);
+      return Optional.of(new HashSet<>(Arrays.asList(new String(bytes, StandardCharsets.UTF_8).split(";"))));
+    } else {
+      return Optional.empty();
+    }
+  }
+
+  public static DescriptorsProtos.ModuleDescriptor moduleDescriptor(Set<Descriptor> descriptors) {
```
gyula-sallai-sonarsource, line 86: Does this need to be public?

python-frontend/src/test/java/org/sonar/python/caching/CachingTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+import static org.sonar.python.PythonTestUtils.parseWithoutSymbols;
+import static org.sonar.python.PythonTestUtils.pythonFile;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.sonar.python.caching.Caching.PROJECT_SYMBOL_TABLE_CACHE_KEY_PREFIX;
+
+public class CachingTest {
+
+  @Test
+  public void testwriteProjectLevelSymbolTableEntry() throws InvalidProtocolBufferException {
```
gyula-sallai-sonarsource, line 43: _minor_ I think we can remove the `test` prefix from the test name.

gyula-sallai-sonarsource (changes requested):
Nice job! I've left some comments (the bigger ones we already discussed)

python-frontend/src/main/java/org/sonar/python/caching/Caching.java
guillaume-dequenne-sonarsource: Actually I think we can log something here and return null. The caller will know the cache entry is missing/corrupted through the null return.

python-frontend/src/main/java/org/sonar/python/caching/Caching.java
guillaume-dequenne-sonarsource: No particular reason to use an optional here, I believe returning null will be simpler indeed, thanks!

python-frontend/src/main/java/org/sonar/python/caching/Caching.java
```diff
@@ -0,0 +1,109 @@
...
+    for (Descriptor descriptor : descriptors) {
+      if (descriptor.kind().equals(Descriptor.Kind.CLASS)) {
+        classDescriptors.add(((ClassDescriptor) descriptor).toProtobuf());
+      } else if (descriptor.kind().equals(Descriptor.Kind.FUNCTION)) {
+        functionDescriptors.add(((FunctionDescriptor) descriptor).toProtobuf());
+      } else if (descriptor.kind().equals(Descriptor.Kind.VARIABLE)) {
+        varDescriptors.add(((VariableDescriptor) descriptor).toProtobuf());
+      } else {
+        ambiguousDescriptors.add(((AmbiguousDescriptor) descriptor).toProtobuf());
+      }
```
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: Using a switch here was a couple more lines (due to the breaks which, although not strictly needed here, I didn't like to do without), so I kind of preferred using a simple `if/else if` construct here. I refactored it slightly to make it a bit more readable. Please let me know if you still prefer a switch, as I don't really have a strong opinion on this.

python-frontend/src/test/java/org/sonar/python/index/ClassDescriptorTest.java
```diff
@@ -43,6 +48,7 @@ public void classDescriptor() {
     FunctionDescriptor fooMethod = ((FunctionDescriptor) classDescriptor.members().iterator().next());
     assertThat(fooMethod.name()).isEqualTo("foo");
     assertThat(fooMethod.fullyQualifiedName()).isEqualTo("package.mod.A.foo");
+    assertClassDescriptors(classDescriptor, fromProtobuf(toProtobuf(classDescriptor)));
```
gyula-sallai-sonarsource, line 51: I am not sure about having this test here: it seems to test `DescriptorsToProtobuf` as opposed to `ClassDescriptor`. I think it's fine for now, but we might want to move this before merging the feature branch.

python-frontend/src/test/java/org/sonar/python/index/FunctionDescriptorTest.java
```diff
@@ -116,6 +131,26 @@ public void method() {
       "class A:",
       "  def foo(self): ...");
     assertThat(functionDescriptor.isInstanceMethod()).isTrue();
+    assertFunctionDescriptors(functionDescriptor, fromProtobuf(toProtobuf(functionDescriptor)));
+  }
+
+  @Test
+  public void protobufSerializationWithoutLocationAndWithAnnotatedReturnType() {
+    // FIXME: Annotated type name is never set for regular project decorators (only Typeshed).
```
gyula-sallai-sonarsource, line 139: Would it make sense to have a ticket number here?

python-frontend/src/main/java/org/sonar/python/index/DescriptorUtils.java
```diff
@@ -224,4 +227,16 @@ private static void setParameterType(FunctionSymbolImpl.ParameterImpl parameter,
     }
     parameter.setDeclaredType(declaredType);
   }
+
+
```
gyula-sallai-sonarsource, line 231: _nitpick_ Excess newline

gyula-sallai-sonarsource (approved):
Looks good to me! I've left very few nitpicky comments, I think it's not worth to relaunch the CI pipeline because of these, we can fix them before merging the feature branch.

python-frontend/src/test/java/org/sonar/python/index/ClassDescriptorTest.java
guillaume-dequenne-sonarsource: It's true that strictly speaking we should have dedicated tests for `DescriptorsToProtobuf`. However I felt this was such as simple way to leverage all our existing descriptor tests almost for free that I didn't bother with it. Especially as this really tests what we want to achieve (what we get after a serialization and deserialization is the same as what we had in input).
Of course we can re-discuss this before we merge the feature branch but indeed in the meantime I also feel this is good for now.

PR #1291, opened 2022-11-17
SONARPY-1201 Add integeration tests for incremental PR analysis

gyula-sallai-sonarsource:


guillaume-dequenne-sonarsource (approved):
Looks good! As discussed, maybe we'd like to cover some other edge cases, but we can revisit once we get these running on the actual implementation!
In the meantime, can you maybe squash your commits together so the commit is already ready for me to rebase my branch against it?

guillaume-dequenne-sonarsource:
Unfortunately, after rebasing my PR on this one, it seems that the integration test is not working as expected.
All files are considered to be `ADDED` no matter what the scenario.
The implementation here is very similar to the one we have in DBD Java, although I'm wondering whether the status of the file in DBD is retrieved through the maven scanner, whereas here we use the generic sonar-scanner?

.cirrus.yml
```diff
@@ -174,11 +174,32 @@ extended_ruling_task:
...
 
+pr_analysis_qa_task:
+  depends_on:
+    - build
+  <<: *ONLY_SONARSOURCE_QA
+  gke_container:
+    <<: *CONTAINER_DEFINITION
+    cpu: 4
+    memory: 8G
+  env:
```
Godin, line 185:
```suggestion
  env:
      GITHUB_TOKEN: VAULT[development/github/token/licenses-ro token]
```

PR #1294, opened 2022-11-28
Compute main files FQNs after setting the project base dir path

guillaume-dequenne-sonarsource:
The computation of fully qualified names requires that `projectBaseDirAbsolutePath` is set, so that the base directory of the project is not considered as part of fully qualified names within the project even when containing an `__init__.py` file.
This PR will fix an inconsistency in FQNs computation that may cause NullPointerException in some cases.


gyula-sallai-sonarsource (approved):
Looks good to me! Thanks for taking care of this!

PR #1297, opened 2022-11-29
SONARPY-1179 Update S5747 (RaiseOutsideExceptCheck) to not report an issue inside a 'except*'

rudy-regazzoni-sonarsource:
Slight modification on the rule : add the new kind for except group.
Added a test to expressly test it.


andreaguarino (approved):
LGTM

PR #1298, opened 2022-11-29
SONARPY-1184 Rule S6468 : ExceptionGroup and BaseExceptionGroup should not be caught in `except*` clauses

rudy-regazzoni-sonarsource:
I didn't find the issue message in RSPEC, so I defined my own, free to be challenged.
I also choose a class/rule name, I'm not really good at this and it's a bit long (ExceptGroupShouldNotCaughtExceptionGroupCheck), feel free to propose something ;)
I generate rule metadata from the PR that is not yet merged, to be updated first once it's done.


python-checks/src/main/java/org/sonar/python/checks/ExceptGroupShouldNotCaughtExceptionGroupCheck.java
```diff
@@ -0,0 +1,68 @@
...
+import org.sonar.plugins.python.api.tree.Name;
+import org.sonar.plugins.python.api.tree.Tree;
+import org.sonar.plugins.python.api.tree.Tuple;
+
+import static org.sonar.plugins.python.api.tree.Tree.Kind.EXCEPT_GROUP_CLAUSE;
+
+@Rule(key = "S6468")
+public class ExceptGroupShouldNotCaughtExceptionGroupCheck extends PythonSubscriptionCheck {
+  private static final Set<String> EXCEPTION_GROUP = Set.of("ExceptionGroup", "BaseExceptionGroup");
+  private static final String MESSAGE = "Exception group cannot be caught by an except*.";
```
andreaguarino, line 36: We usually prefer issue message to be actionable: "Avoid catching {ExceptionGroup|BaseExceptionGroup} exception with 'except*'"

python-checks/src/main/java/org/sonar/python/checks/ExceptGroupShouldNotCaughtExceptionGroupCheck.java
```diff
@@ -0,0 +1,68 @@
...
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(EXCEPT_GROUP_CLAUSE, ctx -> {
+      ExceptClause exceptClause = (ExceptClause) ctx.syntaxNode();
+
+      if (hasAnyExceptionGroup(exceptClause.exception())) {
+        ctx.addIssue(exceptClause.exceptKeyword(), MESSAGE);
```
andreaguarino, line 44: Nitpick: I think we should include `*` in the primary location as well

python-checks/src/main/java/org/sonar/python/checks/ExceptGroupShouldNotCaughtExceptionGroupCheck.java
```diff
@@ -0,0 +1,68 @@
...
+          return true;
+        }
+      }
+      return false;
+    }
+    return false;
+  }
+
+  public boolean isExceptionGroup(Expression exception) {
+    return exception.is(Tree.Kind.NAME) && EXCEPTION_GROUP.contains(((Name) exception).name());
```
andreaguarino, line 66: I think this is fine for the time being, however let's put a "TODO" to change this logic to use type or symbol once we will have it from typeshed.

Also, can you please document the FP we might have when overriding symbol with a unit test?

python-checks/src/test/java/org/sonar/python/checks/ExceptGroupShouldNotCaughtExceptionGroupCheckTest.java
```diff
@@ -0,0 +1,32 @@
+/*
```
andreaguarino, line 1: Because this is a BUG rule, we should add it to `PythonExtendedRuling`

andreaguarino (changes requested):
Nice work, I left some suggestions

python-checks/src/main/java/org/sonar/python/checks/ExceptGroupShouldNotCaughtExceptionGroupCheck.java
rudy-regazzoni-sonarsource: You mean once we will have it from the typeshed version we are going to upgrade in this sprint ?
I will add the TODO + create a ticket to have this fixed.

rudy-regazzoni-sonarsource:
@andreaguarino Code smell about `starToken()` method that may return null, but in our case that can never happen.
How would recommend to "fix" this ?
If I add a defensive check, I will have some uncovered code.
I can also probably hide it with some `Optional.ifPresent`, but I'm not a big fan of this ;)

python-checks/src/main/java/org/sonar/python/checks/ExceptionGroupCheck.java
```diff
@@ -0,0 +1,66 @@
...
+    });
+  }
+
+  private static boolean isExceptionGroup(Expression exception) {
+    // TODO : replace this logic by using type/symbol from typeshed
+    return exception.is(Tree.Kind.NAME) && EXCEPTION_GROUP.contains(((Name) exception).name());
+  }
+
+  private static void raiseIssue(SubscriptionContext ctx, ExceptClause exceptClause, Expression exception) {
+    ctx.addIssue(exceptClause.exceptKeyword(), exceptClause.starToken(), String.format(MESSAGE, ((Name) exception).name()));
```
andreaguarino, line 64: Thinking again about location, I think it would probably be better to raise the issue on the expression `exception`

andreaguarino (approved):
LGTM, just one minor comment about primary location

andreaguarino:
> @andreaguarino Code smell about `starToken()` method that may return null, but in our case that can never happen. How would recommend to "fix" this ? If I add a defensive check, I will have some uncovered code. I can also probably hide it with some `Optional.ifPresent`, but I'm not a big fan of this ;)

Generally speaking, I think we could have accepted this code smell as won't fix. For this specific case, I think it's better to have the primary location on the `exception`

PR #1299, opened 2022-11-29
SONARPY-1217 Update Python API to expose data related to caching

gyula-sallai-sonarsource:


sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonScanner.java
```diff
@@ -142,7 +144,11 @@ public boolean scanFileWithoutParsing(InputFile inputFile) {
       if (!isCheckApplicable(check, fileType)) {
         continue;
       }
-      if (!check.scanWithoutParsing(inputFile)) {
+
+      PythonFile pythonFile = SonarQubePythonFile.create(inputFile);
+      PythonInputFileContext inputFileContext = new PythonInputFileContext(pythonFile, context.fileSystem().workDir(), CacheContextImpl.of(context));
```
guillaume-dequenne-sonarsource, line 149: Skimming through this, I think we can use `indexer.cacheContext()` here, right?

python-frontend/src/main/java/org/sonar/plugins/python/api/PythonInputFileContext.java
```diff
@@ -0,0 +1,51 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.plugins.python.api;
+
+import java.io.File;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nullable;
+import org.sonar.plugins.python.api.caching.CacheContext;
+
+public class PythonInputFileContext {
```
guillaume-dequenne-sonarsource, line 27: To make sure that SonarSecurity is able to write the Typeshed UCFGs even when no Python file changed (in which case we shouldn't do any analysis at all, but this won't be the case right now I believe), we'd need to add a `stubFilesSymbols` method here as well. Wdyt @gyula-sallai-sonarsource and @andreaguarino ?

python-checks-testkit/src/main/java/org/sonar/python/checks/utils/PythonCheckVerifier.java
```diff
@@ -81,7 +81,7 @@ private static MultiFileVerifier createVerifier(List<File> files, PythonCheck ch
     MultiFileVerifier multiFileVerifier = MultiFileVerifier.create(files.get(0).toPath(), UTF_8);
     for (File file : files) {
       PythonVisitorContext context = baseDir != null
-        ? TestPythonVisitorRunner.createContext(file, null, pythonPackageName(file, baseDir.getAbsolutePath()), projectLevelSymbolTable)
+        ? TestPythonVisitorRunner.createContext(file, null, pythonPackageName(file, baseDir.getAbsolutePath()), projectLevelSymbolTable, null)
```
guillaume-dequenne-sonarsource, line 84: Would it make sense to have a dummy implementation of a `CacheContext` here?

python-checks/src/test/java/org/sonar/python/checks/quickfix/PythonQuickFixVerifier.java
```diff
@@ -115,7 +117,7 @@ private static List<PreciseIssue> getIssuesWithQuickFix(PythonCheck check, Strin
 
     PythonVisitorContext visitorContext = new PythonVisitorContext(parse,
       pythonFile, null, "",
-      ProjectLevelSymbolTable.empty());
+      ProjectLevelSymbolTable.empty(), null);
```
guillaume-dequenne-sonarsource, line 120: If you have a dummy implementation for `CacheContext` in the `python-checks-testkit`, I believe you will be able to use it here as well, which would also avoid to pass `null` here.

edit: In one of my next comments, I suggest to have such an implementation in the `frontend` instead, which could be reused here.

python-frontend/src/main/java/org/sonar/plugins/python/api/PythonVisitorContext.java
```diff
@@ -23,53 +23,45 @@
...
 
   private final FileInput rootTree;
-  private final PythonFile pythonFile;
-  private File workingDirectory = null;
   private final RecognitionException parsingException;
   private List<PreciseIssue> issues = new ArrayList<>();
 
-
   public PythonVisitorContext(FileInput rootTree, PythonFile pythonFile, @Nullable File workingDirectory, @Nullable String packageName) {
+    super(pythonFile, workingDirectory, null);
```
guillaume-dequenne-sonarsource, line 40: I don't like that we have to use null here as well. Especially since `CacheContext` has a method `isCacheEnabled`, it feels wrong to me to have it nullable (as then we'd have to check for null, then check for this boolean to be able to use the cache context, which is too much in my opinion).

Maybe an implementation of a dummy cache could exist inside the frontend? Besides, having one here could mean we don't need one in the `python-checks-testkit` and you can use the one from the frontend instead to address my other 2 comments?

guillaume-dequenne-sonarsource (changes requested):
Looks good!
As mentioned in one of the comments, it feels wrong to have a nullable `CacheContext` since we have a method `isCacheEnabled` in it. To me that means that a context for the cache should always be available, even when the cache itself is not.

I'm not sure whether the existing implementation of `CacheContextImpl`/`DummyCache` that lives in the `plugin` can be moved to the `frontend` or if we can have another implementation of `CacheContext` itself that can be used as a placeholder for cases where we don't expect to use the cache, but I feel it would be a bit cleaner that way.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonScanner.java
```diff
@@ -142,7 +144,11 @@ public boolean scanFileWithoutParsing(InputFile inputFile) {
       if (!isCheckApplicable(check, fileType)) {
         continue;
       }
-      if (!check.scanWithoutParsing(inputFile)) {
+
```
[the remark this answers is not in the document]
gyula-sallai-sonarsource: I would prefer to have the empty line here (to have a logical separation), but I agree about the empty line at line 150, I've reverted that.

PR #1301, opened 2022-11-30
SONARPY-1181 S5708 (CaughtExceptionsCheck) should report on except*

rudy-regazzoni-sonarsource:
Small adaptation required to apply the rule. Again I didn't find difference with except use cases.
By default I duplicated all tests with the except* for the sake of completeness with almost no difference, as from the point of view of the code, except and except* are two different Kind for which we apply the same checks.
Not sure if that's a good idea in the end, it may just add noise/make it harder to maintain.


python-checks/src/test/resources/checks/caughtExceptions.py
```diff
@@ -61,3 +61,57 @@ def foo():
...
+
+def except_star():
+  try:
+    ...
+  except* CustomException: # Noncompliant {{Change this expression to be a class deriving from BaseException or a tuple of such classes.}}
+    ...
+  except* (None, BaseException):  # Noncompliant
+    ...
+  except* list():  # Noncompliant
+    ...
```
andreaguarino, line 73: I think we don't need to duplicate everything, only one or two test cases are enough

andreaguarino (approved):
LGTM, Small suggestion before merging

PR #1303, opened 2022-11-30
SONARPY-1183 S5713 (ChildAndParentExceptionCaughtCheck) should report on except*

rudy-regazzoni-sonarsource:
As usual, slightly adapted to make it work with Python 3.11.
Duplicated test code, I didn't normalize the indent of the test file to focus only on real modification, but I can still do it If you think it's a good idea.


andreaguarino (approved):
LGTM

PR #1304, opened 2022-12-02
Avoid trying to save descriptors to cache when their computation failed

guillaume-dequenne-sonarsource:


sonar-python-plugin/src/test/java/org/sonar/plugins/python/indexer/SonarQubePythonIndexerTest.java
```diff
@@ -326,6 +326,17 @@ public void test_regular_scan_when_scan_without_parsing_fails() {
...
 
+  @Test
+  public void test_no_data_in_cache_for_parse_error() {
+    file1 = createInputFile(baseDir, "parse_error.py", InputFile.Status.ADDED, InputFile.Type.MAIN);
+
+    List<InputFile> inputFiles = new ArrayList<>(List.of(file1));
+
+    pythonIndexer = new SonarQubePythonIndexer(inputFiles, cacheContext);
+    pythonIndexer.buildOnce(context);
+    assertThat(writeCache.getData().get(projectSymbolTableCacheKey("moduleKey:parse_error.py"))).isNull();
```
gyula-sallai-sonarsource, line 337: Am I correct that the issue was that we did put `null` into the cache? I think it would make sense to have an assertion that checks that the cache does not contain `null` for this file.

gyula-sallai-sonarsource (approved):
Looks good to me, I've left one small suggestion regarding the test.

PR #1305, opened 2022-12-02
SONARPY-1193 FP on rule S6304: 'Granting access to all resources' should not be raised on actions without resource-level permissions

rudy-regazzoni-sonarsource:
I made some assumption/choice, probably some are wrong, please let me know :
- I added the list of invalid key as a Json file, in src/main/resources/..., I had also to increase a bit the maximum requireFilesSize
- I added a FileResourceUtils to help load this file (in the end this class is not really big)
- I handled the case where this file cannot be loaded for some reasons, in that case we log an error and "disable" the rule. I didn't add test on this part as I'm not sure if that's a good thing and we may decide to go in a different way. It's probably better to throw this error upward and fail/stop the execution, but by default I still put in place a defensive approach
- I load the JSON file in static, which imply some more operations during load of the class by the classloader (loading a JSON array of 9000 values and put them in a HashSet). Another approach could be to lazy load at the very last moment we need it (check some detected action value), and OFC still load this file only once

In any case there will be decision and changes to make :D please let me know what you think.


andreaguarino (changes requested):
Good work, I left some comments. Feel free to call me if you want to discuss more about those

andreaguarino:
Also , it looks like there is one difference in ruling results

rudy-regazzoni-sonarsource:
For the resource loading, I still put some check / specific exception.
Not sure if it's the right way, as technically it will never happen, OFC in result it show some uncovered code.

Other solution I see would be to have a try/catch in `loadResourceAsSet`, and do nothing in the catch (or eventually throw a Runtime like exception), I didn't think comfortable doing so as it seems a bit counterintuitive code style (catch any error and ignore it).
WDYT ?

python-checks/src/main/java/org/sonar/python/checks/utils/FileResourceUtils.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.Charset;
+import java.util.HashSet;
+import java.util.Set;
+
+public class FileResourceUtils {
```
andreaguarino, line 30: Do we need a separate utils? I think this is mostly useful for `ResourceAccessPolicyCheck`, so I would suggest to move this method there.

python-checks/src/main/java/org/sonar/python/checks/cdk/ResourceAccessPolicyCheck.java
```diff
@@ -34,27 +37,37 @@ public class ResourceAccessPolicyCheck extends AbstractIamPolicyStatementCheck {
 
   private static final String MESSAGE = "Make sure granting access to all resources is safe here.";
   private static final String SECONDARY_MESSAGE = "Related effect";
+  private static final Set<String> SENSITIVE_AWS_ACTIONS;
+  private static final String FILEPATH_SENSITIVE_AWS_ACTIONS = "/org/sonar/python/checks/cdk/sensitiveAwsActions.txt";
```
andreaguarino, line 41: I think `sensitiveAwsActions.txt` should be sufficient if you inline `FileResourceUtils.loadResourceAsSet` method

python-checks/src/main/java/org/sonar/python/checks/utils/FileResourceUtils.java
```diff
@@ -0,0 +1,58 @@
...
+        String line;
+        while((line = br.readLine()) != null) {
+          result.add(line);
+        }
+        return result;
+      }
+    }
+  }
+
+  public static class MissingResourceException extends RuntimeException {
```
andreaguarino, line 50: As discussed, there is no need of having this exception

python-checks/src/main/java/org/sonar/python/checks/cdk/ResourceAccessPolicyCheck.java
```diff
@@ -34,27 +37,37 @@ public class ResourceAccessPolicyCheck extends AbstractIamPolicyStatementCheck {
...
+
+  static {
+    try {
+      SENSITIVE_AWS_ACTIONS = FileResourceUtils.loadResourceAsSet(FILEPATH_SENSITIVE_AWS_ACTIONS, StandardCharsets.UTF_8);
+    } catch (IOException e) {
+      throw new FileResourceUtils.MissingResourceException(e);
+    }
```
andreaguarino, line 48: As discussed, we can instead have `SENSITIVE_AWS_ACTIONS  = emptySet()` when an exception is thrown, to avoid halting the scanner analysis

andreaguarino (changes requested):

rudy-regazzoni-sonarsource:
I restored the principle of handling the case of missing resources.
To be able to test it properly, I had to make some compromise :
- switched attributes from `private static final` to `package-private non-static non-final` (I'm supposing the check object is being instantiated only once, I didn't put any guardrail about it yet)
- have an init method being call on the initialize
- switched some method to non-static

Let me know what you think, not totally sure it's worth, still we have a nice code coverage now :D

python-checks/src/main/java/org/sonar/python/checks/cdk/ResourceAccessPolicyCheck.java
```diff
@@ -32,29 +44,64 @@
...
+  @Override
+  public void initialize(SubscriptionCheck.Context context) {
+    super.initialize(context);
+    init();
+  }
+
+  static List<String> loadResourceAsSet(String resourceName, Charset charset) throws IOException {
+    try (InputStream is = ResourceAccessPolicyCheck.class.getResourceAsStream(resourceName)) {
+      if (is == null) {
+        throw new IOException("Cannot find resource file '" + resourceName + "'");
```
andreaguarino, line 71: Minor: I feel like we can avoid throwing this exception. We could instead directly call a method:
```java
void handleMissingResource() {
      sensitiveAwsActions = new HashSet<>();
      LOG.error("Couldn't load resource '" + resourceNameSensitiveAwsActions + "', rule [S6304] ResourceAccessPolicyCheck will be disabled.", e);
}
```

Anyway, feel free to ignore this comment if you prefer otherwise

andreaguarino (approved):
LGTM, I left some minor comments!

python-checks/src/main/java/org/sonar/python/checks/cdk/ResourceAccessPolicyCheck.java
rudy-regazzoni-sonarsource: Even if I remove this throw exception, there is still possibles `IOException` with the `instanciation` of the `InputStreamReader` / `BufferedReader`, and also with the `readLine()` call.

andreaguarino: sure, I'm fine with `loadResourceAsSet` still raising the exception for `BufferedReader` and `readLine()` calls. However it feels a bit odd to me creating a new exception when the `InputStream` is null, just to be then caught at the call site a few lines above

andreaguarino: Btw I just noticed `loadResourceAsSet` returns a `List`. Maybe we can just rename it `loadResource` to avoid confusion

PR #1310, opened 2022-12-07
SONARPY-1222 Disable cache to prepare for SQ 9.8 release

guillaume-dequenne-sonarsource:
Goal is to disable caching for SonarPython 3.21 which will be integrated into SonarQube 9.8 release.
This feature could break SonarSecurity and Dataflow bug detection for Python, since the version of these analyzers to be integrated in 9.8 do not implement the required caching APIs.


andreaguarino (approved):
LGTM

PR #1312, opened 2022-12-07
SONARPY-1221 Invalidate cache when the analyzer version changes

guillaume-dequenne-sonarsource:
Two comments on this PR:
* I left the `cacheContext` as a constructor parameter to `SonarQubePythonIndexer` although it is no longer necessary (since it's only built from the `SensorContext` which is now a constructor parameter as well). The reason for that is to not hinder the testability of the `SonarQubePythonIndexer`, in which we currently inject some test cache contexts. I'm not 100% sure it's the right way to do this, though.
* In unit tests, we don't have an implementation version available. I used the `unknown` default value as if it were a valid value. The behavior of the cache versioning itself is tested in the ITs where the implementation version is actually available.

### DIFF
--- a/its/ruling/src/test/java/org/sonar/python/it/PythonPrAnalysisTest.java
+++ b/its/ruling/src/test/java/org/sonar/python/it/PythonPrAnalysisTest.java
@@ -53,7 +53,6 @@ public class PythonPrAnalysisTest {
 
   private static final String PR_ANALYSIS_PROJECT_KEY = "prAnalysis";
   private static final String INCREMENTAL_ANALYSIS_PROFILE = "incrementalPrAnalysis";
-  private static final String EXPECTED_CACHE_VERSION = "3.21-SNAPSHOT";
 
   @Rule
   public TemporaryFolder temporaryFolder = new TemporaryFolder();
@@ -148,7 +147,7 @@ public class PythonPrAnalysisTest {
     assertThat(result.getLogs())
       .contains(expectedRecomputedLog)
       .contains(expectedRegularAnalysisLog)
-      .contains(String.format("Cache version still up to date: \"%s\".", EXPECTED_CACHE_VERSION));
+      .contains(String.format("Cache version still up to date"));
   }
 
   private void analyzeAndAssertBaseCommit(File tempFile, File litsDifferencesFile) throws IOException {
--- a/its/ruling/src/test/java/org/sonar/python/it/PythonPrAnalysisTest.java
+++ b/its/ruling/src/test/java/org/sonar/python/it/PythonPrAnalysisTest.java
@@ -53,6 +53,7 @@ public class PythonPrAnalysisTest {
 
   private static final String PR_ANALYSIS_PROJECT_KEY = "prAnalysis";
   private static final String INCREMENTAL_ANALYSIS_PROFILE = "incrementalPrAnalysis";
+  private static final String EXPECTED_CACHE_VERSION = "3.21-SNAPSHOT";
 
   @Rule
   public TemporaryFolder temporaryFolder = new TemporaryFolder();
@@ -146,7 +147,8 @@ public class PythonPrAnalysisTest {
 
     assertThat(result.getLogs())
       .contains(expectedRecomputedLog)
-      .contains(expectedRegularAnalysisLog);
+      .contains(expectedRegularAnalysisLog)
+      .contains(String.format("Cache version still up to date: \"%s\".", EXPECTED_CACHE_VERSION));
   }
 
   private void analyzeAndAssertBaseCommit(File tempFile, File litsDifferencesFile) throws IOException {
@@ -177,7 +179,8 @@ public class PythonPrAnalysisTest {
       .setProperty("sonar.cpd.exclusions", "**/*")
       .setProperty("sonar.lits.differences", litsDifferencesFile.getAbsolutePath())
       .setProperty("sonar.internal.analysis.failFast", "true")
-      .setEnvironmentVariable("SONAR_RUNNER_OPTS", "-Xmx2000m");
+      .setDebugLogs(true)
+      .setEnvironmentVariable("SONAR_SCANNER_OPTS", "-Xdebug");
   }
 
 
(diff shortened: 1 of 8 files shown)


its/ruling/src/test/java/org/sonar/python/it/PythonPrAnalysisTest.java
```diff
@@ -146,7 +147,8 @@ private void assertPrAnalysisLogs(BuildResult result) {
 
     assertThat(result.getLogs())
       .contains(expectedRecomputedLog)
-      .contains(expectedRegularAnalysisLog);
+      .contains(expectedRegularAnalysisLog)
+      .contains(String.format("Cache version still up to date: \"%s\".", EXPECTED_CACHE_VERSION));
```
gyula-sallai-sonarsource, line 151: This does not feel very scalable - we would have to update the constant after reach release and it does not work on the CI due to the changed build number. I would recommend to only search for a string starting with ""Cache version still up to date:". In this test we are only interested in knowing that the cache is still fresh, so we do not need the version number.

its/ruling/src/test/java/org/sonar/python/it/PythonPrAnalysisTest.java
```diff
@@ -177,7 +179,8 @@ private SonarScanner prepareScanner(File path, String projectKey, String scenari
       .setProperty("sonar.cpd.exclusions", "**/*")
       .setProperty("sonar.lits.differences", litsDifferencesFile.getAbsolutePath())
       .setProperty("sonar.internal.analysis.failFast", "true")
-      .setEnvironmentVariable("SONAR_RUNNER_OPTS", "-Xmx2000m");
+      .setDebugLogs(true)
+      .setEnvironmentVariable("SONAR_SCANNER_OPTS", "-Xdebug");
```
gyula-sallai-sonarsource, line 183: _minor_ Do we need both?

gyula-sallai-sonarsource (changes requested):
Looks good in general, but please note the failure on the CI.

its/ruling/src/test/java/org/sonar/python/it/PythonPrAnalysisTest.java
guillaume-dequenne-sonarsource: No we don't, and in fact I should've kept the `.setEnvironmentVariable("SONAR_RUNNER_OPTS", "-Xmx2000m");` option. I mixed up the two when trying various ways to enable debug logs.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/indexer/SonarQubePythonIndexer.java
```diff
@@ -54,21 +56,21 @@ public class SonarQubePythonIndexer extends PythonIndexer {
...
+  public SonarQubePythonIndexer(List<InputFile> inputFiles, CacheContext cacheContext, SensorContext context) {
+    this.projectBaseDirAbsolutePath = context.fileSystem().baseDir().getAbsolutePath();
+    this.caching = new Caching(cacheContext, getCacheVersion(context));
     inputFiles.forEach(f -> {
       if (f.type().equals(InputFile.Type.MAIN)) {
         mainFiles.add(f);
       } else {
         testFiles.add(f);
       }
     });
```
andreaguarino, line 68: I would suggest to avoid the foreach at line 69 and move the `put` operation while iterating over `inputFiles`
```suggestion
    inputFiles.forEach(f -> {
      if (f.type().equals(InputFile.Type.MAIN)) {
        mainFiles.add(f);
        inputFileToFQN.put(f, SymbolUtils.fullyQualifiedModuleName(packageName(f), f.filename()));
      } else {
        testFiles.add(f);
      }
    });
```

sonar-python-plugin/src/main/java/org/sonar/plugins/python/indexer/SonarQubePythonIndexer.java
```diff
@@ -183,4 +187,18 @@ public boolean canBeScannedWithoutParsing(InputFile inputFile) {
...
+
+  private static String getCacheVersion(SensorContext context) {
+    String implementationVersion = getImplementationVersion(SonarQubePythonIndexer.class);
+    return context.config().get(PYTHON_VERSION_KEY).map(v -> implementationVersion + ";" + v).orElse(implementationVersion);
+  }
+
+  private static String getImplementationVersion(Class<?> cls) {
+    String implementationVersion = cls.getPackage().getImplementationVersion();
+    if (implementationVersion == null) {
+      LOG.warn("Implementation version of the Python plugin not found. Cached data may not be invalidated properly.");
```
andreaguarino, line 199: Maybe we could add "Cached data may not be invalidated properly , hence analysis results may be inaccurate."

sonar-python-plugin/src/main/java/org/sonar/plugins/python/indexer/SonarQubePythonIndexer.java
```diff
@@ -183,4 +187,18 @@ public boolean canBeScannedWithoutParsing(InputFile inputFile) {
...
+      return "unknown";
```
andreaguarino, line 200: Nitpick: what about `unknownPluginVersion` (or `unknown-plugin-version`)?

its/ruling/src/test/java/org/sonar/python/it/PythonPrAnalysisTest.java
```diff
@@ -180,6 +178,7 @@ private SonarScanner prepareScanner(File path, String projectKey, String scenari
       .setProperty("sonar.cpd.exclusions", "**/*")
       .setProperty("sonar.lits.differences", litsDifferencesFile.getAbsolutePath())
       .setProperty("sonar.internal.analysis.failFast", "true")
+      .setDebugLogs(true)
```
andreaguarino, line 181: As discussed, we can remove this since it's not used anymore

sonar-python-plugin/src/test/java/org/sonar/plugins/python/indexer/SonarQubePythonIndexerTest.java
```diff
@@ -250,14 +255,40 @@ public void test_no_file_modified_missing_descriptors() {
       .contains("2/2 source files have been analyzed");
   }
```
andreaguarino, line 257: As discussed, we can have a test asserting that a change in `sonar.python.version` invalidates the cache

andreaguarino (changes requested):
Overall looks good, I left some minor comments. The only important change is about adding test(s) to assert that changing the Python version invalidates the cache

andreaguarino (approved):
LGTM

PR #1314, opened 2022-12-08
SONARPY-872 S930 should not raise on updated jinja2 filters api

andreaguarino:


rudy-regazzoni-sonarsource (approved):
LGTM !

I downloaded the last version of jinja2 and I can see several others differences
- new methods : `do_min`, `do_max`, ect...
- parameters changing name for existing methods : `do_xmlattr(_eval_ctx)` -> `do_xmlattr(eval_ctx)`, ect...
- new parameters for existing methods : `do_dictsort` has new parameter `reverse`, ect...

I guess it's not really our job to update our stubs according to the last versions of each library (aka maintaining typeshed), but I see we still update on demand when the community report some FP.
So our current compromise is to try to stay up-to-date on our typeshed version + react on user feedback ?

andreaguarino:
> I guess it's not really our job to update our stubs according to the last versions of each library (aka maintaining typeshed), but I see we still update on demand when the community report some FP.
So our current compromise is to try to stay up-to-date on our typeshed version + react on user feedback ?

Indeed, this is a compromise to only fix what have been reported by users. However we should aim at updating to the latest version all pallets project (including Jinja) maybe by generating ourselves the corresponding stubs or by relying on mypy APIs (still under discussions).

PR #1327, opened 2022-12-22
SONARPY-1235 Prevent executing the same rule again in case of successful scanWithoutParsing

guillaume-dequenne-sonarsource:


gyula-sallai-sonarsource (approved):
Looks good to me! Thanks for taking care of this :)

PR #1329, opened 2022-12-23
SONARPY-1216 PR Analysis: Avoid running regular rules for transitively impacted files

guillaume-dequenne-sonarsource:
Note that because merging this will impact what is logged during PR analysis, this will break security and DBD ITs, as they make assertions on these logs.


sonar-python-plugin/src/main/java/org/sonar/plugins/python/indexer/PythonIndexer.java
```diff
@@ -89,6 +89,10 @@ public boolean canBeScannedWithoutParsing(InputFile inputFile) {
     return false;
   }
 
+  public boolean canBeFullyScannedWithoutParsing(InputFile inputFile) {
+    return false;
+  }
```
gyula-sallai-sonarsource, line 94: As discussed: having both a `canBeScannedWithoutParsing` and `canBeFullyScannedWithoutParsing` is a bit confusing for me. Maybe we could rename `canBeFullyScannerWithoutParsing` to something like `dependsOnChangedFile` and the flip the usage?

gyula-sallai-sonarsource, line 94: Also, I think we should add a comment distinguishing the two cases:
- The file is unchanged and does not depend on a changed file -> Python checks are not executed, UCFGs and IR are retrieved from the cache.
- The file is unchanged but depends on a changed file -> Python checks are not exceuted, but UCFGS and IRs need to be regenerated.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonScanner.java
```diff
@@ -147,21 +147,32 @@ protected void scanFile(InputFile inputFile) {
   @Override
   public boolean scanFileWithoutParsing(InputFile inputFile) {
     InputFile.Type fileType = inputFile.type();
+    boolean result = true;
     for (PythonCheck check : checks.all()) {
       if (!isCheckApplicable(check, fileType)) {
         continue;
       }
-
+      if (!indexer.canBeFullyScannedWithoutParsing(inputFile) && !check.getClass().getPackageName().startsWith("org.sonar.python.checks")) {
```
gyula-sallai-sonarsource, line 155: I would recommend to extract this into a separate method

gyula-sallai-sonarsource (approved):
Looks good to me! I left a couple of small suggestions, mostly naming and comments.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/indexer/PythonIndexer.java
guillaume-dequenne-sonarsource: As discussed, I kind of like the fact that the notion of `partiallySkippable` and `fullySkippable` introduces a bit of terminology (even if it's admittedly a bit of a goofy one) that helps make communication a bit clearer.

I renamed the `canBeScannedWithoutParsing` to `canBePartiallyScanWithoutParsing` and added comments explaining the meaning of those concepts so that it's hopefully less confusing.

PR #1334, opened 2023-01-03
Enable build of dogfood branches

guillaume-dequenne-sonarsource:


nils-werner-sonarsource (approved):
This way we enabled dogfooding for PHP in the past. However, your branches have to be named `dogfood-on-...`. I saw you only have one of two named like this.

PR #1336, opened 2023-01-09
SONARPY-1238 Deduplicate strings in CPD token serialization

gyula-sallai-sonarsource:


sonar-python-plugin/src/test/java/org/sonar/plugins/python/PythonSensorTest.java
```diff
@@ -730,9 +729,11 @@ public void test_using_cache() throws IOException {
     writeCache.bind(readCache);
 
     byte[] serializedSymbolTable = toProtobufModuleDescriptor(Set.of(new VariableDescriptor("x", "main.x", null))).toByteArray();
+    var cpdTokens = CpdSerializer.toBytes(Collections.emptyList());
```
guillaume-dequenne-sonarsource, line 732: _Minor opinionated nitpick:_ I don't like using `var` too much in contexts like these: given the method is called `toBytes`, I had to re-read this to understand that it returns a `SerializationResult` instead of a `bytes[]`.
Of course feel free to ignore, just leaving this here as it caught my attention. (If you choose to change it though, keep in mind it's repeated across a few other tests).

guillaume-dequenne-sonarsource (approved):
Looks good! Thanks for handling this.
Please note that I'm not familiar with the underlying algorithm used, I did my review assuming it worked for PHP already.

sonar-python-plugin/src/test/java/org/sonar/plugins/python/PythonSensorTest.java
gyula-sallai-sonarsource: Fair point! Honestly, I do not really like that the method is called `toBytes` as it no longer returns bytes, but a structure. Maybe names like `CpdSerializer#serialize` and `deserialize` would also help understandibility better?

guillaume-dequenne-sonarsource: Indeed, that actually makes more sense than what I suggested.

PR #1338, opened 2023-01-10
SONARPY-1227 Fix parse error when multiple unpack in for loop

andreaguarino:


python-frontend/src/main/java/org/sonar/python/tree/ForStatementImpl.java
```diff
@@ -126,8 +131,23 @@ public Token asyncKeyword() {
 
   @Override
   public List<Tree> computeChildren() {
-    return Stream.of(Arrays.asList(asyncKeyword, forKeyword), expressions, Collections.singletonList(inKeyword), testExpressions,
+    List<Tree> expressionsWithSeparator = addCommas(expressions, expressionsCommas);
+    List<Tree> testExpressionsWithSeparator = addCommas(testExpressions, testExpressionsCommas);
```
guillaume-dequenne-sonarsource, line 135: I'd suggest renaming these `expressionsWithCommas` and `testExpressionsWithCommas`, because `Separator` (as in `StatementWithSeparator`) is used to describe the `;` or the `newline` token that separates two statements, which we probably want to distinguish from regular commas.

guillaume-dequenne-sonarsource (approved):
LGTM, left a minor naming suggestion.

PR #1342, opened 2023-01-23
SONARPY-863 Fix FP on S3516 when return values are different kind of comparisons

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource (approved):
LGTM!

PR #1343, opened 2023-01-23
[IN PROGRESS] SONARPY-1241 Migrate to Gradle from Maven to speedup the build PoC

mstachniuk:


johann-beleites-sonarsource (commented):
Overall, this looks relatively similar to some other projects we have configured in the company. I am not sure about some of the constructs we use but I don't have an instant better solution either. I've been wanting to give our gradle projects some love for a long time, to update some of the likely outdated ways of doing some things - I've not been able to find the time yet, though. 

In any case, configurations similar to this work pretty reliably in our projects at the moment, so I don't think there are any major roadblocks here. I would at least clean up dependency management and possibly property definitions before you merge to master, though.

Seeing all these similar build setups and duplicated functionality, we should probably think about whether we want to share some of this code between projects instead of copy+pasting it all over the place. Just a thought for the future, though, I wouldn't expect you to come up with a solution for that problem within the scope of moving sonar-python to gradle.

Out of curiosity, how come you chose Groovy over Kotlin DSL?

.cirrus.yml
```diff
@@ -106,8 +115,9 @@ ws_scan_task:
     folder: ${CIRRUS_WORKING_DIR}/.m2/repository
   whitesource_script:
     - source cirrus-env QA
-    - source set_maven_build_version $BUILD_NUMBER
-    - mvn clean install -DskipTests -DskipTypeshed -Denforcer.skip=true -Dskip.its=true
+#   TODO: MS This can be more optimized, let's identify what exact task needs to be run
+    - ./gradlew --no-daemon --info --stacktrace --console plain --no-daemon build -x test -x runTypeshed
+    - source ./export_ws_variables_from_gradle
     - source ws_scan.sh
```
alban-auzeill, line 121: I think there is a workaround in `wss-unified-agent.config` to ask mend to also analyze the .m2 folder:
`maven.m2RepositoryPath=.m2/repository`
This workaround is not ideal and generates a lot of false positives in Mend.
But it is here because Mend was not able to find dependencies otherwise.
And now with the gradle build, I think the `.m2/repository` workaround does not work anymore with Mend.

.cirrus.yml
```diff
@@ -106,8 +115,9 @@ ws_scan_task:
     folder: ${CIRRUS_WORKING_DIR}/.m2/repository
   whitesource_script:
     - source cirrus-env QA
-    - source set_maven_build_version $BUILD_NUMBER
-    - mvn clean install -DskipTests -DskipTypeshed -Denforcer.skip=true -Dskip.its=true
+#   TODO: MS This can be more optimized, let's identify what exact task needs to be run
+    - ./gradlew --no-daemon --info --stacktrace --console plain --no-daemon build -x test -x runTypeshed
+    - source ./export_ws_variables_from_gradle
```
alban-auzeill, line 120: My understanding of `_script` block is from [this documentation](https://cirrus-ci.org/guide/writing-tasks/#script-instruction)
> Note: Each script instruction is executed in a newly created process, therefore environment variables are not preserved between them.

So perhaps `PROJECT_VERSION` exported in `export_ws_variables_from_gradle` is not visible for `ws_scan.sh`

andreaguarino:
Closing this PR as the corresponding [initiative](https://discuss.sonarsource.com/t/how-to-improve-the-slow-ci-pipeline-of-sonar-python/12322) reached a conclusion.
Note that a link to this PR has been logged into [SONARPY-975](https://sonarsource.atlassian.net/browse/SONARPY-975) for future reference.
Thanks for your work!

PR #1344, opened 2023-01-23
SONARPY-1087 Add quick fixes for S2772 (NeedlessPassCheck)

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/NeedlessPassCheck.java
```diff
@@ -46,10 +49,18 @@ public void initialize(Context context) {
...
       }
+
       statements.stream()
         .filter(st -> st.is(PASS_STMT))
         .findFirst()
-        .ifPresent(st -> ctx.addIssue(st, MESSAGE));
+        .ifPresent(st -> {
+          var issue = (IssueWithQuickFix) ctx.addIssue(st, MESSAGE);
+          var quickFix = PythonQuickFix
+            .newQuickFix("Remove the \"pass\" statement at the end of a function or method")
```
guillaume-dequenne-sonarsource, line 59: I think it would be better to replace this message with a more simple `Remove the \"pass\" statement`, as the full message as it is would look a bit odd.

I see this message was in the`Approach` part of the ticket. FYI, that part is meant for us to know which approach to take when providing a quick fix but it wasn't meant to be the actual message (we actually didn't define any, it seems...).

guillaume-dequenne-sonarsource (changes requested):
Looks good! I just left a suggestion about issue message.

guillaume-dequenne-sonarsource (changes requested):
Looks good!
Left a minor comment: I have the feeling we don't need what looks like twice the same test.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1346, opened 2023-01-24
SONARPY-1255 S5886 should not raise issues when the yield expression is part of an assignment statement

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource (approved):
LGTM!

PR #1347, opened 2023-01-24
SONARPY-1100 S1045 should raise an issue when the same unknown exception is caught twice

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/UnreachableExceptCheck.java
```diff
@@ -77,23 +79,39 @@ private static void handleExceptClause(SubscriptionContext ctx, Map<String, Expr
 
   private static void handleExceptionExpression(SubscriptionContext ctx, Map<String, Expression> caughtTypes,
                                          Expression exceptionExpression, Map<String, Expression> caughtInExceptClause) {
-    if (exceptionExpression instanceof HasSymbol) {
-      Symbol symbol = ((HasSymbol) exceptionExpression).symbol();
-      if (symbol != null && symbol.kind().equals(Symbol.Kind.CLASS)) {
-        ClassSymbol classSymbol = (ClassSymbol) symbol;
-        List<Expression> handledExceptions = retrieveAlreadyHandledExceptions(classSymbol, caughtTypes);
+    Optional.of(exceptionExpression)
```
guillaume-dequenne-sonarsource, line 82: I don't really mind using a more functional style, especially for the newly added methods. On the other hand, I feel like changes are easier to write and review when we keep the style consistent (when it makes sense). In this case, I feel like an implementation like the following would have maintained the idiomatic style of the existing method:
```
private static void handleExceptionExpression(SubscriptionContext ctx, Map<String, Expression> caughtTypes,
                                         Expression exceptionExpression, Map<String, Expression> caughtInExceptClause) {
    if (!(exceptionExpression instanceof HasSymbol)) {
      return;
    }
    Symbol symbol = ((HasSymbol) exceptionExpression).symbol();
    if (symbol == null) {
      return;
    }
    List<Expression> handledExceptions = symbol.kind().equals(Symbol.Kind.CLASS) ? retrieveAlreadyHandledExceptionsByClass((ClassSymbol) symbol, caughtTypes)
      : retrieveAlreadyHandledExceptionsByFullyQualifiedName(symbol, caughtTypes);
    if (!handledExceptions.isEmpty()) {
      PreciseIssue issue = ctx.addIssue(exceptionExpression, "Catch this exception only once; it is already handled by a previous except clause.");
      handledExceptions.forEach(h -> issue.secondary(h, SECONDARY_MESSAGE));
    }
    caughtInExceptClause.put(symbol.fullyQualifiedName(), exceptionExpression);
  }
```

This approach would also have the benefit of avoiding the IDE warning about `symbol` being nullable on line 88 (despite the check on the `Objects.nonNull(expression.symbol())`, we still have the warning as we call the API again with the `CheckForNull` annotation) and also avoiding the somewhat redundant cast to `Expression` on line 94 (Since we started with an `Expression`, it should still be that despite the cast in the mapping operation - but the compiler doesn't know that).

I let you decide if you want to update the implementation. If you do, don't forget to remove the `java.util.Objects` import which would be unused as a result.

guillaume-dequenne-sonarsource (approved):
The main logic looks good!
Please consider my comment about code style and IDE warning/redundant casting that could be avoided, as well as my more minor comment on test method naming.
When merging the PR, please squash and merge to keep only a single commit.

python-checks/src/test/resources/checks/unreachableExcept.py
```diff
@@ -131,3 +134,40 @@ def oserror_hierarchy():
...
+        #  ^^^^^^^^^^^^^^^^^^^^^^^^> {{Exceptions will be caught here.}}
+        x = 1
+    except unknown.UnknownException as e: # Noncompliant
+        #  ^^^^^^^^^^^^^^^^^^^^^^^^
+        x = 2
+    except:
+        x = 3
+    return x
+
+def test_fn(a, b):
```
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: done

python-checks/src/main/java/org/sonar/python/checks/UnreachableExceptCheck.java
maksim-grebeniuk-sonarsource: done

PR #1348, opened 2023-01-25
SONARPY-1080 Add quick fixes for S3626 (RedundantJumpCheck)

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/RedundantJumpCheck.java
```diff
@@ -49,22 +56,40 @@ public void initialize(Context context) {
...
+
+  private static void addQuickFix(Tree lastElement, PreciseIssue issue) {
+    if (!(lastElement instanceof Statement)) {
+      return;
     }
+    Optional.of(issue)
+      .filter(IssueWithQuickFix.class::isInstance)
+      .map(IssueWithQuickFix.class::cast)
+      .ifPresent(i -> {
+          var quickFix = PythonQuickFix
```
guillaume-dequenne-sonarsource, line 87: There is a code smell on that line due to inconsistent indentation.

python-checks/src/test/java/org/sonar/python/checks/RedundantJumpCheckTest.java
```diff
@@ -29,4 +30,16 @@ public void test() {
...
+  @Test
+  public void quickFixTest() {
+    String input = "def redundant_jump(x):\n" +
+      "  if x == 1:\n" +
+      "    print(True)\n" +
+      "    return";
+    String expected = "def redundant_jump(x):\n" +
+      "  if x == 1:\n" +
+      "    print(True)\n";
+    PythonQuickFixVerifier.verify(new RedundantJumpCheck(), input, expected);
```
guillaume-dequenne-sonarsource, line 42: Similar to what was done in other tests, can you also add an assertion on the quick fix message?

python-checks/src/main/java/org/sonar/python/checks/RedundantJumpCheck.java
```diff
@@ -49,22 +56,40 @@ public void initialize(Context context) {
...
-      return;
-    }
-    for (CfgBlock cfgBlock : cfg.blocks()) {
-      if (cfgBlock.successors().size() == 1 && cfgBlock.successors().contains(cfgBlock.syntacticSuccessor())) {
+    Optional.ofNullable(cfg)
+      .map(ControlFlowGraph::blocks)
+      .stream()
+      .flatMap(Collection::stream)
+      .filter(cfgBlock -> cfgBlock.successors().size() == 1 && cfgBlock.successors().contains(cfgBlock.syntacticSuccessor()))
+      .forEach(cfgBlock -> {
```
guillaume-dequenne-sonarsource, line 64: To be honest, I don't understand the need to rewrite this in a functional style.

Calling the new `addQuickFix` method after the `PreciseIssue preciseIssue = ctx.addIssue(lastElement, message(lastElement));` would have achieved the same result with minimal change, making this trivial to review.

In general, I'm of the opinion that we should try to avoid mixing functional and stylistic changes together, especially when there is no particular reason to do it. And I don't think we have such a strong reason to want to move this particular snippet to a functional style, although I know that this way of writing code is generally more appealing.

Also, I think that one drawback of this code style that we should be aware of is the reduced testability. In this case, we lose coverage information on the `cfg == null` case because it's hidden in the optional chain and we don't have an `orElse` clause.

This is also true for the `ifPresent` clause in the `addQuickFix` method: we know it will always actually always be present, and the case where it's not is untested (which is normal, this is defensive code and it's impossible to test), but we still manage to reach 100% coverage. I think it's okay, but I also think it's good to be conscious of this.

Note that I'm not asking you to rollback the change, but this is a more general feedback that I think it's good to be purposeful with this kind of changes, as I feel doing this consistently could make the review process more difficult in more complex cases, and that despite being more modern, the functional style has some limitations that I think is good to be aware of (sometimes, plain old boring Java is all we need).

guillaume-dequenne-sonarsource (changes requested):
Looks good!
Can you please fix the indentation code smell and add a small test for the quick fix message?

python-checks/src/main/java/org/sonar/python/checks/RedundantJumpCheck.java
maksim-grebeniuk-sonarsource: done

python-checks/src/test/java/org/sonar/python/checks/RedundantJumpCheckTest.java
maksim-grebeniuk-sonarsource: done

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1349, opened 2023-01-25
SONARPY-1049 Add quick-fix for S1110 (UselessParenthesisCheck)

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/UselessParenthesisCheck.java
```diff
@@ -24,21 +24,29 @@
...
+import org.sonar.python.quickfix.PythonQuickFix;
+import org.sonar.python.quickfix.PythonTextEdit;
 
 @Rule(key = UselessParenthesisCheck.CHECK_KEY)
 public class UselessParenthesisCheck extends PythonSubscriptionCheck {
 
   public static final String CHECK_KEY = "S1110";
 
   private static final String MESSAGE = "Remove those useless parentheses.";
+  public static final String QUICK_FIX_MESSAGE = MESSAGE;
```
guillaume-dequenne-sonarsource, line 37: I'd suggest `Remove the redundant parentheses` for the quickfix message.

I also think we should update the regular issue message to `Remove those redundant parentheses` as the word `useless` is quite strong and we should probably favor the word `redundant` instead (this is a general feedback we received in the past, but it seems it was not addressed).

Changing the issue message has other implications, so let's do it in a separate PR.

python-checks/src/main/java/org/sonar/python/checks/UselessParenthesisCheck.java
```diff
@@ -24,21 +24,29 @@
...
 
   @Override
   public void initialize(Context context) {
     context.registerSyntaxNodeConsumer(Tree.Kind.PARENTHESIZED, ctx -> {
       ParenthesizedExpression parenthesized = (ParenthesizedExpression) ctx.syntaxNode();
       Expression expression = parenthesized.expression();
-      if (expression.is(Tree.Kind.PARENTHESIZED) || expression.is(Tree.Kind.TUPLE) || expression.is(Tree.Kind.GENERATOR_EXPR)) {
-        ctx.addIssue(parenthesized.leftParenthesis(), MESSAGE).secondary(parenthesized.rightParenthesis(), null);
+      if (expression.is(Tree.Kind.PARENTHESIZED, Tree.Kind.TUPLE, Tree.Kind.GENERATOR_EXPR)) {
```
guillaume-dequenne-sonarsource, line 44: Nice catch!

guillaume-dequenne-sonarsource (approved):
Looks good!
As mentioned, I kind of have the feeling the wording of the issue message is a bit too strong, so we probably shouldn't use the same for the quick fix (I created a separate [ticket](https://sonarsource.atlassian.net/browse/SONARPY-1259) for that, no need to handle this here).

PR #1352, opened 2023-01-26
SONARPY-1096 Add secondary message for S3923 (AllBranchesAreIdenticalCheck)

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/AllBranchesAreIdenticalCheck.java
```diff
@@ -72,17 +74,17 @@ private static void handleIfStatement(IfStatement ifStmt, SubscriptionContext ct
...
+    issue.secondary(issueLocation(ifStmt.body(), SECONDARY_MESSAGE));
+    ifStmt.elifBranches().forEach(e -> issue.secondary(issueLocation(e.body(), SECONDARY_MESSAGE)));
+    issue.secondary(issueLocation(elseBranch.body(), SECONDARY_MESSAGE));
     if (!hasSideEffect(ifStmt)) {
       issue.addQuickFix(computeQuickFixForIfStatement(ifStmt, elseBranch));
     }
   }
 
-  private static IssueLocation issueLocation(StatementList body) {
+  private static IssueLocation issueLocation(StatementList body, @Nullable String message) {
```
guillaume-dequenne-sonarsource, line 85: It seems that this method only deals with secondary locations. We could consider renaming it `secondaryLocation` instead (as you prefer). Also, you can probably remove the `message` parameter and inline `SECONDARY_MESSAGE`, since the value of the parameter is always the same.

python-checks/src/main/java/org/sonar/python/checks/AllBranchesAreIdenticalCheck.java
```diff
@@ -48,6 +49,7 @@ public class AllBranchesAreIdenticalCheck extends PythonSubscriptionCheck {
   private static final String CONDITIONAL_MESSAGE = "This conditional expression returns the same value whether the condition is \"true\" or \"false\".";
 
   private static final List<ConditionalExpression> ignoreList = new ArrayList<>();
+  public static final String SECONDARY_MESSAGE = "Redundant statements.";
```
guillaume-dequenne-sonarsource, line 52: Thinking about it, I think `Duplicated statements.` would probably make more sense, given the statements themselves are not redundant (the conditional branching is). Wdyt?

guillaume-dequenne-sonarsource (approved):
Looks good! I left some minor suggestions. Please consider them before merging.

python-checks/src/main/java/org/sonar/python/checks/AllBranchesAreIdenticalCheck.java
maksim-grebeniuk-sonarsource: done

python-checks/src/main/java/org/sonar/python/checks/AllBranchesAreIdenticalCheck.java
maksim-grebeniuk-sonarsource: done

PR #1353, opened 2023-01-27
SONARPY-1164 Resolve unpacking expression when analyzing call arguments

maksim-grebeniuk-sonarsource:


python-checks/src/test/resources/checks/cdk/unencryptedEbsVolume.py
```diff
@@ -38,3 +38,40 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
+        a = False
+        volume_args = {"encrypted": a}
+        Volume(self, **volume_args) # Noncompliant
+
+class DictValueLinkedToVariableLinkedToVariable(Stack):
+    def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
+        b = False
+        a = b
+        volume_args = {"encrypted": a}
+        Volume(self, **volume_args) # Noncompliant
```
nils-werner-sonarsource, line 65: Primary and secondary locations in this test case would show, that the trace is correctly represented.

python-checks/src/test/resources/checks/cdk/unencryptedEbsVolume.py
```diff
@@ -38,3 +38,46 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
+        a = b
+        b = a
+        volume_args = {"encrypted": b}
+        Volume(self, **volume_args)
+
+class DictValueLinkedToVariable(Stack):
+    def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
+        a = False
+#       ^^^^^^^^^> 1 {{Propagated setting.}}
+        volume_args = {"encrypted": a}
```
nils-werner-sonarsource, line 60:
```
- 060:         volume_args = {"encrypted": a}
- 060:         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^> {{Propagated setting.}}
```
I guess this is what @guillaume-dequenne-sonarsource was mentioning. Here a step is missing.

PR #1356, opened 2023-01-31
SONARPY-1164 Resolve unpacking expression when analyzing call arguments

nils-werner-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/cdk/CdkUtils.java
```diff
@@ -215,17 +227,29 @@ private ExpressionFlow(SubscriptionContext ctx, Deque<Expression> locations) {
...
+
+    protected static ExpressionFlow build(SubscriptionContext ctx, Expression expression, @Nullable String argumentName) {
       Deque<Expression> locations = new LinkedList<>();
-      resolveLocations(expression, locations);
+      resolveLocations(ctx, expression, locations, argumentName);
       return new ExpressionFlow(ctx, locations);
     }
 
-    static void resolveLocations(Expression expression, Deque<Expression> locations) {
+    static void resolveLocations(SubscriptionContext ctx, Expression expression, Deque<Expression> locations, @Nullable String argumentName) {
```
maksim-grebeniuk-sonarsource, line 239: probably it is better to use LinkedHashSet here since there is a lot of `contains` checks. 
And then construct `LinkedList` out of `Set` results and pass it to the `ExpressionFlow` constructor.

nils-werner-sonarsource: Sounds reasonable!

python-checks/src/main/java/org/sonar/python/checks/cdk/CdkUtils.java
```diff
@@ -87,30 +88,41 @@ public static Optional<DictionaryLiteral> getDictionary(Expression expression) {
...
 
-    if (argument.isEmpty() && arguments.stream().anyMatch(UnpackingExpression.class::isInstance)) {
-      return Optional.of(new UnresolvedExpressionFlow(ctx));
+    // Check for unpacking expression
+    if (argument.isEmpty()) {
+      argument = getUnpackingExpressions(arguments)
+        .map(UnpackingExpression::expression)
+        .map(expression -> ExpressionFlow.build(ctx, expression, argumentName))
+        .findFirst();
     }
```
maksim-grebeniuk-sonarsource, line 115: this two optional chains could be combined with `or` method like these:
```java
Optional<ExpressionFlow> argument = Optional.ofNullable(regArgument)
      .map(expression -> ExpressionFlow.build(ctx, expression))
      // Check for unpacking expression
      .or(() -> getUnpackingExpressions(arguments)
        .map(UnpackingExpression::expression)
        .map(expression -> ExpressionFlow.build(ctx, expression, argumentName))
        .findFirst());
```

python-checks/src/main/java/org/sonar/python/checks/cdk/CdkUtils.java
```diff
@@ -87,30 +88,41 @@ public static Optional<DictionaryLiteral> getDictionary(Expression expression) {
...
     // ctx not required as long as we don't try to raise an issue with ExpressionFlow
-    return getArgument(null, callExpression, TreeUtils.argumentByKeyword(argumentName, callExpression.arguments()));
+    return getArgument(null, callExpression, argumentName);
   }
 
-  private static Optional<ExpressionFlow> getArgument(SubscriptionContext ctx, CallExpression callExpression, RegularArgument regArg) {
+  /**
+   * Resolve a particular argument of a call by keyword or get an empty optional if the argument is not set nor resolvable.
+   */
+  protected static Optional<ExpressionFlow> getArgument(SubscriptionContext ctx, CallExpression callExpression, String argumentName) {
```
maksim-grebeniuk-sonarsource, line 100: As code analysis says that `ctx` should be marked with `@javax.annotation.Nullable` annotation

maksim-grebeniuk-sonarsource (approved):
Approve

PR #1357, opened 2023-01-31
SONARPY-1098 S1854: Add secondary locations on the variable re-assignments

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/DeadStoreCheck.java
```diff
@@ -95,13 +106,47 @@ private static void raiseIssue(SubscriptionContext ctx, DeadStoreUtils.Unnecessa
       issue = ctx.addIssue(element.firstToken(), lastRelevantToken, message);
     }
 
+    unnecessaryAssignment.symbol.usages().stream()
+      .filter(Usage::isBindingUsage)
+      .map(Usage::tree)
+      // skip initial issue binding
+      .filter(tree -> tree != element && TreeUtils.firstAncestor(tree, parent -> parent == element) == null)
+      //skip assignments before
+      .filter(tree -> getTreeByPositionComparator().compare(tree, element) == 1)
```
guillaume-dequenne-sonarsource, line 115: I see why we need this, but I think we're missing a test case for it (if I remove the line, all tests still succeed). Can you add a small test covering this case?
Also, IntelliJ suggests changing the comparison to `> 0` instead.

python-checks/src/main/java/org/sonar/python/checks/DeadStoreCheck.java
```diff
@@ -76,7 +84,9 @@ public void initialize(Context context) {
   private static void verifyBlock(SubscriptionContext ctx, CfgBlock block, LiveVariablesAnalysis.LiveVariables blockLiveVariables,
     Set<Symbol> readSymbols, FunctionDef functionDef) {
 
-    DeadStoreUtils.findUnnecessaryAssignments(block, blockLiveVariables, functionDef)
+    var unnecessaryAssignments = DeadStoreUtils.findUnnecessaryAssignments(block,
+      blockLiveVariables, functionDef);
+    unnecessaryAssignments
```
guillaume-dequenne-sonarsource, line 89: Do we need this change? I don't have the feeling this makes the overall code clearer, and we don't use the `unnecessaryAssignments` variable anywhere else.

guillaume-dequenne-sonarsource (changes requested):
Looks good overall!
Please have a look at my review comments.

guillaume-dequenne-sonarsource (approved):
LGTM!

PR #1359, opened 2023-02-01
SONARPY-1242 Modify rule S5332: False positive for LoadBalancer construct with listeners set as dict (AWS CDK)

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource (approved):
LGTM.

PR #1360, opened 2023-02-01
SONARPY-1115 S1854 Fix FPs on import statements

maksim-grebeniuk-sonarsource:


python-checks/src/test/resources/checks/deadStore.py
```diff
@@ -400,3 +400,22 @@ def multiple_issues(a):
     print(b)
 
 
+def nested_packages_import():
+    import sample # FN
+    import sample.first # FN
```
guillaume-dequenne-sonarsource, line 405: Can you create a ticket to fix these false negatives and reference it here as well?

python-checks/src/test/resources/checks/deadStore.py
```diff
@@ -400,3 +400,22 @@ def multiple_issues(a):
     print(b)
 
 
+def nested_packages_import():
```
guillaume-dequenne-sonarsource, line 403: I think it would be clearer to split this into 2 dedicated tests: one for the FP we want to fix, and one for the FN we accept. Here, both are a bit mixed and I think it will be difficult to understand outside the context of this ticket.

I'd suggest having something like:

```
def imports_of_different_submodules_no_fp():
    import sample.first
    import sample.second
    b = sample.first.First()
    c = sample.second.Second()


def nested_packages_imports_fn():
    import sample # FN
    import sample.first # FN
    import sample.first.third

    a = sample.Sample()
    b = sample.first.First()
    d = sample.first.third.Third()
```

guillaume-dequenne-sonarsource (changes requested):
Looking good. Please have a look at my comments.

python-checks/src/test/resources/checks/deadStore.py
```diff
@@ -400,16 +400,20 @@ def multiple_issues(a):
     print(b)
 
 
-def nested_packages_import():
-    import sample # FN
-    import sample.first # FN
+def nested_packages_import_no_fn():
```
guillaume-dequenne-sonarsource, line 403: This should say `no_fp` instead.

guillaume-dequenne-sonarsource (approved):
LGTM.
Can you please rename the test before merging?

PR #1361, opened 2023-02-01
SONARPY-794 FN on S4830 when “ssl._create_unverified_context()” is provided as keyword argument

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/VerifiedSslTlsCertificateCheck.java
```diff
@@ -318,6 +320,12 @@ private static void standardSslCheck(SubscriptionContext subscriptionContext) {
       }));
   }
 
+  private static void standardSslCheckForRegularArgument(SubscriptionContext subscriptionContext) {
+    var argument = (RegularArgument) subscriptionContext.syntaxNode();
+    searchRhsForVulnerableMethod(argument.expression())
```
guillaume-dequenne-sonarsource, line 325: _Nitpick:_ I think we can rename this method to `isVulnerableMethodCall` since it is no longer only looking for RHS of assignments.

guillaume-dequenne-sonarsource (approved):
Looks good to me!
Left a small renaming suggestion.

PR #1368, opened 2023-02-06
SONARPY-1268 Add quick fix for S5796 (IdentityComparisonWithNewObjectCheck)

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/IdentityComparisonWithNewObjectCheck.java
```diff
@@ -78,16 +83,30 @@ private static boolean checkOperand(Expression operand, IsExpression isExpr, Sub
     Optional<List<Tree>> secondariesOpt = findIssueForOperand(operand);
     secondariesOpt.ifPresent(secondaryLocations -> {
       Token notToken = isExpr.notToken();
-      PreciseIssue preciseIssue = notToken == null ?
-        ctx.addIssue(isExpr.operator(), MESSAGE_IS) :
-        ctx.addIssue(isExpr.operator(), notToken, MESSAGE_IS_NOT);
+      var issue = notToken == null ?
```
guillaume-dequenne-sonarsource, line 86: We seem to be testing twice for the presence of the `not` token (once in the issue creation and one in the quick fix creation), maybe we could only do it once and create the issue and quick fix at the same time? (Basically like in the PR for SONARPY-1084).

python-checks/src/main/java/org/sonar/python/checks/IdentityComparisonWithNewObjectCheck.java
```diff
@@ -46,6 +49,8 @@
 @Rule(key = "S5796")
 public class IdentityComparisonWithNewObjectCheck extends PythonSubscriptionCheck {
   private static final String MESSAGE_IS = "Replace this \"is\" operator with \"==\".";
+  public static final String IS_QUICK_FIX_MESSAGE = "Replace with \"==\".";
+  public static final String IS_NOT_QUICK_FIX_MESSAGE = "Replace with \"!=\".";
```
guillaume-dequenne-sonarsource, line 53: Same as in the other PR, I think we should always avoid the final `.` in quick fix messages (and my bad if there was any in the tickets!)

guillaume-dequenne-sonarsource (approved):
Looks good, left minor comments.

python-checks/src/main/java/org/sonar/python/checks/IdentityComparisonWithNewObjectCheck.java
maksim-grebeniuk-sonarsource: done

python-checks/src/main/java/org/sonar/python/checks/IdentityComparisonWithNewObjectCheck.java
maksim-grebeniuk-sonarsource: done

PR #1369, opened 2023-02-07
SONARPY-1083 Add quick fix for S5712 (NotImplementedErrorInOperatorMethodsCheck)

maksim-grebeniuk-sonarsource:


python-checks/src/test/java/org/sonar/python/checks/NotImplementedErrorInOperatorMethodsCheckTest.java
```diff
@@ -20,11 +20,25 @@
...
+    var check = new NotImplementedErrorInOperatorMethodsCheck();
+    String codeWithIssue = "class MyClass:\n" +
+      "    def __lt__(self, other):\n" +
+      "        raise NotImplementedError()";
+    String codeFixed = "class MyClass:\n" +
+      "    def __lt__(self, other):\n" +
+      "        return NotImplemented";
+    PythonQuickFixVerifier.verify(check, codeWithIssue, codeFixed);
+    PythonQuickFixVerifier.verifyQuickFixMessages(check, codeWithIssue, NotImplementedErrorInOperatorMethodsCheck.QUICK_FIX_MESSAGE);
+  }
```
guillaume-dequenne-sonarsource, line 43: Could you add another test documenting that something like:
```
def __mul__(self, other):
  ex = NotImplementedError()
  raise ex # Noncompliant
```
Would be quick fixed as:

```
def __mul__(self, other):
  ex = NotImplementedError()
  return NotImplemented
```

I think it's fine, but it's probably good to document that the quick fix will not take care of the now redundant variable assignment.

guillaume-dequenne-sonarsource (approved):
Looks good!
Left a suggestion for an additional test we could have.

python-checks/src/test/java/org/sonar/python/checks/NotImplementedErrorInOperatorMethodsCheckTest.java
maksim-grebeniuk-sonarsource: done

PR #1371, opened 2023-02-07
SONARPY-1089 Add quick fix for S108 (EmptyNestedBlockCheck)

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/EmptyNestedBlockCheck.java
```diff
@@ -19,20 +19,28 @@
...
+import org.sonar.python.api.PythonTokenType;
+import org.sonar.python.quickfix.IssueWithQuickFix;
+import org.sonar.python.quickfix.PythonQuickFix;
+import org.sonar.python.quickfix.PythonTextEdit;
 import org.sonar.python.tree.TreeUtils;
 
 @Rule(key = EmptyNestedBlockCheck.CHECK_KEY)
 public class EmptyNestedBlockCheck extends PythonSubscriptionCheck {
   public static final String CHECK_KEY = "S108";
+  public static final String QUICK_FIX_MESSAGE = "Add an empty TODO comment.";
```
guillaume-dequenne-sonarsource, line 41: I'd suggest a simpler `Add a TODO comment`

python-checks/src/main/java/org/sonar/python/checks/EmptyNestedBlockCheck.java
```diff
@@ -52,11 +60,19 @@ public void initialize(Context context) {
...
-        }
+        var passTreeElement = Optional.of(statementListTree)
+          .map(StatementList::statements)
+          .map(Collection::stream)
+          .flatMap(Stream::findFirst)
+          .map(Tree.class::cast)
+          .orElseGet(statementListTree::firstToken);
+
+        var quickFix = PythonQuickFix.newQuickFix(QUICK_FIX_MESSAGE)
+          .addTextEdit(PythonTextEdit.insertLineBefore(passTreeElement, TODO_COMMENT_TEXT))
```
guillaume-dequenne-sonarsource, line 71: I think there is one problem here if the `pass` statement is on the same line as the parent block.
```
def foo():
  for i in range(3): pass
```

Currently it seems to produce something like (looking at the unit tests - haven't confirmed in an actual IDE):

```
def foo():
  for i in range(3): # TODO: Add implementation
                              pass
```

So maybe there is a problem with the method `insertLineBefore` as well, because I would have instinctively expected it to produce:

```
def foo():
  # TODO: Add implementation
  for i in range(3): pass
```
(Note that this wouldn't work either, as I think this snippet would also raise an issue)


Ideally we probably want to produce:

```
def foo():
  for i in range(3): 
    # TODO: Add implementation
    pass
```

And it may be okay to have a separate ticket for this.

guillaume-dequenne-sonarsource (changes requested):
For the common case this looks good.
However I think there is an edge case where the quick fix could break the code formatting.
It could be okay to consider this out of scope (I don't expect it will bring much value to cover this case). Happy to discuss it.

python-checks/src/main/java/org/sonar/python/checks/EmptyNestedBlockCheck.java
maksim-grebeniuk-sonarsource: done

python-checks/src/main/java/org/sonar/python/checks/EmptyNestedBlockCheck.java
maksim-grebeniuk-sonarsource: A separate issue created SONARPY-1277

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1373, opened 2023-02-07
SONARPY-1090 Add quick fixes for S5754 (IgnoredSystemExitCheck)

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/IgnoredSystemExitCheck.java
```diff
@@ -49,13 +53,15 @@ public class IgnoredSystemExitCheck extends PythonSubscriptionCheck {
 
   private static final String SYSTEM_EXIT_FUNCTION_NAME = "sys.exit";
   private static final String SYSTEM_EXC_INFO_NAME = "sys.exc_info";
+  public static final String NOT_RAISED_CAUGHT_EXCEPTION_QUICK_FIX_MESSAGE = "Replace \"pass\" statement with \"raise\"";
```
guillaume-dequenne-sonarsource, line 56: I'd suggest to use `Propagate the exception` as a message.

python-checks/src/main/java/org/sonar/python/checks/IgnoredSystemExitCheck.java
```diff
@@ -154,7 +165,14 @@ private static boolean handleCaughtException(SubscriptionContext ctx, Expression
     }
 
     if (SYSTEM_EXIT_EXCEPTION_NAME.equals(caughtExceptionName)) {
-      ctx.addIssue(caughtException, MESSAGE_NOT_RERAISED_CAUGHT_EXCEPTION);
+      var issue = (IssueWithQuickFix) ctx.addIssue(caughtException, MESSAGE_NOT_RERAISED_CAUGHT_EXCEPTION);
+      if (visitor.passStatement != null) {
+        var quickFix = PythonQuickFix.newQuickFix(NOT_RAISED_CAUGHT_EXCEPTION_QUICK_FIX_MESSAGE)
+          .addTextEdit(PythonTextEdit.replace(visitor.passStatement, "raise"))
+          .build();
+        issue.addQuickFix(quickFix);
```
guillaume-dequenne-sonarsource, line 173: I think in all cases where we raise an issue, a `SystemExit` is ignored. Thus, I think it's valid to suggest that quick fix for the other scenarios where we raise an issue.

python-checks/src/main/java/org/sonar/python/checks/IgnoredSystemExitCheck.java
```diff
@@ -154,7 +165,14 @@ private static boolean handleCaughtException(SubscriptionContext ctx, Expression
     }
 
     if (SYSTEM_EXIT_EXCEPTION_NAME.equals(caughtExceptionName)) {
-      ctx.addIssue(caughtException, MESSAGE_NOT_RERAISED_CAUGHT_EXCEPTION);
+      var issue = (IssueWithQuickFix) ctx.addIssue(caughtException, MESSAGE_NOT_RERAISED_CAUGHT_EXCEPTION);
+      if (visitor.passStatement != null) {
```
guillaume-dequenne-sonarsource, line 169: I think this condition is not robust enough. We could fail to suggest a quick fixes in some cases where we probably should be able to provide one easily. Also, this can break the code in some (admittedly contrived) cases. For example, the following snippet:

```
try:
  foo()
except SystemExit:
  pass
  x = 42
  print("system exit caught")
  ...
```
would be remplaced with:
```
try:
  foo()
except SystemExit:
  raise
  x = 42
  print("system exit caught")
  ...
```
and would change execution.

guillaume-dequenne-sonarsource (changes requested):
Please have a look at my comments.

python-checks/src/main/java/org/sonar/python/checks/IgnoredSystemExitCheck.java
maksim-grebeniuk-sonarsource: done

python-checks/src/main/java/org/sonar/python/checks/IgnoredSystemExitCheck.java
maksim-grebeniuk-sonarsource: done

python-checks/src/main/java/org/sonar/python/checks/IgnoredSystemExitCheck.java
maksim-grebeniuk-sonarsource: done

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1374, opened 2023-02-08
SONARPY-1266 Add quick fix for S5708 (CaughtExceptionsCheck)

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/CaughtExceptionsCheck.java
```diff
@@ -64,10 +73,43 @@ private static void checkExceptClause(SubscriptionContext ctx) {
...
+      .map(Symbol::usages)
+      .stream()
+      .flatMap(Collection::stream)
+      .filter(Usage::isBindingUsage)
+      .findFirst()
+      .map(Usage::tree)
+      .map(Tree::parent)
+      .filter(ClassDef.class::isInstance)
+      .map(ClassDef.class::cast)
+      .ifPresent(classDef -> {
```
guillaume-dequenne-sonarsource, line 96: Somehow I have a feeling this could be simpler. I also have the feeling we are searching a second time for a class definition when it has probably been done when evaluating `(expression instanceof HasSymbol) && !inheritsFromBaseException(((HasSymbol) expression).symbol()))` of line 74. 

Perhaps we could reuse that a bit (e.g when we determined that the `classSymbol.canBeOrExtend(BASE_EXCEPTION)` is false, then we search for the definition of that symbol and create the quick fix from there).

Saying this, I'm really not sure making such change would lead to better code, happy to discuss it!

guillaume-dequenne-sonarsource (changes requested):
Looking good. Please have a look at my comments.

guillaume-dequenne-sonarsource (approved):
LGTM.
As we discussed, please reorganize the commits before merging to separate the ticket from the new API.

PR #1375, opened 2023-02-08
SONARPY-1265 Add quick fix for S5714 (BooleanExpressionInExceptCheck)

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource (changes requested):
I think in the current state, we'd risk producing invalid code.
I also have the feeling we could simplify the implementation a bit (even if that means explicitly excluding the most complex cases from having a quick fix).

python-checks/src/main/java/org/sonar/python/checks/BooleanExpressionInExceptCheck.java
```diff
@@ -19,16 +19,29 @@
...
 import org.sonar.plugins.python.api.tree.Tree.Kind;
+import org.sonar.python.quickfix.IssueWithQuickFix;
+import org.sonar.python.quickfix.PythonQuickFix;
+import org.sonar.python.quickfix.PythonTextEdit;
 
 @Rule(key = "S5714")
 public class BooleanExpressionInExceptCheck extends PythonSubscriptionCheck {
 
+  public static final String MESSAGE = "Rewrite this \"except\" expression as a tuple of exception classes.";
+  public static final String QUICK_FIX_MESSAGE = "Replace boolean expression with tuple";
```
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: done

python-checks/src/main/java/org/sonar/python/checks/BooleanExpressionInExceptCheck.java
```diff
@@ -37,10 +57,62 @@ public void initialize(Context context) {
...
+    if (expression.is(Kind.OR, Kind.AND)) {
+      var binaryExpression = (BinaryExpression) expression;
+      var leftExceptions = collectNames(binaryExpression.leftOperand());
+      var rightExceptions = collectNames(binaryExpression.rightOperand());
+      var result = new ArrayList<String>();
+      result.addAll(leftExceptions);
+      result.addAll(rightExceptions);
+      return result;
+    } else if (expression.is(Kind.PARENTHESIZED)) {
+      var parenthesizedExpression = (ParenthesizedExpression) expression;
```
guillaume-dequenne-sonarsource, line 80: You can actually use `Expressions#removeParentheses` here as well (result will be exactly the same).
You could probably do that as soon as you receive the `expression` in the method so you don't have to deal with it in the `if / else if` chain.

python-checks/src/main/java/org/sonar/python/checks/BooleanExpressionInExceptCheck.java
```diff
@@ -37,10 +57,62 @@ public void initialize(Context context) {
...
+  }
+
+  private static void addQuickFix(IssueWithQuickFix issue, Tree expression) {
+    expression = Objects.requireNonNullElse(TreeUtils.firstAncestorOfKind(expression, Kind.PARENTHESIZED), expression);
+
+    List<String> names;
+    try {
+      names = collectNames(expression);
+    } catch (IllegalArgumentException e) {
+      LOG.error("Could not add quick fix", e);
```
guillaume-dequenne-sonarsource, line 102: I think logging an error here is a tad too strong. We actually expect this will happen in cases a call expression is involved, for example.
In such case, we simply don't want to raise a quick fix. I think it's enough to add a comment here saying that the expression contains subexpressions that are out of scope for quick fixing and that we don't provide one, and avoid logging anything.

guillaume-dequenne-sonarsource (changes requested):
Looking good, minor change suggestions.

python-checks/src/main/java/org/sonar/python/checks/BooleanExpressionInExceptCheck.java
maksim-grebeniuk-sonarsource: done

python-checks/src/main/java/org/sonar/python/checks/BooleanExpressionInExceptCheck.java
maksim-grebeniuk-sonarsource: done

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1376, opened 2023-02-08
SONARPY-1263 Add quick fix for S5915 (AssertAfterRaiseCheck)

maksim-grebeniuk-sonarsource:


python-checks/src/test/java/org/sonar/python/checks/tests/AssertAfterRaiseCheckTest.java
```diff
@@ -42,4 +43,20 @@ public void testWithAnotherLibraryUnittest() {
   public void testImportPytestAs() {
     PythonCheckVerifier.verify("src/test/resources/checks/tests/assertAfterRaiseImportPytestAs.py", new AssertAfterRaiseCheck());
   }
+
```
guillaume-dequenne-sonarsource, line 46: Can you also add a small test asserting we don't provide quick fixes for single statement bodies?

Also, probably a silly edge case as well, but what if all the statements are on the same line?
```
import pytest
def test_smth():
  with pytest.raises(ZeroDivisionError): foo(); assert bar() == 42
```
Nobody writes code like this, but if anyone ever does, I think it'd be better to not provide any quick fixes than to possibly break the code.

guillaume-dequenne-sonarsource (changes requested):
Looks good.
I think an additional test for single-statement bodies could be nice.
Also, a test for the edge case of statements on the same line could be nice as well.

python-checks/src/test/java/org/sonar/python/checks/tests/AssertAfterRaiseCheckTest.java
maksim-grebeniuk-sonarsource: done

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1378, opened 2023-02-09
SONARPY-1277 Add quick fix for S108 (EmptyNestedBlockCheck) in case of inline statements

maksim-grebeniuk-sonarsource:


python-frontend/src/main/java/org/sonar/python/tree/TreeUtils.java
```diff
@@ -248,6 +248,41 @@ public static List<FunctionDef> topLevelFunctionDefs(ClassDef classDef) {
     return visitor.functionDefs;
   }
 
+  public static int findIndent(Tree tree) {
```
andreaguarino, line 251: I would suggest to rename it to `findIndentationSize`

python-frontend/src/test/java/org/sonar/python/tree/TreeUtilsTest.java
```diff
@@ -441,6 +441,49 @@ public void test_getTreeByPositionComparator() {
...
+    assertThat(indent).isZero();
+  }
+
+  @Test
+  public void test_findIndentEmpty() {
+    FileInput fileInput = PythonTestUtils.parse("");
+    var indent = TreeUtils.findIndent(fileInput);
+    assertThat(indent).isZero();
+  }
+
```
andreaguarino, line 486: Would be nice adding an example with inconsistent indentation, for example in same cases it's 2 and in other cases it's 4

python-checks/src/main/java/org/sonar/python/checks/EmptyNestedBlockCheck.java
```diff
@@ -70,11 +70,21 @@ public void initialize(Context context) {
...
+          if (indent > 0) {
+            var offset = parent.firstToken().column() + indent;
+            var textToInsert = "\n" + " ".repeat(offset) + TODO_COMMENT_TEXT + "\n" + " ".repeat(offset);
+            var quickFix = PythonQuickFix.newQuickFix(QUICK_FIX_MESSAGE,
+              PythonTextEdit.insertBefore(passTreeElement, textToInsert));
+            issue.addQuickFix(quickFix);
+          }
         }
 
+
```
andreaguarino, line 87: Unnecessary change?

andreaguarino (approved):
LGTM, please have a look at my comments before merging

python-frontend/src/test/java/org/sonar/python/tree/TreeUtilsTest.java
maksim-grebeniuk-sonarsource: done

python-frontend/src/main/java/org/sonar/python/tree/TreeUtils.java
maksim-grebeniuk-sonarsource: done

python-checks/src/main/java/org/sonar/python/checks/EmptyNestedBlockCheck.java
maksim-grebeniuk-sonarsource: removed

PR #1380, opened 2023-02-10
SONARPY-1092 Add quick fixes for S5806 (BuiltinShadowingAssignmentCheck)

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/BuiltinShadowingAssignmentCheck.java
```diff
@@ -86,15 +93,35 @@ private void checkAnnotatedAssignment(SubscriptionContext ctx) {
...
+        }
+      });
+  }
+
+  private static PythonQuickFix createQuickFix(Symbol symbol) {
+    var edits = symbol.usages()
+      .stream()
+      .map(Usage::tree)
+      .map(Tree::firstToken)
+      .sorted(Comparator.comparing(Token::line).reversed().thenComparing(Token::column).reversed())
```
pynicolas, line 117: Why use the `reversed` method? (it's used twice on this line)
Do we need to sort? (maybe we do)

python-checks/src/main/java/org/sonar/python/checks/BuiltinShadowingAssignmentCheck.java
```diff
@@ -86,15 +93,35 @@ private void checkAnnotatedAssignment(SubscriptionContext ctx) {
...
-      PreciseIssue existingIssue = variableIssuesRaised.get(variableSymbol);
-      if (existingIssue != null) {
-        existingIssue.secondary(variable, REPEATED_VAR_MESSAGE);
-      } else {
-        variableIssuesRaised.put(variableSymbol, ctx.addIssue(variable, MESSAGE));
-      }
-    }
+    Optional.ofNullable(variable.symbol())
+      .filter(symbol -> symbol.usages().stream().map(Usage::kind).noneMatch(Usage.Kind.GLOBAL_DECLARATION::equals))
+      .ifPresent(symbol -> {
```
pynicolas, line 98: I'm not sure I get why using `Optional` and a functional style is better here.
It seems to me that it has at least one drawback: it may hide test coverage issues, e.g. we may not see that we never test a case where `variable.symbol()` is null.

python-checks/src/main/java/org/sonar/python/checks/BuiltinShadowingAssignmentCheck.java
```diff
@@ -86,15 +93,35 @@ private void checkAnnotatedAssignment(SubscriptionContext ctx) {
...
+      });
+  }
+
+  private static PythonQuickFix createQuickFix(Symbol symbol) {
+    var edits = symbol.usages()
+      .stream()
+      .map(Usage::tree)
+      .map(Tree::firstToken)
+      .sorted(Comparator.comparing(Token::line).reversed().thenComparing(Token::column).reversed())
+      .map(token -> PythonTextEdit.insertBefore(token, "_"))
```
pynicolas, line 118: What if another variable already uses that name?
```
def foo():
  max = 1
  _max = 2
  print(max)
```
I know, it's probably a corner case.
If it's not easy to fix, maybe we can create a separate ticket to handle it later.

pynicolas (approved):

python-checks/src/main/java/org/sonar/python/checks/BuiltinShadowingAssignmentCheck.java
maksim-grebeniuk-sonarsource: removed, I did several tests including calling quick fix in VS code, and didn't find that it makes sense. Good catch

python-checks/src/main/java/org/sonar/python/checks/BuiltinShadowingAssignmentCheck.java
maksim-grebeniuk-sonarsource: Some protection was added to prevent renaming to already used value. It will not help in 100% of cases. If you have any ideas on how to make it in a more proper way - let's discuss.

python-checks/src/main/java/org/sonar/python/checks/BuiltinShadowingAssignmentCheck.java
maksim-grebeniuk-sonarsource: That's true but in case we don't need to react somehow on a null case there - it keeps code more clear for me instead of using two `if` statements (null check and usages filtering)

python-checks/src/main/java/org/sonar/python/checks/BuiltinShadowingAssignmentCheck.java
```diff
@@ -85,16 +97,62 @@ private void checkAnnotatedAssignment(SubscriptionContext ctx) {
...
+    return Optional.ofNullable(TreeUtils.firstAncestorOfKind(tree, Tree.Kind.FUNCDEF))
+      .map(funcDef -> findChildren(funcDef, t -> t.is(Tree.Kind.NAME)))
+      .stream()
+      .flatMap(Collection::stream)
+      .map(Name.class::cast)
+      .map(Name::name)
+      .collect(Collectors.toSet());
+  }
+
+  private static List<Tree> findChildren(Tree tree, Predicate<Tree> predicate) {
```
pynicolas, line 110: Can't we use `TreeUtils.hasDescendant` instead of reimplementing that?

pynicolas (approved):

PR #1384, opened 2023-02-14
SONARPY-1280 Fix NPE for VerboseRegexCheck

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/regex/VerboseRegexCheck.java
```diff
@@ -50,7 +50,7 @@ public void checkRegex(RegexParseResult regexParseResult, CallExpression regexFu
 
   @Override
   public PreciseIssue addIssue(RegexSyntaxElement regexTree, String message, @Nullable Integer cost, List<RegexIssueLocation> secondaries) {
-    return Optional.of(super.addIssue(regexTree, message, cost, secondaries))
+    return Optional.ofNullable(super.addIssue(regexTree, message, cost, secondaries))
```
pynicolas, line 53: The `NullPointerException` we're fixing illustrates rather well the comment I wrote on another PR: https://github.com/SonarSource/sonar-python/pull/1380#discussion_r1104692880

pynicolas (approved):

PR #1387, opened 2023-02-16
SONARPY-1074 comment added

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/ModifiedParameterValueCheck.java
```diff
@@ -150,6 +150,8 @@ private static String getAssignedValueString(Expression defaultValue) {
     var firstLine = defaultValue.firstToken().line() - 1;
     var lastLine = defaultValue.lastToken().line() - 1;
 
+    //Multiline parameter value assignment doesn't supported
+    // cause it requires indents calculation for place where the value should be copied.
```
andreaguarino, line 154: I'd suggest to update the comment:
```suggestion
    // We decided to not support multiline default parameters because it requires indents calculation for place where the value should be copied.
```

andreaguarino (approved):
LGTM, please have a look at my suggestion before merging

PR #1393, opened 2023-02-22
 SONARPY-1287 Introduce a new sensor for IPython notebook files 

guillaume-dequenne-sonarsource:


sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPynb.java
```diff
@@ -0,0 +1,38 @@
...
+  public static final String KEY = "ipynb";
+
+  private static final String[] DEFAULT_FILE_SUFFIXES = { ".ipynb" };
+
+  public IPynb() {
+    super(KEY, "IPynb");
+  }
+
+  @Override
+  public String[] getFileSuffixes() {
```
guillaume-dequenne-sonarsource, line 35: Should we allow for custom file suffixes like for Python?
In my opinion, no: since we expect to receive files that were pre-processed by SonarLint, we probably want to avoid dealing with unexpected files that may have bypassed SonarLint processing.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPynb.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.sonar.plugins.python;
+
+import org.sonar.api.resources.AbstractLanguage;
+
+public class IPynb extends AbstractLanguage {
+
+  public static final String KEY = "ipynb";
+
+  private static final String[] DEFAULT_FILE_SUFFIXES = { ".ipynb" };
```
guillaume-dequenne-sonarsource, line 28: Should the default file suffix be `.ipynb`, `ipynb` or something else?
Still unclear at this point.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPynbProfile.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package org.sonar.plugins.python;
+
+import org.sonar.api.server.profile.BuiltInQualityProfilesDefinition;
+import org.sonar.python.checks.CheckList;
+import org.sonarsource.analyzer.commons.BuiltInQualityProfileJsonLoader;
+
+import static org.sonar.plugins.python.PythonRuleRepository.RESOURCE_FOLDER;
+
+public class IPynbProfile implements BuiltInQualityProfilesDefinition {
```
guillaume-dequenne-sonarsource, line 28: Same as `PythonProfile` for now, without DBD and security rules. [SONARPY-1290](https://sonarsource.atlassian.net/browse/SONARPY-1290) should remove the identified noisy rules.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPynbSensor.java
```diff
@@ -0,0 +1,79 @@
...
+    descriptor
+      .onlyOnLanguage(IPynb.KEY)
+      .name("IPython Notebooks Sensor");
+  }
+
+  @Override
+  public void execute(SensorContext context) {
+    List<InputFile> pythonFiles = getInputFiles(context);
+    Optional<String> pythonVersionParameter = context.config().get(PYTHON_VERSION_KEY);
+    pythonVersionParameter.ifPresent(value -> ProjectPythonVersion.setCurrentVersions(PythonVersionUtils.fromString(value)));
```
maksim-grebeniuk-sonarsource, line 67: maybe is could be rewritten like:
```java
context.config().get(PYTHON_VERSION_KEY)
  .map(PythonVersionUtils:fromString)
  .ifPresent(ProjectPythonVersion:setCurrentVersions)
```

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPynbSensor.java
```diff
@@ -0,0 +1,79 @@
...
+    PythonScanner scanner = new PythonScanner(context, checks, fileLinesContextFactory, noSonarFilter, indexer);
+    scanner.execute(pythonFiles, context);
+  }
+
+  private static List<InputFile> getInputFiles(SensorContext context) {
+    FilePredicates p = context.fileSystem().predicates();
+    Iterable<InputFile> it = context.fileSystem().inputFiles(p.and(p.hasLanguage(IPynb.KEY)));
+    List<InputFile> list = new ArrayList<>();
+    it.forEach(list::add);
+    return Collections.unmodifiableList(list);
```
maksim-grebeniuk-sonarsource, line 77: could be simplified:
```java
return StreamSupport.stream(context.fileSystem().inputFiles(p.and(p.hasLanguage(IPynb.KEY))).spliterator(), false)
      .collect(Collectors.toUnmodifiableList());
```

sonar-python-plugin/src/test/resources/org/sonar/plugins/python/ipynb/file1.ipynb
```diff
@@ -0,0 +1,24 @@
+# Integers
```
andreaguarino, line 1: Do we need all this code snippet? can we maybe simplify it a bit?

andreaguarino (approved):
LGTM, please have a look at my comments before merging

sonar-python-plugin/src/test/resources/org/sonar/plugins/python/ipynb/file1.ipynb
guillaume-dequenne-sonarsource: It's a copy of the one used for the `PythonSensorTest`. I want to ensure the `IPynbSensor` works correctly on regular Python files. I think it would make sense to update this snippet (or even have a separate file) to have some IPython specific syntax once we start to support it.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPynbSensor.java
```diff
@@ -0,0 +1,79 @@
...
+  @Override
+  public void execute(SensorContext context) {
+    List<InputFile> pythonFiles = getInputFiles(context);
+    Optional<String> pythonVersionParameter = context.config().get(PYTHON_VERSION_KEY);
+    pythonVersionParameter.ifPresent(value -> ProjectPythonVersion.setCurrentVersions(PythonVersionUtils.fromString(value)));
+    PythonScanner scanner = new PythonScanner(context, checks, fileLinesContextFactory, noSonarFilter, indexer);
+    scanner.execute(pythonFiles, context);
+  }
+
+  private static List<InputFile> getInputFiles(SensorContext context) {
```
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: As we discussed, there is no obvious place to put such shared method and creating a `SensorUtils` class just for this feels a bit overkill. Since I don't think we'll duplicate this method again, I suggest to live with the duplication for now (and revisit if we have need for a `SensorUtils` class or if we duplicate the method again).

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPynbSensor.java
guillaume-dequenne-sonarsource: If you're okay, I'll keep the original version, which I find a bit clearer.

PR #1395, opened 2023-02-27
SONARPY-1290 Remove noisy rules from IPython quality profile

guillaume-dequenne-sonarsource:


sonar-python-plugin/src/test/java/org/sonar/plugins/python/IPynbProfileTest.java
```diff
@@ -38,5 +38,8 @@ public void profile() {
     assertThat(profile.rules()).extracting("repoKey").containsOnly("ipython");
     assertThat(profile.rules()).hasSizeGreaterThan(25);
     assertThat(profile.rules()).extracting(BuiltInQualityProfilesDefinition.BuiltInActiveRule::ruleKey).contains("S100");
+    assertThat(profile.rules()).extracting(BuiltInQualityProfilesDefinition.BuiltInActiveRule::ruleKey).doesNotContain("S905");
```
andreaguarino, line 41: what about having a test similar to `IPynbSensorTest` where we assert issues for those rules are not reported?

andreaguarino (approved):

sonar-python-plugin/src/test/java/org/sonar/plugins/python/IPynbProfileTest.java
guillaume-dequenne-sonarsource: I'm not sure I understand. In our sensor tests we generally manually activate the rules we want to enable. If the rule is enabled, I would expect it to raise an issue. Here, we simply test that the rule is left out of the default quality profile.
Basically what I'm testing is that the default quality profile for IPython is the same as the one for Python, except for those rules we filter out. My assumption is that, for now, we will not deliver new rules in IPython that would be excluded from Python (making IPython profile a strict subset of the Python one).

andreaguarino: I was thinking having a test where we activate all default quality profile rules and assert no issue is raised for those 3 rules. I don't think it's strictly necessary however it can makes things more explicit, wdyt?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPynbProfile.java
```diff
@@ -29,11 +33,23 @@ public class IPynbProfile implements BuiltInQualityProfilesDefinition {
...
   static final String PROFILE_NAME = "Sonar way";
   static final String PROFILE_LOCATION = RESOURCE_FOLDER + "/Sonar_way_profile.json";
+  static final Set<String> DISABLED_RULES = Set.of("S905", "S2201", "S5754");
 
   @Override
   public void define(Context context) {
     NewBuiltInQualityProfile profile = context.createBuiltInQualityProfile(PROFILE_NAME, IPynb.KEY);
     BuiltInQualityProfileJsonLoader.load(profile, CheckList.IPYTHON_REPOSITORY_KEY, PROFILE_LOCATION);
+    Collection<NewBuiltInActiveRule> activeRules = profile.activeRules();
+    List<NewBuiltInActiveRule> disabledRules = activeRules.stream().filter(IPynbProfile::isDisabled).collect(Collectors.toList());
```
maksim-grebeniuk-sonarsource, line 43: could be replaced with
```java
profile.activeRules().removeIf(IPynbProfile::isDisabled);
```

sonar-python-plugin/src/test/java/org/sonar/plugins/python/IPynbProfileTest.java
```diff
@@ -38,5 +38,8 @@ public void profile() {
     assertThat(profile.rules()).extracting("repoKey").containsOnly("ipython");
```
maksim-grebeniuk-sonarsource, line 38: I see that "ipython" uses in many different places. Probably it is good to have constant for it? Like as in `CheckList.IPYTHON_REPOSITORY_KEY`

guillaume-dequenne-sonarsource: Good point. I've updated some of the usages. In tests, I think it makes sense to use the literal for explicit expectations, though.

sonar-python-plugin/src/test/java/org/sonar/plugins/python/IPynbProfileTest.java
guillaume-dequenne-sonarsource: I am under the impression this would not be so easy to do outside of an integration test. In our regular sensor tests, we build the check factory manually, with the rules activated manually as well.

PR #1396, opened 2023-02-28
SONARPY-975 Compute checksum to know if typeshed_serializer should be run

andreaguarino:


python-frontend/typeshed_serializer/tox_runner.py
```diff
@@ -0,0 +1,49 @@
...
+
+def update_checksum():
+    with open(CHECKSUM_FILE, 'w') as file:
+        file.write(compute_checksum())
+
+
+def main():
+    previous_checksum = read_previous_checksum()
+    current_checksum = compute_checksum()
+    if previous_checksum is None or previous_checksum != current_checksum:
```
maksim-grebeniuk-sonarsource, line 39: Is it needed to check for `None` here? Can `previous_checksum` and `current_checksum` be `None`? If not, then we can compare them.

andreaguarino: only `previous_checksum` can be `None` and that might happen when there is not "checksum" file to be found. 
It's true that `current_checksum` can never be `None` so we might simplify the code to `if previous_checksum != current_checksum` however in Python is good practice to use `is` / `is not` operators with `None` instead of `==` / `!=`
Does it make sense?

andreaguarino:
Closing this PR as it was meant to be a POC. SONARPY-975 is the ticket to implement a production-ready version of it.

PR #1397, opened 2023-03-01
SONARPY-1291 Support dynamic object information syntax

maksim-grebeniuk-sonarsource:


python-frontend/src/main/java/org/sonar/plugins/python/api/tree/Tree.java
```diff
@@ -246,7 +246,10 @@ enum Kind {
     KEY_VALUE_PATTERN(KeyValuePattern.class),
     TOKEN(Token.class),
     LINE_MAGIC(LineMagic.class),
-    LINE_MAGIC_STATEMENT(LineMagicStatement.class);
+    LINE_MAGIC_STATEMENT(LineMagicStatement.class),
+    DYNAMIC_OBJECT_INFO_STATEMENT(DynamicObjectInfoStatement.class),
```
guillaume-dequenne-sonarsource, line 250: Probably you can replace the comma here by semicolon and remove the extra one from line 252.

python-frontend/src/main/java/org/sonar/python/api/PythonPunctuator.java
```diff
@@ -83,7 +83,9 @@ public enum PythonPunctuator implements TokenType {
   LEFT_ASSIGN("<<="),
   MUL_MUL_ASSIGN("**="),
   MATRIX_MULT_ASSIGN("@="),
-  WALRUS_OPERATOR(":=");
+  WALRUS_OPERATOR(":="),
+  QUESTION_MARK("?")
```
guillaume-dequenne-sonarsource, line 87: I think this will make us lex `?` as a valid Python punctuator even outside of IPython context.
Can we avoid that?
If we can't avoid it easily, I'd suggest separating this from the other punctuators and to add a comment that this is used for IPython dynamic object information.

python-frontend/src/main/java/org/sonar/python/tree/DynamicObjectInfoStatementImpl.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.sonar.plugins.python.api.tree.DynamicObjectInfoStatement;
+import org.sonar.plugins.python.api.tree.Tree;
+import org.sonar.plugins.python.api.tree.TreeVisitor;
+
+public class DynamicObjectInfoStatementImpl extends PyTree implements DynamicObjectInfoStatement {
+
+
+  private final List<Tree> questionMarksBefore;
```
guillaume-dequenne-sonarsource, line 31: I think for the question mark tokens we can be more specific and say it's a list of `Token`.

guillaume-dequenne-sonarsource (changes requested):
Looks good.
Left some minor comments. Also I'd like to explore the possibility of not declaring a new punctuator (`?`) for "standard" Python, if possible.

python-frontend/src/main/java/org/sonar/python/api/PythonPunctuator.java
maksim-grebeniuk-sonarsource: done

python-frontend/src/main/java/org/sonar/plugins/python/api/tree/Tree.java
maksim-grebeniuk-sonarsource: done

python-frontend/src/main/java/org/sonar/python/tree/DynamicObjectInfoStatementImpl.java
maksim-grebeniuk-sonarsource: done

python-frontend/src/main/java/org/sonar/python/api/IPythonGrammarBuilder.java
```diff
@@ -79,7 +102,8 @@ protected void simpleStatement(LexerfulGrammarBuilder b) {
       IMPORT_STMT,
       GLOBAL_STMT,
       NONLOCAL_STMT,
-      LINE_MAGIC_STATEMENT));
+      LINE_MAGIC_STATEMENT
+      ));
```
guillaume-dequenne-sonarsource, line 106: This change looks unnecessary.

python-frontend/src/main/java/org/sonar/python/tree/DynamicObjectInfoStatementImpl.java
```diff
@@ -0,0 +1,61 @@
...
+import org.sonar.plugins.python.api.tree.Token;
+import org.sonar.plugins.python.api.tree.Tree;
+import org.sonar.plugins.python.api.tree.TreeVisitor;
+
+public class DynamicObjectInfoStatementImpl extends PyTree implements DynamicObjectInfoStatement {
+
+
+  private final List<Token> questionMarksBefore;
+  private final List<Tree> children;
+  private final List<Token> questionMarksAfter;
```
guillaume-dequenne-sonarsource, line 34: Since we won't do anything with those question marks for now, and we don't provide API to access them, I think it's okay not to differentiate them from the other tokens and just to maintain a `List<Token>` here. This would also make the creation of the object much simpler from the `IPythonTreeMaker` (just collect and add all tokens).

guillaume-dequenne-sonarsource (approved):
Looks good, but I think we can make things a bit simpler (I don't really see the need to separate the question marks from the other tokens for now).

python-frontend/src/main/java/org/sonar/python/api/IPythonGrammarBuilder.java
maksim-grebeniuk-sonarsource: done

PR #1399, opened 2023-03-02
SONARPY-1294 Support parsing of cell-magic commands

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/api/IPythonGrammarBuilder.java
```diff
@@ -40,20 +45,32 @@
...
+  public Grammar create() {
+    LexerfulGrammarBuilder b = LexerfulGrammarBuilder.create();
+
+    b.rule(FILE_INPUT).is(b.zeroOrMore(b.firstOf(CELL, MAGIC_CELL, NEWLINE, IPYNB_CELL_DELIMITER)), EOF);
+
+    setupIPythonNotebooksRules(b);
+    setupCommonRules(b);
+
+    b.setRootRule(FILE_INPUT);
+    return b.buildWithMemoizationOfMatchesForAllRules();
```
maksim-grebeniuk-sonarsource, line 67: probably it is better to extend `setupFileInput` method instead of override whole `create` method with all the code almost the same as in the parent builder

python-frontend/src/main/java/org/sonar/python/api/IPythonGrammarBuilder.java
```diff
@@ -40,20 +45,32 @@
...
   }
 
-  protected void lineMagic(LexerfulGrammarBuilder b) {
+  protected void setupIPythonNotebooksRules(LexerfulGrammarBuilder b) {
```
maksim-grebeniuk-sonarsource, line 70: in another PR there is a separate method for another IPython-specific grammar, I think is better to keep methods responsibility more atomic instead of one method for everything

python-frontend/src/main/java/org/sonar/python/lexer/IPynbCellDelimiterChannel.java
```diff
@@ -0,0 +1,79 @@
...
+import com.sonar.sslr.impl.Lexer;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.sonar.python.api.PythonTokenType;
+import org.sonar.sslr.channel.Channel;
+import org.sonar.sslr.channel.CodeReader;
+
+public class IPynbCellDelimiterChannel extends Channel<Lexer> {
+
+  private final StringBuilder tmpBuilder = new StringBuilder();
```
maksim-grebeniuk-sonarsource, line 32: it is a personal preference but in my opinion, if we have a constructor, it is better to initialize all the fields in it, to have a common place for initialization.

python-frontend/src/main/java/org/sonar/python/tree/IPythonTreeMaker.java
```diff
@@ -20,18 +20,44 @@
...
+    return pyFileInputTree;
+  }
+
+  private void addStatementsFromCell(List<Statement> statements, AstNode cell) {
+    if (cell.is(IPythonGrammar.CELL)) {
+      getStatements(cell).stream().map(this::statement).forEach(statements::add);
+    } else {
+      statements.add(cellMagicStatement(cell.getFirstChild(IPythonGrammar.CELL_MAGIC_STATEMENT)));
+    }
+  }
```
maksim-grebeniuk-sonarsource, line 59: could be simplified:
```java
@Override
  public FileInput fileInput(AstNode astNode) {
    var statementList = astNode.getChildren(IPythonGrammar.CELL, IPythonGrammar.MAGIC_CELL)
      .stream()
      .flatMap(this::getStatementsFromCell)
      .collect(Collectors.collectingAndThen(Collectors.toList(), l -> l.isEmpty()? null : new StatementListImpl(l)));
    Token endOfFile = toPyToken(astNode.getFirstChild(GenericTokenType.EOF).getToken());
    FileInputImpl pyFileInputTree = new FileInputImpl(statementList, endOfFile, DocstringExtractor.extractDocstring(statementList));
    setParents(pyFileInputTree);
    pyFileInputTree.accept(new ExceptGroupJumpInstructionsCheck());
    return pyFileInputTree;
  }

  private Stream<Statement> getStatementsFromCell(AstNode cell) {
    if (cell.is(IPythonGrammar.CELL)) {
      return getStatements(cell).stream().map(this::statement);
    } else {
      return Stream.of(cellMagicStatement(cell.getFirstChild(IPythonGrammar.CELL_MAGIC_STATEMENT)));
    }
  }
```

python-frontend/src/main/java/org/sonar/python/tree/IPythonTreeMaker.java
```diff
@@ -51,6 +77,16 @@ protected Expression annotatedRhs(AstNode annotatedRhs) {
     return super.annotatedRhs(annotatedRhs);
   }
 
+  private static CellMagicStatementImpl cellMagicStatement(AstNode astNode) {
```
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: also probably it could be non static

python-frontend/src/main/java/org/sonar/python/api/IPythonGrammarBuilder.java
guillaume-dequenne-sonarsource: I kind of wanted to group IPython rules to a method, similar to how we already group rules for `statements`, `expressions`, etc.. in one method per category rather than one method per rule.
Although it's true that here, the various elements from the `IPython` category could be broken down further (in cells, statements and expressions for example), but I don't think we'll have so much more rules, and that should not affect parsing results either.

python-frontend/src/main/java/org/sonar/python/lexer/IPynbCellDelimiterChannel.java
guillaume-dequenne-sonarsource: That's a fair point, and I must admit I don't have a strong opinion on it. Though in this case to be honest I just went with consistency with other lexer channels (like `IndentationChannel`, `StringLiteralsChannel`, etc...)

python-frontend/src/main/java/org/sonar/python/tree/IPythonTreeMaker.java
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: Good catch for the interface!
As for static, it's actually our recommendation (SQ raises a code smell) to make methods static when possible, that's why I changed it (I initially had declared it non-static ).

python-frontend/src/main/java/org/sonar/python/api/IPythonGrammarBuilder.java
guillaume-dequenne-sonarsource: As discussed, I went with a slightly different approach and just decided to define all rules in a `setupRules` method and to overload that. I also created a `fileInput` method to overload as well.

maksim-grebeniuk-sonarsource:
LGTM!

maksim-grebeniuk-sonarsource (approved):

PR #1400, opened 2023-03-02
SONARPY-1293 Support line continuation in line-magic commands

maksim-grebeniuk-sonarsource:


python-frontend/src/main/java/org/sonar/python/api/PythonPunctuator.java
```diff
@@ -83,8 +83,8 @@ public enum PythonPunctuator implements TokenType {
   LEFT_ASSIGN("<<="),
   MUL_MUL_ASSIGN("**="),
   MATRIX_MULT_ASSIGN("@="),
-  WALRUS_OPERATOR(":=");
-
+  WALRUS_OPERATOR(":="),
+  BACK_SLASH("\\")
```
guillaume-dequenne-sonarsource, line 87: I'm a bit surprised we need to declare the backslash as a Python punctuator.
This line continuation logic already exists for regular Python, I feel we should be able to make it work for line magic as well somehow, without declaring new tokens or punctuators.

guillaume-dequenne-sonarsource (changes requested):
There is a failing test regarding punctuator list size.
Also, I have some doubts about the approach, I feel that we should be able to make this work without introducing new punctuators. I haven't looked at this in detail though, we can probably discuss this.

python-frontend/src/main/java/org/sonar/python/api/PythonPunctuator.java
maksim-grebeniuk-sonarsource: done

python-frontend/src/test/java/org/sonar/python/tree/IPythonTreeMakerTest.java
```diff
@@ -39,21 +40,23 @@ public void line_magic() {
     var parse = parseIPython("print(b)\n" +
       "a = %alias showPath pwd && ls -a\n", treeMaker::fileInput);
     var statementList = findFirstChildWithKind(parse, Tree.Kind.STATEMENT_LIST);
```
guillaume-dequenne-sonarsource, line 42: By the way, instead of this, you can access the statement list simply with `parse.statements().statements()` which I think is a bit more explicit.

python-frontend/src/test/java/org/sonar/python/tree/IPythonTreeMakerTest.java
```diff
@@ -39,21 +40,23 @@ public void line_magic() {
...
-    assertThat(statementList.children()).hasSize(2);
-    assertThat(statementList.children().get(0).getKind()).isEqualTo(Tree.Kind.EXPRESSION_STMT);
-    assertThat(statementList.children().get(1).getKind()).isEqualTo(Tree.Kind.ASSIGNMENT_STMT);
-    var lineMagic = findFirstChildWithKind(statementList.children().get(1), Tree.Kind.LINE_MAGIC);
+    var statements = statementList.children();
```
guillaume-dequenne-sonarsource, line 43: (To be honest, I kind of would have preferred to see this change on the initial PR which introduced these tests - I think it's okay to keep it as it is, but just to let you know).

python-frontend/src/test/java/org/sonar/python/tree/IPythonTreeMakerTest.java
```diff
@@ -93,36 +98,43 @@ public void line_magic_statement() {
...
+    assertThat(statements.get(0).getKind()).isEqualTo(Tree.Kind.EXPRESSION_STMT);
+    assertThat(statements.get(1).getKind()).isEqualTo(Tree.Kind.LINE_MAGIC_STATEMENT);
+    assertThat(statements.get(2).getKind()).isEqualTo(Tree.Kind.COMPOUND_ASSIGNMENT);
+    assertThat(statements.get(3).getKind()).isEqualTo(Tree.Kind.ASSIGNMENT_STMT);
+    assertThat(findFirstChildWithKind(statements.get(3), Tree.Kind.MODULO)).isNotNull();
 
     parse = parseIPython("print(b)\n" +
-      "%timeit a = foo(b); b = 2\n" +
-      "a += b", treeMaker::fileInput);
+      "%timeit a =\\\n" +
```
guillaume-dequenne-sonarsource, line 110: I don't like that we change existing tests to introduce more complex use cases in them. I think there is value in having multiple tests testing a single functionality (closer to the spirit of a unit test). 

I also think it makes things more difficult to review, especially since there are a lot of minor refactoring in the same commit. Adding a single simple test would make the review trivial, whereas here it's more complex than it should.

guillaume-dequenne-sonarsource (changes requested):
This looks good, but please have a look at my comments as I think we can make tests a bit clearer by keeping them separate.

python-frontend/src/test/java/org/sonar/python/tree/IPythonTreeMakerTest.java
maksim-grebeniuk-sonarsource: great! done)

python-frontend/src/test/java/org/sonar/python/tree/IPythonTreeMakerTest.java
maksim-grebeniuk-sonarsource: agreed

python-frontend/src/test/java/org/sonar/python/tree/IPythonTreeMakerTest.java
maksim-grebeniuk-sonarsource: done, slit up to two tests

python-frontend/src/test/java/org/sonar/python/tree/IPythonTreeMakerTest.java
```diff
@@ -121,14 +115,11 @@ public void line_magic_statement() {
 
   @Test
   public void system_shell_access() {
-    var parse = parseIPython("print(b)\n" +
+    var statements = parseIPython("print(b)\n" +
```
guillaume-dequenne-sonarsource, line 118: Sorry, it wasn't clear in my previous comment, but here as well there is a test that got complexified. I think it could be nice to keep the previous test as well.

guillaume-dequenne-sonarsource (approved):
Just left a last comment about another test that I think we should reintroduce (sorry that it was not explicit from previous review).
Feel free to squash and merge after.

PR #1401, opened 2023-03-03
MMF-3039

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/plugins/python/api/tree/LineMagic.java
```diff
@@ -0,0 +1,24 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.plugins.python.api.tree;
+
+public interface LineMagic extends Expression {
```
andreaguarino, line 22: I would expect this to be an `ExpressionStatement`, is this done intentionally?

python-frontend/src/main/java/org/sonar/plugins/python/api/tree/BaseTreeVisitor.java
```diff
@@ -561,4 +561,19 @@ public void visitKeywordPattern(KeywordPattern keywordPattern) {
...
+
+  @Override
+  public void visitLineMagicStatement(LineMagicStatement lineMagicStatement) {
+    // noop
+  }
+
+  @Override
+  public void visitCellMagicStatement(CellMagicStatement lineMagicStatement) {
+    // noop
+  }
```
andreaguarino: I'm not sure if it's a good idea exposing `IPython` nodes in the public API for the time being, especially because we don't intend to use them for anything useful so far.
I would instead suggest to:
- either remove those strongly typed node from the AST as long as we don't need them
- or create a private `IPythonTree` extending `Tree` which contains the `IPython` specific visitor, nodes, ...

python-frontend/src/main/java/org/sonar/python/parser/PythonParser.java
```diff
@@ -69,18 +76,19 @@ private static class SslrPythonParser extends Parser<Grammar> {
...
+      this.lexer = lexer;
     }
 
     @Override
     public AstNode parse(String source) {
       lexerState.reset();
       lexer.lex(source);
-      return super.parse(tokens());
+      List<Token> tokens = tokens();
+      return super.parse(tokens);
```
andreaguarino, line 91: why this change?

python-frontend/src/main/java/org/sonar/python/api/IPythonGrammarBuilder.java
```diff
@@ -0,0 +1,103 @@
...
+    b.rule(FILE_INPUT).is(b.zeroOrMore(b.firstOf(CELL, MAGIC_CELL, NEWLINE, IPYNB_CELL_DELIMITER)), EOF);
+  }
+
+  @Override
+  protected void setupRules(LexerfulGrammarBuilder b) {
+    ipythonRules(b);
+    super.setupRules(b);
+  }
+
+  protected void ipythonRules(LexerfulGrammarBuilder b) {
```
andreaguarino, line 71:
```suggestion
  protected void iPythonRules(LexerfulGrammarBuilder b) {
```

andreaguarino (commented):
Overall looks good, my main concern is about exposing those newly created nodes in the public API

python-frontend/src/main/java/org/sonar/plugins/python/api/tree/LineMagic.java
maksim-grebeniuk-sonarsource: Line magic could be part of an expression as well as a statement like:
```python
# line magic statement
%time a = foo()
# lime magic expression
a = %time foo()
```

That's why I've implemented it as LineMagic which is Expression and LineMagicStatement which is a statement containing an expression

python-frontend/src/main/java/org/sonar/python/parser/PythonParser.java
maksim-grebeniuk-sonarsource: this change was done to make it easier to debug parsing, but could be removed for sure if you think is required

python-frontend/src/main/java/org/sonar/plugins/python/api/tree/LineMagic.java
guillaume-dequenne-sonarsource: As we discussed, we don't actually have nodes extending `ExpressionStatement` (except for `ExpressionStatementImpl`). The usual way we've dealt with expression statements (like `CallExpression`) is by having them defined as `Expression` and nesting them into an `ExpressionStatement`.
I expect this is what we should do here as well (and thus, ditch the concept of `LineMagicStatement`).

python-frontend/src/main/java/org/sonar/plugins/python/api/tree/BaseTreeVisitor.java
guillaume-dequenne-sonarsource: As we discussed, we can do without the visitors. We also can flag the interfaces for the strongly typed nodes as `Beta`. It's still useful to have them (for tests, for example), and we suspect we'll need them in the future, so if that's okay, I'd suggest keeping them as `Beta` for now.

python-frontend/src/main/java/org/sonar/python/parser/PythonParser.java
guillaume-dequenne-sonarsource: (We chose to leave this unchanged)

andreaguarino (approved):

PR #1405, opened 2023-03-06
Fix to MMF-3039 after review

guillaume-dequenne-sonarsource:


python-frontend/src/test/java/org/sonar/python/tree/IPythonTreeMakerTest.java
```diff
@@ -302,6 +288,16 @@ public void assignmentRhs() {
     assertThat(assignments.get(2).assignedValue().getKind()).isEqualTo(Tree.Kind.MODULO);
   }
 
+  private static void assertLineMagic(Statement statement) {
```
guillaume-dequenne-sonarsource, line 291: _Nitpick:_ Despite removing the concept of `LineMagicStatement`, I think we could name this test `assertLineMagicStatement`, because it doesn't actually assert line magics as expression. It only assert them when they are used as statement.

guillaume-dequenne-sonarsource (commented):
Looks good, just a small comment.
Also, I can't approve the PR myself, since I opened it 😄.

maksim-grebeniuk-sonarsource (approved):
LGTM

PR #1412, opened 2023-03-20
SONARPY-1301 Regex patterns following a possessive quantifier should not always fail

maksim-grebeniuk-sonarsource:


python-checks/src/test/resources/checks/regex/possessiveQuantifierContinuationCheck.py
```diff
@@ -0,0 +1,7 @@
+import re
```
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: done

python-checks/src/test/resources/checks/regex/possessiveQuantifierContinuationCheck.py
```diff
@@ -1,7 +1,27 @@
 import re
 
-match1 = re.match(r"a++abc", "aaaabc", re.DOTALL) # Noncompliant {{Change this impossible to match sub-pattern that conflicts with the previous possessive quantifier.}}
-match2 = re.match(r"\d*+[02468]", "1234", re.DOTALL) # Noncompliant {{Change this impossible to match sub-pattern that conflicts with the previous possessive quantifier.}}
+def foo():
+    match1 = re.match(r"a++abc", "aaaabc", re.DOTALL) # Noncompliant {{Change this impossible to match sub-pattern that conflicts with the previous possessive quantifier.}}
```
guillaume-dequenne-sonarsource, line 4: Could you also add an assertion on issue location for one of the issues?
In general, it's important that we have at least one assertion on the issue message and one for the primary (and secondary when applicable) location for each rule test.

python-checks/src/test/resources/checks/regex/possessiveQuantifierContinuationCheck.py
```diff
@@ -1,7 +1,27 @@
...
+    match2 = re.match(r"\d*+[02468]", "1234", re.DOTALL) # Noncompliant {{Change this impossible to match sub-pattern that conflicts with the previous possessive quantifier.}}
 
-match3 = re.match(r"aa++bc", "aaaabc", re.DOTALL) # Compliant, for example it can match "aaaabc"
-match4 = re.match(r"\d*+(?<=[02468])", "1234", re.DOTALL) # Compliant, for example, it can match an even number like "1234"
+    match3 = re.match(r"aa++bc", "aaaabc", re.DOTALL) # Compliant, for example it can match "aaaabc"
+    match4 = re.match(r"\d*+(?<=[02468])", "1234", re.DOTALL) # Compliant, for example, it can match an even number like "1234"
+
+    re.match(r"a+abc", "aaaabc", re.DOTALL)
```
guillaume-dequenne-sonarsource, line 10: Do we need to have this `re.DOTALL` flag in all tests (as well as in the rule description examples)?
In my opinion, if we don't need it, we shouldn't have it (at least not everywhere), as it adds a bit of noise.

I don't think it matters much and we can keep it like this, but this is more of a general point that we should try to keep examples as minimal as possible.

guillaume-dequenne-sonarsource (approved):
Looks good, I left some last comments, not that important for this particular PR, but more of an overall thing to keep in mind when writing rules tests.
Sorry I didn't mention these point on the 1st review.

PR #1415, opened 2023-03-20
SONARPY-1306 Add quick fix for S6397 (SingleCharCharacterClassCheck)

maksim-grebeniuk-sonarsource:


python-checks/src/test/java/org/sonar/python/checks/regex/SingleCharCharacterClassCheckTest.java
```diff
@@ -28,5 +29,16 @@ public class SingleCharCharacterClassCheckTest {
...
+
+  @Test
+  public void quickFixTest() {
+    var before = "import re\n" +
+      "changed = re.match(r\"[B]\", input)";
+    var after = "import re\n" +
+      "changed = re.match(r\"B\", input)";
+    var check = new SingleCharCharacterClassCheck();
+    PythonQuickFixVerifier.verify(check, before, after);
+    PythonQuickFixVerifier.verifyQuickFixMessages(check, before, "Replace this character class by the character itself");
```
guillaume-dequenne-sonarsource, line 41: I think `Replace this character class with the character itself` is more correct.

maksim-grebeniuk-sonarsource: done

PR #1417, opened 2023-03-20
SONARPY-1310 Add support of possessive quantifiers in regex parser for Python

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/regex/ReluctantQuantifierCheck.java
```diff
@@ -21,15 +21,29 @@
...
+
+  private static class PythonReluctantQuantifierFinder extends ReluctantQuantifierFinder {
+
+    public PythonReluctantQuantifierFinder(RegexIssueReporter.ElementIssue regexElementIssueReporter) {
+      super(regexElementIssueReporter);
+    }
+
+    @Override
+    protected String makePossessiveOrGreedy(Quantifier quantifier, boolean possessive) {
+      return super.makePossessiveOrGreedy(quantifier, false);
```
guillaume-dequenne-sonarsource, line 45: One alternative could be to check whether the Python version set for the analyzer is >= 3.11 (in which case we could actually make the quantifier possessive).
No need to do it now, but maybe we could have a comment for that.

pom.xml
```diff
@@ -93,7 +93,7 @@
     <sonar.version>9.7.1.62043</sonar.version>
     <sonar.orchestrator.version>3.40.0.183</sonar.orchestrator.version>
     <sonar.api.version>9.11.0.290</sonar.api.version>
-    <sonar-analyzer-commons.version>2.1.0.1111</sonar-analyzer-commons.version>
+    <sonar-analyzer-commons.version>2.4.0.1285</sonar-analyzer-commons.version>
```
guillaume-dequenne-sonarsource, line 96: I know this will be changed to a released version, but before merging this PR, can you update the version to the latest master build of analyzer commons (PR making `makePossessiveOrGreedy`overridable just got merged)?

guillaume-dequenne-sonarsource (approved):
Looks good.
Before merging, can you create a task ticket to update the `sonar-analyzer-commons` dependency to a released version?

python-frontend/src/test/java/org/sonar/python/regex/PythonAnalyzerRegexSourceTest.java
```diff
@@ -82,7 +84,20 @@ public void test_triple_quote_literal() {
   }
 
   @Test
-  public void multiline_string_literal() {
+  public void possessiveQuantifiers() {
```
guillaume-dequenne-sonarsource, line 87: I think it would also be nice to have a test for atomic groups (could be done as part of the other ticket, or as part of this one)

PR #1418, opened 2023-03-20
SONARPY-1303 Rule S6001: Back references in regular expressions should only refer to capturing groups that are matched before the reference

maksim-grebeniuk-sonarsource:


python-checks/src/test/resources/checks/regex/impossibleBackReferenceCheck.py
```diff
@@ -0,0 +1,9 @@
+import re
+
+pattern = re.compile(r"\1(.)", re.DOTALL) # Noncompliant {{Fix this backreference, so that it refers to a group that can be matched before it.}}
```
guillaume-dequenne-sonarsource, line 3: It would be good to have some more tests (especially as we found that some finders behaviors could be different from one language to the other like with the `\w`) and to have at least one assertion on location.

guillaume-dequenne-sonarsource (changes requested):
Implementation looks good. The description should be updated for Python behavior (cf rspec PR). Also, it would be good to have more tests.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1419, opened 2023-03-21
SONARPY-1307 Add character range validation to S6353 (VerboseRegexCheck)

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/regex/VerboseRegexCheck.java
```diff
@@ -41,15 +44,16 @@ public class VerboseRegexCheck extends AbstractRegexCheck {
   private static final String ISSUE_MESSAGE_PATTERN = ".+syntax '(.+)' instead of.+";
   private static final Pattern issueMessagePattern = Pattern.compile(ISSUE_MESSAGE_PATTERN);
   public static final String QUICK_FIX_FORMAT = "Replace with \"%s\"";
+  public static final String REDUNDANT_RANGE_MESSAGE = "Use simple character '%s' instead of '%s'.";
```
guillaume-dequenne-sonarsource, line 47: I'd suggest `Replace with '%s'` for the issue message.

maksim-grebeniuk-sonarsource: discussed, not needed

PR #1421, opened 2023-03-21
SONARPY-1314 Rule S6538: Function returns should have type hints

joke1196:


python-checks/src/main/java/org/sonar/python/checks/MandatoryFunctionReturnTypeHintCheck.java
```diff
@@ -0,0 +1,48 @@
...
+  private static final String MESSAGE = "Add a return type hint to this function declaration.";
+  private static final String CONSTRUCTOR_MESSAGE = "Annotate the return type of this constructor with `None`.";
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.FUNCDEF, ctx -> {
+      FunctionDef functionDef = (FunctionDef) ctx.syntaxNode();
+      if (functionDef.returnTypeAnnotation() == null) {
+        Name functionName = functionDef.name();
+        if ("__init__".equals(functionName.name())) {
```
andreaguarino, line 40: I think we might have an incorrect message when `__init__` is not a method (although that should be pretty rare).

```python
def __init__():
  print("hello")
```

I would suggest to check if `TreeUtils.getFunctionSymbolFromDef(functionDef).isInstanceMethod()`

python-checks/src/main/java/org/sonar/python/checks/MandatoryFunctionReturnTypeHintCheck.java
```diff
@@ -0,0 +1,48 @@
...
+          ctx.addIssue(functionName, CONSTRUCTOR_MESSAGE);
+        } else {
+          ctx.addIssue(functionName, MESSAGE);
```
andreaguarino, line 43: As discussed I think it would be nice to suggest a quick fix for this case.
Suggestion for a future ticket: we might collect inferred types of `ReturnStatement` and suggest a quickfix

andreaguarino (changes requested):
Nice job!
Please have a look at my comment.
Note that the one about quick fix can also be done in a separate PR / ticket

andreaguarino (approved):
LGTM, nice job !

PR #1422, opened 2023-03-22
SONARPY-1313 Rule S6537: Octal escape sequences should not be used in regular expressions

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/regex/OctalEscapeCheck.java
```diff
@@ -0,0 +1,48 @@
...
+import org.sonar.check.Rule;
+import org.sonar.plugins.python.api.tree.CallExpression;
+import org.sonarsource.analyzer.commons.regex.RegexParseResult;
+import org.sonarsource.analyzer.commons.regex.ast.CharacterTree;
+import org.sonarsource.analyzer.commons.regex.ast.RegexBaseVisitor;
+
+@Rule(key = "S6537")
+public class OctalEscapeCheck extends AbstractRegexCheck {
+
+  public static final String MESSAGE = "Replace with hexadecimal code";
```
guillaume-dequenne-sonarsource, line 32: I'd suggest `Consider replacing this octal escape sequence with a Unicode or hexadecimal sequence instead.`, or at least something that doesn't just suggest to replace with hexadecimal code (mainly because it's not the only fix, but also because it would be unclear for users if they actually meant to have a backreference).

python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S6537.html
```diff
@@ -0,0 +1,25 @@
+<p>Octal escapes are included in a limited form. If the first digit is a 0, or if there are three octal digits, it is considered an octal escape.
+Otherwise, it is a group reference. As for string literals, octal escapes are always at most three digits in length.</p>
```
guillaume-dequenne-sonarsource, line 2: After re-reading this part, I think it would be better for it to be a short description of the rule (octal escapes should not be used), rather than explaining what an octal escape it. I'll leave a comment/suggestion on the rspec PR, but just to let you know that I think we should update this.

guillaume-dequenne-sonarsource (changes requested):
I think we need more tests to make sure we don't raise false positives.
I also think we need some minor updates to the RSPEC and issue message.

PR #1424, opened 2023-03-22
SONARPY-1315 Rule S6540: Function parameters should have type hints

joke1196:


python-checks/src/test/resources/checks/mandatoryFunctionParameterTypeHint.py
```diff
@@ -0,0 +1,70 @@
...
+                         #^^^^^
+        pass
+
+def nested(param1: str):
+    def foo(param2): # Noncompliant {{Add a type hint to this function parameter.}}
+           #^^^^^^
+        pass
+    print("the end")
+
+def dynamic_param_list(*args, **kwargs): # Noncompliant 2
```
gyula-sallai-sonarsource, line 39: I think we should also add test cases where we have some "regular" parameters next to `*args` and `**kwargs`, and when they are not together.

```python
def foo(a, b, *args, **kwargs): ...
def bar(a, *args): ...
def baz (a, **kwargs): ...
```

Also  some parameters with a default value:
```python
def foo(x=None):
   pass
```

python-checks/src/main/java/org/sonar/python/checks/MandatoryFunctionParameterTypeHintCheck.java
```diff
@@ -0,0 +1,77 @@
...
+  private boolean isAnInstanceMethod(FunctionDefImpl functionDefImpl) {
+    return Optional.ofNullable(functionDefImpl.functionSymbol())
+      .map(FunctionSymbol::isInstanceMethod).orElse(false);
+  }
+
+  private boolean isAClassMethodParameter(Parameter parameter, FunctionDefImpl functionDefImpl) {
+    return hasName("cls", parameter) && hasAClassMethodDecorator(functionDefImpl);
+  }
+
+  private boolean hasAClassMethodDecorator(FunctionDefImpl functionDefImpl) {
```
gyula-sallai-sonarsource, line 73: We do not need to pass `FunctionDefImpl` through all these functions, as we only need the function symbol. We can just get the symbol inside the lambda in `initialize` and pass it there.

python-checks/src/main/java/org/sonar/python/checks/MandatoryFunctionParameterTypeHintCheck.java
```diff
@@ -0,0 +1,77 @@
...
+public class MandatoryFunctionParameterTypeHintCheck extends PythonSubscriptionCheck {
+
+  private static final String MESSAGE = "Add a type hint to this function parameter.";
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.FUNCDEF, ctx -> {
+      FunctionDef functionDef = (FunctionDef) ctx.syntaxNode();
+      Optional.ofNullable(functionDef.parameters())
+        .ifPresent(parameterList -> removeSelfAndClassParameters(parameterList, (FunctionDefImpl) functionDef).forEach(parameter -> {
```
gyula-sallai-sonarsource, line 43: I think we can do the casing to `(FunctionDefImpl)` in the line above.

gyula-sallai-sonarsource (changes requested):
Nice job, looks good!

Please note that we do not use merge commits, could you rebase the branch instead. Also I've left a couple of smaller comments to handle.

gyula-sallai-sonarsource (approved):
Looks good to me, nice job!

PR #1426, opened 2023-03-24
SONARPY-1321 S6546: Union type expressions should be used over "typing.Union" in type hints

gyula-sallai-sonarsource:


python-checks/src/test/resources/checks/unionTypeExpression_versions.py
joke1196, line 1: Is this file used anywhere? I could not find a reference to it

python-frontend/src/main/java/org/sonar/plugins/python/api/SubscriptionContext.java
```diff
@@ -46,6 +47,12 @@ public interface SubscriptionContext {
 
   PythonFile pythonFile();
 
+  /**
+   * List of Python versions this project is compatible with.
+   */
+  @Beta
+  Set<PythonVersionUtils.Version> currentPythonVersions();
```
andreaguarino, line 54: Maybe we can call it `sourcePythonVersions`?

PR #1428, opened 2023-03-27
SONARPY-1315 Rule S6540 Function parameters should have type hints

joke1196:


python-checks/src/main/java/org/sonar/python/checks/MandatoryFunctionParameterTypeHintCheck.java
```diff
@@ -41,28 +44,41 @@ public void initialize(Context context) {
...
+    return functionSymbol != null && "__new__".equals(functionSymbol.name()) && isFirstParameter;
+  }
+
+  private static final List<String> specialTokenParams = Arrays.asList("*", "/");
+
+  private static boolean isSpecialCharParameter(Parameter parameter) {
+    Name parameterName = parameter.name();
+    Token token = parameter.starToken();
+    return (hasName("_", parameter)) ||
+      (parameterName == null && token != null && specialTokenParams.contains(token.value()));
```
gyula-sallai-sonarsource, line 81: If you want to increase coverage here, you can use the `Optional.ofNullable` trick and map to the value, but I think it's fine as it is as well.

gyula-sallai-sonarsource (approved):
Looks good to me! Please take care of the comments below before merging.

PR #1431, opened 2023-03-29
SONARPY-1233 Allow import of mypy reports

gyula-sallai-sonarsource:


sonar-python-plugin/src/main/java/org/sonar/plugins/python/mypy/MypySensor.java
```diff
@@ -0,0 +1,127 @@
...
+  private static final Logger LOG = Loggers.get(MypySensor.class);
+
+  public static final String LINTER_NAME = "Mypy";
+  public static final String LINTER_KEY = "mypy";
+  public static final String REPORT_PATH_KEY = "sonar.python.mypy.reportPaths";
+
+  private static final String FALLBACK_RULE_KEY = "unknown_mypy_rule";
+
+  // Pattern -> Location ': ' Severity ':' Message '['Code']'
+  // Location -> File ':' StartLine
```
joke1196, line 50: Small detail but we could add the StartColumn for completeness

joke1196 (approved):
LGTM!

PR #1432, opened 2023-03-30
SONARPY-1318 Rule S6545: Built-in generic types should be preferred over the typing module in type hints

joke1196:


its/ruling/src/test/java/org/sonar/python/it/PythonRulingTest.java
```diff
@@ -53,7 +53,7 @@ public class PythonRulingTest {
   public static void prepare_quality_profile() {
     ProfileGenerator.RulesConfiguration parameters = new ProfileGenerator.RulesConfiguration()
       .add("CommentRegularExpression", "message", "The regular expression matches this comment")
-      .add("S1451","headerFormat" , "# Copyright 2004 by Harry Zuzan. All rights reserved.");
+      .add("S1451", "headerFormat", "# Copyright 2004 by Harry Zuzan. All rights reserved.");
```
sallaigy, line 56: Thanks for fixing the format here :)

python-checks/src/main/java/org/sonar/python/checks/BuiltinGenericsOverTypingModuleCheck.java
```diff
@@ -0,0 +1,97 @@
...
+import org.sonar.plugins.python.api.tree.SubscriptionExpression;
+import org.sonar.plugins.python.api.tree.Tree;
+import org.sonar.plugins.python.api.tree.TypeAnnotation;
+
+import static java.util.Map.entry;
+
+@Rule(key = "S6545")
+public class BuiltinGenericsOverTypingModuleCheck extends PythonSubscriptionCheck {
+
+  private static final String MESSAGE = "Use the built-in generic type: `%s`, instead of its typing counterpart.";
```
sallaigy, line 40: _super minor, subjective_ I think we can phrase it without the `:`:
```suggestion
  private static final String MESSAGE = "Use the built-in generic type `%s` instead of its typing counterpart.";
```

Feel free to ignore this if it doesn't make sense to you.

sallaigy (approved):
Looks good to me! I've left one very minor comment, but feel free to ignore if it doesn't make sense to you.

python-checks/src/main/java/org/sonar/python/checks/BuiltinGenericsOverTypingModuleCheck.java
joke1196: Sure I think it is clear enough with the back ticks only

PR #1434, opened 2023-04-04
SONARPY-1320 Rule S6538: Add quick fix for return type hints

joke1196:


python-checks/src/test/java/org/sonar/python/checks/MandatoryFunctionReturnTypeHintCheckTest.java
```diff
@@ -20,11 +20,170 @@
...
+  @Test
+  public void testQuickFixGenericTypes() {
+    String codeWithIssue = code(
+      "def foo(alice: dict[str, int]):",
+      "    bob = {}",
+      "    for k, v in alice.items():",
+      "        bob[k] = str(v)",
+      "    return bob");
+    String codeFixed = code(
+      "def foo(alice: dict[str, int]) -> dict:",
```
joke1196, line 101: Not sure if this is the way to go. As Rule S6543 would then ask the user to provide a type parameter for `dict`. It feels a bit weird to quickly fix something which still needs fixing.

sallaigy: Indeed, I agree. I think we can be conservative here and only offer the quick fix if the type is primitive (or at least non-generic) or None.

We could use a logic similar to generics without type argument check to see if the type is generic.

python-checks/src/test/java/org/sonar/python/checks/MandatoryFunctionReturnTypeHintCheckTest.java
```diff
@@ -20,11 +20,170 @@
...
+import static org.sonar.python.checks.utils.CodeTestUtils.code;
+
 public class MandatoryFunctionReturnTypeHintCheckTest {
   @Test
   public void test() {
     PythonCheckVerifier.verify("src/test/resources/checks/mandatoryFunctionReturnType.py", new MandatoryFunctionReturnTypeHintCheck());
   }
+
+  @Test
+  public void testQuickFixAddReturnTypeLiteral() {
```
sallaigy, line 35: _minor_ The usual :)
```suggestion
  public void quick_fix_add_return_type_literal() {
```

sallaigy (changes requested):
Looks good to me, I added one major comment around types for which we should not offer the fix. Also please rebase on the current master.

python-checks/src/test/java/org/sonar/python/checks/MandatoryFunctionReturnTypeHintCheckTest.java
joke1196: Sadly, it is not as easy with the inferred type to detect if it could be a generic type. I think I will have to approach this by only taking care of nongeneric built-in types.

python-checks/src/test/java/org/sonar/python/checks/MandatoryFunctionReturnTypeHintCheckTest.java
joke1196: Sadly, it is not as easy with the inferred type to detect if it could be a generic type. I think I will have to approach this by only taking care of nongeneric built-in types.

python-checks/src/main/java/org/sonar/python/checks/MandatoryFunctionReturnTypeHintCheck.java
```diff
@@ -55,6 +73,66 @@ private static void raiseIssueForConstructor(SubscriptionContext ctx, Name funct
...
+    PreciseIssue issue = ctx.addIssue(functionName, MESSAGE);
+    ReturnStatementVisitor returnStatementVisitor = new ReturnStatementVisitor();
+    functionDef.body().accept(returnStatementVisitor);
+    if (!returnStatementVisitor.returnStatements.isEmpty()) {
+      addQuickFixForReturnType(issue, functionDef, returnStatementVisitor.returnStatements);
+    } else {
+      if (returnStatementVisitor.yieldStatements.isEmpty()) {
+        addQuickFixForNoneType(issue, functionDef);
+      }
+    }
```
sallaigy, line 89: _minor_ We can simplify this a bit:
```suggestion
    } else if (returnStatementVisitor.yieldStatements.isEmpty()) {
        addQuickFixForNoneType(issue, functionDef);
    }
```

sallaigy (approved):
Looks good to me! See one minor comment before merging.

PR #1436, opened 2023-04-05
SONARPY-1331 Rule S6545: Add quick fix to replace typing module type hints with built-in types

joke1196:


python-checks/src/test/java/org/sonar/python/checks/BuiltinGenericsOverTypingModuleCheckTest.java
```diff
@@ -36,4 +40,139 @@ public void checkCollections() {
...
+      "from typing import defaultdict",
+      "def foo() -> tuple[str, int]:",
+      "    def bar(test: collections.defaultdict[str, int]) -> str:",
+      "        return \"bar\"",
+      "    return (\"foo\", list())");
+    BuiltinGenericsOverTypingModuleCheck check = new BuiltinGenericsOverTypingModuleCheck();
+    PythonQuickFixVerifier.verify(check, codeWithIssue, codeFixed);
+    String specificMessage = String.format(MESSAGE, "collections.defaultdict");
+    PythonQuickFixVerifier.verifyQuickFixMessages(check, codeWithIssue, specificMessage);
+  }
```
sallaigy, line 177: It's a special case that came to my mind now, but could we test the case when we have an obsolete generic type without type arguments and that we do not raise an issue (and of course offer no quick fix):
```
def f() -> typing.List:
    pass
```

sallaigy, line 177: Also, could we have a test with a weird formatting?
```
def f() -> typing.Tuple[str,
    int]:
    pass
```

sallaigy (changes requested):
Looks mostly good, but I let's not offer quick fixes that would require an import in order to avoid breaking code.

python-checks/src/test/java/org/sonar/python/checks/BuiltinGenericsOverTypingModuleCheckTest.java
joke1196: Yes, good call I will add this to the main test.

python-checks/src/main/java/org/sonar/python/checks/BuiltinGenericsOverTypingModuleCheck.java
```diff
@@ -70,22 +72,38 @@ private static void checkForTypingModule(SubscriptionContext subscriptionContext
...
+
+  private static void raiseIssueForGenerics(SubscriptionContext context, SubscriptionExpression expression, String preferredGenerics) {
+    String specificMessage = String.format(MESSAGE, preferredGenerics);
+    PreciseIssue preciseIssue = context.addIssue(expression, specificMessage);
+    addQuickFix(preciseIssue, expression, preferredGenerics, specificMessage);
+  }
+
+  private static void addQuickFix(PreciseIssue issue, SubscriptionExpression expression, String preferredGenerics, String message) {
+    // Ignoring quick fix if the change would require an import
+    if (!preferredGenerics.contains(".")) {
```
sallaigy, line 90: Smart solution! :)

sallaigy (approved):
Looks good to me!

PR #1439, opened 2023-04-12
SONARPY-1279 S1144 should take into account usages apart from self

sallaigy:


python-checks/src/main/java/org/sonar/python/checks/UnreadPrivateMethodsCheck.java
```diff
@@ -61,4 +64,29 @@ protected boolean isException(Symbol symbol) {
       .flatMap(Collection::stream)
       .anyMatch(Predicate.not(BuiltinSymbols.STATIC_AND_CLASS_METHOD_DECORATORS::contains));
   }
+
+  @Override
+  protected boolean hasAmbiguousUsage(Symbol symbol, ClassDef classDef) {
+    CallNamesVisitor visitor = new CallNamesVisitor(symbol.name());
+    classDef.accept(visitor);
+
+    // There is always at least on usage - the declaration itself.
```
maksim-grebeniuk-sonarsource, line 73: probably "...at least one usage..."?

maksim-grebeniuk-sonarsource (approved):
Looks good just one thing to correct in comments

PR #1441, opened 2023-04-13
SONARPY-1326 UnusedLocalVariableCheck(1481) should raise with sequence unpacking

maksim-grebeniuk-sonarsource:


python-checks/src/test/java/org/sonar/python/checks/UnusedLocalVariableCheckTest.java
```diff
@@ -35,4 +36,19 @@ public void custom() {
     check.format = "(_|myignore)";
     PythonCheckVerifier.verify("src/test/resources/checks/unusedLocalVariableCustom.py", check);
   }
+
+  @Test
+  public void quickFixTest() {
```
sallaigy, line 41: _minor_ Test naming is all over the place in the code base, but we should try to stick to `snake_case`, and maybe indicate that the quick fix is only for sequence unpacking.
```suggestion
  public void tuple_quick_fix() {
```

sallaigy, line 41: Also, there is another edge case with the quick fix, where it slightly changes the semantics of the program:
```python
def using_tuples():
    x, y = (1, 2)
    y = 5
    print(x)
```
Replacing `y` with `_` is not erroneous, but it somewhat changes program semantics. I think we can document this case with another test.

python-checks/src/test/resources/checks/unusedLocalVariable.py
```diff
@@ -47,7 +47,8 @@ def function_with_lambdas():
 #             ^
 
 def using_tuples():
-    x, y = (1, 2)
+    x, y = (1, 2) # Noncompliant {{Replace unused local variable "y" with "_".}}
```
sallaigy, line 50: Note that the rule description also needs to be updated, see the "Exceptions" section https://sonarsource.github.io/rspec/#/rspec/S1481/python 

```
for _ in range(10):
    do_something()
username, login, password = auth # This now raises
do_something_else(username, login)
```

sallaigy (changes requested):
Looks good to me! Could you also update the rule description before merging?

python-checks/src/test/java/org/sonar/python/checks/UnusedLocalVariableCheckTest.java
maksim-grebeniuk-sonarsource: done

python-checks/src/test/resources/checks/unusedLocalVariable.py
maksim-grebeniuk-sonarsource: done

sallaigy (approved):
Looks good to me!

PR #1442, opened 2023-04-13
SONARPY-1328 UnusedLocalVariable(S1481) add quick fixes for redundant exception bindings

maksim-grebeniuk-sonarsource:


python-frontend/src/main/java/org/sonar/plugins/python/api/tree/ExceptClause.java
```diff
@@ -54,4 +55,7 @@ public interface ExceptClause extends Tree {
   Expression exceptionInstance();
 
   StatementList body();
+
+  @Beta
+  Token colon();
```
sallaigy, line 60: _minor_ I would move this above the body, just to be more in-line with the flow of the grammar, as the colon precedes the body.

sallaigy (approved):
Looks good to me! I've left one minor comment regarding the ordering in the API change.

PR #1444, opened 2023-04-13
SONARPY-975: Avoid running typeshed serializer in CI when not needed 

joke1196:


python-frontend/typeshed_serializer/tests/test_tox_runner.py
```diff
@@ -0,0 +1,219 @@
+#
```
andreaguarino, line 1: It looks like coverage has not been taken into account by SonarQube analysis.
I think that's because the configuration in the `python-frontend/pom.xml` is the following:
```xml
    <sonar.sources>pom.xml,src/main/java,typeshed_serializer/serializer</sonar.sources>
    <sonar.python.coverage.reportPaths>typeshed_serializer/cov.xml</sonar.python.coverage.reportPaths>
    <sonar.python.version>3.9</sonar.python.version>
```

So I think we should:
- either update `sonar.sources` to include `typeshed_serializer/runners`
- or move `runners` inside `typeshed_serializer/serializer`

python-frontend/typeshed_serializer/tests/test_tox_runner.py
```diff
@@ -0,0 +1,219 @@
...
+            assert text == self.FILE_CONTENT
+
+    def test_read_file(self):
+        file_bytes = bytes("\ntest end\n", 'utf-8')
+        with mock.patch(f'{self.MODULE_NAME}.Path.read_bytes') as mock_read_bytes:
+            mock_read_bytes.return_value = file_bytes
+            text = tox_runner.read_file("test")
+            assert text == file_bytes
+
+    def test_compute_checksum(self):
```
andreaguarino, line 150: I would like to see a test where we call twice actual files from `tests/resources` to `compute_checksum` and assert the return value is the same between the two calls

andreaguarino (changes requested):
Build time is taking 8 minutes, so I don't think the runner is working yet.
I guess that's because `checksum` file has not been commited

python-frontend/typeshed_serializer/runners/tox_runner.py
```diff
@@ -0,0 +1,145 @@
...
+        binary_file_names = fetch_binary_file_names()
+        current_binaries_checksum = compute_checksum(binary_file_names, read_file)
+        logger.info("STARTING TYPESHED BINARY FILES CHECKSUM COMPUTATION")
+        logger.info(f"Previous binaries checksum {previous_binaries_checksum}")
+        logger.info(f"Current binaries checksum {current_binaries_checksum}")
+        logger.info(f"Checksum is computed over {len(binary_file_names)} files")
+        if previous_binaries_checksum != current_binaries_checksum:
+            raise RuntimeError('INCONSISTENT BINARY CHECKSUMS')
+        logger.info("SKIPPING TYPESHED SERIALIZATION")
+        subprocess.run(['tox','-e', 'testenv'])
```
andreaguarino, line 141: As discussed, this should run `tox -e py39` instead. In MacOS `testenv` doesn't seem to be a valid environment

andreaguarino, line 141: Can you add a comment explaining why we need to run unit tests? (i.e. to avoid breaking quality gate when we modify python files)

andreaguarino (approved):
LGTM.
As discussed, can you please create the following two tickets:
* Improve unit tests to avoid serializing typeshed
* Investigate why a few protobuf files differ between different runs

PR #1445, opened 2023-04-13
SONARPY-1338 Fix FP on S5655 when the argument is a valid TypedDict

sallaigy:


python-checks/src/main/java/org/sonar/python/checks/FunctionReturnTypeCheck.java
```diff
@@ -146,7 +146,8 @@ public void visitReturnStatement(ReturnStatement returnStatement) {
         Expression expression = expressions.get(0);
         InferredType inferredType = expression.type();
         if (returnType.mustBeOrExtend("typing.TypedDict")) {
```
maksim-grebeniuk-sonarsource, line 148: Probably it is better to use the functionality implemented in InferredTypes to check is return type is dictionary like 
`InferredTypes.DICT.isCompatibleWith(returnType)`. 
What do you think?

maksim-grebeniuk-sonarsource: also to keep a more flat code structure probably it is better to move this if statement to the parent if-else chain like
```java
} else if (InferredTypes.DICT.isCompatibleWith(returnType)) {
  return;
} else {
  Expression expression = expressions.get(0);
  ...
}
```

maksim-grebeniuk-sonarsource (approved):
Looks good to me, just two minor short comments

python-checks/src/main/java/org/sonar/python/checks/FunctionReturnTypeCheck.java
sallaigy: The two solutions are not the same, consider the following test case:
```
  def my_dict2() -> dict:
    return [] # Noncompliant
```

With the solution above, this would be a FN. As the code is not part of this ticket scope, I would prefer just to leave it as it is (with the TODO comment). What do you think?

PR #1446, opened 2023-04-13
SONARPY-1285 Fix FP on S1451 when shebang head lines are used

maksim-grebeniuk-sonarsource:


python-checks/src/test/java/org/sonar/python/checks/FileHeaderCopyrightCheckTest.java
```diff
@@ -141,4 +141,20 @@ public void test_searchPattern_exception() {
...
+    fileHeaderCopyrightCheck.headerFormat = "# Copyright FOO\n";
+    PythonCheckVerifier.verifyNoIssue("src/test/resources/checks/fileHeaderCopyright/shebangCopyright.py", fileHeaderCopyrightCheck);
+  }
+
+  @Test
+  public void shebangPatternTest() {
+    var fileHeaderCopyrightCheck = new FileHeaderCopyrightCheck();
+    fileHeaderCopyrightCheck.isRegularExpression = true;
+    fileHeaderCopyrightCheck.headerFormat = "# Copyright[ ]20[0-9]{2}";
+    PythonCheckVerifier.verifyNoIssue("src/test/resources/checks/fileHeaderCopyright/searchPatternShebangCopyright.py",
```
sallaigy, line 157: I would also add a test where we use the shebang line, but the copyright header is noncompliant.

sallaigy (approved):
Looks good to me! Please add a test where we the shebang line is present but the header is noncompliant.

PR #1448, opened 2023-04-26
SONARPY-1342 Improve typeshed unit test to avoid serializing Typeshed

guillaume-dequenne-sonarsource:
All this PR does is using some sample files of `Typeshed` for the unit tests, instead of the full thing.
On my machine, this reduces the time taken by the unit tests from ~50s to ~10s. And I don't think we lose any meaningful confidence in the implementation by using the subset of `Typeshed` rather than the complete library.


python-frontend/typeshed_serializer/tests/conftest.py
```diff
@@ -31,6 +31,8 @@
 
 @pytest.fixture(scope="session")
 def typeshed_stdlib():
+    typeshed_serializer.STDLIB_PATH = "resources/mock_stdlib"
```
joke1196, line 34: This is the way, with the fixtures :+1:

python-frontend/typeshed_serializer/tests/conftest.py
```diff
@@ -31,6 +31,8 @@
...
+    typeshed_serializer.CURRENT_PATH = os.path.dirname(__file__)
```
joke1196, line 35: Here can we use the CURRENT PATH constant instead? Genuine question not sure when the fixture is invoked.

joke1196 (approved):
LGTM! Just added a small question. Thank you for this; the speed upgrade feels really nice!

python-frontend/typeshed_serializer/tests/conftest.py
guillaume-dequenne-sonarsource: Thanks, good catch! I somehow missed it.

PR #1449, opened 2023-04-27
SONARPY-1334 Rule S6554: Django models should define a "__str__" method

maksim-grebeniuk-sonarsource:


python-checks/src/test/resources/checks/django/djangoModelStrMethodCheck.py
```diff
@@ -0,0 +1,14 @@
+from django.db import models
```
guillaume-dequenne-sonarsource, line 1: I feel we need more tests.
We need:
* One test when we inherit from a valid model, but do not override the `__str__` method (should not raise FP)
* One test when we inherit from an invalid model (should we raise the issue twice here? On the parent and on the child? This may be noisy, but probably makes sense - other option would be to raise only on whatever directly inherits from `django.db.models.Model`, but we risk raising on abstract classes whose children are correctly defined). Whatever we do should probably be documented with tests, though.
* One test when we inherit from an unknown model (should not raise FPs)

its/ruling/src/test/java/org/sonar/python/it/RulingHelper.java
```diff
@@ -143,6 +142,7 @@ static List<String> bugRuleKeys() {
       "S6323",
       "S6328",
       "S6468",
+      "S6554",
```
guillaume-dequenne-sonarsource, line 145: We probably don't want to enable this rule on extended ruling since it's not a bug rule.
I'm torn on this one though, as it's a rule that is quite reliant on our semantics, and the extended ruling is about good semantic resolutions.

guillaume-dequenne-sonarsource (changes requested):
I have the feeling this rule may be noisy and/or controversial.

I'd like to make sure we don't raise FPs on cases where the semantic is not properly resolved by the analyzer.
I'd also like to consider whether we really want this rule on the default profile, as it seems to raise quite a few issues on the ruling, some of which could be considered FPs.

python-checks/src/test/resources/checks/django/djangoModelStrMethodCheck.py
maksim-grebeniuk-sonarsource: done

python-checks/src/main/resources/org/sonar/l10n/py/rules/python/Sonar_way_profile.json
```diff
@@ -189,6 +189,7 @@
     "S6463",
     "S6468",
     "S6537",
-    "S6546"
+    "S6546",
+    "S6554"
```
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: done

its/ruling/src/test/java/org/sonar/python/it/RulingHelper.java
maksim-grebeniuk-sonarsource: done

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1450, opened 2023-04-27
SONARPY-1333 Rule S6559: Fields of a Django ModelForm should be defined explicitly

maksim-grebeniuk-sonarsource:


its/ruling/src/test/java/org/sonar/python/it/RulingHelper.java
```diff
@@ -143,6 +142,7 @@ static List<String> bugRuleKeys() {
       "S6323",
       "S6328",
       "S6468",
+      "S6559",
```
guillaume-dequenne-sonarsource, line 145: I admit this is not properly documented (and it's not even very well designed at all) but this `RulingHelper` is meant to list only our bug rules, to be run on the extended ruling.

The rationale was that we thought our existing ruling was not sufficient to ensure proper confidence for bug rules, and yet we didn't want to bloat it and have too many projects on it.

We explicitly decided to only run bug rules on this extended ruling, so maybe it doesn't make sense to add this one here.

This whole approach can be challenged and we may want to do something about it, but for the time being I think we could stay consistent with this approach.

python-checks/src/test/resources/checks/django/djangoModelFormFieldsCheck.py
```diff
@@ -0,0 +1,29 @@
+from django import forms
+
+class ExcludeFieldsForm(forms.ModelForm):
+    class Meta:
+        model = MyModel
+        exclude = ['field1', 'field2']  # Noncompliant {{Set the fields of this form explicitly instead of using "exclude".}}
```
guillaume-dequenne-sonarsource, line 6: __Nitpick:__ we could have one assertion on location as well as message, for the sake of completeness.

maksim-grebeniuk-sonarsource: done

python-checks/src/test/resources/checks/django/djangoModelFormFieldsCheck.py
```diff
@@ -0,0 +1,29 @@
+from django import forms
```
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: done

its/ruling/src/test/java/org/sonar/python/it/RulingHelper.java
maksim-grebeniuk-sonarsource: done

PR #1451, opened 2023-04-28
SONARPY-1335 Rule S6553: "null=True": should not be used on string-based fields in Django models

maksim-grebeniuk-sonarsource:


python-checks/src/test/resources/checks/django/djangoModelStringFieldCheck.py
```diff
@@ -0,0 +1,16 @@
...
+
+class NullBlankFieldsModel(models.Model):
+    name = models.CharField(max_length=50, null=True, blank=True) # Noncompliant {{Remove this "null=True" flag.}}
+    desc = models.TextField(max_length=50, null=True, blank=True) # Noncompliant
+
+class BlankFieldsModel(models.Model):
+    name = models.CharField(max_length=50, blank=True)
+    desc = models.TextField(max_length=50, blank=True)
+
+
```
guillaume-dequenne-sonarsource, line 16: One case that seems to be missing is when both `unique=True` and `blank=True`, where we should not raise an issue.

python-checks/src/main/java/org/sonar/python/checks/django/DjangoModelStringFieldCheck.java
```diff
@@ -0,0 +1,142 @@
...
+      .filter(FIELD_TYPES_FQN::contains)
+      .isPresent();
+  }
+
+  private static void validateTextFieldArguments(SubscriptionContext ctx, CallExpression call) {
+    var nullArg = getCallArgumentByName(call, "null")
+      .filter(DjangoModelStringFieldCheck::isArgumentSetAsTrue);
+    var blankArg = getCallArgumentByName(call, "blank")
+      .filter(DjangoModelStringFieldCheck::isArgumentSetAsTrue);
+
```
guillaume-dequenne-sonarsource, line 83: We can probably look for `unique=True` here and if it's present as well as `blankArg`, we can abort.

python-checks/src/main/java/org/sonar/python/checks/django/DjangoModelStringFieldCheck.java
```diff
@@ -0,0 +1,142 @@
...
+  private static Optional<PythonQuickFix> createRemoveArgQuickFix(CallExpression call, Tree arg) {
+    return Optional.ofNullable(call.argumentList())
+      .map(Tree::children)
+      .map(args -> {
+        var count = args.size();
+        var index = args.indexOf(arg);
+        var removeFrom = arg;
+        Tree removeTo;
+        if (index == 0) {
```
guillaume-dequenne-sonarsource, line 128: I don't like too much when we do this kind of index arithmetic. If we can't simplify, maybe it's good to have at least some comments that clarify what we're doing in each branch.

guillaume-dequenne-sonarsource (changes requested):
Looking good.
There is one exception case that isn't covered that we should probably cover, though.
I left some other minor suggestions, nothing big though.

python-checks/src/test/java/org/sonar/python/checks/django/DjangoModelStringFieldCheckTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+    var after = "from django.db import models\n" +
+      "class NullFieldsModel(models.Model):\n" +
+      "    name = models.CharField(max_length=50, blank=True)";
+    PythonQuickFixVerifier.verify(check, before, after);
+    PythonQuickFixVerifier.verifyQuickFixMessages(check, before, "Replace with \"blank=True\"");
+  }
+
+  @Test
+  public void removeQuickFixTest() {
```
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: by our tests I can see that it removes the line

python-checks/src/main/java/org/sonar/python/checks/django/DjangoModelStringFieldCheck.java
maksim-grebeniuk-sonarsource: done

python-checks/src/main/java/org/sonar/python/checks/django/DjangoModelStringFieldCheck.java
maksim-grebeniuk-sonarsource: done

python-checks/src/test/resources/checks/django/djangoModelStringFieldCheck.py
maksim-grebeniuk-sonarsource: done

python-checks/src/main/java/org/sonar/python/checks/django/DjangoModelStringFieldCheck.java
```diff
@@ -0,0 +1,142 @@
...
+import org.sonar.python.quickfix.TextEditUtils;
+import org.sonar.python.tree.TreeUtils;
+
+@Rule(key = "S6553")
+public class DjangoModelStringFieldCheck extends PythonSubscriptionCheck {
+
+  private static final String REPLACE_MESSAGE = "Replace this \"null=True\" flag with \"blank=True\".";
+  private static final String REMOVE_MESSAGE = "Remove this \"null=True\" flag.";
+  private static final String REPLACE_QUICK_FIX_MESSAGE = "Replace with \"blank=True\"";
+  private static final String REMOVE_QUICK_FIX_MESSAGE = "Remove \"null\" argument";
```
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: done

python-checks/src/main/java/org/sonar/python/checks/django/DjangoModelStringFieldCheck.java
guillaume-dequenne-sonarsource: Thanks!

guillaume-dequenne-sonarsource (approved):
LGTM!
It seems you need to rebase before merging.

PR #1452, opened 2023-04-28
SONARPY-1336 Rule S6552: Django signal handler functions should have the '@receiver' decorator on top of all other decorators

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource (approved):
Looks good. Minor suggestion on some tests that can be added.

Also, I think a quick fix could be easily implemented for this rule. If you agree, could you  create a new ticket/subtask for this?
I think we should merge this PR first, though, as it's ready to merge (so if we have issues with the quick fix we can still drop this branch).

PR #1453, opened 2023-04-28
SONARPY-1330 Rule S6556: 'locals()' should not be passed to a Django 'render()' function

joke1196:


python-checks/src/test/resources/checks/djangoRenderContext.py
```diff
@@ -0,0 +1,78 @@
+from django.shortcuts import render
+
+def using_locals_from_var(request):
+    username = "alice"
+    password = "p@ssw0rd"
+    context = locals()
+    return render(request, "my_template.html", context) # Noncompliant {{Use an explicit context instead of passing "locals()" to this Django "render" call.}} [[secondary=-1]]
```
joke1196, line 7: I could not find a proper way to test for secondary messages. I would like to verify that it is displayed correctly with the correct variable name.

guillaume-dequenne-sonarsource: From the top of my head, I believe this can be done using the syntax:
`    #      <^^^^ {{Secondary message here.}}` as done for a normal primary location.
With a `>` or `<` sign to point towards the corresponding primary location.

joke1196: Perfect thank you!

python-checks/src/main/java/org/sonar/python/checks/DjangoRenderContextCheck.java
```diff
@@ -0,0 +1,107 @@
...
+        .map(Usage::tree)
+        .map(usage -> TreeUtils.firstAncestorOfKind(usage, Tree.Kind.ASSIGNMENT_STMT))
+        .collect(Collectors.toList());
+      if (assignmentStmts.size() == 1) {
+        raiseIssueLocalsIsAssigned(ctx, contextArg, assignmentStmts);
+      }
+    }
+  }
+
+  private static void raiseIssueLocalsIsAssigned(SubscriptionContext ctx, RegularArgument contextArg, List<Tree> assignmentStmts) {
```
maksim-grebeniuk-sonarsource, line 92: I guess it is possible to make this part recursive just to support multiple assignments chain like:
```python
a = locals()
b = a
c = b
return render(request, "my_template.html", content_type=None, context=c) # Issue 
```
Or add a test for this and mark it as FN

maksim-grebeniuk-sonarsource (approved):
Looks good, just have one suggestion to support multiple assignments

python-checks/src/main/java/org/sonar/python/checks/DjangoRenderContextCheck.java
joke1196: I added the recursive call but with a maximum amount of reassignment of 5. I don't it make that much sense to go any further up the tree.

PR #1454, opened 2023-05-01
SONARPY-1332 Rule S6560: The safe flag should be set to False when serializing non-dictionary objects in Django JSON-encoded responses

joke1196:


python-checks/src/main/java/org/sonar/python/checks/DjangoNonDictSerializationCheck.java
```diff
@@ -0,0 +1,114 @@
...
+      if (symbol != null && JSON_RESPONSE_FUNCTION_NAME.equals(symbol.fullyQualifiedName())) {
+        checkForDictSerialization(ctx, callExpression);
+      }
+    });
+  }
+
+  private static void checkForDictSerialization(SubscriptionContext ctx, CallExpression callExpression) {
+    RegularArgument safe = TreeUtils.nthArgumentOrKeyword(2, "safe", callExpression.arguments());
+    if (safe == null || (safe.expression().is(Tree.Kind.NAME) && "True".equals(((Name) safe.expression()).name()))) {
+      RegularArgument dataArg = getFirstArgument(callExpression.arguments());
```
guillaume-dequenne-sonarsource, line 62: Is the data argument positional only, or should we use `nthArgumentOrKeyword` here as well? (I didn't double check, this is a legitimate question)

python-checks/src/main/java/org/sonar/python/checks/DjangoNonDictSerializationCheck.java
```diff
@@ -0,0 +1,114 @@
...
+      if (dataArg != null
+        && (!dataArg.expression().is(Tree.Kind.DICTIONARY_LITERAL))
+        && (!dataArg.expression().is(Tree.Kind.CALL_EXPR)
+          && !isDictAssignedToExpression(dataArg.expression()))) {
+        ctx.addIssue(dataArg, MESSAGE);
+      }
```
guillaume-dequenne-sonarsource, line 68: Could we use type inference here instead, searching for type compatibility with dictionary types? (Might be prone to FPs though - I don't mind the current approach, this is just a suggestion).

That said, what if the data argument is a qualified expression like `my_obj.dict_field`? I have the feeling `isDictAssignedToExpression` would return false in that case, and we could raise a false positive. We'd probably want to have a test with this.
We also probably want to ensure we don't raise FPs with dictionary comprehensions.

In general, when in doubt, we should err on the side of not raising an issue, so maybe `isDictAssignedToExpression` could be renamed `couldExpressionBeADict` and return true whenever we don't know (so that if we forgot to consider an AST type, we don't raise FPs for it - and this would be consistent with the other method `couldDictBeAssignedToDataArg`).

python-checks/src/test/resources/checks/djangoNonDictSerialization.py
```diff
@@ -0,0 +1,65 @@
...
+  success6 = JsonResponse(data)
+  a = data
+  b = a
+  c = b
+  d = c
+  e = d
+  f = e
+  g = f
+  reassigned_data = g
+  success7 = JsonResponse(reassigned_data)
```
guillaume-dequenne-sonarsource, line 57: Are you testing the recursion limit here? Wouldn't it make more sense to test it with a non dictionary object then? (Here `data` is initially a dict, so it wouldn't raise even without the recursion limit, I believe).

python-checks/src/main/java/org/sonar/python/checks/DjangoNonDictSerializationCheck.java
```diff
@@ -0,0 +1,114 @@
...
+        .filter(usage -> usage.kind() == Usage.Kind.ASSIGNMENT_LHS)
+        .map(Usage::tree)
+        .map(usage -> TreeUtils.firstAncestorOfKind(usage, Tree.Kind.ASSIGNMENT_STMT))
+        .collect(Collectors.toList());
+      if (assignmentStmts.size() == 1) {
+        AssignmentStatement assignment = (AssignmentStatement) assignmentStmts.get(0);
+        // We do not yet support checking the types from call expressions
+        if (assignment.assignedValue().is(Tree.Kind.DICTIONARY_LITERAL)
+          || assignment.assignedValue().is(Tree.Kind.CALL_EXPR)
+          || assignment.assignedValue().is(Tree.Kind.QUALIFIED_EXPR)) {
```
guillaume-dequenne-sonarsource, line 103: Consistently with the other comment where we don't consider all AST nodes that could be `dict`, I'd suggest to extract this logic in a dedicated method that checks compatibility with a dictionary.
Potentially this could also make testing the type inference approach easier.

guillaume-dequenne-sonarsource (changes requested):
Looks good.
Left some minor suggestions to ensure we don't risk raising FPs.

python-checks/src/main/java/org/sonar/python/checks/DjangoNonDictSerializationCheck.java
joke1196: Exactly the data arg is positional only 
```
class JsonResponse(data, encoder=DjangoJSONEncoder, safe=True, json_dumps_params=None, **kwargs)
```

guillaume-dequenne-sonarsource: I don't think this signature makes `data` positional only. It just means that there is no default value for the parameter (positional only arguments should be before a `/` delimiter argument, like: `def my_func(pos_only, /, keyword_or_pos, *, keyword_only): ...` - although it's true that some Typeshed stubs for builtins don't follow this convention).

In this case I could actually create a JsonResponse with the following call:
```
>>> response = JsonResponse(safe=False, data=42 )
>>> response
<JsonResponse status_code=200, "application/json">
```
So I think we should actually consider this order as being possible (granted it's very unlikely to happen in real cases).

PR #1456, opened 2023-05-02
SONARPY-1347 Add quick fix for S6552 (DjangoReceiverDecoratorCheck)

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/django/DjangoReceiverDecoratorCheck.java
```diff
@@ -22,15 +22,18 @@
...
 import org.sonar.plugins.python.api.tree.FunctionDef;
 import org.sonar.plugins.python.api.tree.Tree;
+import org.sonar.python.quickfix.TextEditUtils;
 import org.sonar.python.tree.TreeUtils;
 
 @Rule(key = "S6552")
 public class DjangoReceiverDecoratorCheck extends PythonSubscriptionCheck {
 
   private static final String MESSAGE = "Move this '@receiver' decorator to the top of the other decorators.";
+  private static final String QUICK_FIX_MESSAGE = "Move this '@receiver' decorator to the top of the other decorators.";
```
guillaume-dequenne-sonarsource, line 36: I'd suggest `Move the '@receiver' decorator to the top` for the QF message, to avoid the demonstrative pronoun and be a bit more concise.

guillaume-dequenne-sonarsource (approved):
Looks good! Minor suggestion on the quick fix message.

python-checks/src/main/java/org/sonar/python/checks/django/DjangoReceiverDecoratorCheck.java
maksim-grebeniuk-sonarsource: done

PR #1459, opened 2023-05-03
SONARPY-1334 Prevent S6559 from raising issues on abstract models

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/django/DjangoModelStrMethodCheck.java
```diff
@@ -50,6 +57,13 @@ public void initialize(Context context) {
     });
   }
 
+  private static boolean isAbstractModel(ClassDef classDef) {
+    return getMetaClass(classDef)
+      .flatMap(metaClass -> getFieldAssignment(metaClass, "abstract"))
+      .filter(assignmentStatement -> Expressions.isTruthy(assignmentStatement.assignedValue()))
```
maksim-grebeniuk-sonarsource, line 63: Probably it is good idea to support (or at least check that it it not fails) for the cases when we have multiple values assignment like:
```python
class A(models.Model):
    name = models.CharField(max_length=100)
    class Meta:
        abstract, fields = True, '__all__'
```

guillaume-dequenne-sonarsource: Good catch! I didn't consider this case.

Luckily, in this case the assigned value will be a tuple, which is considered truthy, and we won't raise by default.
I added some tests to document the behavior. In particular, that means we have a FN if the value for `abstract` is `False` in the tuple. I don't see that happening in practice (and it's a FN anyway), so I didn't try to cover this case.

maksim-grebeniuk-sonarsource (approved):
LGTM!

PR #1460, opened 2023-05-03
SONARPY-1344 Fix FP on S1721: Avoid raising an issue on all single element tuples

joke1196:


guillaume-dequenne-sonarsource (approved):
LGTM.
The scope is a bit different than what the ticket implies: the ticket only mentions avoiding raising issues for `return`, but the change applies to all keywords.

I think it makes sense to keep it like that, but could you update the title of the ticket and the description so that it says that we exclude all single element tuples from the rule (not just with `return`) before merging?

PR #1461, opened 2023-05-03
SONARPY-1254 Fix false negative for S1720 on missing docstrings in methods

mbyrnepr2:
https://sonarsource.atlassian.net/browse/SONARPY-1254.

FYI @guillaume-dequenne-sonarsource


guillaume-dequenne-sonarsource:
Hello @mbyrnepr2 and sorry for the late reaction to this PR.

The suggested change looks sensible according to the ticket. Thank you for contributing to sonar-python!

We are however a bit worried at the impact this would have on existing projects (lots of new issues popping up overnight might not be a good experience - despite the fact they are true positives according to the description of the rule) so we want to consider other approaches before fixing this ticket.

I'll therefore close the PR for now (sorry about that!).

mbyrnepr2:
No worries at all @guillaume-dequenne-sonarsource! Thanks for the rationale.

PR #1463, opened 2023-05-04
SONARPY-1341 Fix broken formatting when applying quick fix for S2772

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/NeedlessPassCheck.java
```diff
@@ -49,21 +51,37 @@ public void initialize(Context context) {
...
+      var removeFrom = TreeUtils.getTreeSeparatorOrLastToken(previous);
+      var removeTo = TreeUtils.getTreeSeparatorOrLastToken(toRemove);
+      return TextEditUtils.removeRange(
+        removeFrom.line(),
+        removeFrom.column(),
+        removeTo.line(),
+        removeTo.column());
+    } else {
+      return TextEditUtils.removeUntil(toRemove, statements.get(removeIndex + 1));
+    }
```
joke1196, line 82: This is out of the scope of this ticket but I was wondering if this issue would not also appear in other quick fixes: where removing a statement is not removing the blank line. Maybe at some point, this logic could be part of the utils.

PR #1464, opened 2023-05-04
Document supported Python and IPython versions

nils-werner-sonarsource:
This PR is based on the concept to document the supported versions for all languages. The embedded documentation will be used as a reference for the Documentation Team to update the documentation for SonarQube and SonarCloud.

### DIFF
--- a/sonar-python-plugin/src/main/resources/static/documentation.md
+++ b/sonar-python-plugin/src/main/resources/static/documentation.md
@@ -9,8 +9,12 @@ key: python
 
 
 ## Supported Versions
-* Python 3.X
-* Python 2.X
+
+| Version           | Status    |
+|-------------------|-----------|
+| Python 3.0 - 3.11 | Supported |
+| Python 2.x        | Supported |
+| IPython 7.x       | Supported |
 
 ## Language-Specific Properties
 


nils-werner-sonarsource:
> Looking to your PR, I have one doubt about IPython support
since we added basic support for Jupiter notebooks just for sonarlint for now, as far as I understand
there are special JupiterNotebooks parsing done on the lint side and our IPython parser already works with Python only code cells (as a single plain python file)
so I guess it should not work on sonarqube side for example
(Conversation with Maksim)

guillaume-dequenne-sonarsource:
> > Looking to your PR, I have one doubt about IPython support
> > since we added basic support for Jupiter notebooks just for sonarlint for now, as far as I understand
> > there are special JupiterNotebooks parsing done on the lint side and our IPython parser already works with Python only code cells (as a single plain python file)
> > so I guess it should not work on sonarqube side for example
> > (Conversation with Maksim)

Indeed!
We support some IPython syntax in the specific context of Jupyter Notebooks files analyzed through SonarLint VSCode. Any other scenario (SQ, SC or even IPython syntax part of a non-notebook file in SonarLint VSCode) will not be supported. Therefore I think we shouldn't mention IPython here.

PR #1465, opened 2023-05-04
SONARPY-1325 Fix FP on S5727 when annotated return type is object

joke1196:
This PR removes some FPs but also adds some FNs.

Because I think we would want to raise an issue on:
```
obj = object()
if obj is not None: pass # Remove this identity check; it will always be True.
```
Is it ok to have a bit more FNs, by assuming that the comparison between object and None is always meaningful?


guillaume-dequenne-sonarsource (approved):
LGTM.
Good point about the FNs, I think they are acceptable. Besides, `None` is technically an object in Python, so from a contract perspective they're not even really false negatives.

PR #1475, opened 2023-05-12
Typeshed mypy importer random commit

guillaume-dequenne-sonarsource:


python-checks/src/test/resources/checks/hotspots/httpOnlyCookie.py
```diff
@@ -73,13 +73,13 @@ def flask_tests():
     from flask import Response, make_response, redirect
 
     response1 = Response('OK')
-    response1.set_cookie('c1', 'value') # Noncompliant
+    response1.set_cookie('c1', 'value') # FN
```
guillaume-dequenne-sonarsource, line 76: `set_cookie` is no longer resolved to be `flask.wrappers.Response.set_cookie` nor `werkzeug.wrappers.BaseResponse.set_cookie` but instead `werkzeug.sansio.response.Response.set_cookie`. It is accurate, as this is where the method is inherited from, but possibly problematic when trying to define sensitive methods.

python-checks/src/test/resources/checks/hotspots/secureCookie.py
```diff
@@ -73,13 +73,13 @@ def flask_tests():
     from flask import Response, make_response, redirect
 
     response1 = Response('OK')
-    response1.set_cookie('c1', 'value') # Noncompliant
+    response1.set_cookie('c1', 'value') # FN
```
guillaume-dequenne-sonarsource, line 76: Same here.

python-checks/src/test/resources/checks/verifiedSslTlsCertificateCheck.py
```diff
@@ -34,51 +34,51 @@ def requestsTests():
   # requests-tests.py
   import requests
 
-  requests.request('GET', 'https://example.domain', verify=False) # Noncompliant {{Enable server certificate validation on this SSL/TLS connection.}}
```
guillaume-dequenne-sonarsource, line 37: `requests` is missing from serialization

python-checks/src/test/resources/checks/hotspots/cors/cors.py
```diff
@@ -93,14 +93,14 @@ def flask_response_headers():
 
 def werkzeug_headers():
     from werkzeug.datastructures import Headers
-    Headers({"Access-Control-Allow-Origin": "*"}) # Noncompliant
-#           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    Headers({"Access-Control-Allow-Origin": "*"}) # FN
```
guillaume-dequenne-sonarsource, line 96: FQN seems to be `werkzeug.datastructures.headers.Headers` instead of expected `werkzeug.datastructures.Headers`

python-checks/src/test/resources/checks/verifiedSslTlsCertificateCheck.py
joke1196: Adding `types-requests` in the requirements helped solved this issue

joke1196: I understood it mypy option `--install-types` should check for type info in typeshed or download types-{libraryName} or get the type information if provided. But [this does not seem to work with requests ](https://stackoverflow.com/questions/65947076/why-is-mypy-finding-no-type-hints-or-library-stubs-for-any-of-my-imports) and maybe other libraries

PR #1476, opened 2023-05-19
 SONARPY-1354 Use mypy to generate stubs for typed Python libraries

guillaume-dequenne-sonarsource:
Some information that may help for the review:
* The first commit of this PR is a general refactoring of the Typeshed serializer meant to remove some technical debt and to make the following change easier. No functional changes are expected from it.
* The second commit is where meaningful changes are located: third party libraries are directly serialized from `mypy` building the semantic model for them in an environment where these libraries are installed
* The remaining commits segregate the resulting impacts of those changes in separate commits (protobuf, test expectations, checksum/plugin size).

The following libraries are still missing:
```
backports
contextvar
fb303
nmap
pymssql
scribes
```


python-frontend/typeshed_serializer/requirements.txt
```diff
@@ -6,3 +6,36 @@ toml==0.10.2
...
+geoip2==4.7.0
+ipaddress==1.0.23
+itsdangerous==2.1.2
+Jinja2==3.1.2
+kazoo==2.9.0
+MarkupSafe==2.1.2
+maxminddb==2.3.0
+orjson==3.8.12
+pathlib==1.0.1
+PyJWT==2.7.0
```
joke1196, line 34: Here using pymssql==2.2.7 worked

python-frontend/typeshed_serializer/requirements.txt
```diff
@@ -6,3 +6,36 @@ toml==0.10.2
...
+freezegun==1.2.2
+frozenlist==1.3.3
+futures==3.0.5
+geoip2==4.7.0
+ipaddress==1.0.23
+itsdangerous==2.1.2
+Jinja2==3.1.2
+kazoo==2.9.0
+MarkupSafe==2.1.2
+maxminddb==2.3.0
```
joke1196, line 31: The package to import nmap is called python-nmap==0.7.1

python-frontend/typeshed_serializer/requirements.txt
```diff
@@ -6,3 +6,36 @@ toml==0.10.2
...
+atomicwrites==1.4.1
+backports-abc==0.5
+certifi==2023.5.7
+characteristic==14.3.0
+click==8.1.3
+cryptography==36.0.1
+dataclasses==0.6
+enum34==1.1.10
+filelock==3.4.0
+first==2.0.2
```
joke1196, line 20: fb303 would be part of thrift, I think we would need to import it from thrift.Thrift.fb303 but maybe the easiest would be to add types-fb303 to the requirements. Not sure at all.

joke1196 (approved):
LGTM! I just added a few comments for the requirements.txt
I am not sure how much the change in the imports will impact the detection at the end but it could be interesting to see the impact on the generated protobufs. What do you think? Is it worth it?

python-frontend/typeshed_serializer/requirements.txt
guillaume-dequenne-sonarsource: Unfortunately it seems that it doesn't work out of the box on M1 Macs. I think I actually managed to solve this during the PoC, but I feel like it would make the build quite more complex in the process. I think we can keep this one out of now and see whether there is a more suitable solution (e.g Docker, or getting back to stubs) if we don't want to lose this library.

python-frontend/typeshed_serializer/requirements.txt
guillaume-dequenne-sonarsource: Somehow, adding this didn't change the generated protobuf, so I removed `nmap` altogether.

PR #1477, opened 2023-05-23
SONARPY-1355: Update Typeshed revision

joke1196:


python-checks/src/test/resources/checks/itemOperationsTypeCheck/itemOperations_delitem.py
```diff
@@ -21,7 +21,7 @@ def builtin_types_not_supporting_delitem():
   del mydict.items()[0]  # Noncompliant
 
   # iterators
-  del iter(mylist)[0]  # Noncompliant
+  del iter(mylist)[0]  # FN Noncompliant
```
guillaume-dequenne-sonarsource, line 24: From what I see, `iter` was always ambiguous. It used to have 3 possible definitions that all returned an `Iterator`. Now there is a new definition that returns a `_SupportsNextT` as well.

python-checks/src/test/resources/checks/itemOperationsTypeCheck/itemOperations_delitem.py
```diff
@@ -82,10 +82,10 @@ def standard_library():
   del math[0]  # FN: Symbol "OTHER" Any type
 
   # File
-  del open("foo.py")[0]  # Noncompliant
+  del open("foo.py")[0]  # FN Noncompliant
 
   import os
-  del os.popen('ls')[0]  # Noncompliant
+  del os.popen('ls')[0]  # FN Noncompliant
```
guillaume-dequenne-sonarsource, line 88: On master, the type is an union between `typing.IO` and `os._wrap_close`.

python-checks/src/test/resources/checks/incompatibleOperands/arithmetic.py
```diff
@@ -59,12 +59,12 @@ def __neg__(self):
 
 def builtin_noncompliant():
   1 + "1"  # Noncompliant
-  "1" + 1  # Noncompliant
+  "1" + 1  # FN Noncompliant
```
guillaume-dequenne-sonarsource, line 62: Here, `str.__add__` is ambiguous and has 2 possible symbols. It is not known to have incompatible type.
`int.__radd__` is not ambiguous and known to have incompatible type.

python-checks/src/test/resources/checks/incompatibleOperands/arithmetic.py
```diff
@@ -59,12 +59,12 @@ def __neg__(self):
...
   1 + [1]  # Noncompliant
   1 + {1}  # Noncompliant
   1 + (1,)  # Noncompliant
   1 + {'a': 1}  # Noncompliant
-  [1] + (1,)  # Noncompliant
+  [1] + (1,)  # FN Noncompliant
```
guillaume-dequenne-sonarsource, line 67: This is the same problem here, with same fix.

python-checks/src/test/resources/checks/incompatibleOperands/comparison.py
```diff
@@ -2,7 +2,7 @@ def builtin_noncompliant(p):
   1 < "1"  # Noncompliant {{Fix this invalid "<" operation between incompatible types (int and str).}}
 #   ^
   complex(1) < complex(1)  # Noncompliant
-  [1] < (1,)  # Noncompliant
+  [1] < (1,)  # FN Noncompliant
```
guillaume-dequenne-sonarsource, line 5: The problem here seems to be a bit different, as for some reason `list` and `tuple` are considered compatible with regard to the `__gt__` method of `tuple`, not sure how though...

python-checks/src/test/resources/checks/argumentType.py
```diff
@@ -59,25 +59,27 @@ def stdlib_functions():
   A = UnexpectedClass()
   acos(A) # Noncompliant
   B = datetime.tzinfo()
-  B.tzname(42) # Noncompliant {{Change this argument; Function "tzname" expects a different type}}
-#          ^^
+  B.tzname(42) # FN Noncompliant {{Change this argument; Function "tzname" expects a different type}}
```
guillaume-dequenne-sonarsource, line 62: We seem to resolve that the method is originally an abstract method (has the `abstractmethod` decorator), which we consider unknown, and we bail because of that.

python-checks/src/test/resources/checks/argumentType.py
```diff
@@ -59,25 +59,27 @@ def stdlib_functions():
...
+##          ^^
   select.select([],[],[], 0)
   time.sleep(1) # OK
   x = time.gmtime(int(time.time()))
   x = time.gmtime(secs = int(time.time()))
   time.sleep(True) # OK, converted to 1
   time.sleep(1j) # FN, considered duck type compatible
   genericpath.isfile("some/path")
-  genericpath.isfile(42) # Noncompliant
+  genericpath.isfile(42) # FN Noncompliant
```
guillaume-dequenne-sonarsource, line 71: The expected parameter is now an union between the following types: `bytes`, ``int`, `str` and `os.PathLike`.
Interestingly, I don't get a `TypeError` when executing this, it simply returns `False` (Python 3.9).
We should look if this API has changed. From a first look, this looks like a legitimate change.

python-checks/src/test/resources/checks/argumentType.py
```diff
@@ -59,25 +59,27 @@ def stdlib_functions():
...
   my_list = [1,2,3]
   _heapq.heapify(42) # Noncompliant {{Change this argument; Function "heapify" expects a different type}}
 #                ^^
   _heapq.heapify(my_list)
   imap4 = imaplib.IMAP4()
-  imap4.setannotation(42) # FN, we do not handle variadic parameters
+  # imap4 is not part of our protobufs anymore but it should be an FN, we do not handle variadic parameters
+  imap4.setannotation(42) # Noncompliant
```
guillaume-dequenne-sonarsource, line 78: I'm not sure how we resolve this then? We resolve a call to `imaplib.IMAP4.setannotation`, but we do not see that the argument is variadic for some reason.

python-checks/src/test/resources/checks/argumentType.py
```diff
@@ -86,8 +88,8 @@ def builtin_functions():
   number = 42
   number.__add__("27") # Noncompliant
   number.__add__(unexpected) # Noncompliant
-  number.__add__(x = unexpected) # Noncompliant {{Change this argument; Function "__add__" expects a different type}}
-#                ^^^^^^^^^^^^^^
+  number.__add__(x = unexpected) # FN Noncompliant {{Change this argument; Function "__add__" expects a different type}}
```
guillaume-dequenne-sonarsource, line 91: Here what happens is that we don't recognize `x` as being the name of the parameter (it's now empty).

python-checks/src/test/resources/checks/argumentType.py
joke1196: This is solved with 6c16fd10dd402bc04e629a6aa3fdfda877f2fee1

python-checks/src/test/resources/checks/argumentNumber.py
```diff
@@ -181,7 +181,7 @@ def method2(self, a):
 def flask_send_file():
     # make sure no FPs are raised on flask.send_file
     from flask import send_file
-    return send_file(
+    return send_file( # Noncompliant
```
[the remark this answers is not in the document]
joke1196: This is solved with 6c16fd10dd402bc04e629a6aa3fdfda877f2fee1

python-checks/src/test/resources/checks/argumentType.py
joke1196: Not much has changed in this method. It uses os.stat which when provided with an int throws an OSError. The error is then caught by genericpath.isfile and returns None. So I would agree here the program is safe but it should always return False if we pass an int to genericpath.isfile.

its/ruling/src/test/resources/expected/python-S5607.json
```diff
@@ -9,6 +9,9 @@
 'project:buildbot-0.8.6p1/buildbot/status/build.py':[
 104,
 ],
+'project:buildbot-0.8.6p1/buildbot/status/web/waterfall.py':[
+571,
```
guillaume-dequenne-sonarsource, line 13: Looks like a TP (TypeError in Python 3.9)

its/ruling/src/test/resources/expected/python-S5644.json
```diff
@@ -1,17 +1,49 @@
...
+251,
+],
+'project:docker-compose-1.24.1/compose/cli/main.py':[
+1158,
+],
 'project:mypy-0.782/mypy/bogus_type.py':[
 22,
 24,
 ],
-'project:twisted-12.1.0/twisted/python/runtime.py':[
```
guillaume-dequenne-sonarsource, line 10: This change is unrelated to the other API changes (still to be investigated).

its/ruling/src/test/resources/expected/python-S930.json
```diff
@@ -1,6 +1,10 @@
 {
-'project:django-2.2.3/django/db/backends/sqlite3/schema.py':[
-44,
+'project:buildbot-0.8.6p1/buildbot/db/pool.py':[
+46,
+],
+'project:buildbot-0.8.6p1/contrib/svn_watcher.py':[
+119,
+139,
```
joke1196, line 7: These are now FPs. As we are now detecting the correct number of arguments

its/ruling/src/test/resources/expected/python-S930.json
```diff
@@ -22,9 +26,6 @@
 'project:pecos/examples/qp2q/utils/analyse_eval_results.py':[
 241,
 ],
-'project:tornado-2.3/tornado/escape.py':[
-137,
```
joke1196, line 26: This one was an FP but we are now detecting the correct number of expected arguments

its/ruling/src/test/resources/expected/python-S5713.json
```diff
@@ -2,13 +2,6 @@
...
 816,
 ],
-'project:tornado-2.3/tornado/platform/common.py':[
-75,
-84,
-],
-'project:twisted-12.1.0/twisted/internet/selectreactor.py':[
-115,
-],
 'project:twisted-12.1.0/twisted/web/http.py':[
```
joke1196, line 5: These are deprecated Error APIs. So they are somewhat FNs as now IOError does not exist anymore (since python 3.3)  instead it is now OsError and socket.error is an alias to OsError.

guillaume-dequenne-sonarsource: I had a very quick look at this one, it seems that `IOError` and `socket.error` refer to the same symbol (`OSError` ?) this is indeed FN, but quite acceptable I feel.

its/ruling/src/test/resources/expected/python-S2638.json
```diff
@@ -39,9 +39,6 @@
 'project:buildbot-0.8.6p1/buildbot/pbutil.py':[
 75,
 ],
-'project:buildbot-0.8.6p1/buildbot/util/__init__.py':[
-166,
```
joke1196, line 43: This is now an FN we do not detect anymore that tzname needs a dt object as parameter

its/ruling/src/test/resources/expected/python-S2638.json
```diff
@@ -73,6 +70,9 @@
 151,
 193,
 ],
+'project:django-2.2.3/django/utils/text.py':[
+284,
```
joke1196, line 74: Somewhat of an FP. So the detection is correct here read should have a second parameter called size but the message says the parameter name is `null`

its/ruling/src/test/resources/expected/python-S2638.json
```diff
@@ -91,6 +91,9 @@
 'project:numpy-1.16.4/numpy/ma/core.py':[
 6170,
 ],
+'project:numpy-1.16.4/tools/c_coverage/c_coverage_report.py':[
+34,
```
joke1196, line 95: Pygments API changed since version 2.12 where the outline parameter was dropped. The previous version where the outline parameter existed is 2.11.2 and was released in January 2022. Not sure is we should do the same as with PyYAML or as this library is less used we could accept this FP.

its/ruling/src/test/resources/expected/python-S2638.json
guillaume-dequenne-sonarsource: I guess this is due to the fact the parameter starts with `__` ?
We should probably have a small ticket to fix this and update the message in such case.
It would be nice to have this case as a unit test case in S2638 unit tests.

its/ruling/src/test/resources/expected/python-S107.json
```diff
@@ -91,9 +91,6 @@
 510,
 519,
 ],
-'project:numpy-1.16.4/numpy/distutils/extension.py':[
-24,
```
joke1196, line 95: We can remove this FP as we should only raise an issue on the parent method and not the overridden one.

its/ruling/src/test/resources/expected_extended/salt/python-S930.json
```diff
@@ -29,7 +32,22 @@
...
+216,
+218,
+],
+'salt:salt/modules/win_useradd.py':[
+334,
+705,
+],
+'salt:salt/spm/__init__.py':[
+1114,
+1117,
```
guillaume-dequenne-sonarsource, line 45: These are within an `except` block and are valid in Python 2.
The correct Python 3 call is in the `try` block so this seems to be an intended fallback (FP).
We should probably exclude `except` blocks from this rule.

its/ruling/src/test/resources/expected_extended/indico/python-S930.json
```diff
@@ -0,0 +1,5 @@
+{
+'indico:indico/util/string.py':[
+475,
```
guillaume-dequenne-sonarsource, line 3: This is a FP due to the fact that the [latest](https://bleach.readthedocs.io/en/latest/clean.html#bleach.clean) version of the method doesn't allow for a `styles` parameter, but [previous](https://bleach.readthedocs.io/en/v4.0.0/_modules/bleach.html#clean) ones do.

python-checks/src/test/resources/checks/use_of_empty_return_value.py
```diff
@@ -23,3 +23,9 @@ def passed_as_argument():
 
 ls = [1, 2, 3]
 result = ls.append(42)  # FN - no type inference outside of functions
+
+import win32pdh
+
+# FP due to the stubs of win32 being incomplete
```
guillaume-dequenne-sonarsource, line 29: In this case it's not a problem, but generally it's better to keep test cases within functions, as type inference may have an influence and isn't enabled at module level.

its/ruling/src/test/resources/expected_extended/nltk/python-S930.json
```diff
@@ -1,7 +1,4 @@
 {
-'nltk:nltk/ccg/lexicon.py':[
-77,
```
guillaume-dequenne-sonarsource, line 3: This symbol was only defined in our Python 2 stubs. It's fair that we lose the issue in this context.
Not sure whether it was TP/FP though.

its/ruling/src/test/resources/expected/python-S2638.json
joke1196: Not sure about the `__`, the parameter seems to be called `size`. Still [I have created this issue to investigate.](https://sonarsource.atlassian.net/browse/SONARPY-1363)

its/ruling/src/test/resources/expected/python-S930.json
```diff
@@ -28,7 +28,16 @@
...
+285,
+],
 'project:twisted-12.1.0/twisted/lore/slides.py':[
 260,
 ],
+'project:twisted-12.1.0/twisted/persisted/aot.py':[
+224,
+],
+'project:twisted-12.1.0/twisted/python/htmlizer.py':[
+81,
```
joke1196, line 42: Those are all somewhat TPs as the tokenize methods expects 
 now 1 argument. 2 arguments were needed only in python 2.7

its/ruling/src/test/resources/expected/python-S2638.json
```diff
@@ -55,6 +55,9 @@
 'project:django-2.2.3/django/contrib/gis/db/backends/spatialite/operations.py':[
 105,
 ],
+'project:django-2.2.3/django/core/management/commands/loaddata.py':[
+346,
```
joke1196, line 56: This is a TPs as read should have 2 arguments.

its/ruling/src/test/resources/expected/python-S1172.json
```diff
@@ -1071,6 +1071,9 @@
 'project:numpy-1.16.4/numpy/distutils/ccompiler.py':[
 730,
 ],
+'project:numpy-1.16.4/numpy/distutils/command/build_ext.py':[
+479,
```
joke1196, line 1075: TPs swig sources should have 2 arguments

its/ruling/src/test/resources/expected/python-S2638.json
```diff
@@ -82,6 +85,9 @@
 'project:numpy-1.16.4/numpy/distutils/fcompiler/pg.py':[
 131,
 ],
+'project:numpy-1.16.4/numpy/distutils/command/build_ext.py':[
+284,
```
joke1196, line 89: TPs swig sources should have 2 arguments

its/ruling/src/test/resources/expected/python-S3699.json
```diff
@@ -13,4 +13,7 @@
 297,
 297,
 ],
+'project:tensorflow/tools/pip_package/setup.py':[
+141,
```
joke1196, line 20: TP finalize_options does not return anything

its/ruling/src/test/resources/expected/python-S1172.json
joke1196, line 1075: Unused parameter c_sources

its/ruling/src/test/resources/expected/python-S5655.json
```diff
@@ -1,4 +1,7 @@
 {
+'project:buildbot-0.8.6p1/contrib/svn_watcher.py':[
+119,
```
joke1196, line 3: win32.popen expects a str not a sequence

its/ruling/src/test/resources/expected/python-S5607.json
```diff
@@ -3,6 +3,9 @@
 562,
 610,
 ],
+'project:buildbot-0.8.6p1/buildbot/process/buildstep.py':[
+657,
```
joke1196, line 7: TP, should compare 2 maps

its/ruling/src/test/resources/expected_extended/calibre/python-S930.json
```diff
@@ -23,4 +23,7 @@
 'calibre:src/calibre/library/prefs.py':[
 105,
 ],
+'calibre:src/calibre/utils/certgen.py':[
+63,
```
guillaume-dequenne-sonarsource, line 27: Looks like a TP, the function only takes 1 parameter in both Python 2 and Python 3.

its/ruling/src/test/resources/expected_extended/calibre/python-S5607.json
```diff
@@ -0,0 +1,5 @@
+{
+'calibre:src/calibre/utils/exim.py':[
+412,
```
guillaume-dequenne-sonarsource, line 3: Looks like a TP.

python-checks/src/test/resources/checks/argumentType.py
```diff
@@ -221,3 +221,22 @@ def special_form_types():
...
+
+def fp_win32pdh(machine, object, instance, inum, counter):
+    import win32pdh
+    path = win32pdh.MakeCounterPath((machine, object, instance, None, inum, counter))
+    hq = win32pdh.OpenQuery()
+    # FP: stubs for win32pdh are incomplete
+    hc = win32pdh.AddCounter(hq, path)  # Noncompliant
+
+
+def fn_len():
```
guillaume-dequenne-sonarsource, line 240: SONARPY-1368 was created for this FN.

python-checks/src/test/resources/checks/argumentNumber.py
```diff
@@ -148,7 +148,7 @@ def builtin_method():
     l = list[int](42, 43)
     l.append(1, 2) # Noncompliant
     import math
-    math.acos(0, 1) # Noncompliant
+    math.acos(0, 1) # FN Noncompliant
```
guillaume-dequenne-sonarsource, line 152: [SONARPY-1370](https://sonarsource.atlassian.net/browse/SONARPY-1370)

python-checks/src/test/resources/checks/hotspots/cors/cors.py
```diff
@@ -43,40 +43,40 @@ def django_response():
...
-    @cross_origin(origins=["*"]) # Noncompliant
+    @cross_origin() # FN Noncompliant
+##   ^^^^^^^^^^^^^^^
+    @cross_origin # FN Noncompliant
+##   ^^^^^^^^^^^^^
+    @cross_origin(origins="*") # FN Noncompliant
+##   ^^^^^^^^^^^^^^^^^^^^^^^^^^
+    @cross_origin(origins=r".*") # FN Noncompliant
+    @cross_origin(origins=r".+") # FN Noncompliant
+    @cross_origin(origins=["*"]) # FN Noncompliant
```
joke1196, line 79: This is documented in the following ticket : [SONARPY-1373](https://sonarsource.atlassian.net/browse/SONARPY-1373)

python-checks/src/test/resources/checks/changeMethodContract.py
```diff
@@ -66,7 +66,7 @@ class UnresolvedParent(OtherClass):
   def my_method(self): ... # OK
 
 class MyString(str):
-  def capitalize(self, p1): ... # Noncompliant {{Remove parameter p1 or provide default value. This method overrides str.capitalize.}}
+  def capitalize(self, p1): ... # FN Noncompliant {{Remove parameter p1 or provide default value. This method overrides str.capitalize.}}
```
guillaume-dequenne-sonarsource, line 69: [SONARPY-1375](https://sonarsource.atlassian.net/browse/SONARPY-1375)

guillaume-dequenne-sonarsource (approved):
All known regressions should have tickets associated with them now, and the commit history is somewhat clean, given a PR of this size. I think we can now merge this PR.

PR #1486, opened 2023-06-09
SONARPY-1378 Rule S5549: Fix emoji checks to fit the new API

joke1196:


guillaume-dequenne-sonarsource (approved):
LGTM.
Minor comment: when rewriting commit history, don't forget to look at the aggregated commit message. It's generally generated from the commits you squashed and may or may not make sense. In this case for the first commit it says:
```
SONARPY-1378 Rule S5549: Import emoji library for up-to-date stubs ge…

…neration

Corrected ruling expectations

Corrected ruling expectations
```

I guess those "Corrected ruling expectations" don't make sense anymore.
In general, I think we have a tendency to completely remove the content of those commit messages (though in some cases it might be helpful to see what was squashed, like during this MMF).

PR #1488, opened 2023-06-14
SONARPY-1384 Ensure submodule imports doesn't prevent parent module import

guillaume-dequenne-sonarsource:
The missing condition of coverage is an old defensive check. To be honest, I don't remember what it's about but I'd rather just keep it as a defensive check and accept the missing piece of coverage.

### DIFF
--- a/python-frontend/src/test/java/org/sonar/python/semantic/SymbolTableBuilderTest.java
+++ b/python-frontend/src/test/java/org/sonar/python/semantic/SymbolTableBuilderTest.java
@@ -309,7 +309,6 @@ public class SymbolTableBuilderTest {
 
   @Test
   public void importing_submodule_as() {
-    // SONARPY-1384
     FunctionDef functionDef = functionTreesByName.get("importing_submodule_as");
     Map<String, Symbol> symbolByName = getSymbolByName(functionDef);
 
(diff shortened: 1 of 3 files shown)


python-frontend/src/test/java/org/sonar/python/semantic/SymbolTableBuilderTest.java
```diff
@@ -387,16 +391,21 @@ public void importing_parent_after_submodule_2() {
 
     assertThat(symbolByName).containsOnlyKeys("werkzeug");
     SymbolImpl werkzeug = (SymbolImpl) symbolByName.get("werkzeug");
-    assertThat(werkzeug.usages()).extracting(Usage::kind).containsExactly(Usage.Kind.IMPORT, Usage.Kind.IMPORT, Usage.Kind.IMPORT, Usage.Kind.OTHER);
+    assertThat(werkzeug.usages()).extracting(Usage::kind).containsExactly(Usage.Kind.IMPORT, Usage.Kind.IMPORT, Usage.Kind.IMPORT, Usage.Kind.OTHER, Usage.Kind.OTHER);
     Map<String, Symbol> childrenSymbolByName = werkzeug.getChildrenSymbolByName();
-    assertThat(childrenSymbolByName.values()).extracting(Symbol::name).contains("datastructures");
+    assertThat(childrenSymbolByName.values()).extracting(Symbol::name).contains("datastructures", "Response");
```
joke1196, line 396: Shouldn't we get here `Request` as well? No need to add it just curious.

guillaume-dequenne-sonarsource: Yes, I just picked one randomly that we couldn't get from simply importing `werkzeug.datastructures.csp` or `werkzeug.datastructures` to ensure the `import werkzeug` was correctly resolved on top of that.

PR #1489, opened 2023-06-14
SONARPY-1387 Rule S5144 and S5135: python2 stubs as custom stubs, to mitigate security FNs on urllib2 and Cookie libraries

joke1196:


python-frontend/typeshed_serializer/resources/custom/urllib2/__init__.pyi
```diff
@@ -0,0 +1,67 @@
...
+    def get_selector(self): ...
+    def set_proxy(self, host, type) -> None: ...
+    def has_proxy(self) -> bool: ...
+    def get_origin_req_host(self) -> str: ...
+    def is_unverifiable(self) -> bool: ...
+    def add_header(self, key: str, val: str) -> None: ...
+    def add_unredirected_header(self, key: str, val: str) -> None: ...
+    def has_header(self, header_name: str) -> bool: ...
+    def get_header(self, header_name: str, default: Optional[str] = ...) -> str: ...
+    def header_items(self): ...
```
joke1196, line 39: I have kept the Request and urlopen stubs even if they are not directly linked to the issue but I think it would be nice to not raise the TPs on the Security side because of a lucky match of the FQNs.

guillaume-dequenne-sonarsource (approved):
Looks good! I suggest we keep `SimpleCookie` as well, if it doesn't conflict with anything.

PR #1497, opened 2023-06-20
Update rules metadata

guillaume-dequenne-sonarsource:


python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S3776.html
```diff
@@ -1,7 +1,7 @@
 <h2>Why is this an issue?</h2>
-<p>Cognitive Complexity is a measure of how hard the control flow of a function is to understand. Functions with high Cognitive Complexity will be
-difficult to maintain.</p>
-<h2>Resources</h2>
+<p><a href="https://www.sonarsource.com/docs/CognitiveComplexity.pdf">Cognitive Complexity</a> is a measure of how hard the control flow of a function
+is to understand. Functions with high Cognitive Complexity will be difficult to maintain. == Resources</p>
```
joke1196, line 3: Noting it here but I guess this `== Resource` is missing a line return in the original rule

joke1196 (approved):
LGTM! I just noticed a formatting change we could make on the rspec S3776

python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S3776.html
guillaume-dequenne-sonarsource: Thanks! Good catch! We changed this on RSPEC side but apparently the fix was not correct.
I updated this manually for now and we can fix it on RSPEC side during next sprint.

PR #1498, opened 2023-06-20
SONARPY-1345 Fix stack overflow when a nested class inherits from a class with the same name

guillaume-dequenne-sonarsource:


joke1196 (approved):
LGTM!

PR #1513, opened 2023-07-24
SONARPY-1402 Rule S6658: Special methods should have an expected return type

anton-haubner-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/SpecialMethodReturnTypeCheck.java
```diff
@@ -0,0 +1,361 @@
...
+    public List<Token> getYieldKeywords() {
+      return yieldKeywords;
+    }
+
+    public boolean raisesWhitelistingException() {
+      return raisesWhitelistingException;
+    }
+
+    @Override
+    public void visitReturnStatement(ReturnStatement pyReturnStatementTree) {
```
andreaguarino, line 308: Minor: can you please remove the "py" prefix from parameter names? I know the suggestion comes from the interface, however, I don't think it really adds information.

python-checks/src/main/java/org/sonar/python/checks/SpecialMethodReturnTypeCheck.java
```diff
@@ -0,0 +1,361 @@
...
+            }
+
+            // Sometimes users just raise an exception class without instantiating it.
+            // Even in these cases we should not trigger the rule, so we check for raised expressions which are just an exception class
+            // symbol.
+            //
+            // Example:
+            // raise NotImplementedError()
+            // vs
+            // raise NotImplementedError
```
andreaguarino, line 350: Btw, we should already have a rule to catch this 😄

python-checks/src/main/java/org/sonar/python/checks/SpecialMethodReturnTypeCheck.java
```diff
@@ -0,0 +1,361 @@
...
+   *     elif x == 'False':
+   *       x = False
+   *   return x
+   * </pre>
+   *
+   * Here, the type of {@code x} is identified to be {@code Union[bool, str]} (simplified).
+   * In theory, this code may return a string but likely not in practice.
+   * Hence, we can not raise an issue just because a returned union type contains a type which is not a subtype of the expected type.
+   * (As long as the union type still contains another type that would be a valid return type.)
+   */
```
andreaguarino, line 185: I think this comment is not super relevant here.
You're describing the fact that sonar-python type inference is not path-sensitive and also that we prefer to under-approximate while raising an issue (i.e. we don't want to raise when the inferred type is "Any").

Those concepts are valid for all rules, so that's why I tend to think having a comment in the rule is not appropriate.

At the same time, I realize that from the eyes of a new maintainer, this comment would be useful.
So, I would suggest generalizing it a bit more and moving it to the `InferredType` interface.

python-checks/src/main/java/org/sonar/python/checks/SpecialMethodReturnTypeCheck.java
```diff
@@ -0,0 +1,361 @@
...
+      ctx.addIssue(firstElement.firstToken(), secondElement.lastToken(), INVALID_GETNEWARGSEX_TUPLE_MESSAGE);
+    }
+  }
+
+  private static InferredType getReturnStmtType(ReturnStatement returnStatement) {
+    final List<Expression> returnedExpressions = returnStatement.expressions();
+
+    if (returnedExpressions.isEmpty()) {
+      return InferredTypes.NONE;
+    }
```
andreaguarino, line 255: I think this is redundant. You're already checking for the return expressions to be empty at line 188.
Also removing those lines doesn't change anything in unit tests

python-checks/src/main/java/org/sonar/python/checks/SpecialMethodReturnTypeCheck.java
```diff
@@ -0,0 +1,361 @@
...
+   *
+   * ...all returned expressions are marked as the source of the issue if the return statement contains such expressions
+   * ...the return keyword is marked as the source of the issue if the return statement does not contain any expressions
+   */
+  private static void addIssueOnReturnedExpressions(SubscriptionContext ctx, ReturnStatement returnStatement, String message) {
+    final List<Expression> returnedExpressions = returnStatement.expressions();
+
+    if (returnedExpressions.isEmpty()) {
+      ctx.addIssue(returnStatement.returnKeyword(), message);
+    } else {
```
andreaguarino, line 275: Same here, I think this condition is not needed because you early return al line 190 if `returnedExpressions.isEmpty()`

python-checks/src/test/resources/checks/specialMethodReturnTypeCheck.py
```diff
@@ -0,0 +1,319 @@
...
+
+    def __bool__(self):
+        return self.condition # Compliant but potential for FNs: Technically, it is possible that this may not be a boolean but there is no explicit indication for it. Marking this may lead to FPs
+
+class BoolMethodCheck05:
+    def __init__(self, condition):
+        self.condition = condition
+
+    def __bool__(self):
+        return self.condition # FN: Requires better analysis tooling, see guaranteed invocation below
```
andreaguarino, line 35: Yes and at the same time, even with a better analysis, the primary location should not be here, but at the call site, i.e. at line 37, because that specific instance of `BoolMethodCheck05` is problematic. Of course, line 35 will still be part of the issue reporting, as a secondary location.
You'll see that we take this approach in DBD, where we do support inter-procedural analysis.

python-checks/src/main/java/org/sonar/python/checks/SpecialMethodReturnTypeCheck.java
```diff
@@ -0,0 +1,361 @@
...
+    }
+
+    @Override
+    public void visitLambda(LambdaExpression pyLambdaExpressionTree) {
+      // We do not visit nested lambda definitions as they may contain irrelevant yield expressions
+    }
+
+    @Override
+    public void visitRaiseStatement(RaiseStatement pyRaiseStatementTree) {
+      raisesWhitelistingException |= pyRaiseStatementTree
```
andreaguarino, line 334: Looking at ruling issues, I think it probably makes sense to avoid reporting an issue when the method implementation has only "raise statements" as exit points, regardless of the exception raised. This should also simplify the logic here below.

E.g.

```python
    def __hash__(self):
        self.hashCalled = True
        raise RuntimeError('not a TypeError!') # Compliant - done intentionally
```

python-checks/src/main/java/org/sonar/python/checks/SpecialMethodReturnTypeCheck.java
anton-haubner-sonarsource: In this case, I would like to keep the check for an empty expression list, if possible.
The method is used a lot in the code, and it should be safe to call if it is ever applied to a non-empty list of expressions in the future.

python-checks/src/main/java/org/sonar/python/checks/SpecialMethodReturnTypeCheck.java
anton-haubner-sonarsource: As discussed, I moved this function to the ReturnStatement interface and added testing for all cases.
I am not very confident whether the placement of the tests in `TypeInferenceTest.java` is correct, though, see a92a48cb5eb29542f55e5464f8a72284a3885e1e.

python-frontend/src/main/java/org/sonar/plugins/python/api/tree/ReturnStatement.java
```diff
@@ -36,4 +38,22 @@ public interface ReturnStatement extends Statement {
 
   @Beta
   List<Token> commas();
+
+  /**
+   * Infers the type of the returned value that would result from the execution of this return statement.
+   */
+  @Beta
+  default InferredType type() {
```
andreaguarino, line 46: To differentiate with `Expression#type()` and to make explicit this is about the return value, I would maybe call it `returnedType` or `returnValueType`

andreaguarino (approved):
LGTM, I suggest to rename `ReturnStatement#type()` API

PR #1515, opened 2023-07-25
SONARPY-1417 Serialize symbols for Python 3.11

guillaume-dequenne-sonarsource:


joke1196:
I was under the impression that we had a fallback logic where if the configured Python version was not in our list we would just use all the stubs. I cannot remember where this was.

joke1196 (approved):
LGTM. Just left a small comment,

guillaume-dequenne-sonarsource:
> I was under the impression that we had a fallback logic where if the configured Python version was not in our list we would just use all the stubs. I cannot find this logic anymore.

We have it in `PythonVersionUtils`. The issue is that we added `3.11` here without actually serializing `3.11` stubs...

PR #1517, opened 2023-07-26
SONARPY-1413 Rule S6662: Set members and dictionary keys should be ha…

andreaguarino:
…shable


python-checks/src/test/resources/checks/hashMethod.py
```diff
@@ -0,0 +1,39 @@
...
+    {some_func()}
+    {None}
+
+def dictionary_keys():
+    {EmptyClass(): 42} # OK
+    my_list = [1, 2, 3]
+    {my_list: 42} # Noncompliant
+    {1: "one", 2: "two"}
+    {ClassWithHash(): 42}
+    {None: 42}
```
anton-haubner-sonarsource, line 35: Minor: Maybe we should add some cases involving dictionary comprehension / set comprehension.
E.g.:

```python
{k: 42 for k in [1, 2]}
sequence_of_unhashable: Sequence[List] = [[], []]
{k: 42 for k in sequence_of_unhashable} # Noncompliant (?) - Example 1
{[]: v for v in [1, 2]} # Noncompliant - Example 2
```

Although, I am not sure if our tooling is currently sufficient for inferring the type of the key expression in many cases.
I did a quick experiment where I inspected a `DictCompExpression` generated for Example 1, and it was not able to infer the type of k, even when I added the type annotations.

For Example 2 it is able to determine the type.

anton-haubner-sonarsource: Ah, also: The key value expressions in dictionary comprehensions are not parsed as KEY_VALUE_PAIR nodes. Hence, they are currently not checked by the implementation.
Instead, we have to register a consumer for `DICT_COMPREHENSION`. The same probably applies for set comprehensions.

anton-haubner-sonarsource (approved):
The implementation looks good! :) I have some minor comments and suggestions for additional test cases.

python-checks/src/test/resources/checks/hashMethod.py
andreaguarino: Thanks, good point! I will create a separate ticket in the backlog to handle dict and set comprehensions!

python-checks/src/main/java/org/sonar/python/checks/HashMethodCheck.java
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   Note that, by default "object" itself defines a __hash__ method, and it is removed from classes whenever one of those two things happen (see Python
+   <a href="https://docs.python.org/3/reference/datamodel.html#object.__hash__">documentation</a>):
+   <ol>
+   <li>A class implements __eq__ without implementing __hash__</li>
+   <li>__hash__ is manually set to None (common in mutable sequences like list)</li>
+   </ol>
+
+   The implementation will only focus on 1., because 2. is causing too many FPs for typeshed symbols (e.g. int)
```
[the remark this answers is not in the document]
andreaguarino: Indeed! Thanks a lot for spotting this!

python-checks/src/main/java/org/sonar/python/checks/HashMethodCheck.java
```diff
@@ -0,0 +1,65 @@
...
+   */
+  private static void checkHashMethod(Expression expression, SubscriptionContext ctx) {
+    InferredType type = expression.type();
+    boolean hashIsManuallySetToNone = type.resolveMember("__hash__")
+      .filter(symbol -> ((SymbolImpl) symbol).inferredType().canOnlyBe(NONE_TYPE))
```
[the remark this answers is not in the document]
andreaguarino: You're right: sonar-python type inference doesn't infer types on class attributes. It only does for standard library types thanks to typeshed type annotations.
I'll document this case and create a ticket

PR #1518, opened 2023-07-26
SONARPY-1410 Rule S6659: 'startswith' or 'endswith' methods should be used instead of string slicing in condition expressions

anton-haubner-sonarsource:


andreaguarino (approved):
Very nice work! The implementation is very clean.
I let you rebase and solve the conflict before merging!

PR #1522, opened 2023-07-28
SONARPY-1404 Rule S5642: "in" and "not in" operators should be used on objects supporting them

anton-haubner-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/MembershipTestSupportCheck.java
```diff
@@ -0,0 +1,168 @@
...
+    // Try to render rhs into the message, but only if it is not multiline to avoid generating messages that are too long.
+    String inTarget = TreeUtils.treeToString(rhs, false);
+    String message;
+    if (inTarget == null) {
+      message = PRIMARY_MESSAGE_MULTILINE;
+    } else {
+      message = String.format(PRIMARY_MESSAGE, inTarget);
+    }
+
+    Symbol typeSymbol = rhsType.runtimeTypeSymbol();
```
andreaguarino, line 80: Here you could use `InferredTypes.typeName(rhsType)` to get the name to show in the message

python-checks/src/main/java/org/sonar/python/checks/MembershipTestSupportCheck.java
```diff
@@ -0,0 +1,168 @@
...
+  private static void checkInExpression(SubscriptionContext ctx, InExpression inExpression) {
+    Expression rhs = inExpression.rightOperand();
+    InferredType rhsType = rhs.type();
+
+    if (canSupportMembershipProtocol(rhsType)) {
+      return;
+    }
+
+    PreciseIssue primaryLocation = addIssueOnInAndNotIn(ctx, inExpression, genPrimaryMessage(rhs, rhsType));
+    primaryLocation.secondary(inExpression.rightOperand(), SECONDARY_MESSAGE);
```
andreaguarino, line 67: Minor: I would inline `primaryLocation` variable here:
```suggestion
    addIssueOnInAndNotIn(ctx, inExpression, genPrimaryMessage(rhs, rhsType))
      .secondary(inExpression.rightOperand(), SECONDARY_MESSAGE);
```

python-checks/src/test/resources/checks/membershipTestSupportCheck.py
```diff
@@ -0,0 +1,225 @@
...
+    x in () # Compliant
+    x in {} # Compliant
+    x in {v for v in "abc"} # Compliant
+    x in frozenset()
+    x in dict() # Compliant
+    x in {k: 42 for k in "abc"} # Compliant
+    x in produces_container() # Compliant
+    x in produces_iterable() # Compliant
+    x in produces_sequence() # Compliant
+    # TODO: Be more strict regarding declared types?
```
andreaguarino, line 66: For declared types, we have a rule called `ConfusingTypeChecking` which keeps the logic from the type-checking BUG rules but only watches for declared types.
Note that `ConfusingTypeChecking` raises a "Code smell" instead of a "bug".

So I would suggest editing the TODO comment with: Add this to `ConfusingTypeChecking` rule. Also, I think you can create a ticket in the backlog for another sprint.

andreaguarino (approved):
LGTM, I left some minor comments / suggestions

python-checks/src/test/resources/checks/membershipTestSupportCheck.py
anton-haubner-sonarsource: I updated the comment and created a ticket here: https://sonarsource.atlassian.net/browse/SONARPY-1420?atlOrigin=eyJpIjoiMjI2MmNlMDYzY2JmNGFlOThkYzNmZTZmMTM1OGFmZTEiLCJwIjoiaiJ9

PR #1523, opened 2023-07-31
SONARPY-1402 Rule S6658: Fix - Raise only for methods

anton-haubner-sonarsource:
This is a minor change to not raise any issues for functions whose name matches one of the special methods but which are not actual methods.

Example:

```python
class C:
    def __bool__(self):
        return "Hello" # Noncompliant

def __bool__():
    return "Hello" # Compliant: This function is not part of a class definition
```

Hence, this fixes potential (albeit very unlikely) FPs


andreaguarino (approved):
LGTM!

PR #1524, opened 2023-08-01
SONARPY-1406 (+ SONARPY-1402): Addressing FNs and FPs on Peach

anton-haubner-sonarsource:
Testing on Peach revealed the following issues for SONARPY-1406 which are addressed by this PR:

* FPs: The rule should not raise on `Protocol` classes (https://peach.sonarsource.com/project/issues?issues=AYmfnLQ7Jo9HYjaF5VTd&open=AYmfnLQ7Jo9HYjaF5VTd&id=sentry). Here, we should also consider `zope.interfaces.Interface` as it predates `typing.Protocol` and is also used in Peach projects. 

* FNs: The type analysis does not resolve the type of `self`, in which case we can not detect FNs like this one: https://peach.sonarsource.com/code?id=python%3Apypy&selected=python%3Apypy%3Alib-python%2F2.7%2Fwsgiref%2Fvalidate.py&line=266
   
  In iterator implementations, `self` is often used as the return value of `__iter__`, see also https://docs.python.org/3/library/stdtypes.html#iterator.__iter__
  Furthermore, iterator implementations tend to implement the incrementation method with the name `next()` instead of `__next__`, see for example ruling tests.
  Hence, we were missing many invalid iterator implementations in the ruling tests, and also in peach.

* FPs: Some projects annotate methods that raise exceptions and do not return with the bottom types `typing.Never` and `typing.NoReturn`. Some tested code on Peach was calling such methods instead of raising exceptions in iterator implementations: https://peach.sonarsource.com/code?id=python%3Azulip&selected=python%3Azulip%3Azerver%2Flib%2Fvalidator.py&line=658

Since SONARPY-1402 shares a lot of code with SONARPY-1406, it is also affected by this PR. Tests have been extended accordingly.


python-checks/src/test/resources/checks/iterMethodReturnTypeCheck.py
```diff
@@ -202,6 +206,69 @@ def __iter__(self):
         else:
             raise NotImplementedError()
 
+def noReturn01(sth) -> NoReturn:
+    ...
```
andreaguarino, line 210: to better explain why we're trusting the type annotation, I would replace `...` with an actual raise, e.g. `raise ValueError()`

python-checks/src/test/resources/checks/checkUtils/isSelfTest.py
```diff
@@ -0,0 +1,12 @@
+class C:
+    def returnsSelf01(self):
```
andreaguarino, line 2: For completeness, you could probably add a test with `@staticmethod` and `@classmethod` decorators.
In those cases the first parameter wouldn't be "self"

python-checks/src/test/java/org/sonar/python/checks/CheckUtilsTest.java
```diff
@@ -170,6 +223,17 @@ private static FileInput parseFile(String path) throws IOException {
...
 
+  private static FileInput parseFileWithSymbols(String path) throws IOException {
+    var file = new File(path);
+    var baseDirFile = new File(file.getParent());
+
+    var projectLevelSymbolTable = TestPythonVisitorRunner.globalSymbols(List.of(file), baseDirFile);
+    var context = TestPythonVisitorRunner.createContext(file, null, pythonPackageName(file, baseDirFile.getAbsolutePath()), projectLevelSymbolTable,
+      CacheContextImpl.dummyCache());
+
+    return context.rootTree();
```
andreaguarino, line 234: For those tests, I don't think a project symbol table is needed. I would rewrite it as:
```suggestion
return TestPythonVisitorRunner.createContext(new File(path)).rootTree();
```

andreaguarino (approved):
LGTM, I left a few minor comments

python-checks/src/main/java/org/sonar/python/checks/CheckUtils.java
```diff
@@ -163,4 +182,30 @@ public static boolean isAbstract(FunctionDef funDef) {
       .map(decorator -> TreeUtils.decoratorNameFromExpression(decorator.expression()))
       .anyMatch(foundDeco -> ABC_ABSTRACTMETHOD_DECORATORS.stream().anyMatch(abcDeco -> abcDeco.equals(foundDeco)));
   }
+
+  /**
+   * Simple check whether the given expression is the "self" name expression.
+   * It does not check whether the name is actually referring to a method parameter!
+   */
```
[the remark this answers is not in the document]
anton-haubner-sonarsource: You are absolutely right, doing a manual string comparison here is not the best approach when we can check for the symbol.

`isSelf` and `findSelfParameterSymbol` are actually methods from existing checks that I factored out, added tests for, and reused. I didn't want to change them too much so that I do not affect the behavior of the other checks.

I will try to add the TODO note, however, I believe the SonarQube code analysis check on GitHub will complain about new TODO comments in Java code. I'll see what happens.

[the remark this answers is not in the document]
andreaguarino: I think we can accept the code smell related to the TODO comment, as this is intentional :)

PR #1527, opened 2023-08-02
Rename S6658 to S935

anton-haubner-sonarsource:
See also https://sonarsource.slack.com/archives/CFUS31LRE/p1690534750237959


anton-haubner-sonarsource:
Closed, as it would be better to PR the rule rename and rule api update changes separately

PR #1532, opened 2023-08-15
SONARPY-1422: Add support for importing Ruff reports

joke1196:


sonar-python-plugin/src/main/java/org/sonar/plugins/python/ruff/RuffJsonReportReader.java
```diff
@@ -0,0 +1,111 @@
...
+import org.sonarsource.analyzer.commons.internal.json.simple.JSONObject;
+import org.sonarsource.analyzer.commons.internal.json.simple.parser.JSONParser;
+import org.sonarsource.analyzer.commons.internal.json.simple.parser.ParseException;
+
+public class RuffJsonReportReader {
+
+  private final Consumer<Issue> consumer;
+  private final JSONParser jsonParser = new JSONParser();
+
+  public static class Issue {
```
maksim-grebeniuk-sonarsource, line 42: do we actually need to make this class nested?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/ruff/RuffJsonReportReader.java
```diff
@@ -0,0 +1,111 @@
...
+
+import org.sonarsource.analyzer.commons.internal.json.simple.JSONArray;
+import org.sonarsource.analyzer.commons.internal.json.simple.JSONObject;
+import org.sonarsource.analyzer.commons.internal.json.simple.parser.JSONParser;
+import org.sonarsource.analyzer.commons.internal.json.simple.parser.ParseException;
+
+public class RuffJsonReportReader {
+
+  private final Consumer<Issue> consumer;
+  private final JSONParser jsonParser = new JSONParser();
```
maksim-grebeniuk-sonarsource, line 40: since you have constructor it could be good to initialize everything in one place

sonar-python-plugin/src/main/java/org/sonar/plugins/python/ruff/RuffJsonReportReader.java
```diff
@@ -0,0 +1,111 @@
...
+  private static Integer correctEndLocationCol(Object value, int startLocationCol){
+    Integer endLocationCol = toInteger(value);
+    if(endLocationCol != null){
+      if(endLocationCol > startLocationCol) {
+        return endLocationCol - 1;
+      } else {
+        return startLocationCol;
+      }
+    }
+    return null;
```
maksim-grebeniuk-sonarsource, line 101: why can't just return `startLocationCol` if `endLocationCol ` is `null` or  less than `startLocationCol`?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/ruff/RuffRulesDefinition.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.sonar.api.server.rule.RulesDefinition;
+import org.sonar.plugins.python.Python;
+import org.sonarsource.analyzer.commons.ExternalRuleLoader;
+
+public class RuffRulesDefinition implements RulesDefinition {
+
+  private static final String RULES_JSON = "org/sonar/plugins/python/ruff/rules.json";
+
+  private static final ExternalRuleLoader RULE_LOADER = new ExternalRuleLoader(RuffSensor.LINTER_KEY, RuffSensor.LINTER_NAME, RULES_JSON, Python.KEY);
```
maksim-grebeniuk-sonarsource, line 30: do we need to make this static?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/ruff/RuffSensor.java
```diff
@@ -0,0 +1,131 @@
...
+      .type(RuleType.CODE_SMELL)
+      .severity(Severity.MAJOR)
+      .remediationEffortMinutes(DEFAULT_CONSTANT_DEBT_MINUTES);
+
+    NewIssueLocation primaryLocation = newExternalIssue.newLocation()
+      .message(issue.message)
+      .on(inputFile);
+
+    if (issue.startLocationRow != null ){
+      if(isValidEndLocation(issue, inputFile)) {
```
maksim-grebeniuk-sonarsource, line 103: fix formatting

sonar-python-plugin/src/main/java/org/sonar/plugins/python/ruff/RuffJsonReportReader.java
joke1196: I think it make sense as  it is only a data type it is nice to have it (namespaced) in the json reader

sonar-python-plugin/src/main/java/org/sonar/plugins/python/ruff/RuffJsonReportReader.java
joke1196: If it is null I do not  want to return startLocationCol I want to return null

sonar-python-plugin/src/main/java/org/sonar/plugins/python/ruff/RuffRulesDefinition.java
joke1196: No, good catch! Thanks

sonar-python-plugin/src/main/java/org/sonar/plugins/python/ruff/RuffRulesDefinition.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.sonar.api.server.rule.RulesDefinition;
+import org.sonar.plugins.python.Python;
+import org.sonarsource.analyzer.commons.ExternalRuleLoader;
+
+public class RuffRulesDefinition implements RulesDefinition {
+
+  private static final String RULES_JSON = "org/sonar/plugins/python/ruff/rules.json";
+
+  private final ExternalRuleLoader RULE_LOADER = new ExternalRuleLoader(RuffSensor.LINTER_KEY, RuffSensor.LINTER_NAME, RULES_JSON, Python.KEY);
```
maksim-grebeniuk-sonarsource, line 30: There is code smell on this line because since it is not static and should be named in regular fields Java naming strategy.

maksim-grebeniuk-sonarsource (approved):
LGTM, just one field to rename before merge.

PR #1538, opened 2023-08-17
SONARPY-1376 Fix FP on S5886 when the function is a context manager

maksim-grebeniuk-sonarsource:


joke1196 (approved):
LGTM. Maybe we could add a small test when importing the contextmanager directly i.e. `from contextlib import contextmanager` and the decorator `@contextmanager`. I think the FQN should be correctly resolved, so I let you decide if itis worth it or not.

PR #1541, opened 2023-08-17
SONARPY-1394 S6553: Avoid reporting when the "managed" flag is set to False

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/django/DjangoModelStringFieldCheck.java
```diff
@@ -67,6 +77,36 @@ public void initialize(Context context) {
...
+      // lookup for managed field assignment
+      .filter(fieldAssignment -> fieldAssignment.lhsExpressions()
+        .stream()
+        .map(lhs -> TreeUtils.firstChild(lhs, Name.class::isInstance)
+          .map(Name.class::cast)
+          .orElse(null))
+        .filter(Objects::nonNull)
+        .map(Name::name)
+        .anyMatch("managed"::equals)
+      )
```
joke1196, line 101: This statement is quite long. Maybe we could at least refactor this filter into another function.

joke1196 (approved):
LGTM. Just added a small comment for a refactoring nothing major.

PR #1549, opened 2023-08-21
Update typeshed dependencies

joke1196:


maksim-grebeniuk-sonarsource (approved):
LGTM

guillaume-dequenne-sonarsource (commented):
Why are we doing this update?
Is this because of the WhiteSource alert?

I'm not sure it's 100% a valid reason why we should update the dependencies.
Our goal is to declare the most commonly used version of the dependencies, and that's not a problem if they're vulnerable, as we're not actually using them.

That said, I didn't check the libraries in question and if they are actively considered deprecated and most people are on a newer version, then for sure we should update them to a more common version. It's not really a fix though, more of a simple update of our stubs.

joke1196:
Yes, the main goal was to satisfy the WhiteSource issue. I was told the issue came from Typeshed that's why I taught this would solve the problem. If we can ignore this check in WhiteSource then it would be preferable to do that, as we know our stubs are working correctly currently. For the update of the libraries, they are just regular updates (only cryptography is a major one).

joke1196:
We are only using these dependencies to generate stubs. White source alarm are FPs as there are not risk involved. We can keep the dependencies unchanged

PR #1553, opened 2023-09-05
SONARPY-1447 Rule S6727: The abs_tol parameter should be provided when using math.isclose to compare values to 0

joke1196:


python-checks/src/main/java/org/sonar/python/checks/IsCloseAbsTolCheck.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  private static boolean anyArgumentIsZero(CallExpression call) {
+    RegularArgument firstArg = TreeUtils.nthArgumentOrKeyword(0, "a", call.arguments());
+    RegularArgument secondArg = TreeUtils.nthArgumentOrKeyword(1, "b", call.arguments());
+    return (firstArg != null && isLitteralZeroOrAssignedZero(firstArg.expression())) ||
+        (secondArg != null && isLitteralZeroOrAssignedZero(secondArg.expression()));
+  }
+
+  private static boolean isLitteralZeroOrAssignedZero(Expression expression) {
```
guillaume-dequenne-sonarsource, line 61: _Nitpick:_ typo on `literal`.

python-checks/src/main/java/org/sonar/python/checks/IsCloseAbsTolCheck.java
```diff
@@ -0,0 +1,76 @@
...
+    context.registerSyntaxNodeConsumer(Tree.Kind.CALL_EXPR,
+        ctx -> checkForIsCloseAbsTolArgument(ctx, (CallExpression) ctx.syntaxNode()));
+  }
+
+  private static void checkForIsCloseAbsTolArgument(SubscriptionContext ctx, CallExpression call) {
+    Optional.ofNullable(call.calleeSymbol())
+        .filter(symbol -> "math.isclose".equals(symbol.fullyQualifiedName()) &&
+            anyArgumentIsZero(call) &&
+            TreeUtils.argumentByKeyword("abs_tol", call.arguments()) == null)
+        .ifPresent(s -> ctx.addIssue(call, MESSAGE));
```
guillaume-dequenne-sonarsource, line 51: I'm wondering whether we should have a secondary location on the argument that evaluates to 0 which would say something like "this argument evaluates to 0." Wdyt?
I don't think it's strictly necessary, but I feel it could add clarity to the rule.
If we make that change, we may also consider moving the primary location to the callee rather than to the full call expression.

python-checks/src/main/java/org/sonar/python/checks/IsCloseAbsTolCheck.java
```diff
@@ -0,0 +1,76 @@
...
+
+@Rule(key = "S6727")
+public class IsCloseAbsTolCheck extends PythonSubscriptionCheck {
+
+  private static final String MESSAGE = "Provide the abs_tol parameter when using math.isclose to compare a value to 0";
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.CALL_EXPR,
+        ctx -> checkForIsCloseAbsTolArgument(ctx, (CallExpression) ctx.syntaxNode()));
```
guillaume-dequenne-sonarsource, line 43: What about offering a quick fix for this rule?

We could consider adding a quick fix that provide an arbitrary value for the `abs_tol`. If numpy provides `1e-8` as a default, I guess we could have that, or `1e-9` (not sure what is best here - I think `rel_tol` defaults to `1e-9`). Then the quickfix would simply add `abs_tol=1e-9` as an extra argument to the call.
I don't think it's strictly necessary, but I think it could add some value to the rule.

Wdyt?

python-checks/src/main/java/org/sonar/python/checks/IsCloseAbsTolCheck.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  private static void checkForIsCloseAbsTolArgument(SubscriptionContext ctx, CallExpression call) {
+    Optional.ofNullable(call.calleeSymbol())
+        .filter(symbol -> "math.isclose".equals(symbol.fullyQualifiedName()) &&
+            anyArgumentIsZero(call) &&
+            TreeUtils.argumentByKeyword("abs_tol", call.arguments()) == null)
```
guillaume-dequenne-sonarsource, line 50: I see the `abs_tol` parameter is keyword-only. That wasn't clear to me initially.
Maybe we could have a test that highlights that, with a call such as `math.isclose(0, x, 1, 2)` and a comment that says we acknowledge that an issue is raised because the call is not legit anyway.

python-checks/src/main/java/org/sonar/python/checks/IsCloseAbsTolCheck.java
```diff
@@ -0,0 +1,76 @@
...
+        (secondArg != null && isLitteralZeroOrAssignedZero(secondArg.expression()));
+  }
+
+  private static boolean isLitteralZeroOrAssignedZero(Expression expression) {
+    return isZero(expression) || isAssignedZero(expression);
+  }
+
+  private static boolean isAssignedZero(Expression expression) {
+    if (expression.is(Tree.Kind.NAME)) {
+      Expression assignedValue = Expressions.singleAssignedValue((Name) expression);
```
guillaume-dequenne-sonarsource, line 67: Are you aware of the `ReachingDefinitionsAnalysis` class?

This is just for information, because I feel like we could eventually use that by default rather than relying on `Expressions.singleAssignedValue` in such scenarios. It would require some refactoring to avoid running `ReachingDefinitionsAnalysis` multiple times (currently each check that invokes it runs the analysis again), but eventually I think it would improve our tracking of assigned values.

Not asking for change in this case as I think `Expressions.singleAssignedValue` does the job just fine here, just wanted to share this.

guillaume-dequenne-sonarsource (approved):
Looks good to me! As far as the rule is concerned, there are just a few typos to address.

I think we could improve the reporting a bit, and a quick fix would be a nice addition to this rule. If you agree with that, I let you decide whether to do that as part of this PR (in which case please submit a review again) or as part of a separate PR (in which case please create a ticket for it).

jdodinh:
Just a detail: the rule number is 'S6727' as opposed to the 'S6527' used in the title of the PR. The commit message is good though.

python-checks/src/main/java/org/sonar/python/checks/IsCloseAbsTolCheck.java
joke1196: I did not know about it. I will try using it in the future!

python-checks/src/main/java/org/sonar/python/checks/IsCloseAbsTolCheck.java
```diff
@@ -21,21 +21,28 @@
...
 import org.sonar.plugins.python.api.tree.Tree;
+import org.sonar.python.quickfix.TextEditUtils;
 import org.sonar.python.tree.TreeUtils;
 
 @Rule(key = "S6727")
 public class IsCloseAbsTolCheck extends PythonSubscriptionCheck {
 
-  private static final String MESSAGE = "Provide the abs_tol parameter when using math.isclose to compare a value to 0";
+  private static final String MESSAGE = "Provide the abs_tol parameter when using math.isclose to compare a value to 0.";
+  private static final String SECOND_LOCATION = "This argument evaluates to zero.";
```
guillaume-dequenne-sonarsource, line 44: _Nitpick:_ Probably it would make more sense to name this `SECONDARY_LOCATION_MESSAGE`.

python-checks/src/main/java/org/sonar/python/checks/IsCloseAbsTolCheck.java
```diff
@@ -44,21 +51,35 @@ public void initialize(Context context) {
...
-    RegularArgument secondArg = TreeUtils.nthArgumentOrKeyword(1, "b", call.arguments());
-    return (firstArg != null && isLitteralZeroOrAssignedZero(firstArg.expression())) ||
-        (secondArg != null && isLitteralZeroOrAssignedZero(secondArg.expression()));
+  private static Optional<RegularArgument> checkArgumentExistsAndIsZero(@Nullable RegularArgument argument) {
+    return Optional
+        .ofNullable(argument)
+        .filter(arg -> isLiteralZeroOrAssignedZero(arg.expression()));
+  }
+
+  private static void addIssueAndQuickFix(RegularArgument arg, SubscriptionContext ctx, CallExpression call) {
```
guillaume-dequenne-sonarsource, line 72: This looks good but it's missing a unit test: we have a `PythonQuickFixVerifier` class that allows you to test before / after snippets of code when applying quick fixes, so that we're sure that everything displays properly.

guillaume-dequenne-sonarsource (changes requested):
Looking good, I think however that we need a unit test for the quick fix.

python-checks/src/main/java/org/sonar/python/checks/IsCloseAbsTolCheck.java
joke1196: Thank you, yes, I went too fast!

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1554, opened 2023-09-08
SONARPY-1443 Rule S6725: Equality checks should not be made against numpy.nan

jdodinh:


python-checks/src/test/resources/checks/numpyIsNan.py
```diff
@@ -0,0 +1,14 @@
+import numpy as np
+
+
+def foo(x):
+    if x == np.nan: print(1)  # Noncompliant
+    #  ^^^^^^^^^^^
+    if np.nan == x: print(1)  # Noncompliant
```
joke1196, line 7: Looks really good, great job! 

I was thinking, maybe we could add a small test with for example:

```
import numpy

def foo():
    if x == numpy.nan
``` 
Or instead of `import numpy as np` `import numpy as something_else`. 
WDYT?

joke1196 (approved):
Great work. I left a small suggestion, no changes are required! It is up to you!

python-checks/src/test/resources/checks/numpyIsNan2.py
```diff
@@ -0,0 +1,14 @@
+import numpy as xx
+
+
+def foo(x):
+    if x == xx.nan: print(1)  # Noncompliant
```
joke1196, line 5: Sorry, I missed it the first time. We are missing the check for the noncompliant message 
````
if x == xx.nan: print(1) # Noncompliant {{Equality checks should not be made against \"numpy.nan\". Use numpy.isnan() instead.}}
```

python-checks/src/test/resources/checks/numpyIsNan.py
jdodinh: Thanks for the tip, and for raising the issue. Turns out that my check didn't work is cases where we do `from numpy import nan`. I have now addressed the change in `b80fd05f99ef4c801e5dbdb6b3be08d39b8966b5`.

python-checks/src/test/resources/checks/numpyIsNan2.py
jdodinh: No need to be sorry. You didn't miss anything, b80fd05f99ef4c801e5dbdb6b3be08d39b8966b5 was a response to your comment :)

joke1196: Yes agree, I do not think we should check for this. I think it is more than enough what you already did, it covers all the main cases.

jdodinh: Addressed in 51ce44553971f9b57e200745beb6fcfea817d406

python-checks/src/main/java/org/sonar/python/checks/NumpyIsNanCheck.java
```diff
@@ -0,0 +1,64 @@
...
+    if (operand.is(Tree.Kind.QUALIFIED_EXPR)) {
+      cast = (QualifiedExpression) operand;
+    } else if (operand.is(Tree.Kind.NAME)) {
+      cast = (Name) operand;
+    } else {
+      return;
+    }
+    Symbol symbol = cast.symbol();
+    if (symbol != null && "numpy.nan".equals(symbol.fullyQualifiedName())) {
+      ctx.addIssue(be, "Equality checks should not be made against \"numpy.nan\". Use numpy.isnan() instead.");
```
guillaume-dequenne-sonarsource, line 61: Small feedback: the issue message should be imperative. The description part is then explained in the "Why is this an issue" section.
I'd suggest to simply use `Don't perform an equality check against "numpy.nan".`
When implemented, the quick fix could then read `Use numpy.isnan`

Still, good job and kudos on your first merged PR!

jdodinh: Addressed in 42964f1f0e484919e3c122e61ab1337e964bbfb8

PR #1555, opened 2023-09-11
SONARPY-1452 S930: Fix FN on math.acos calls

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/ArgumentNumberCheck.java
```diff
@@ -55,19 +57,54 @@ public class ArgumentNumberCheck extends PythonSubscriptionCheck {
...
+        .collect(Collectors.toList());
+
+      var isEqualAmountOfArguments = functionSymbols.stream()
+        .map(FunctionSymbol::parameters)
+        .filter(Objects::nonNull)
+        .map(List::size)
+        .distinct()
+        .count() == 1;
+
+      if (isEqualAmountOfArguments) {
```
guillaume-dequenne-sonarsource, line 92: Argument count may be the same, but what if they have different keywords? 
I guess we could then raise a false positive due to that.
Maybe the simplest approach would be to make sure they also have the same keywords and also positional/keyword-only properties. (and we probably need some tests for that as well)

guillaume-dequenne-sonarsource (changes requested):
I think the approach is good but I'm afraid we might have some false positives by not taking into account the argument names. I think it would be worth looking into.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1556, opened 2023-09-11
SONARPY-1363 S2638 ChangeMethodContract: Should properly state the missing parameter name and not null.

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/ChangeMethodContractCheck.java
```diff
@@ -89,15 +89,24 @@ private static void checkMethodContract(SubscriptionContext ctx, FunctionSymbol
...
-      reportIssue(ctx, "Add missing parameters " + missingParameters.trim() + ".", method.definitionLocation(), overriddenMethod);
+//      If at least one parameter missing name - set message as "Add missing N parameters" where N is amount of missing parameters
+      reportIssue(ctx, getMissingParametersMessage(missingParameters), method.definitionLocation(), overriddenMethod);
     }
   }
 
+  private static String getMissingParametersMessage(List<String> missingParameters) {
+    if (missingParameters.contains(null)) {
+      return missingParameters.size() == 1 ?
+        "Add missing parameter." : ("Add missing " + missingParameters.size()  + " parameters.");
```
guillaume-dequenne-sonarsource, line 105: I feel we could change a bit the message when we don't give a name, to be clearer.
`Add 1 missing parameter.` and `Add X missing parameters.` may sound a bit more natural, wdyt?

guillaume-dequenne-sonarsource (approved):
Looks good!
I think the message could be improved just a bit in case no name is available, but this is pretty minor.

PR #1557, opened 2023-09-12
SONARPY-1375 Fix FN: S2638 should report on ambiguous symbols when no definition contract is respected

maksim-grebeniuk-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/SymbolUtils.java
```diff
@@ -252,6 +251,42 @@ public static Optional<FunctionSymbol> getOverriddenMethod(FunctionSymbol functi
     return Optional.empty();
   }
 
+  public static Optional<FunctionSymbol> getFunctionSymbol(@Nullable Symbol symbol) {
```
guillaume-dequenne-sonarsource, line 254: See my comment [here](https://github.com/SonarSource/sonar-python/pull/1559/files#r1323188284) as I reviewed this part in the second PR, which I happened to review before this one.

guillaume-dequenne-sonarsource (commented):
Changes look good. I think the refactoring can be improved, maybe here or in a separate PR. Let's discuss this.

python-frontend/src/main/java/org/sonar/python/semantic/SymbolUtils.java
```diff
@@ -252,6 +251,42 @@ public static Optional<FunctionSymbol> getOverriddenMethod(FunctionSymbol functi
...
+      return Optional.of(symbol).map(FunctionSymbol.class::cast);
+    }
+
+    if (symbol.is(Symbol.Kind.AMBIGUOUS)) {
+      var ambiguousSymbol = (AmbiguousSymbol) symbol;
+
+      var functionSymbols = ambiguousSymbol
+        .alternatives()
+        .stream()
+        .filter(alternative -> alternative.is(Symbol.Kind.FUNCTION))
```
maksim-grebeniuk-sonarsource, line 269: instead of filtering `FUNCTION`, we need to make sure that every alternative is a `FUNCTION`

python-frontend/src/main/java/org/sonar/python/semantic/SymbolUtils.java
```diff
@@ -230,34 +230,32 @@ public static int firstParameterOffset(FunctionSymbol functionSymbol, boolean is
...
+      .map(c -> c.resolveMember(functionSymbol.name())
+        .map(SymbolUtils::getFunctionSymbols)
+        .orElseGet(List::of))
+      .filter(Predicate.not(List::isEmpty))
+      .findFirst()
+      .orElseGet(List::of);
   }
 
-  public static Optional<FunctionSymbol> getFunctionSymbol(@Nullable Symbol symbol) {
+  public static List<FunctionSymbol> getFunctionSymbols(@Nullable Symbol symbol) {
```
guillaume-dequenne-sonarsource, line 252: I kind of feel like we should keep a singular version of this method as well as `getOverridenMethod` because from an API perspective it feels a bit surprising to expect a list of function symbols from a symbol, knowing you'd get a list only if there is an ambiguous symbol with only functions that match whatever disambiguation strategy is selected.

python-frontend/src/main/java/org/sonar/python/semantic/SymbolUtils.java
```diff
@@ -230,34 +230,32 @@ public static int firstParameterOffset(FunctionSymbol functionSymbol, boolean is
     return 0;
   }
 
-  public static Optional<FunctionSymbol> getOverriddenMethod(FunctionSymbol functionSymbol) {
+  public static List<FunctionSymbol> getOverriddenMethods(FunctionSymbol functionSymbol) {
```
guillaume-dequenne-sonarsource, line 233: While the change in behavior makes sense for what we need, I think we still want to keep the original behavior accessible.

What I had in mind was to keep the methods
```
public static Optional<FunctionSymbol> getOverriddenMethod(FunctionSymbol functionSymbol)
Optional<FunctionSymbol> getFunctionSymbol(@Nullable Symbol symbol)
```
and then have methods that would do:
```
public static Optional<FunctionSymbol> getOverriddenMethod(FunctionSymbol functionSymbol, Predicate<AmbiguousSymbol> disambiguationPredicate)
Optional<FunctionSymbol> getFunctionSymbol(@Nullable Symbol symbol, Predicate<AmbiguousSymbol> disambiguationPredicate)
```

With the default predicate being just something that always returns false and then a `isSameArguments` that would behave as we expect here.

So that by default the basic methods always return a single value iff there is only one, and the callee doesn't have to call anything else.

python-checks/src/main/java/org/sonar/python/checks/TooManyParametersCheck.java
```diff
@@ -87,7 +87,7 @@ private void checkFunctionDef(SubscriptionContext ctx) {
 
   private boolean isAlreadyReportedInParent(FunctionSymbol functionSymbol) {
     // If the rule would already raise an issue on a parent class, don't raise the issue twice
-    return SymbolUtils.getOverriddenMethod(functionSymbol).map(f -> {
+    return SymbolUtils.getOverriddenMethods(functionSymbol).stream().findFirst().map(f -> {
```
guillaume-dequenne-sonarsource, line 90: Here the change of behavior is okay because we might just remove FPs thanks to the fact `getOverriddenMethods` might return more than its counterpart, but I'm not sure it's really intentional, and it could have been different in another rule.

guillaume-dequenne-sonarsource (changes requested):
I still think we can improve the API in the refactoring to make sure it's easier to use.
Open to challenge though, as this PR shows that there are not many impacted rules and so far it would just remove FPs.

python-frontend/src/test/java/org/sonar/python/semantic/SymbolUtilsTest.java
```diff
@@ -192,10 +198,53 @@ void get_overridden_method() {
     FunctionSymbolImpl foo10 = new FunctionSymbolImpl(functionDef, "mod.foo", pythonFile("mod.py"));
     foo10.setOwner(new SymbolImpl("some", "some"));
     assertThat(SymbolUtils.canBeAnOverridingMethod(foo10)).isFalse();
+  }
 
+  @Test
+  void getFunctionSymbolTest() {
```
joke1196, line 204: _Nitpick_: I think the test name should be getFunctionSymbol**s**Test.

joke1196 (approved):
LGTM. I have just left a super minor comment. Everything looks good!

PR #1558, opened 2023-09-12
SONARPY-1449 Rule S1244: Floating point numbers should not be tested for equality

joke1196:


its/ruling/src/test/resources/expected/python-S1244.json
```diff
@@ -0,0 +1,186 @@
...
+925,
+982,
+983
+],
+"project:numpy-1.16.4/numpy/f2py/crackfortran.py": [
+2157,
+2157
+],
+"project:numpy-1.16.4/numpy/lib/function_base.py": [
+420,
```
joke1196, line 61: There are quite a few issues raised. I am not sure about all  the `np.sum(a == 0.0)` it seems like a TP for me.

python-checks/src/main/java/org/sonar/python/checks/FloatingPointEqualityCheck.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  private static boolean isFloat(Expression expression) {
+    return expression.is(Tree.Kind.NUMERIC_LITERAL) && expression.type().equals(InferredTypes.FLOAT);
+  }
+
+  private boolean isAssignedFloat(Expression expression) {
+    if (expression.is(Tree.Kind.NAME)) {
+      Set<Expression> values = reachingDefinitionsAnalysis.valuesAtLocation((Name) expression);
+      if (values.size() == 1) {
```
guillaume-dequenne-sonarsource, line 83: I guess we could also report if all values are float when there are multiples.

guillaume-dequenne-sonarsource (changes requested):
Looking good overall.
I think we need to change the message a little, and to ensure that the quick fix doesn't break the code. Happy to discuss this.

python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S1244.json
```diff
@@ -0,0 +1,21 @@
+{
+  "title": "Floating point numbers should not be tested for equality",
+  "type": "BUG",
+  "code": {
+    "impacts": {
+      "RELIABILITY": "MEDIUM"
```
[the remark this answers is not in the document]
joke1196: Yes sure, it makes sense!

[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: Sorry, after looking back, it seems to have been set at medium during the clean code taxonomy effort, for all languages in which this rule is implemented.
I guess it makes sense to stay consistent then.

python-checks/src/main/java/org/sonar/python/checks/FloatingPointEqualityCheck.java
```diff
@@ -38,20 +43,29 @@
...
+
+  private String importedModuleForIsClose;
+  private Name importedAlias;
 
   @Override
   public void initialize(Context context) {
     context.registerSyntaxNodeConsumer(Tree.Kind.FILE_INPUT,
         ctx -> reachingDefinitionsAnalysis = new ReachingDefinitionsAnalysis(ctx.pythonFile()));
 
+    context.registerSyntaxNodeConsumer(Tree.Kind.IMPORT_NAME,
```
guillaume-dequenne-sonarsource, line 66: Do we need to monitor `IMPORT_FROM` as well? (I don't think so, it's unlikely the `isclose` function has already been imported alone, and worst case we just reimport the whole module next to it - I don't think avoiding that is worth the extra effort)

python-checks/src/main/java/org/sonar/python/checks/FloatingPointEqualityCheck.java
```diff
@@ -94,15 +108,41 @@ private boolean isBinaryOperationWithFloat(Expression expression) {
...
+    }
+    if(importedModuleForIsClose != null){
+      return importedModuleForIsClose;
+    }
+    return "math";
+  }
+
+  private void addImportedName(AliasedName aliasedName) {
+    List<Name> importNames = aliasedName.dottedName().names();
+    if (importedModuleForIsClose == null && importNames.size() == 1 && SUPPORTED_IS_CLOSE_MODULES.contains(importNames.get(0).name())) {
```
guillaume-dequenne-sonarsource, line 143: Why the condition on `importNames.size() == 1 ` ?
It's unlikely that someone does `import numpy, math`, but who knows?
Also, the current logic will use whatever was imported last, maybe we can have some preference ordering? Although that point is pretty minor.

python-checks/src/main/java/org/sonar/python/checks/FloatingPointEqualityCheck.java
```diff
@@ -38,20 +43,29 @@
...
+  private static final String QUICK_FIX_IMPORTED_MODULE = "%s%s.isclose(%s, %s, rtol=1e-09, atol=1e-09)";
 
   private static final Tree.Kind[] BINARY_OPERATION_KINDS = { Tree.Kind.PLUS, Tree.Kind.MINUS, Tree.Kind.MULTIPLICATION,
       Tree.Kind.DIVISION };
 
   private ReachingDefinitionsAnalysis reachingDefinitionsAnalysis;
+  private static final List<String> SUPPORTED_IS_CLOSE_MODULES = Arrays.asList("math", "numpy", "torch");
+
+  private String importedModuleForIsClose;
+  private Name importedAlias;
```
guillaume-dequenne-sonarsource, line 59: Note that the same Check object is used throughout the analysis: if those names are set during the analysis of one file, they will not be reinitialized in the next file, and we might erroneously assume names are imported when they're not.

python-checks/src/main/java/org/sonar/python/checks/FloatingPointEqualityCheck.java
joke1196: Yes while doing this I noticed that  `dotteName().names()` is always size 1 even with the test `import some_module, math` both imports are considered separately, maybe we could change the API here but I will investigate that later on. Still, I have changed the function to be a bit more robust. For the priority I find it difficult to choose this is why I arbitrarily picked the first one. I could prioritize torch or numpy before math.

python-checks/src/main/java/org/sonar/python/checks/FloatingPointEqualityCheck.java
joke1196: I do not think we do. I feel the rule is quite complex only for the quick fix and handling all the cases can be difficult. I also think it would be ok to just put the import at the top.

python-checks/src/main/java/org/sonar/python/checks/FloatingPointEqualityCheck.java
```diff
@@ -0,0 +1,165 @@
...
+
+    String quickFixText = MATH_MODULE.equals(isCloseModuleName) ? QUICK_FIX_MATH : QUICK_FIX_IMPORTED_MODULE;
+    String quickFixTextWithModuleName = String.format(quickFixText, notToken, isCloseModuleName,
+        TreeUtils.treeToString(binaryExpression.leftOperand(), false),
+        TreeUtils.treeToString(binaryExpression.rightOperand(), false));
+
+    quickFix.addTextEdit(TextEditUtils.replace(binaryExpression, quickFixTextWithModuleName));
+
+    if (MATH_MODULE.equals(isCloseModuleName) && !isMathImported) {
+      quickFix.addTextEdit(TextEditUtils.insertAtPosition(0, 0, "import math\n"));
```
guillaume-dequenne-sonarsource, line 133: One thing that may be annoying is that the user may not realize we're adding this import (the focus will not move as the quick fix gets added), and this may mess with stuff like license headers.

I don't really have a better solution: adding the import right before the `isclose` call may be better at getting the user attention (and then they can just move the import to where it belongs), but it's not ideal either to be honest.

guillaume-dequenne-sonarsource (approved):
LGTM.
The way we add the import may be problematic in some cases, but I think we can proceed with this and readjust later if we find it to be an actual problem, as I don't have a better suggestion for now.

python-checks/src/main/java/org/sonar/python/checks/FloatingPointEqualityCheck.java
joke1196: I agree this is not ideal. I am not sure as well how we could tackle this. Finding the first global import and inserting before or finding the last header line.

PR #1559, opened 2023-09-12
SONARPY-1364 S2638 Argument number check should correctly detect tzname number of parameters

maksim-grebeniuk-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/SymbolUtils.java
```diff
@@ -252,6 +251,42 @@ public static Optional<FunctionSymbol> getOverriddenMethod(FunctionSymbol functi
     return Optional.empty();
   }
 
+  public static Optional<FunctionSymbol> getFunctionSymbol(@Nullable Symbol symbol) {
```
guillaume-dequenne-sonarsource, line 254: While I agree with the idea of having this as common logic, I'm not sure this is the best approach.

In rules such as S2638 where we care about the number and names of parameters, it makes sense to consider ambiguous symbols alternatives as equivalent when their parameters have same number and names.

However, for some other rules, we might care about the type hint of the parameters, or the return type, in which case this method might not behave as expected.

Probably we could have some disambiguation strategies that are clearly defined by the caller (either by having multiple methods, some argument or some predicate) that would return a single function / class symbol from an ambiguous symbol only if the criteria are met. And maybe we don't need to do it in this PR and have a separate one for that (or even a ticket). Happy to discuss this.

its/ruling/src/test/resources/expected/python-S2638.json
```diff
@@ -82,6 +85,13 @@
 'project:django-cms-3.7.1/cms/tests/test_permmod.py': [
 665
 ],
+'project:mypy-0.782/mypy/semanal.py': [
+4682,
```
guillaume-dequenne-sonarsource, line 89: This looks like a false positive. In the rule description we say this is okay:
* Extend the ways a parameter can be provided, i.e. change a keyword-only or positional-only parameter to a keyword-or-positional parameter. This is only true if the order of positional parameters doesn’t change. New positional parameters should be placed at the end

In this case it seems that `code` goes from keyword-only to keyword-or-positional with no other change, so it looks like we shouldn't raise the issue.
It doesn't look related to the ticket itself so maybe creating a separate ticket would make sense.

its/ruling/src/test/resources/expected/python-S2638.json
```diff
@@ -82,6 +85,13 @@
...
+4749
```
guillaume-dequenne-sonarsource, line 90: This one looks like a TP but it was hard for me to see why: the parameter `t` got renamed `typ`.
I'm not sure if there is something we could do to improve the reporting here, but it's not ideal (not asking for change here, and not even sure we should create a ticket because I can imagine the cost to benefit would be pretty bad). Just wanted to share this, happy to discuss.

guillaume-dequenne-sonarsource (changes requested):
The fix itself looks good but I'm not sure about the refactoring into `SymbolUtils`.
There is one FP in the ruling as well that we probably should address, most likely in a separate ticket.

python-checks/src/test/resources/checks/changeMethodContract.py
```diff
@@ -118,3 +118,9 @@ def read(self): ... # Noncompliant {{Add 1 missing parameter. This method overri
 
 class MyTextIOWrapper(TextIOWrapper):
   def seek(self): ... # Noncompliant {{Add 2 missing parameters. This method overrides io.TextIOWrapper.seek.}}
+
+import datetime
+
+class MyTZInfo(datetime.tzinfo):
+  def tzname(self): # Noncompliant
+    ...
```
joke1196, line 126: I was thinking, shouldn't this raise as well a ArgumentNumber issue? Or is the test on line 71 of argumentNumber.py at the wrong place?

joke1196 (approved):
LGTM. I have just added a small question related to a similar test in the rule S930. Maybe you have already sorted that out, but it is just to be sure.

PR #1560, opened 2023-09-12
SONARPY-1389 Fix parse error when an unpacking expression is used as subscript

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/tree/PythonTreeMaker.java
```diff
@@ -1373,7 +1373,12 @@ private Expression subscriptionOrSlicing(Expression expr, Token leftBracket, Ast
     for (AstNode subscript : subscriptList.getChildren(PythonGrammar.SUBSCRIPT)) {
       AstNode colon = subscript.getFirstChild(PythonPunctuator.COLON);
       if (colon == null) {
```
maksim-grebeniuk-sonarsource, line 1375: the `getFirstChild` method is already able to return the child of the first matching given type.
So the whole section could be implemented like:
```java
var slice = colon == null ?
        expression(subscript.getFirstChild(PythonGrammar.NAMED_EXPR_TEST, PythonGrammar.STAR_EXPR))
        : sliceItem(subscript);
slices.add(slice);
```
or instead of the loop could this could be done with stream API:
```java
List<Tree> slices = subscriptList.getChildren(PythonGrammar.SUBSCRIPT)
      .stream()
      .map(subscript -> subscript.getFirstChild(PythonPunctuator.COLON) == null ?
          expression(subscript.getFirstChild(PythonGrammar.NAMED_EXPR_TEST, PythonGrammar.STAR_EXPR))
          : sliceItem(subscript))
      .collect(Collectors.toList());
```

maksim-grebeniuk-sonarsource (approved):
Looks good, just one small hint how to clean up the code for a bit

PR #1562, opened 2023-09-13
SONARPY-1448 Rule S6730: Deprecated NumPy aliases of built-in types should not be used

joke1196:


python-checks/src/main/java/org/sonar/python/checks/DeprecatedNumpyTypesCheck.java
```diff
@@ -0,0 +1,78 @@
...
+import org.sonar.plugins.python.api.tree.Tree;
+import org.sonar.python.quickfix.TextEditUtils;
+
+@Rule(key = "S6730")
+public class DeprecatedNumpyTypesCheck extends PythonSubscriptionCheck {
+
+  private static final String MESSAGE = "Replace this deprecated numpy type alias with the builtin type %s.";
+  private static final String QUICK_FIX_MESSAGE = "Replace with %s.";
+
+  private static final Map<String, String> typesToCheck = Map.of(
```
maksim-grebeniuk-sonarsource, line 42: I guess it is good to use a constant naming strategy here like `TYPES_TO_CHECK`

python-checks/src/main/java/org/sonar/python/checks/DeprecatedNumpyTypesCheck.java
```diff
@@ -0,0 +1,78 @@
...
+        DeprecatedNumpyTypesCheck::checkForDeprecatedTypesNames);
+  }
+
+  private static void checkForDeprecatedTypesNames(SubscriptionContext ctx) {
+    QualifiedExpression expression = (QualifiedExpression) ctx.syntaxNode();
+    Symbol symbol = expression.symbol();
+    checkForNumpyType(symbol, expression, ctx);
+  }
+
+  private static void checkForNumpyType(@Nullable Symbol symbol, Tree expression, SubscriptionContext ctx) {
```
maksim-grebeniuk-sonarsource, line 64: Minor: I think there is not that much sense to have the separate method here. I guess the whole logic could be done in the `checkForDeprecatedTypeNames` method like:
```java
Optional.of(ctx.syntaxNode)
  .map(QualifiedExpression.class::cast)
  .ifPresent(expression -> Optional.ofNullable(expression.symbol)
    .map(typesToCheck::get)
    .ifPresent(type -> raiseIssue(expression, type, ctx)));
```

maksim-grebeniuk-sonarsource (approved):
LGTM, see a couple of minor hints, feel free to ignore

python-checks/src/main/java/org/sonar/python/checks/DeprecatedNumpyTypesCheck.java
joke1196: Yes, good catch! Thank you. This was a left over from some old implementation.

PR #1566, opened 2023-09-14
Rule S6729: np.nonzero should be preferred over np.where when only the condition parameter is set.

jdodinh:


python-checks/src/test/resources/checks/numpyWhereOneCondition.py
```diff
@@ -0,0 +1,47 @@
+def import_1():
+    import numpy as np
+    arr = np.array([1, 2, 3, 4])
+
+    empty = np.where()
+    args = (arr > 1, arr + 1, arr)
+    result = np.where(*args)
+    indices = np.nonzero(arr > 1)
```
joke1196, line 8: I think this test `indices = np.nonzero(arr > 1)` repeats itself quite a few times. Having it only once should be enough.

python-checks/src/test/resources/checks/numpyWhereOneCondition.py
```diff
@@ -0,0 +1,47 @@
...
+    import numpy as xx
+    arr = xx.array([1, 2, 3, 4])
+
+    result = xx.where(arr > 2)  # Noncompliant
+    #        ^^^^^^^^^^^^^^^^^
+
+    result = xx.where(arr > 1, arr + 1, arr)
+    indices = xx.nonzero(arr > 1)
+
+    result = np.where(arr > 2)
```
joke1196, line 32: One missing test could be when the user provides the parameter keyword as well. I don't think we want to raise the issue if the user does weird stuff like `np.where(x=arr+1)` it will be confusing to raise the issue with the message about using `np.nonzero` because it does not seem like it is the intention. Feel free to challenge here as I am not 100% sure it is worth the effort.

python-checks/src/test/resources/checks/numpyWhereOneCondition.py
```diff
@@ -0,0 +1,47 @@
...
+    from numpy import where, nonzero, array
+    arr = array([1, 2, 3, 4])
+
+    result = where(arr > 2)  # Noncompliant {{Use "np.nonzero" when only the condition parameter is provided to "np.where".}}
+    #        ^^^^^^^^^^^^^^
+
+    result = where(arr > 1, arr + 1, arr)
+    indices = nonzero(arr > 1)
+
+    result = xx.where(arr > 2)
```
joke1196, line 46: Not sure if we need this test again about the non-available `xx` or `np`.

python-checks/src/test/resources/checks/numpyWhereOneCondition.py
```diff
@@ -0,0 +1,47 @@
+def import_1():
+    import numpy as np
+    arr = np.array([1, 2, 3, 4])
+
+    empty = np.where()
+    args = (arr > 1, arr + 1, arr)
+    result = np.where(*args)
```
joke1196, line 7: Maybe we can add a test with the tuple `(arr > 1)` to see if the issue is still raised. I think that would be quite nice as it is a less obvious mistake.

python-checks/src/main/java/org/sonar/python/checks/NumpyWhereOneConditionCheck.java
```diff
@@ -0,0 +1,54 @@
...
+
+@Rule(key = "S6729")
+public class NumpyWhereOneConditionCheck extends PythonSubscriptionCheck {
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.CALL_EXPR, NumpyWhereOneConditionCheck::checkNumpyWhereCall);
+  }
+
+  private static void checkNumpyWhereCall(SubscriptionContext ctx) {
+    // To be implemented.
```
joke1196, line 38: I think we can remove this comment :)

python-checks/src/main/java/org/sonar/python/checks/NumpyWhereOneConditionCheck.java
```diff
@@ -0,0 +1,54 @@
...
+    CallExpression ce = (CallExpression) ctx.syntaxNode();
+    Symbol symbol = ce.calleeSymbol();
+    if (symbol != null && hasOneParameter(symbol, ce)) {
```
joke1196, line 41: Not useful here, but just for you to know there are a few helpful methods in the `TreeUtils` class, for example, to grab specific arguments by keyword or position.

python-checks/src/main/java/org/sonar/python/checks/NumpyWhereOneConditionCheck.java
```diff
@@ -0,0 +1,54 @@
...
+      ctx.addIssue(ce, "Use \"np.nonzero\" when only the condition parameter is provided to \"np.where\".");
+    }
+  }
+
+  private static boolean hasOneParameter(Symbol symbol, CallExpression ce) {
+    ArgList argList = ce.argumentList();
```
joke1196, line 47: Here we could simplify the check by using `ce.arguments()` this way you don't have to check for `null` but only for the size of the list.

joke1196 (changes requested):
Looks good! I added a few comments mainly to remove things here and there. Good job!

python-checks/src/main/java/org/sonar/python/checks/NumpyWhereOneConditionCheck.java
```diff
@@ -0,0 +1,69 @@
...
+      return false;
+    }
+
+    RegularArgument regArg = (RegularArgument) argList.get(0);
+    Name keywordArgument = regArg.keywordArgument();
+    if (keywordArgument == null) {
+      return true;
+    }
+    Optional<String> optName = Optional.ofNullable(keywordArgument.name()).filter(name -> "condition".equals(keywordArgument.name()));
+    return optName.isPresent();
```
joke1196, line 67: Here we don't need to assign a new variable we could just return the Optional statement directly.

python-checks/src/main/java/org/sonar/python/checks/NumpyWhereOneConditionCheck.java
```diff
@@ -0,0 +1,69 @@
...
+  private static void checkNumpyWhereCall(SubscriptionContext ctx) {
+    CallExpression ce = (CallExpression) ctx.syntaxNode();
+    Symbol symbol = ce.calleeSymbol();
+    if (symbol == null) {
+      return;
+    }
+    if (!"numpy.where".equals(symbol.fullyQualifiedName())) {
+      return;
+    }
+    if (hasOneParameter(ce)) {
```
joke1196, line 50: Feel free to ignore, as it is more a preference, but we could merge this into a single if statement i.e.
```
if(symbol != null && "numpy.where".equals(symbol.fullyQualifiedName()) && hasOneParameter(ce)){
 raiseTheIssue ...
}
```

joke1196 (changes requested):
LGTM! Just a minor change, and a small suggestion.

joke1196 (approved):
LGTM! Good job!

PR #1567, opened 2023-09-15
SONARPY-1446 Rule S6711: numpy.random.Generator should be preferred to numpy.random.RandomState

joke1196:


joke1196 (commented):
So I have decided to raise an issue on all the legacy random generator methods. It is a bit different than what is shown in the RSPEC. In the RSPEC we do not raise any issue for numpy.random.seed.

python-checks/src/main/java/org/sonar/python/checks/NumpyRandomStateCheck.java
```diff
@@ -0,0 +1,104 @@
...
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.CALL_EXPR, NumpyRandomStateCheck::checkNumpyRandomState);
+  }
+
+  private static void checkNumpyRandomState(SubscriptionContext ctx) {
+    CallExpression call = (CallExpression) ctx.syntaxNode();
+    Optional.ofNullable(call.calleeSymbol())
+        .map(Symbol::fullyQualifiedName)
+        .filter(fqn -> fqn.startsWith(LEGACY_MODULE_NAME) || LEGACY_RANDOM_FUNCTIONS.contains(fqn))
+        .ifPresent(symbol -> ctx.addIssue(call.callee(), MESSAGE));
```
jdodinh, line 101: Very nice and clean way to add the issue! Will definitely make more use of the Java functional interface in the future.

python-checks/src/test/resources/checks/numpyRandomStateCheck.py
```diff
@@ -0,0 +1,38 @@
...
+    import numpy as np
+    generator = np.random.default_rng(42)
+    generator.standard_normal()
+
+    generator.f(a, b)
+    generator.poisson()
+
+    from numpy.random import Generator, PCG64
+    rng = Generator(PCG64())
+    rng.standard_normal()
```
jdodinh, line 38: Looks good to me.

jdodinh (approved):

python-checks/src/main/java/org/sonar/python/checks/NumpyRandomStateCheck.java
jdodinh, line 101: Maybe `symbol` is a bit misleading in this context. I would suggest using `fqn` to stay consistent with the output of the filer function.

guillaume-dequenne-sonarsource (commented):
Implementation looks good, I suggest simplifying the message a bit.

Seeing the issue displayed in SonarQube, I think we need to update the RSPEC to map the alternatives that should be used instead of the legacy functions, otherwise addressing the issues may be frustrating.

python-checks/src/main/java/org/sonar/python/checks/NumpyRandomStateCheck.java
```diff
@@ -0,0 +1,104 @@
...
+import org.sonar.plugins.python.api.tree.CallExpression;
+import org.sonar.plugins.python.api.tree.Tree;
+
+@Rule(key = "S6711")
+public class NumpyRandomStateCheck extends PythonSubscriptionCheck {
+
+  private static final String MESSAGE = "Do not use legacy functions that relies on \"numpy.random.RandomState\", use their \"numpy.random.Generator\" counterpart instead.";
+
+  private static final String LEGACY_MODULE_NAME = "numpy.random.RandomState";
+  private static final List<String> LEGACY_RANDOM_FUNCTIONS = List.of(
```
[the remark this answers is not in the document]
joke1196: Sure, makes sense. I will remove `numpy.random.seed` from the raised issues, and once the rspec is modified I will remove other functions if necessary.

PR #1568, opened 2023-09-18
SONARPY-1445 Rule S6709: Results that depend on random number generation should be reproducible.

joke1196:


jdodinh (commented):
Everything LGTM here. The test cases are comprehensive, and the argument checking is very thorough.

python-checks/src/main/java/org/sonar/python/checks/NumpyRandomSeedCheck.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  private boolean isSeedArgumentAbsentOrNone(String argName, List<Argument> args) {
+    RegularArgument arg = TreeUtils.nthArgumentOrKeyword(0, argName, args);
+    return arg == null || arg.expression().is(Tree.Kind.NONE) || isAssignedNone(arg.expression());
+  }
+
+  private boolean isAssignedNone(Expression exp) {
+    if(exp.is(Tree.Kind.NAME)){
+      Set<Expression> values = this.reachingDefinitionsAnalysis.valuesAtLocation((Name)exp);
```
maksim-grebeniuk-sonarsource, line 80: is `this` needed here? also there is one space missing between `(Name)` and `exp`.
This is my personal preference but since you already using stream API here I think the whole method could be implemented in the following way:
```java
  private boolean isAssignedNone(Expression exp) {
    return Optional.of(exp)
      .map(reachingDefinitionsAnalysis::valuesAtLocation)
      .filter(Predicate.not(Set::isEmpty))
      .filter(values -> values.stream().allMatch(value -> value.is(Tree.Kind.NONE)))
      .isPresent();
  }
```

maksim-grebeniuk-sonarsource (approved):
LGTM, just one minor formatting issue

python-checks/src/main/java/org/sonar/python/checks/NumpyRandomSeedCheck.java
joke1196: Great thank you very much! I did not know about this `toOptionalInstanceOfMapper` quite useful here.

PR #1569, opened 2023-09-18
SONARPY-1374 Fix FN on S5607: list and tuple are not duck-type compatible

guillaume-dequenne-sonarsource:
I did not see a clean and easy way to represent this other than by hardcoding it.
By our definition, a list can be used as a substitute for a tuple and they are therefore compatible types in that regard, as they define the same members.
We know it's actually not the case and I therefore hardcoded that information.

### DIFF
--- a/python-frontend/src/main/java/org/sonar/python/types/InferredTypes.java
+++ b/python-frontend/src/main/java/org/sonar/python/types/InferredTypes.java
@@ -91,6 +91,7 @@ public class InferredTypes {
   private static final String BYTES = "bytes";
   // https://github.com/python/mypy/blob/e97377c454a1d5c019e9c56871d5f229db6b47b2/mypy/semanal_classprop.py#L16-L46
   private static final Map<String, Set<String>> HARDCODED_COMPATIBLE_TYPES = new HashMap<>();
+  private static final Set<String> HARDCODED_INCOMPATIBLE_TYPES = Set.of("tuple", "list");
 
   static {
     HARDCODED_COMPATIBLE_TYPES.put(BuiltinTypes.INT, new HashSet<>(Arrays.asList(BuiltinTypes.FLOAT, BuiltinTypes.COMPLEX)));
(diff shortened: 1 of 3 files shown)


python-frontend/src/main/java/org/sonar/python/types/InferredTypes.java
```diff
@@ -364,14 +365,26 @@ static boolean areSymbolsCompatible(Symbol actual, Symbol expected) {
...
+    boolean areHardcodedIncompatible = areHardCodedIncompatible(actualTypeClass, expectedTypeClass);
+    return (areHardcodedCompatible || isDuckTypeCompatible || canBeOrExtend) && !areHardcodedIncompatible;
   }
 
   private static boolean areHardcodedCompatible(ClassSymbol actual, ClassSymbol expected) {
     Set<String> compatibleTypes = HARDCODED_COMPATIBLE_TYPES.getOrDefault(actual.fullyQualifiedName(), Collections.emptySet());
     return compatibleTypes.stream().anyMatch(expected::canBeOrExtend);
   }
 
+  private static boolean areHardCodedIncompatible(ClassSymbol actual, ClassSymbol expected) {
```
maksim-grebeniuk-sonarsource, line 377: what if we'll need to support other sets of hardcoded incompatible types?
I think a solution like below could cover it and be a bit cleaner
```java
private static final Set<Set<String>> INCOMPATIBLE_TYPES = Set.of(Set.of("tuple", "list"));

  private static boolean areHardCodedIncompatible(ClassSymbol actual, ClassSymbol expected) {
    return INCOMPATIBLE_TYPES.contains(
      Stream.of(actual, expected)
        .map(ClassSymbol::fullyQualifiedName)
        .filter(Objects::nonNull)
        .collect(Collectors.toSet())
    );
  }
```

maksim-grebeniuk-sonarsource (changes requested):
Looks good. Left one comment to be changed and note that there is a coverage gate failed.

python-frontend/src/main/java/org/sonar/python/types/InferredTypes.java
guillaume-dequenne-sonarsource: I honestly hope we won't have to use this trick for other incompatible types, but I agree this approach looks better, thanks!

maksim-grebeniuk-sonarsource (approved):
LGTM!

PR #1570, opened 2023-09-19
SONARPY-1371 S5644: Fix FN on iter calls

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource (changes requested):
The implementation looks like a good baby step for supporting those typevars but the ruling differences make me suspect we might need to exclude and document some more cases.

Also, we probably want to document the ticket number for the `Self` support at the same place we are doing special handling for it, so that it's easier to understand.

guillaume-dequenne-sonarsource (approved):
LGTM, I expect we should not wait too long to fix the TypeVar issue in a more durable way, as there is probably room for other FPs in the current implementationl.

PR #1571, opened 2023-09-19
SONARPY-750 S1192 (StringLiteralDuplicationCheck) shouldn't exclude capitalized strings

maksim-grebeniuk-sonarsource:


its/ruling/src/test/resources/expected/python-S1192.json
```diff
@@ -1,43 +1,43 @@
 {
```
guillaume-dequenne-sonarsource, line 1: Seeing the changes on the ruling, I'm not sure about whether this change is beneficial.

It seems that there are issues raised on strings that actually look like formatting patterns and I'm not sure we want that, despite the regex technically being fixed.
Wdyt?

guillaume-dequenne-sonarsource (commented):
The regex change looks good considering the ticket but I'm not sure the resulting change will be that beneficial, knowing this is already a noisy rule.
Let's discuss this.

guillaume-dequenne-sonarsource (approved):
As discussed, it seems to make sense to fix this particular ticket, despite the fact we might want to revisit how the rule is implemented altogether. Since we have a separate ticket for that, let's move on with this one.
LGTM.

PR #1573, opened 2023-09-19
SONARPY-1437 Rule S6714: Passing a list to np.array should be preferred over passing a generator.

jdodinh:


python-checks/src/test/resources/checks/numpyListOverGenerator.py
```diff
@@ -0,0 +1,35 @@
...
+    arr = np.array(x ** 2 for x in range(10))  # Noncompliant {{Pass a list to "np.array" instead of passing a generator.}}
+    #     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+    arr = np.array([x ** 2 for x in range(10)])  # Compliant: a list of 10 elements is passed to the np.array function.
+
+    arr = np.array((x ** 2 for x in range(10)), dtype=str)  # Noncompliant
+
+    arr = np.array()
+
+    gen_4 = (x * 2 for x in range(5))
```
joke1196, line 34: I think we can remove this test as it was already done on line `8`.

python-checks/src/test/resources/checks/numpyListOverGenerator.py
```diff
@@ -0,0 +1,35 @@
...
+    import numpy as np
+    from typing import Any
+
+    gen_1 = (x * 2 for x in range(5))
+    arr = np.array(gen_1)  # Noncompliant {{Pass a list to "np.array" instead of passing a generator.}}
+    #     ^^^^^^^^^^^^^^^
+
+    gen_2: Any = (x * 2 for x in range(5))  # This is a false negative. ReachingDefinitionsAnalysis does not find the variable when it has
+    # been type annotated. Let's construct a ticket for this.
+    arr = np.array(gen_2)
```
joke1196, line 14: I think we should change a bit the comment here. Because we do not want to raise an issue when the generator is type-hinted to `Any`. We should maybe add another line with another `gen_2` being assigned without a type hint. for example:
```
  gen_2 = (x * 2 for x in range(5))  # Here would be the real FN
  arr = np.array(gen_2)
``` 
Because of the constraints of ReachingDefinitionAnalysis we do not raise an issue here.
Adding this test will help when fixing the issue because it will fail if ReachingDefinitionAnalysis can handle the type hints. 

I think we can also remove the comment about the ticket, we can just create it. Here is the ticket:(https://sonarsource.atlassian.net/browse/SONARPY-1470).

python-checks/src/main/java/org/sonar/python/checks/NumpyListOverGeneratorCheck.java
```diff
@@ -0,0 +1,97 @@
...
+  private boolean isNamedGeneratorExpression(Expression expression) {
+    return Optional.of(expression)
+      .filter(expr -> expr.is(Tree.Kind.NAME))
+      .map(Name.class::cast)
+      .map(name -> this.reachingDefinitionsAnalysis.valuesAtLocation(name))
+      .filter(this::checkSetProperties)
+      .isPresent();
+  }
+
+  private boolean checkSetProperties(Set<Expression> set) {
```
joke1196, line 94: Here we could change the method to a static one. SonarQube will detect this issue, you can check the code smells on PR.

joke1196 (changes requested):
LGTM. Added a few comments mainly for the tests. But looks quite good! Nice job!

python-checks/src/test/resources/checks/numpyListOverGenerator.py
jdodinh: Removed in 51dee63374166f9f2257b621a2a6f82cbe43808c.

python-checks/src/test/resources/checks/numpyListOverGenerator.py
jdodinh: Addressed in 05740032e5e18f3cbf6e22f821f61d8f5c9f7f1f.

python-checks/src/main/java/org/sonar/python/checks/NumpyListOverGeneratorCheck.java
jdodinh: Corrected in 0f9c655b3d898b729177f761f78d055a58a8f31d.

joke1196 (approved):
LGTM! Thank you!

python-checks/src/main/java/org/sonar/python/checks/NumpyListOverGeneratorCheck.java
```diff
@@ -0,0 +1,100 @@
...
+      .map(name -> this.reachingDefinitionsAnalysis.valuesAtLocation(name))
+      .filter(NumpyListOverGeneratorCheck::checkSetProperties)
+      .isPresent();
+  }
+
+  private static boolean checkSetProperties(Set<Expression> set) {
+    return !set.isEmpty() && set.stream().allMatch(NumpyListOverGeneratorCheck::getSetCondition);
+  }
+
+  private static boolean getSetCondition(Expression expression) {
```
joke1196, line 97: Here, I think the name could be a bit more explicit i.e isGeneratorAndParentHasType or something like that.

joke1196 (approved):
LGTM. I have just added a small comment on the naming. Otherwise, it looks good. Thank you!

python-checks/src/main/java/org/sonar/python/checks/NumpyListOverGeneratorCheck.java
jdodinh: Changed in 3fb81d1c3966f5db1a9807c217e1f7a94173e340.

PR #1575, opened 2023-09-20
SONARPY-1348 Support type inference in presence of augmented assignments

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/types/TypeInference.java
```diff
@@ -191,6 +193,25 @@ public void visitAssignmentStatement(AssignmentStatement assignmentStatement) {
     assignmentsByLhs.computeIfAbsent(symbol, s -> new HashSet<>()).add(assignment);
   }
 
+  @Override
+  public void visitCompoundAssignment(CompoundAssignmentStatement compoundAssignment) {
```
maksim-grebeniuk-sonarsource, line 197: `visitCompoundAssignment` and `visitAssignmentStatement` have common part so it could be extended like this:
```java
@Override
  public void visitAssignmentStatement(AssignmentStatement assignmentStatement) {
    super.visitAssignmentStatement(assignmentStatement);
    if (assignmentStatement.lhsExpressions()
      .stream()
      .map(ExpressionList::commas)
      .allMatch(List::isEmpty)) {
      return;
    }
    List<Expression> lhsExpressions = assignmentStatement.lhsExpressions().stream()
      .flatMap(exprList -> exprList.expressions().stream())
      .collect(Collectors.toList());
    if (lhsExpressions.size() != 1) {
      return;
    }

    Expression rhs = assignmentStatement.assignedValue();
    Expression lhsExpression = lhsExpressions.get(0);
    processAssignment(assignmentStatement, lhsExpression, rhs);
  }

  @Override
  public void visitCompoundAssignment(CompoundAssignmentStatement compoundAssignment) {
    super.visitCompoundAssignment(compoundAssignment);
    Expression lhsExpression = compoundAssignment.lhsExpression();
    Expression rhs = compoundAssignment.rhsExpression();
    processAssignment(compoundAssignment, lhsExpression, rhs);
  }

  private void processAssignment(Statement assignmentStatement, Expression lhsExpression, Expression rhs) {
    Optional.of(lhsExpression)
      .ifPresent(lhs -> Optional.of(lhs)
        .map(HasSymbol::symbol)
        .filter(SymbolImpl.class::isInstance)
        .map(SymbolImpl.class::cast)
        .ifPresent(symbol -> {
          Assignment assignment = new Assignment(symbol, lhs, rhs);
          assignmentsByAssignmentStatement.put(assignmentStatement, assignment);
          assignmentsByLhs.computeIfAbsent(symbol, s -> new HashSet<>()).add(assignment);
        }));
  }

```

maksim-grebeniuk-sonarsource (approved):
LGTM, but I think common part could be extended for the changes in `TypeInference` class. Feel free to ignore.

python-frontend/src/main/java/org/sonar/python/types/TypeInference.java
guillaume-dequenne-sonarsource: I agree that it could be mutualized a bit, but at the same time the current implementation is quite straightforward so I don't really feel the refactoring brings that much. For sure if we have to duplicate the code again, it will be time to do such mutualization, but as of now I'd rather keep it as it is.

PR #1576, opened 2023-09-21
SONARPY-1462 Rule S6725: Add quick fix for equality checks against np.nan.

jdodinh:


python-checks/src/test/resources/checks/numpyIsNan.py
jdodinh, line 1: Moved the content of the previous 3 numpyIsNan functions into one.

python-checks/src/main/java/org/sonar/python/checks/NumpyIsNanCheck.java
```diff
@@ -48,17 +57,63 @@ private static void checkForIsNan(SubscriptionContext ctx) {
...
+      getQuickFix(issue, operand, be.rightOperand(), be);
+    } else {
+      getQuickFix(issue, operand, be.leftOperand(), be);
     }
-    Symbol symbol = cast.symbol();
-    if (symbol != null && "numpy.nan".equals(symbol.fullyQualifiedName())) {
-      ctx.addIssue(be, "Don't perform an equality check against \"numpy.nan\".");
+  }
+
+  private static void getQuickFix(PreciseIssue issue, Expression nanOperand, Expression otherOperand, BinaryExpression be) {
```
maksim-grebeniuk-sonarsource, line 88: maybe since the method does not actually return anything but add it to the issue - it is better to name it `addQuickFix` as well?

python-checks/src/main/java/org/sonar/python/checks/NumpyIsNanCheck.java
```diff
@@ -48,17 +57,63 @@ private static void checkForIsNan(SubscriptionContext ctx) {
...
+    Optional.of(nanOperand)
+      .filter(op -> op.is(Tree.Kind.QUALIFIED_EXPR))
+      .map(QualifiedExpression.class::cast)
+      .map(QualifiedExpression::qualifier)
+      .filter(quali -> quali.is(Tree.Kind.NAME))
+      .map(Name.class::cast)
```
maksim-grebeniuk-sonarsource, line 94:
```java
.filter(quali -> quali.is(Tree.Kind.NAME))
      .map(Name.class::cast)
```
could be just replaced with
```java
```

python-checks/src/main/java/org/sonar/python/checks/NumpyIsNanCheck.java
```diff
@@ -48,17 +57,63 @@ private static void checkForIsNan(SubscriptionContext ctx) {
...
     }
-    Symbol symbol = cast.symbol();
-    if (symbol != null && "numpy.nan".equals(symbol.fullyQualifiedName())) {
-      ctx.addIssue(be, "Don't perform an equality check against \"numpy.nan\".");
+  }
+
+  private static void getQuickFix(PreciseIssue issue, Expression nanOperand, Expression otherOperand, BinaryExpression be) {
+    Optional.of(nanOperand)
+      .filter(op -> op.is(Tree.Kind.QUALIFIED_EXPR))
+      .map(QualifiedExpression.class::cast)
```
maksim-grebeniuk-sonarsource, line 91:
```java
.filter(op -> op.is(Tree.Kind.QUALIFIED_EXPR))
      .map(QualifiedExpression.class::cast)
```
could be just replaced with
```java
.flatMap(TreeUtils.toOptionalInstanceOfMapper(QualifiedExpression.class))
```

python-checks/src/main/java/org/sonar/python/checks/NumpyIsNanCheck.java
```diff
@@ -48,17 +57,63 @@ private static void checkForIsNan(SubscriptionContext ctx) {
   }
 
   private static void checkOperand(SubscriptionContext ctx, Expression operand, BinaryExpression be) {
-    HasSymbol cast;
-    if (operand.is(Tree.Kind.QUALIFIED_EXPR)) {
-      cast = (QualifiedExpression) operand;
-    } else if (operand.is(Tree.Kind.NAME)) {
-      cast = (Name) operand;
+    Optional.ofNullable(getSymbolFromExpression(operand))
```
maksim-grebeniuk-sonarsource, line 60: since the `symbol` method is declared in the `HasSymbol` interface and both `Name` and `QualifiedExpression` are inherited from it, and the result of the `getSymbolFromExpression` method is used only in `Optional.ofNullable` - I'd suggest removing it and use `TreeUtils.getSymbolFromTree` instead:
```java
TreeUtils.getSymbolFromTree(operand)
      .filter("numpy.nan"::equals)
      .ifPresent(fqn -> {
        PreciseIssue issue = ctx.addIssue(be, MESSAGE);
        addQuickFix(issue, operand, be);
      });
```

python-checks/src/main/java/org/sonar/python/checks/NumpyIsNanCheck.java
jdodinh: Renamed in 6eb115ab593c76efe85e5c5ff00f7676fa81b455.

python-checks/src/main/java/org/sonar/python/checks/NumpyIsNanCheck.java
jdodinh: Simplidied in b480327fad4c01af3a47bedad7ab2ab595588863.

python-checks/src/main/java/org/sonar/python/checks/NumpyIsNanCheck.java
jdodinh: Simplidied in b480327fad4c01af3a47bedad7ab2ab595588863.

python-checks/src/main/java/org/sonar/python/checks/NumpyIsNanCheck.java
jdodinh: Simplified in e9b3f58c73c70c1e6796be327cff00e07ab95025.

maksim-grebeniuk-sonarsource (approved):
LGTM

PR #1577, opened 2023-09-21
SONARPY-1386 Avoid running Typeshed serializer tests when mvn has -DskipTests argument

guillaume-dequenne-sonarsource:


python-frontend/typeshed_serializer/runners/tox_runner.py
```diff
@@ -140,8 +141,17 @@ def main():
...
+        if skip_tests:
+            logger.info("SKIPPING TYPESHED SERIALIZER TESTS")
+            return
         subprocess.run(['tox', '-e', 'py39'], check=True)
 
 
 if __name__ == '__main__':
-    main()
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--skip_tests')
```
joke1196, line 152: This could be simplified with `add_argument("--skip_tests", action="store_true")`. It will return a boolean so no need to check against a string. And the default will be `False`. Only if `--skip_test` is passed as an argument we will get `True`.

python-frontend/typeshed_serializer/runners/tox_runner.py
```diff
@@ -140,8 +141,17 @@ def main():
...
+    args = parser.parse_args()
+    skip_tests = False
+    if args.skip_tests == "true":
+        skip_tests = True
```
maksim-grebeniuk-sonarsource, line 156: why not just:
```python
skip_tests = arg.skip_tests == "true"
```

maksim-grebeniuk-sonarsource (approved):
LGTM, see one small suggestion

PR #1580, opened 2023-09-22
SONARPY-1366 Updating python projects used in ruling ITs

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource:
Closing for now, we probably want to reconsider how our ITs are set up entirely (1 big project vs multiple projects)

PR #1583, opened 2023-09-27
SONARPY-1455 Rule S6741 The 'pandas.DataFrame.to_numpy()' method should be preferred to the 'pandas.DataFrame.values' attribute.

jdodinh:


python-checks/src/main/java/org/sonar/python/checks/PandasDataFrameToNumpyCheck.java
```diff
@@ -0,0 +1,84 @@
...
+    if (expr.qualifier().is(Tree.Kind.NAME)) {
+      this.reachingDefinitionsAnalysis.valuesAtLocation((Name) expr.qualifier())
+        .stream()
+        .filter(exp -> exp.is(Tree.Kind.CALL_EXPR))
+        .map(CallExpression.class::cast)
+        .filter(ce -> ce.callee().is(Tree.Kind.NAME, Tree.Kind.QUALIFIED_EXPR))
+        .map(PandasDataFrameToNumpyCheck::getFullyQualifiedName)
+        .filter(Optional::isPresent)
+        .map(Optional::get)
+        .filter("pandas.DataFrame"::equals).findAny()
```
guillaume-dequenne-sonarsource, line 61: I have the feeling this is a bit too narrow, as there are many calls that may produce a `pandas.DataFrame` object besides the constructor (e.g `read_csv`) and that in the real world, most people won't actually use the constructor itself.

What we would really want to do is to rely on type inference to catch qualified expressions whose qualifier has type `pandas.DataFrame`, but because we don't have Pandas stubs yet, we won't understand which method/function is susceptible from returning this type. Maybe we can discuss considering adding such stubs as a separate ticket for this sprint, though.

jdodinh: Yes, this is true and I agree. I decided to stay in this narrow scope exactly because I wasn't sure about the return type of functions such as `read_csv`. I agree that we should improve this later, and it would be great to have a meeting to discuss this.

guillaume-dequenne-sonarsource: That sounds good. Could you create a ticket to expand this rule to common Pandas methods? We can keep the limited scope as part of this PR, but I think it would be nice to have another that expands on this during the sprint.

python-checks/src/main/java/org/sonar/python/checks/PandasDataFrameToNumpyCheck.java
```diff
@@ -0,0 +1,84 @@
...
+      return;
+    }
+
+    if (expr.qualifier().is(Tree.Kind.NAME)) {
+      this.reachingDefinitionsAnalysis.valuesAtLocation((Name) expr.qualifier())
+        .stream()
+        .filter(exp -> exp.is(Tree.Kind.CALL_EXPR))
+        .map(CallExpression.class::cast)
+        .filter(ce -> ce.callee().is(Tree.Kind.NAME, Tree.Kind.QUALIFIED_EXPR))
+        .map(PandasDataFrameToNumpyCheck::getFullyQualifiedName)
```
guillaume-dequenne-sonarsource, line 58: There might be something simpler to do here, as it seems what you really want to check is that the callee is an instance of `HasSymbol` then retrieve the fully qualified name out of the symbol itself.

Given the fact I feel we'll need to revisit this bit as soon as we want to support more than just `pd.DataFrame`, I'm not sure it needs fixing now though.

guillaume-dequenne-sonarsource (approved):
As mentioned in my other comment, I think the way we search for `pd.DataFrame` is too limited. However, the overall implementation looks good and I think it's ok to merge it as it is and fix this in a later PR, so that we keep things clean. I just left minor comments.

Can you please create the tickets to expand the scope of this rule, as well as one for implementing a quick fix?

python-checks/src/main/java/org/sonar/python/checks/PandasDataFrameToNumpyCheck.java
jdodinh: Yes, I just created the ticket, which can be found [here](https://sonarsource.atlassian.net/browse/SONARPY-1488).

python-checks/src/test/java/org/sonar/python/checks/PandasDataFrameToNumpyCheckTest.java
```diff
@@ -0,0 +1,34 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.python.checks;
+
+import org.junit.jupiter.api.Test;
+import org.sonar.python.checks.utils.PythonCheckVerifier;
+
+class PandasDataFrameToNumpyCheckTest {
+
+  PandasDataFrameToNumpyCheck check = new PandasDataFrameToNumpyCheck();
```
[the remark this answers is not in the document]
jdodinh: Addressed in d6cdd1684d7677b2ff47a5486118bba9915235be.

PR #1587, opened 2023-10-02
SONARPY-1488 Generate protobuf stubs for Pandas

guillaume-dequenne-sonarsource:
Ruling issue differences are on the numpy project, probably due to a mismatch between the project version and the stub versions. Given the special case, I'm tempted to accept them.


its/ruling/src/test/resources/expected/python-S2638.json
```diff
@@ -1,311 +1,325 @@
...
 ],
-'project:numpy-1.16.4/numpy/distutils/mingw32ccompiler.py': [
+"project:numpy-1.16.4/numpy/distutils/mingw32ccompiler.py": [
 176,
 177,
 178,
 179
 ],
-'project:numpy-1.16.4/numpy/ma/core.py': [
+"project:numpy-1.16.4/numpy/ma/core.py": [
```
guillaume-dequenne-sonarsource, line 107: I'm a bit uneasy about accepting these as issues.
I have the feeling this rule could have an exception for "private" components (given away by the `_` in their names), as it could be argued that there is no broken contract if the parent is not meant to be used as-is (it might be challenged though).

If you agree, I can create a ticket to make this exception. 

For the time being though, I'd accept these issues as they document an existing behavior and have the special situation of appearing while analyzing `numpy` using stubs for `numpy`, possibly with mismatching versions (which is another problem...).

python-checks/src/test/resources/checks/pandasDataFrameToNumpy.py
joke1196, line 1: This would work as well with a parameter right? Like 
```python
import pandas as pd
def dataframe_param(df : pd.DataFrame):
    df.values # Noncompliant
```

joke1196 (approved):
LGTM! Thank you for taking care of this.

python-checks/src/test/resources/checks/pandasDataFrameToNumpy.py
guillaume-dequenne-sonarsource: It actually doesn't, because the type annotation makes it a declared type rather than a runtime type (we're not sure a `pd.DataFrame` will actually be provided since the type hint is not enforced).

I'll document the FN though to make that explicit.

jdodinh (approved):
LGTM! Thank goodness that the code is so much simpler now.

PR #1588, opened 2023-10-02
SONARPY-1459 Rule S6735: When using pandas.merge or pandas.join, the parameters on, how and validate should be specified.

jdodinh:


python-checks/src/test/resources/checks/pandasAddMergeParameters.py
```diff
@@ -0,0 +1,66 @@
...
+    _ = merge(age_df, name_df, how="cross", validate="1:1")  # Noncompliant {{The 'on' parameter of the merge should be specified.}}
+
+
+def compliant_1(xx):
+    import pandas as pd
+
+    age_df = pd.DataFrame({"user_id": [1, 2, 4], "age": [42, 45, 35]})
+    name_df = pd.DataFrame({"user_id": [1, 2, 3, 4], "name": ["a", "b", "c", "d"]})
+
+    _ = pd.merge(name_df, on="user_id", how="right", validate="1:1")
```
joke1196, line 64: Here we could add a small test to check if a merge (which is not pandas.merge) does not raise the issue.

python-checks/src/test/resources/checks/pandasAddMergeParameters.py
```diff
@@ -0,0 +1,66 @@
...
+    _ = age_df.merge(name_df, how="right")  # FN Noncompliant {{The 'on' and 'validate' parameters of the merge should be specified.}}
+
+    _ = age_df.merge(name_df, validate="1:1")  # FN Noncompliant {{The 'haw' and 'on' parameters of the merge should be specified.}}
+
+    _ = age_df.merge(name_df, how="right", on="user_id")  # FN Noncompliant {{The 'validate' parameter of the merge should be specified.}}
+
+    _ = age_df.merge(name_df, on="user_id", validate="1:1")  # FN Noncompliant {{The 'how' parameter of the merge should be specified.}}
+
+    _ = age_df.merge(name_df, how="cross", validate="1:1")  # FN Noncompliant {{The 'validate' parameter of the merge should be specified.}}
+
```
joke1196, line 34: we could also add a small test with `pd.merge(age_df, name_df, "inner", "user_id", None, None, False, False, False, ('_x', '_y'), None, False, "1:1")` and check that we do not raise an error

joke1196 (changes requested):
Looks good! I think there could be a few missing tests. Mainly linked to keyword/positional arguments, this means a small change would be required when fetching arguments by keywords only. Great job! Thank you

python-checks/src/test/resources/checks/pandasAddMergeParameters.py
jdodinh: Thanks for the tip. This test case was added in 9517a9d96e8ea9a1c9f6508bc91276fb585ffeb7.

python-checks/src/test/resources/checks/pandasAddMergeParameters.py
jdodinh: I agree, this is a good test case. However I would like to hold off on the implementation until we have the definite information from the stubs, because the position of the keywords will differ depending on whether we call `pandas.merge` or `DataFrame.merge`.

joke1196 (approved):

PR #1589, opened 2023-10-03
Fix Pandas methods FQNs for S6735 and S6740

guillaume-dequenne-sonarsource:


joke1196 (approved):
LGTM! Thank you for the fix

PR #1590, opened 2023-10-03
SONARPY-1456 Rule S6734: inplace=True should not be used when modifying a Pandas DataFrame.

jdodinh:


python-checks/src/main/java/org/sonar/python/checks/PandasModifyInPlaceCheck.java
```diff
@@ -0,0 +1,69 @@
...
+    "pandas.core.frame.DataFrame.dropna",
+    "pandas.core.frame.DataFrame.drop_duplicates",
+    "pandas.core.frame.DataFrame.sort_values",
+    "pandas.core.frame.DataFrame.sort_index",
+    "pandas.core.frame.DataFrame.query",
+    "pandas.core.frame.DataFrame.transpose",
+    "pandas.core.frame.DataFrame.swapaxes",
+    "pandas.core.frame.DataFrame.reindex",
+    "pandas.core.frame.DataFrame.reindex_like",
+    "pandas.core.frame.DataFrame.truncate");
```
jdodinh, line 50: I would just like to mention that I am aware that this list may not be exhaustive. I took it from the RSPEC. Furthermore, some of these functions actually don't have the `inplace` parameter specified in their documentation anymore.

guillaume-dequenne-sonarsource: As far as I remember, some of these actually come from [PDEP-8](https://github.com/pandas-dev/pandas/pull/51466/files), though I see `eval` is missing and some of the methods from this list are actually not referenced in the PDEP. Which methods do you see that doesn't have the parameter anymore? I think it's fine to have them included if they had the parameter in a previous version of Pandas, as some people might still be using an older version anyway.

That said, I think it's fair to start with this list and re-evaluate if need be. A more radical approach could've been to flag every call to a method on a `pandas.DataFrame` object that uses `inplace=True` due to how misleading the parameter is.

guillaume-dequenne-sonarsource: As I re-read the PDEP and the method list, I actually think we probably should stick with "group 4" methods that can never operate in place despite the parameter. If the parameter eventually is deprecated, we can then expand to more methods.

The methods you highlight that don't have the `inplace` parameter, actually never had it and have a `copy` parameter instead. I'm not sure why they made it to the list, but I guess it would make sense to remove them to avoid confusion.

Could you update the list accordingly here and in RSPEC? That means the methods that should be flagged are:
`drop`
`dropna`
`drop_duplicates`
`sort_values`
`sort_index`
`eval`
`query`

Sorry for the confusion!

guillaume-dequenne-sonarsource (changes requested):
LGTM on implementation.
As you highlighted, the list of methods that should be flagged is a bit confusing. Could you update it to match the PDEP-8 group 4 list?

python-checks/src/main/java/org/sonar/python/checks/PandasModifyInPlaceCheck.java
jdodinh: I addressed this in d221ad3af4eb334c1cb4ed62de7f937f27cbf717.

jdodinh: I also made sure that the error was raised on the argument, rather than the entire qualified expression.

python-checks/src/test/resources/pandasModifyInPlace.py
```diff
@@ -0,0 +1,95 @@
+def non_compliant_1():
+    import pandas as pd
+
+    df = pd.DataFrame({'A': [3, 2, 1], 'B': ['x', 'y', 'z']})
+    df2 = pd.DataFrame({'A': [6, 5, 4], 'C': ['a', 'b', 'c']})
```
guillaume-dequenne-sonarsource, line 5: Why do we need `df2` here? It doesn't seem used. Same in the other methods.

python-checks/src/test/resources/pandasModifyInPlace.py
```diff
@@ -0,0 +1,95 @@
...
+    _ = df.sort_values(by=['A'], inplace=True)  # Noncompliant
+    #                            ^^^^^^^^^^^^
+
+    _ = df.sort_index(inplace=True)  # Noncompliant
+    #                 ^^^^^^^^^^^^
+
+    _ = df.eval("C = str(A) + B", inplace=True)  # Noncompliant
+    #                             ^^^^^^^^^^^^
+
+    _ = df.query('A > B', inplace=True)  # Noncompliant
```
guillaume-dequenne-sonarsource, line 27: I think it would be nice to also have a test that shows this work when the data frame is created through something else than the constructor, such as:
```
my_df = pd.read_csv("my_csv.csv")
    _ = my_df.query('A > B', inplace=True)  # Noncompliant
```

guillaume-dequenne-sonarsource (approved):
Looks good!
Left minor suggestions to improve the tests.

python-checks/src/test/resources/pandasModifyInPlace.py
jdodinh: Indeed it was a bit unnecessary. I improved the test cases in a5d9fcc0fc25598cadd33cfd97e2b8d4e3a972f8.

PR #1591, opened 2023-10-04
SONARPY-1458 Rule S6742: pandas.pipe method should be preferred over long chains of instructions

joke1196:


python-checks/src/main/java/org/sonar/python/checks/PandasChainInstructionCheck.java
```diff
@@ -0,0 +1,136 @@
...
+      .noneMatch((DATAFRAME_FQN + ".pipe")::equals);
+
+    boolean isADataFrame = "DataFrame".equals(InferredTypes.typeName(firstQualifiedExpression.qualifier().type()));
+
+    return (isADataFrameMethodCall || isAFunctionReturningADataFrame || isADataFrame) && doesNotContainACallToPipe;
+  }
+
+  private static Optional<Boolean> isReturnTypeADataFrame(Symbol symbol) {
+    return Optional.of(symbol)
+      .filter(s -> s.is(Symbol.Kind.AMBIGUOUS))
```
guillaume-dequenne-sonarsource, line 127: What if the symbol is a `FUNCTION`? I understand most Pandas methods are ambiguous so in practice, this should work, but would we then have FNs if a method only has one definition?

python-checks/src/main/java/org/sonar/python/checks/PandasChainInstructionCheck.java
```diff
@@ -0,0 +1,136 @@
...
+    if (qualifier.object().is(Tree.Kind.CALL_EXPR)) {
+      return Optional.of((CallExpression) qualifier.object());
+    } else if (qualifier.object().is(Tree.Kind.SUBSCRIPTION)) {
+      return ignoreSubscriptionAndGetCallExpr((SubscriptionExpression) qualifier.object());
+    }
+    return Optional.empty();
+  }
+
+  private static boolean isValidPandasCall(List<QualifiedExpression> chain) {
+    QualifiedExpression firstQualifiedExpression = chain.get(chain.size() - 1);
```
guillaume-dequenne-sonarsource, line 102: The assumption is that if the first call returns a data frame, every subsequent call will do that as well? Do we risk FPs with that assumption (Not challenging the assumption, just curious about potential edge case risk).

guillaume-dequenne-sonarsource (approved):
Looks good! Left a few comments/questions. Nothing major, though.

python-checks/src/main/java/org/sonar/python/checks/PandasChainInstructionCheck.java
```diff
@@ -0,0 +1,136 @@
...
+
+  private void checkChainedInstructions(SubscriptionContext ctx) {
+    QualifiedExpression expr = (QualifiedExpression) ctx.syntaxNode();
+    List<QualifiedExpression> chain = visitParent(expr, new ArrayList<>());
+    if (chain.size() >= MAX_CHAIN_LENGTH && isValidPandasCall(chain)) {
+      ctx.addIssue(chain.iterator().next(), MESSAGE);
+    }
+  }
+
+  private List<QualifiedExpression> visitParent(QualifiedExpression expr, List<QualifiedExpression> chain) {
```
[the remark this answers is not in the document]
joke1196: I see what you mean! It makes sense, I'll make the change.

python-checks/src/main/java/org/sonar/python/checks/PandasChainInstructionCheck.java
joke1196: Yes, I think I could add the happy path here :).

python-checks/src/main/java/org/sonar/python/checks/PandasChainInstructionCheck.java
joke1196: Yes maybe you are right. It would be particular cases. For example:

```python
 pd.read_csv("some_csv.csv").axes[1].join(pd.Index([4, 5, 6])).repeat([1,2]).drop_duplicates().insert(1, 42)
```

Here we are working on an `Index` after the `axes[1]` instruction. We would raise an issue here. The problem is that there is no `pipe` method on `Index`, also changing the read_csv to pipe would be less clear. One way to circumvent this would be to check that at least the first two instructions return a `DataFrame`. It would make more sense than to refactor with a pipe method. That said I am not convinced this is a case we would encounter often. I am not 100% sure we would gain much from implementing this use case.  WDYT?

python-checks/src/main/java/org/sonar/python/checks/PandasChainInstructionCheck.java
```diff
@@ -0,0 +1,136 @@
...
+      .orElse(chain);
+  }
+
+  private static Optional<CallExpression> ignoreSubscriptionAndGetCallExpr(SubscriptionExpression qualifier) {
+    if (qualifier.object().is(Tree.Kind.CALL_EXPR)) {
+      return Optional.of((CallExpression) qualifier.object());
+    } else if (qualifier.object().is(Tree.Kind.SUBSCRIPTION)) {
+      return ignoreSubscriptionAndGetCallExpr((SubscriptionExpression) qualifier.object());
+    }
+    return Optional.empty();
```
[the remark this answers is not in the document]
joke1196: I think I have overlooked the case where the qualifier is a name.

```python
df2.set_index("name").T.filter(like='joe', axis=0)[1].mean().head()
```
I think I should raise an issue on such a case.

python-checks/src/main/java/org/sonar/python/checks/PandasChainInstructionCheck.java
guillaume-dequenne-sonarsource: I think we can accept this risk of FP. Maybe we can have such example documented in the tests, though?
So that we could revisit this later if this proves to be problematic.

joke1196: Okay so in the end we would have more FNs than FPs. As I did not implement the support for subscription with name we will not raise issue with such cases. But also we will not raise issues with the following case: 
```python
    pd.read_csv("some_csv.csv").filter(like='joe', axis=0).add(10).add(10).groupby("team")["salary"]["test"].axes[1].unique() # FN
```

python-checks/src/test/resources/checks/pandasChainInstructionCheck.py
```diff
@@ -0,0 +1,41 @@
...
+#   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    pd.read_csv("some_csv.csv").filter(like='joe', axis=0).groupby("team")["salary"]["test"].mean().head() # Noncompliant
+#   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+
+#   Here we do not raise an issue only because we do not support subscription with Name
+#   If support is added for such case we would encounter FPs when the subscription with Name is at the beginning of the chain 
+    pd.read_csv("some_csv.csv").filter(like='joe', axis=0).groupby("team")["salary"]["test"].axes[1].unique() # FN
+
+#   Here we should not raise an issue as the chain is done mainly on an Index object which does not have a pipe methof
```
guillaume-dequenne-sonarsource, line 23: Small typo here: `methof` -> `method`

guillaume-dequenne-sonarsource (approved):
LGTM.

jdodinh (commented):
LGTM! Just a minor repetition at some point.

python-checks/src/main/java/org/sonar/python/checks/PandasChainInstructionCheck.java
```diff
@@ -0,0 +1,142 @@
...
+      .map(AmbiguousSymbol.class::cast)
+      .map(s -> s.alternatives().stream()
+        .filter(a -> a.is(Symbol.Kind.FUNCTION))
+        .map(FunctionSymbol.class::cast)
+        .map(FunctionSymbol::annotatedReturnTypeName)
+        .anyMatch(DATAFRAME_FQN::equals))
+      .or(() -> Optional.of(symbol)
+        .filter(s -> s.is(Symbol.Kind.FUNCTION))
+        .map(FunctionSymbol.class::cast)
+        .map(FunctionSymbol.class::cast)
```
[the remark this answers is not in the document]
joke1196: Oh, it is not at all. Really good catch! Thank you

PR #1593, opened 2023-10-04
SONARPY-1495 Expand the scope of rule S6735 to calls to merge and join on the DataFrame object.

jdodinh:


python-checks/src/main/java/org/sonar/python/checks/PandasAddMergeParametersCheck.java
```diff
@@ -50,24 +88,39 @@ private static void verifyMergeCallParameters(SubscriptionContext ctx) {
...
+      parameters.add(Keywords.ON.getKeyword());
     }
-    if (TreeUtils.argumentByKeyword("validate", callExpression.arguments()) == null) {
-      parameters.add("validate");
+    if (argumentIsMissing(fullyQualifiedName, Keywords.VALIDATE, callExpression.arguments())) {
+      parameters.add(Keywords.VALIDATE.getKeyword());
     }
     if (!parameters.isEmpty()) {
-      ctx.addIssue(callExpression, String.format(messages.get(parameters.size() - 1), parameters.toArray()));
+      parameters.add(fullyQualifiedName.substring(fullyQualifiedName.lastIndexOf('.') + 1));
```
joke1196, line 107: I  think we could make it a bit more explicit on what is happening here. We could refactor the String.format() into its own function where we could pass the arguments and name them properly for example:
```java
private static String generateMessage(String message, List<String> missingKeywords, String functionName){
    ...
}
```
This would help in understanding a bit the code. It is always a bit difficult to see the intent when using index on lists. This is why I would prefer having the arguments named so that if at some point the order of the MESSAGES came to change it would be easier to debug, knowing what `parameters.size() - 2` should be returning.

python-checks/src/main/java/org/sonar/python/checks/PandasAddMergeParametersCheck.java
```diff
@@ -50,24 +88,39 @@ private static void verifyMergeCallParameters(SubscriptionContext ctx) {
...
+      ctx.addIssue(callExpression, String.format(MESSAGES.get(parameters.size() - 2), parameters.toArray()));
```
guillaume-dequenne-sonarsource, line 108: As an additional nitpick, I'm not a fan of the arithmetics to get the list index in `MESSAGES.get(parameters.size() - 2)`.
I'd probably prefer an explicit computation of the number of missing arguments and having the message retrieved from something like a map. I admit it sounds a bit overkill for such a simple case, but I think it would lead to more maintainable code.

guillaume-dequenne-sonarsource (changes requested):
I agree with David that some things could be a bit more explicit. Also, while unrelated to this PR, the issue message should be updated to be imperative.
Looks good otherwise!

python-checks/src/main/java/org/sonar/python/checks/PandasAddMergeParametersCheck.java
```diff
@@ -50,24 +88,39 @@ private static void verifyMergeCallParameters(SubscriptionContext ctx) {
...
     }
+  }
+
+  private static boolean argumentIsMissing(String fullyQualfiedName, Keywords keyword, List<Argument> arguments) {
```
joke1196, line 112: Nitpick: no need to change anything. Usually, when naming a method returning a boolean I prefer starting with the verb so `is`, `has` etc... like `isArgumentMissing`.

python-checks/src/main/java/org/sonar/python/checks/PandasAddMergeParametersCheck.java
```diff
@@ -50,24 +88,39 @@ private static void verifyMergeCallParameters(SubscriptionContext ctx) {
...
+    return Optional.of(keyword)
+      .map(kw -> TreeUtils.nthArgumentOrKeyword(getPosition(fullyQualfiedName, kw), kw.getKeyword(), arguments))
+      .isEmpty();
+  }
 
+  private static int getPosition(String fullyQualifiedName, Keywords parameter) {
```
joke1196, line 118: Nitpick: Maybe we could make the intent of this method clearer with a name like getArgumentPosition.

python-checks/src/main/java/org/sonar/python/checks/PandasAddMergeParametersCheck.java
```diff
@@ -27,19 +27,57 @@
...
+
+    public int getPandasMergePosition() {
+      return pandasMergePosition;
+    }
+
+    final String keyword;
+    final int joinPosition;
+    final int dataFrameMergePosition;
+    final int pandasMergePosition;
+
```
joke1196, line 63: _Nitpick:_ Related to my comments about making things clearer with naming, here I think the fields and their getters could benefit from names like `positionInJoin` as they represent the position of the argument in the join method and not the position of join itself. I personally think it is better to have a longer name that describes the intent clearly rather than shorter names

joke1196 (changes requested):
Look good! Just a few comments on naming mainly and making the intention clearer. Otherwise good job. The tests are quite complete as well, nice work!

python-checks/src/main/java/org/sonar/python/checks/PandasAddMergeParametersCheck.java
jdodinh: I changed the way that it was implemented in 57fbeb2d6285498a551e7fbf581e7c47ab35da99.

python-checks/src/main/java/org/sonar/python/checks/PandasAddMergeParametersCheck.java
```diff
@@ -27,19 +27,57 @@
...
+      this.pandasMergePosition = pandasMergePosition;
+    }
+  }
+
+  private static final Set<String> METHODS = Set.of(
+    "pandas.core.frame.DataFrame.merge",
+    "pandas.core.reshape.merge.merge",
+    "pandas.core.frame.DataFrame.join");
+
+  private static final List<String> MESSAGES = List.of(
```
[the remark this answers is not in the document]
jdodinh: Changes made in 57fbeb2d6285498a551e7fbf581e7c47ab35da99.

python-checks/src/main/java/org/sonar/python/checks/PandasAddMergeParametersCheck.java
jdodinh: I made the changes in 57fbeb2d6285498a551e7fbf581e7c47ab35da99.

python-checks/src/main/java/org/sonar/python/checks/PandasAddMergeParametersCheck.java
jdodinh: Fixed in b62ad8a311e6cf817d950df11b14d0b282135a79.

python-checks/src/main/java/org/sonar/python/checks/PandasAddMergeParametersCheck.java
jdodinh: Changed in 65ecde482e55039b470e0fd514887d0a5d32f9a9.

python-checks/src/main/java/org/sonar/python/checks/PandasAddMergeParametersCheck.java
jdodinh: I agree that this would be better. I changed it in 2669345097d5e46d6ace627f7900ffe63f46d29f.

jdodinh (commented):
Thank you very much for the useful comments! I made the corresponding changes.

joke1196 (approved):
LGTM!

guillaume-dequenne-sonarsource (approved):
LGTM.

PR #1594, opened 2023-10-05
SONARPY-1490: Add a quickfix for rule S6741.

jdodinh:


python-checks/src/main/java/org/sonar/python/checks/PandasDataFrameToNumpyCheck.java
```diff
@@ -23,14 +23,19 @@
...
 import org.sonar.plugins.python.api.tree.QualifiedExpression;
 import org.sonar.plugins.python.api.tree.Tree;
+import org.sonar.python.quickfix.TextEditUtils;
 
 @Rule(key = "S6741")
 public class PandasDataFrameToNumpyCheck extends PythonSubscriptionCheck {
 
+  private static final String DATAFRAME_VALUES_FQN = "pandas.core.frame.DataFrame.values";
   private static final String MESSAGE = "Do not use \"DataFrame.values\".";
+  private static final String QUICK_FIX_MESSAGE = "Replace this with a call to DataFrame.to_numpy()";
```
joke1196, line 37: For the message I think we can go with `Replace .values  with DataFrame.to_numpy()`

joke1196 (changes requested):
LGTM! I have just added a small suggestion on the message for the quick fix to make it more direct. Also, no need to change but I think the last test is testing if the issue is raised not really the Quick fix itself. And you have already done this test, that's why I don't think it is needed, but it does not hurt to keep it.

python-checks/src/main/java/org/sonar/python/checks/PandasDataFrameToNumpyCheck.java
guillaume-dequenne-sonarsource: I'd suggest just `Replace with DataFrame.to_numpy()`
I don't remember where exactly are the quick fixes messages guidelines, but I think they mention that we should generally not compare the states before/after and just focus on the state after the fix.

joke1196: Ok good to know!

python-checks/src/test/java/org/sonar/python/checks/PandasDataFrameToNumpyCheckTest.java
```diff
@@ -20,13 +20,81 @@
...
+      "    import pandas as pd\n" +
+      "\n" +
+      "    my_df = pd.read_csv(\"some_csv.csv\")\n" +
+      "    my_df.to_numpy()";
+    performVerification(non_compliant, compliant);
+  }
+
+  private void performVerification(String non_compliant, String compliant) {
+    PythonQuickFixVerifier.verify(check, non_compliant, compliant);
+    PythonQuickFixVerifier.verifyQuickFixMessages(check, non_compliant, QUICK_FIX_MESSAGE);
```
guillaume-dequenne-sonarsource, line 98: This is a nitpick but I generally prefer to have the quick fix message tested explicitly (i.e having it written as a string literal rather than using the field from the check). This makes things more intentional and protects us against typos.

I know we don't consistently do it and I don't think it matters too much so I'm not asking for a change here, just sharing an opinion on how to approach these tests.

python-checks/src/test/java/org/sonar/python/checks/PandasDataFrameToNumpyCheckTest.java
```diff
@@ -20,13 +20,81 @@
...
+      "\n" +
+      "    _ = pd.DataFrame({\n" +
+      "        'X': ['A', 'B', 'A', 'C'],\n" +
+      "        'Y': [10, 7, 12, 5]\n" +
+      "    }).to_numpy()";
+    performVerification(non_compliant, compliant);
   }
 
+  @Test
+  void quick_fix_test_3() {
```
guillaume-dequenne-sonarsource, line 81: I share David's opinion that here, we're not really testing the quick fix but rather the fact that we raise an issue on the DataFrame obtained through `read_csv`. I don't think it's a problem to keep the test either, though.

If you want additional quick fix tests, generally you can keep simple issues but play with things like context / indentation. For instance, what happens to the quick fix if the `values` attribute access is followed by another call?

For instance, would `df.values.astype(str)` be quick fixed into `df.values.to_numpy().astype(str)`?

joke1196 (approved):
LGTM! Just left a comment on an unused variable.

python-checks/src/test/java/org/sonar/python/checks/PandasDataFrameToNumpyCheckTest.java
```diff
@@ -20,13 +20,81 @@
...
 
 import org.junit.jupiter.api.Test;
+import org.sonar.python.checks.quickfix.PythonQuickFixVerifier;
 import org.sonar.python.checks.utils.PythonCheckVerifier;
 
 class PandasDataFrameToNumpyCheckTest {
 
+  PandasDataFrameToNumpyCheck check = new PandasDataFrameToNumpyCheck();
+
+  private static final String QUICK_FIX_MESSAGE = "Replace this with a call to DataFrame.to_numpy()";
```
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: Both this and the issue message could be inlined indeed (it's still used once), but I guess given we have two messages, having named variables for them help understand the intention behind each of them. No strong opinion here, but I'm tempted to merge as-is.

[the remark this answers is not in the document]
joke1196: Sorry I don't see where this is used.  I thought this was the older Quick fix message. I guess it is already inlined on line 98 this is why I think we don't need it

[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: Oh, my bad, I was looking at the wrong file and went too quickly...
I thought the tests were using the variable defined in the main code as well, which is what happened in the past (and doesn't feel too intentional).

PR #1603, opened 2023-10-11
SONARPY-1484 Support type parameter syntax for classes

maksim-grebeniuk-sonarsource:


python-frontend/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
```diff
@@ -747,6 +749,54 @@ void nonlocalStatement() {
...
+      "   TypeVarWithConstraints: (str, bytes),\n" +
+      "   *SimpleTypeVarTuple,\n" +
+      "   **SimpleParamSpec\n" +
+      "](\n" +
+      "   a: SimpleTypeVar,\n" +
+      "   b: TypeVarWithBound,\n" +
+      "   c: Callable[SimpleParamSpec, TypeVarWithConstraints],\n" +
+      "   *d: SimpleTypeVarTuple,\n" +
+      "): pass");
+    var functionDef = treeMaker.funcDefStatement(astNode);
```
guillaume-dequenne-sonarsource, line 767: I suggest to refactor this way:

```suggestion
    var functionDef = parse("def overly_generic[\n" +
      "   SimpleTypeVar,\n" +
      "   TypeVarWithBound: int,\n" +
      "   TypeVarWithConstraints: (str, bytes),\n" +
      "   *SimpleTypeVarTuple,\n" +
      "   **SimpleParamSpec\n" +
      "](\n" +
      "   a: SimpleTypeVar,\n" +
      "   b: TypeVarWithBound,\n" +
      "   c: Callable[SimpleParamSpec, TypeVarWithConstraints],\n" +
      "   *d: SimpleTypeVarTuple,\n" +
      "): pass", treeMaker::funcDefStatement);
```

When calling `RuleTest#parse`, a few additional things are done to ensure every child is correctly visited and that they are in the correct order. I think it makes sense to have here because currently, removing the sorting of the children of `TypeParams` doesn't crash anything (but it does when using this test method).

python-frontend/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
```diff
@@ -976,6 +1026,49 @@ void classdef_statement() {
...
+  void classdef_statement_type_params() {
+    setRootRule(PythonGrammar.CLASSDEF);
+    var astNode = p.parse("class generic_class[\n" +
+      "   SimpleTypeVar,\n" +
+      "   TypeVarWithBound: int,\n" +
+      "   TypeVarWithConstraints: (str, bytes),\n" +
+      "   *SimpleTypeVarTuple,\n" +
+      "   **SimpleParamSpec\n" +
+      "]: pass");
+    var classDef = treeMaker.classDefStatement(astNode);
```
guillaume-dequenne-sonarsource, line 1039: Same here, we can use:
```suggestion
var classDef = parse("class generic_class[\n" +
      "   SimpleTypeVar,\n" +
      "   TypeVarWithBound: int,\n" +
      "   TypeVarWithConstraints: (str, bytes),\n" +
      "   *SimpleTypeVarTuple,\n" +
      "   **SimpleParamSpec\n" +
      "]: pass", treeMaker::classDefStatement);
```
When doing so, we can see that the `TypeParams` are not registered as children of `ClassDef`.

python-frontend/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
```diff
@@ -747,6 +749,54 @@ void nonlocalStatement() {
...
+    var simpleTypeVarTuple = typeParamsList.get(3);
+    assertThat(simpleTypeVarTuple.name().name()).isEqualTo("SimpleTypeVarTuple");
+    assertThat(simpleTypeVarTuple.starToken()).isNotNull();
+    assertThat(simpleTypeVarTuple.typeAnnotation()).isNull();
+
+    var visitor = Mockito.spy(new BaseTreeVisitor());
+
+    functionDef.accept(visitor);
+    Mockito.verify(visitor, Mockito.times(1)).visitTypeParams(Mockito.any());
+    Mockito.verify(visitor, Mockito.times(5)).visitTypeParam(Mockito.any());
```
guillaume-dequenne-sonarsource, line 797: Once using the `RuleTest#parse` method, maybe this is not needed anymore (same for the `ClassDef` version) and we can remove the `Mockito` import.

python-frontend/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
```diff
@@ -747,6 +749,54 @@ void nonlocalStatement() {
...
+
+    var simpleTypeVar = typeParamsList.get(0);
+    assertThat(simpleTypeVar.name().name()).isEqualTo("SimpleTypeVar");
+    assertThat(simpleTypeVar.starToken()).isNull();
+    assertThat(simpleTypeVar.typeAnnotation()).isNull();
+
+    var typeWithBound = typeParamsList.get(1);
+    assertThat(typeWithBound.name().name()).isEqualTo("TypeVarWithBound");
+    assertThat(typeWithBound.starToken()).isNull();
+    assertThat(typeWithBound.typeAnnotation()).isNotNull();
```
guillaume-dequenne-sonarsource, line 786: _Nitpick:_ Maybe we can have a small assertion to ensure the type annotation is correctly created.

guillaume-dequenne-sonarsource (changes requested):
Left some suggestions to improve the tests, and the children for `ClassDefImpl` need to be updated. Aside from that, this looks good!
One suggestion: it would be nice to merge this PR with 1 commit for each ticket, so if you can organize commits accordingly that would be great. This is not critical, though.

python-frontend/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
maksim-grebeniuk-sonarsource: Without it, it shows a lack of test coverage for these methods.

python-frontend/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
maksim-grebeniuk-sonarsource: Good catch, thanks

python-frontend/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
guillaume-dequenne-sonarsource: When using `p.parse` only instead of `RuleTest#parse`, I can see why. But given that `RuleTest#parse` performs a visit of the tree, I'm confused why coverage would still be missing. Are you sure it's the case even with the updated method?

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1605, opened 2023-10-16
SONARPY-1515 Support function and class type parameters for Symbol table

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource (approved):
LGTM, clean and straightforward.
As far as I can see, I don't think we should expect false positives coming from this.
That said, we might want to add some tests in some specific rules to ensure that. We can discuss it and do it in a separate ticket, though.

PR #1607, opened 2023-10-16
JDD/SONARPY-1510 Rule S6792: Generic classes should be defined using the type parameter syntax.

jdodinh:


python-checks/src/main/java/org/sonar/python/checks/GenericClassTypeParameterCheck.java
```diff
@@ -0,0 +1,101 @@
...
+public class GenericClassTypeParameterCheck extends PythonSubscriptionCheck {
+
+  private static final String MESSAGE = "Use the type parameter syntax to declare this generic class.";
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.CLASSDEF, GenericClassTypeParameterCheck::checkGenericInheritance);
+  }
+
+  private static void checkGenericInheritance(SubscriptionContext ctx) {
```
guillaume-dequenne-sonarsource, line 50: I realize there is one thing we forgot to check: make sure that we are using Python 3.12 to avoid FPs on older projects.

python-checks/src/main/java/org/sonar/python/checks/GenericClassTypeParameterCheck.java
```diff
@@ -0,0 +1,113 @@
...
+    if (!supportsTypeParameterSyntax(ctx)) {
+      return;
+    }
+    ClassDef classDef = (ClassDef) ctx.syntaxNode();
+
+    ArgList args = classDef.args();
+    if (args == null) {
+      return;
```
maksim-grebeniuk-sonarsource, line 59: Minor: Since there is long chain below instead of if-return it could be done as single optional-stream chain:
```java
Optional.of(classDef)
      .map(ClassDef::args)
      .map(ArgList::arguments)
      .stream()
      .flatMap(Collection::stream)
      .map(GenericClassTypeParameterCheck::checkGenericValue)
      .filter(lst -> !lst.isEmpty())
      .findFirst()
      .ifPresent(locations -> {
        PreciseIssue issue = ctx.addIssue(classDef.name(), MESSAGE);
        locations.forEach(issue::secondary);
      });
```

maksim-grebeniuk-sonarsource (approved):
LGTM, see one minor comment, feel free to ignore

python-checks/src/main/java/org/sonar/python/checks/GenericClassTypeParameterCheck.java
jdodinh: Thanks for the tip. I made the change in bb370abfb9270a60dbb5204c63455f6a4d242072.

PR #1609, opened 2023-10-17
SONARPY-1521 S4143: Fix FP when there is different collections value assignment for same key

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/OverwrittenCollectionEntryCheck.java
```diff
@@ -121,6 +123,25 @@ private static CollectionWrite collectionWrite(AssignmentStatement assignment, E
     return null;
   }
 
+  @CheckForNull
+  private static String key(SubscriptionExpression subscriptionExpression) {
```
guillaume-dequenne-sonarsource, line 127: I'm confused that we have to recompute the key part of the subscription here.

As I understand the rule, in `reportOverwrites` we should check for a given key if the writes are made on the same collection. Isn't there where the bug is? It seems that in this method if the 2 collections (not key) are different, we should simply not raise the issue.

guillaume-dequenne-sonarsource (commented):
I see the fix works but I'm confused why it had to be done that way, it seems we're not really fixing the root cause, given how the rule is supposed to work.

Maybe the rule logic itself is flawed and it's not possible to fix while respecting the rule original idea, but in this case I think it'd deserve some comments at least.

guillaume-dequenne-sonarsource (changes requested):
As we discussed, it would make sense to investigate whether we can fix this bug by fixing the way `reportOverwrites` distinguishes between different collections rather than by changing the way keys are computed.

joke1196 (approved):
LGTM! Was there a ticket created for the investigation of the problematic `reportOverwrites`?

PR #1610, opened 2023-10-17
SONARPY-1486 Support generic type alias declaration syntax

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource (changes requested):
Looks good overall!
Left a few small comments, mostly around testing.

python-frontend/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
```diff
@@ -2809,6 +2809,26 @@ void except_group_multiple() {
...
+    var astNode = p.parse("type A[B] = str");
+    var statementWithSeparator = new StatementWithSeparator(astNode, null);
+    var typeAlias = treeMaker.typeAliasStatement(statementWithSeparator);
+
+    assertThat(typeAlias).isNotNull();
+    assertThat(typeAlias.typeKeyword()).isNotNull();
+    assertThat(typeAlias.typeKeyword().value()).isEqualTo("type");
+    assertThat(typeAlias.name()).isNotNull();
+    assertThat(typeAlias.name().name()).isEqualTo("A");
+    assertThat(typeAlias.typeParams()).isNotNull();
```
jdodinh, line 2825: Would it be worth checking that the type parameters are indeed what they were assigned? Or is this test case handled somewhere else?

python-frontend/src/test/java/org/sonar/python/tree/PythonTreeMakerTest.java
```diff
@@ -2812,7 +2812,11 @@ void except_group_multiple() {
   @Test
   void typeAliasStatement() {
     setRootRule(PythonGrammar.TYPE_ALIAS_STMT);
-    var typeAlias = parse("type A[B] = str", treeMaker::typeAliasStatement);
+
+    var astNode = p.parse("type A[B] = str");
```
guillaume-dequenne-sonarsource, line 2816: Because we can't use the `RuleTest#parse` method anymore I guess we lose the visit and the test that makes sure all children are computed in order.
I realize for simple statements we have the `verify_expected_statement` that will do that by putting all simple statements in a file and check the file. Maybe we can add a type statement there?

guillaume-dequenne-sonarsource (approved):
Looks good! Minor suggestion about additional test that we could have.

PR #1612, opened 2023-10-19
SONARPY-1524 S3827: Fix FP when value is referenced in type alias

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource (approved):
LGTM!

PR #1613, opened 2023-10-19
SONARPY-1512 Rule S6796: Generic functions should be defined using the type parameter syntax.

jdodinh:


python-checks/src/main/java/org/sonar/python/checks/GenericFunctionTypeParameterCheck.java
```diff
@@ -0,0 +1,69 @@
...
+
+@Rule(key = "S6796")
+public class GenericFunctionTypeParameterCheck extends PythonSubscriptionCheck {
+
+  private static String MESSAGE = "Use a generic type parameter for this function instead of a TypeVar.";
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.PARAMETER_TYPE_ANNOTATION, GenericFunctionTypeParameterCheck::checkUseOfTypeVarGeneric);
+    context.registerSyntaxNodeConsumer(Tree.Kind.RETURN_TYPE_ANNOTATION, GenericFunctionTypeParameterCheck::checkUseOfTypeVarGeneric);
```
guillaume-dequenne-sonarsource, line 42: As I re-read the specification, I admit this wasn't super clear, but I feel it would be better to have a similar approach to the one for classes. 

That is, we should subscribe to function definitions and collect their parameter and return type annotations, then raise a single issue on the function if some of them are `TypeVar`.  That would avoid raising multiple issues on the same function (it's really one issue: using legacy syntax) (and would make a quick fix possible if we ever decide to pursue that).

python-checks/src/main/java/org/sonar/python/checks/GenericFunctionTypeParameterCheck.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  private static void checkUseOfTypeVarGeneric(SubscriptionContext ctx) {
```
guillaume-dequenne-sonarsource, line 45: Similar to the classes rule, I think we also need secondary locations to where the type vars are assigned.
That will help fixing the issue by providing a quick view of how the type var is defined.
For the secondary location message, we can have `TypeVar is assigned here`

python-checks/src/test/resources/checks/genericFunctionTypeParameter.py
```diff
@@ -0,0 +1,23 @@
+def non_compliant(xx):
```
guillaume-dequenne-sonarsource, line 1: I feel we could have some more tests. For example, having `TypeVar` imported from somewhere else than `typing` or imported as a different name.
Maybe stuff where the `TypeVar` has multiple assignments.
Basically the usual stuff, but I feel it would document how the rule is implemented and should behave.

guillaume-dequenne-sonarsource (commented):
Implementation looks good! I feel we could improve on how the rule actually reports issues, and we can have more tests.

That being said, I wouldn't be against merging this version of the implementation and revisiting it in a different PR, as I believe it'd still provide value as it is now, and we still have other tickets to tackle.
Let's discuss this.

python-checks/src/main/java/org/sonar/python/checks/GenericFunctionTypeParameterCheck.java
```diff
@@ -0,0 +1,69 @@
...
+import org.sonar.plugins.python.api.tree.CallExpression;
+import org.sonar.plugins.python.api.tree.Name;
+import org.sonar.plugins.python.api.tree.Tree;
+import org.sonar.plugins.python.api.tree.TypeAnnotation;
+import org.sonar.python.tree.TreeUtils;
+
+@Rule(key = "S6796")
+public class GenericFunctionTypeParameterCheck extends PythonSubscriptionCheck {
+
+  private static String MESSAGE = "Use a generic type parameter for this function instead of a TypeVar.";
```
guillaume-dequenne-sonarsource, line 37: This can also be `final`.

python-checks/src/main/java/org/sonar/python/checks/GenericFunctionTypeParameterCheck.java
jdodinh: Yes, this was my initial implementation actually. It is a little bit more lengthy, but I think the concept of raising one issue is valid. I will make the changes.

python-checks/src/main/java/org/sonar/python/checks/GenericFunctionTypeParameterCheck.java
jdodinh: Addressed in 10dce7e8e2c4444e7cf8493ede52fc9d98f8e050.

python-checks/src/main/java/org/sonar/python/checks/GenericFunctionTypeParameterCheck.java
jdodinh: I made the implementation change in 10dce7e8e2c4444e7cf8493ede52fc9d98f8e050.

python-checks/src/test/resources/checks/genericFunctionTypeParameter.py
```diff
@@ -0,0 +1,51 @@
...
+    _T = TV("_T", bound=str)
+    #    ^^^^^^^^^^^^^^^^^^^>  {{"TypeVar" is assigned here.}}
+    def func(a: _T, b: int) -> str:  # Noncompliant {{Use a generic type parameter for this function instead of a "TypeVar".}}
+    #   ^^^^    ^^<                                 {{Use of "TypeVar" here.}}
+        ...
+
+def FN_non_compliant():
+    _T = TypeVar("_T", bound=str)
+
+    def func(a: list[_T]) -> set[_T]:  # FN
```
guillaume-dequenne-sonarsource, line 43: Please don't forget to reference a ticket here (you can reuse the same example in the ticket) so that the FN is fully documented and we can eventually do something about it.

python-checks/src/main/java/org/sonar/python/checks/GenericFunctionTypeParameterCheck.java
```diff
@@ -0,0 +1,108 @@
...
+      .ifPresent(secondaryLocations::add);
+
+    if (!secondaryLocations.isEmpty()) {
+      PreciseIssue issue = ctx.addIssue(functionDef.name(), MESSAGE);
+      secondaryLocations.forEach(loc -> issue.secondary(loc, SECONDARY_MESSAGE_USE));
+      getAssignmentLocations(secondaryLocations).forEach(loc -> issue.secondary(loc, SECONDARY_MESSAGE_ASSIGNMENT));
+    }
+  }
+
+  private static Set<Tree> getAssignmentLocations(Set<Tree> secondaryLocations) {
```
guillaume-dequenne-sonarsource, line 81: It is a bit unfortunate to recompute the assigned value location when it's already done in `isGenericTypeAnnotation`. 

I have the feeling that method could probably return a list of `IssueLocation` that would be ready to add on the issue (it would then have to distinguish explicitly between the parameter/assigned value and to ensure no duplicate gets through - and probably would be easier to code imperatively than with the stream).

That said, I think it's okay to keep as it is now, since it won't change much in the grand scheme of things.

guillaume-dequenne-sonarsource (approved):
Looks good!
Left a small comment, not asking to make the change, just a thought to consider for other rules/tickets.

python-checks/src/test/resources/checks/genericFunctionTypeParameter.py
jdodinh: Done in f6edd0239d1600ea93f64f6d104abbc4c17bdb59.

python-checks/src/main/java/org/sonar/python/checks/GenericFunctionTypeParameterCheck.java
jdodinh: Thanks for the tip.

PR #1615, opened 2023-10-20
SONARPY-1523 Configure an extended ruling project as a Python 3.12 project

guillaume-dequenne-sonarsource:
The initial goal of this ticket was to add an actual Python 3.12 project, or a project for which our new rules could raise issues (such as `mypy`).

However, this is not practical for various reasons:
* Extended ruling currently only raises issues on hardcoded bug rules, which is something that should be revisited (SONARPY-984) (our new rules are code smells)
* Reusing `mypy` would require changing the actual ruling as well, for which there is still an open ticket (SONARPY-1366)
* There are not yet many projects that use the Python 3.12 syntax to be integrated in our ruling tests.

Simply setting one of our existing project as a Python 3.12 project allows us to guarantee analysis performs correctly on such project for a fraction of the cost that would be involved should we want an actual Python 3.12 project there.


joke1196 (approved):
LGTM. Thank you for looking into that.

PR #1616, opened 2023-10-20
SONARPY-1525 S5953: fix FP when using generic types for parameters and return type hints

maksim-grebeniuk-sonarsource:


python-checks/src/test/resources/checks/undefinedSymbols/undefinedSymbols.py
```diff
@@ -151,3 +151,8 @@ def getType():
 
     class C[A]():
         ...
+
+def undefined_generic_type[T: T](a: T) -> T:
+    x: T = a
+    a.foo()
+
```
guillaume-dequenne-sonarsource, line 158: I think the fix will do the trick, but as discussed, I'd like to see a test where it produces false negatives/false positives (in this rule or another) with a link to the ticket for the proper fix.
Because with this PR it looks as though everything is fine despite the fact we know it's not actually the correct behavior.

guillaume-dequenne-sonarsource (changes requested):
See my comment.

guillaume-dequenne-sonarsource (approved):
LGTM, thanks

PR #1617, opened 2023-10-23
SONARPY-1511 Rule S6794: Type aliases should be declared with a type statement.

jdodinh:


python-checks/src/test/java/org/sonar/python/checks/TypeAliasAnnotationCheckTest.java
```diff
@@ -0,0 +1,46 @@
...
+class TypeAliasAnnotationCheckTest {
+
+  @Test
+  void test_1() {
+    ProjectPythonVersion.setCurrentVersions(EnumSet.of(PythonVersionUtils.Version.V_312));
+    PythonCheckVerifier.verify("src/test/resources/checks/typeAliasAnnotation.py", new TypeAliasAnnotationCheck());
+  }
+
+  @Test
+  void test_2() {
```
joke1196, line 39: _Nitpick_ No changes required here but I think it would be nice to name the tests meaningfully. As they serve a specific purpose.

joke1196 (approved):
LGTM! Thanks for the good work!

PR #1618, opened 2023-10-23
SONARPY-1529 Standardise rule message formatting.

jdodinh:


joke1196 (approved):
LGTM! Not sure about the reformatting of the String for the Python code in `IsCloseAbsTolCheckTest `, I find it less clear that way. But no need to make a change here! Thank you for taking care of this!

PR #1621, opened 2023-10-26
SONARPY-1533: Ruff report is not using the correct column and row location

joke1196:


sonar-python-plugin/src/main/java/org/sonar/plugins/python/ruff/RuffJsonReportReader.java
```diff
@@ -81,22 +81,22 @@ private void onResult(JSONObject result) {
...
     JSONObject endLocation = (JSONObject) result.get("end_location");
-    issue.endLocationCol = correctEndLocationCol(endLocation.get("column"), issue.startLocationCol);
     issue.endLocationRow = toInteger(endLocation.get("row"));
+    issue.endLocationCol = correctEndLocationCol(endLocation.get("column"), issue.startLocationCol, issue.startLocationRow, issue.endLocationRow);
     consumer.accept(issue);
   }
 
   /*
     Ruff returns the col number of the last char + 1.
     In order to properly read the col number we need to return the number of the last char.
```
jdodinh, line 91: Maybe specify that we need to return the `col number`, not just `number`.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/ruff/RuffJsonReportReader.java
```diff
@@ -81,22 +81,22 @@ private void onResult(JSONObject result) {
...
+  private static Integer correctEndLocationCol(Object value, int startLocationCol, int startLocationRow, int endLocationRow) {
     Integer endLocationCol = toInteger(value);
     if (endLocationCol != null) {
-      if (endLocationCol > startLocationCol) {
-        return endLocationCol - 1;
-      } else {
+      if (endLocationRow == startLocationRow && endLocationCol <= startLocationCol) {
         return startLocationCol;
+      } else {
+        return endLocationCol -1 ;
```
jdodinh, line 99: Nitpick: Maybe we should have a space between the `-` and the `1`, i.e.: `endLocationCol - 1`.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/ruff/RuffJsonReportReader.java
```diff
@@ -81,22 +81,22 @@ private void onResult(JSONObject result) {
...
+    issue.endLocationCol = correctEndLocationCol(endLocation.get("column"), issue.startLocationCol, issue.startLocationRow, issue.endLocationRow);
     consumer.accept(issue);
   }
 
   /*
     Ruff returns the col number of the last char + 1.
     In order to properly read the col number we need to return the number of the last char.
    */
-  private static Integer correctEndLocationCol(Object value, int startLocationCol) {
+  private static Integer correctEndLocationCol(Object value, int startLocationCol, int startLocationRow, int endLocationRow) {
```
jdodinh, line 93: I assume that in this method we already assume that `startLocationRow <= endLocationRow`? Just making sure, as I'm not entirely clear on this.

jdodinh (approved):
LGTM! Left some minor comments.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/ruff/RuffJsonReportReader.java
joke1196: Yes so on the same row we can get the end location column which is either bigger or equal to the start location. For example, on an empty line, we get the same end location and start location

PR #1623, opened 2023-10-27
SONARPY-1509 Rule S6799: f-strings should not be nested too deeply

joke1196:


python-checks/src/main/java/org/sonar/python/checks/FStringNestingLevelCheck.java
```diff
@@ -0,0 +1,110 @@
...
+      }
+    }
+    return false;
+  }
+
+  private static boolean isTheNestingTooDeepInExpression(Expression expression, int updatedCount) {
+    return Optional.of(expression)
+      .flatMap(TreeUtils.toOptionalInstanceOfMapper(StringLiteral.class))
+      .map(StringLiteral::stringElements)
+      .map(elements -> elements.stream().anyMatch(sElement -> isFStringNestedTooDeep(sElement, updatedCount)))
```
jdodinh, line 92: Maybe it would be good to put the `.anyMatch` on a new line.

python-checks/src/test/resources/checks/fstringNestingLevel.py
```diff
@@ -0,0 +1,17 @@
...
+#               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
+
+    my_string = f"hello { f"format specifier: {greeting :{F"1"}.{2}}"}" # Noncompliant
+#               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+def compliant(hello, name):
+    greeting = f"{f"{hello}"},"
+    my_string = f"{greeting} {name}!" # Compliant
+
+    my_string = f"{greeting} {name : { f"1" }.{2}}!"
```
jdodinh, line 17: It would maybe be good to add some more cases, that include more of the edge cases of the rule. Maybe have the use of escaped curly braces (`{{`), or play around with the number of quotes that the fstring implements.

jdodinh (approved):
Overall looks very good, just a minor formatting detali, and some thoughts on test cases.

python-checks/src/test/resources/checks/fstringNestingLevel.py
joke1196: So the number of quotes is something I may not be able to test. In the sense that it should be an implementation detail of the StringElement, I should not have to test if the isInterpolated() method correctly reports if we are working with an interpolated string. This should be done at the level of the StringElement. For the escaped curly braces, it is a bit similar as I am only working with FromattedExpression I do not deal with the literals of the f-string. So the responsibility does not belong here in this case. If you want we can go into more details on this.

python-checks/src/main/java/org/sonar/python/checks/FStringNestingLevelCheck.java
```diff
@@ -0,0 +1,113 @@
...
+      }
+    }
+    return false;
+  }
+
+  private static boolean isTheNestingTooDeepInExpression(Expression expression, int updatedCount) {
+    return Optional.of(expression)
+      .flatMap(TreeUtils.toOptionalInstanceOfMapper(StringLiteral.class))
+      .map(StringLiteral::stringElements)
+      .map(Collection::stream)
```
maksim-grebeniuk-sonarsource, line 93: maybe a more flat chain is better:
```java
...
.flatMap(Collections::stream)
.anyMatch(sElement -> isFStringNestedTooDeep(sElement, updatedCount))
.orElse(false);
```

joke1196: I don't think I can do this `flatMap` on the `Optional` with the stream.

PR #1624, opened 2023-10-30
SONARPY-1536 Fix false positive for Rule S6735.

jdodinh:


joke1196 (approved):
LGTM.  I guess right_on without left_on could also raise an issue. But I guess this would be just out of the scope of this rule.

PR #1627, opened 2023-10-31
SONARPY-1478 Rule S5547: Add support for ssl module.

jdodinh:


python-checks/src/main/java/org/sonar/python/checks/RobustCipherAlgorithmCheck.java
```diff
@@ -63,14 +86,54 @@ public class RobustCipherAlgorithmCheck extends PythonSubscriptionCheck {
...
-        String fqn = calleeSymbol.fullyQualifiedName();
-        if (fqn != null && sensitiveCalleeFqns.contains(fqn)) {
-          subscriptionContext.addIssue(callExpr.callee(), MESSAGE);
-        }
-      }
+      Optional.ofNullable(callExpr)
+        .map(CallExpression::calleeSymbol)
+        .map(Symbol::fullyQualifiedName)
+        .filter(fqn -> sensitiveCalleeFqns.contains(fqn) || SSL_SET_CIPHERS_FQN.equals(fqn))
+        .ifPresent(str -> addMethodSpecificIssue(subscriptionContext, callExpr, str));
```
joke1196, line 93: I think here it would be easier to understand the program flow with branching: There are two case either it is a sensitive call and so we raise the issue or it is a ssl context call with a sensitive algorithm. I think it could be clearer to have your branching directly in the ifPresent.

python-checks/src/main/java/org/sonar/python/checks/RobustCipherAlgorithmCheck.java
```diff
@@ -63,14 +86,54 @@ public class RobustCipherAlgorithmCheck extends PythonSubscriptionCheck {
...
     });
   }
 
+  private static void addMethodSpecificIssue(
```
joke1196, line 97: I feel the naming is not really representative of what happens here. It is not an issue for a specific method. It is just an issue. I think this could be refactored in the previous ifPresent. This would allow the removal of a cognitive overhead with two contradicting `if`s in a row. `!SSL_SET_CIPHERS_FQN.equals(fullyQualifiedName)` and `SSL_SET_CIPHERS_FQN.equals(fqn)`.

python-checks/src/main/java/org/sonar/python/checks/RobustCipherAlgorithmCheck.java
```diff
@@ -63,14 +86,54 @@ public class RobustCipherAlgorithmCheck extends PythonSubscriptionCheck {
...
+    SubscriptionContext subscriptionContext,
+    CallExpression callExpression,
+    String fullyQualifiedName
+  ) {
+    if (!SSL_SET_CIPHERS_FQN.equals(fullyQualifiedName) || argumentSpecifiesRiskyAlgorithm(callExpression)) {
+      subscriptionContext.addIssue(callExpression.callee(), MESSAGE);
+    }
+  }
+
+  private static boolean argumentSpecifiesRiskyAlgorithm(CallExpression callExpression) {
```
joke1196, line 107: It is a good practice to name functions that returns booleans with is or has. for example `hasArgumentWithSensitiveAlgorithm`.

python-checks/src/main/java/org/sonar/python/checks/RobustCipherAlgorithmCheck.java
```diff
@@ -63,14 +86,54 @@ public class RobustCipherAlgorithmCheck extends PythonSubscriptionCheck {
...
+      .map(list -> list.get(0))
+      .flatMap(TreeUtils.toOptionalInstanceOfMapper(RegularArgument.class))
+      .map(RegularArgument::expression)
+      .map(RobustCipherAlgorithmCheck::unpackArgument)
+      .filter(RobustCipherAlgorithmCheck::containsInsecureCipher)
+      .isPresent();
+  }
+
+  @CheckForNull
+  private static String unpackArgument(@CheckForNull Expression expression) {
```
joke1196, line 119: I think in the argument list the annotation should be @Nullable.

python-checks/src/test/resources/checks/robustCipherAlgorithm.py
```diff
@@ -98,3 +114,19 @@ def pycryptoCompliant():
...
+  ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
+  ctx.set_ciphers(ciphers)
+  ciphers2 = ''
+  ctx.set_ciphers(ciphers2)
+  ciphers3 = 'ECDHE:RSA:AES256:GCM:SHA384'
+  ctx.set_ciphers(ciphers3)
+
+  ciphers4 = 1
+  ctx.set_ciphers(ciphers4)
+  ctx.set_ciphers(ciphers5)
```
joke1196, line 129: Here if you are testing with the `None` value it would be better to explicitly state it.

python-checks/src/test/resources/checks/robustCipherAlgorithm.py
```diff
@@ -98,3 +114,19 @@ def pycryptoCompliant():
   from Crypto.Cipher import *
   aes1 = AES.new('This is a key123', AES.MODE_CBC, 'This is an IV456') # Compliant
 
+def pysslCompliant():
```
joke1196, line 117: I see it was not done before but CamelCase is not really Pythonic. I don't think it needs to be changed here, but just noting it.

joke1196 (changes requested):
Looks good! I think a small refactoring can be done to improve readability.

python-checks/src/main/java/org/sonar/python/checks/RobustCipherAlgorithmCheck.java
jdodinh: Thanks for the tip. I refactored this in 1dd32b3d33381b16d208e2d0d833a5b7c2956532.

python-checks/src/main/java/org/sonar/python/checks/RobustCipherAlgorithmCheck.java
jdodinh: I removed this method in 1dd32b3d33381b16d208e2d0d833a5b7c2956532.

python-checks/src/main/java/org/sonar/python/checks/RobustCipherAlgorithmCheck.java
jdodinh: Thanks for the tip. I refactored this in 1dd32b3d33381b16d208e2d0d833a5b7c2956532.

python-checks/src/main/java/org/sonar/python/checks/RobustCipherAlgorithmCheck.java
jdodinh: I removed this method in 1dd32b3d33381b16d208e2d0d833a5b7c2956532.

python-checks/src/main/java/org/sonar/python/checks/RobustCipherAlgorithmCheck.java
jdodinh: You are correct. Made the change in e5c69678b8e41c64fdde93c1c58a30377d138350.

python-checks/src/main/java/org/sonar/python/checks/RobustCipherAlgorithmCheck.java
jdodinh: Updated in a18d2c0663b4df943eb27a458edc0490c7d7ee48.

python-checks/src/test/resources/checks/robustCipherAlgorithm.py
```diff
@@ -81,6 +82,21 @@ def pydesExamples():
...
+  ctx.set_ciphers(ciphers)  # Noncompliant
```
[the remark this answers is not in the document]
jdodinh: Yes I agree. This was done in 244c2dec1f07e424721054503110d6d4dd1919a5.

python-checks/src/test/resources/checks/robustCipherAlgorithm.py
jdodinh: I made the change in 724de8800a0cff9aef4f33ca3d0e813f80acfb00.

python-checks/src/test/resources/checks/robustCipherAlgorithm.py
jdodinh: Ok, thank you. I think it's maybe worth changing these function names.

jdodinh: Fixed in 03af7494fb21355e655e0b166750628ac16418f1.

python-checks/src/test/resources/checks/robustCipherAlgorithm.py
```diff
@@ -81,20 +82,52 @@ def pydesExamples():
...
+  ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
+  ctx.set_ciphers(ciphers)
+  ciphers2 = ''
+  ctx.set_ciphers(ciphers2)
+  ciphers3 = 'ECDHE:RSA:AES256:GCM:SHA384'
+  ctx.set_ciphers(ciphers3)
+
+  ciphers4 = 1
+  ctx.set_ciphers(ciphers4)
+  ctx.set_ciphers(null_value)
```
joke1196, line 130: So here it is not valid Python code. null_value is not defined. Either you want to do the test with `None` or you want to test the fact that passing a value you don't know should not raise an issue. In this case you should have:
```python
def pyssl_compliant(unknown_cipher):
    ...
    ctx.set_ciphers(unknown_cipher)
```

joke1196 (changes requested):
Just one small change on the test otherwise it looks really good the refactoring makes it super readable. Good job!

joke1196 (approved):
LGTM! Great work!

PR #1628, opened 2023-10-31
SONARPY-1542: Fix lexer as raw strings consider backslash as a character except before quotes

joke1196:


maksim-grebeniuk-sonarsource (approved):
LGTM

PR #1633, opened 2023-11-02
SONARPY-1513 Rule S6795: Generic type statement should not use TypeVars

joke1196:


ghislainpiot (changes requested):
The logic looks good to me. I still have some changes on code duplication and the fact it does not build.

Also, there is a typo in the RSPEC : `# Type checker error would be raise`

ghislainpiot (approved):
LGTM!

PR #1635, opened 2023-11-06
SONARPY-1475 Rule S5332: Support `http.server.HTTPServer` and subclasses

joke1196:


python-checks/src/main/java/org/sonar/python/checks/hotspots/ClearTextProtocolsCheck.java
```diff
@@ -47,14 +57,16 @@ public class ClearTextProtocolsCheck extends PythonSubscriptionCheck {
   private static final List<String> SENSITIVE_PROTOCOLS = Arrays.asList("http://", "ftp://", "telnet://");
   private static final Pattern LOOPBACK = Pattern.compile("localhost|127(?:\\.[0-9]+){0,2}\\.[0-9]+$|^(?:0*\\:)*?:?0*1", Pattern.CASE_INSENSITIVE);
   private static final Map<String, String> ALTERNATIVES = new HashMap<>();
+  private static final String SENSITIVE_HTTP_SERVER_CALL = "socketserver.BaseServer.serve_forever";
```
jdodinh, line 60: Are we sure that in this case only the calls to `serve_forever()` originating from instances of `http.server.HTTPServer`, `http.server.ThreadingHTTPServer` and subclasses will resolve to this fully qualified name? Otherwise it could result in nome FPs.

jdodinh (commented):
LGTM! Great work. Just a few minor comments, and a clarification!

python-checks/src/main/java/org/sonar/python/checks/hotspots/ClearTextProtocolsCheck.java
joke1196: Yes in the worst case scenario it will resolve to the BaseServer which is the parent class of `HTTPServer` and `ThreadingServer`

python-checks/src/test/resources/checks/hotspots/clearTextProtocols.py
```diff
@@ -142,3 +142,62 @@ def FP_starttls_in_different_method():
...
+    class MyServer(HTTPServer):
+        def run(self):
+            super(self).serve_forever()  # Compliant
+
+        def server_bind(self):
+            pass
+
+    my_server = MyServer()
+    my_server.serve_forever()
+    my_server.server_bind()
```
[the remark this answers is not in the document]
joke1196: I am not sure what you mean by custom. Here `HTTPServer` is a user-defined class not the one from the import.

python-checks/src/main/java/org/sonar/python/checks/hotspots/ClearTextProtocolsCheck.java
jdodinh: Sounds good.

jdodinh: I meant repeating this test, but with an instance of the user-defined `HTTPServer` class. Such as:
```
my_server = MyServer()
my_server.serve_forever()
my_server.server_bind()
```
Maybe it would be an overkill though. Let me know what you think.

python-checks/src/test/resources/checks/hotspots/clearTextProtocols.py
```diff
@@ -142,3 +142,62 @@ def FP_starttls_in_different_method():
...
+            super(self).serve_forever()  # Noncompliant
+        #   ^^^^^^^^^^^^^^^^^^^^^^^^^
+
+        def server_bind(self):  # Noncompliant
+#       ^^^^^^^^^^^^^^^^^^^^^
+            pass
+
+    my_server = MyServer(('0.0.0.0', 8080), SimpleHTTPRequestHandler)
+    my_server.serve_forever()  # Noncompliant
+#   ^^^^^^^^^^^^^^^^^^^^^^
```
jdodinh, line 167: Just careful with this line. Add a few more `^` to make the tests pass.

python-checks/src/test/resources/checks/hotspots/clearTextProtocols.py
```diff
@@ -142,3 +142,62 @@ def FP_starttls_in_different_method():
...
+    http_server = HTTPServer(('0.0.0.0', 8080), SimpleHTTPRequestHandler)
+    http_server.serve_forever()  # Noncompliant
+#   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    threading_http_server = ThreadingHTTPServer(('0.0.0.0', 8080), SimpleHTTPRequestHandler)
+    threading_http_server.serve_forever()  # Noncompliant
+
+    class MyServer(HTTPServer):
+        def run(self):
+            super(self).serve_forever()  # Noncompliant
+        #   ^^^^^^^^^^^^^^^^^^^^^^^^^
```
jdodinh, line 159: Maybe we should raise the issue on the entire function call, and not just the callee?

jdodinh (changes requested):
It looks good to me, but there is one detail that makes the build fail, and a small detail that doesn't necessarily need to be addressed.

python-checks/src/main/java/org/sonar/python/checks/hotspots/ClearTextProtocolsCheck.java
joke1196: I see maybe we can discuss what will be tested here if I add the tests.

jdodinh (approved):
Amazing! LGTM, thank you very much!

PR #1636, opened 2023-11-06
SONARPY-1468 Rule S6779: Flask secret keys should not be disclosed

jdodinh:
This is a preliminary implementation, that passes the current tests. However, I think some of these should be modified, mainly since I am still not clear on whether we should raise the issue when the assigned values are not known.

### DIFF
--- a/python-checks/src/main/java/org/sonar/python/checks/FlaskHardCodedSecretCheck.java
+++ b/python-checks/src/main/java/org/sonar/python/checks/FlaskHardCodedSecretCheck.java
@@ -22,7 +22,6 @@ package org.sonar.python.checks;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
-import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import org.sonar.check.Rule;
@@ -150,12 +149,6 @@ public class FlaskHardCodedSecretCheck extends PythonSubscriptionCheck {
       expressionList.forEach(expr -> issue.secondary(expr, SECONDARY_MESSAGE));
     }
   }
-
-  private static BiFunction<PreciseIssue, Expression, PreciseIssue> myBiFunction =
-    (issue, expression) -> {
-      issue.secondary(expression, SECONDARY_MESSAGE);
-      return issue;
-    };
 
   private static boolean isSensitiveProperty(Expression expression) {
     if (expression.is(Tree.Kind.SUBSCRIPTION)) {


python-checks/src/main/java/org/sonar/python/checks/FlaskHardCodedSecretCheck.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  private static void verifyAssignmentStatement(SubscriptionContext ctx) {
+    AssignmentStatement assignmentStatementTree = (AssignmentStatement) ctx.syntaxNode();
+    for (ExpressionList lhsExpression : assignmentStatementTree.lhsExpressions()) {
+      boolean isModifyingSensitiveProperty = lhsExpression.expressions().stream()
+      .anyMatch(FlaskHardCodedSecretCheck::isSensitiveProperty);
+      if (isModifyingSensitiveProperty) {
+        ctx.addIssue(lhsExpression, MESSAGE);
+      }
```
joke1196, line 142: Could that be replaced with a `filter` and a `ifPresent`?

python-checks/src/main/java/org/sonar/python/checks/FlaskHardCodedSecretCheck.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+  }
+
+  private static boolean isSensitiveProperty(Expression expression) {
+    if (expression.is(Tree.Kind.SUBSCRIPTION)) {
+      return Optional.of((SubscriptionExpression) expression)
+        .map(SubscriptionExpression::object)
+        .flatMap(TreeUtils.toOptionalInstanceOfMapper(QualifiedExpression.class))
+        .map(QualifiedExpression::symbol)
+        .map(Symbol::fullyQualifiedName)
```
joke1196, line 152: This may be replaced with `TreeUtils.fullyQualifiedNameFromQualifiedExpression`

python-checks/src/main/java/org/sonar/python/checks/FlaskHardCodedSecretCheck.java
```diff
@@ -0,0 +1,183 @@
...
+        .map(list -> list.get(0))
+        .map(FlaskHardCodedSecretCheck::getAssignedValue)
+        .flatMap(TreeUtils.toOptionalInstanceOfMapper(StringLiteral.class))
+        .map(StringLiteral::trimmedQuotesValue)
+        .filter(SECRET_KEY_KEYWORD::equals)
+        .isPresent();
+    } else if (expression.is(Tree.Kind.QUALIFIED_EXPR)) {
+      return Optional.of((QualifiedExpression) expression)
+        .map(QualifiedExpression::symbol)
+        .map(Symbol::fullyQualifiedName)
```
joke1196, line 166: Same as the comment above

python-checks/src/test/resources/checks/flaskHardCodedSecret.py
```diff
@@ -0,0 +1,110 @@
...
+    ))
+
+
+    current_app.config.update(hello(
+        SECRET_KEY='SECRET'  # OK, but not sure if it should be.
+    ))
+    current_app.config.update()
+    current_app.config.update('secret')
+
+    assigned_secret = 'hardcoded_secret'
```
joke1196, line 94: I guess this is already defined above. We could remove this statement

python-checks/src/test/resources/checks/flaskHardCodedSecret.py
```diff
@@ -0,0 +1,110 @@
...
+    x = current_app.config['SECRET'] = 'secret'  # OK
+    current_app.config['SECRET'] = x = 'secret'  # OK
+    current_app.config['SECRET'], _ = 'mysecret', x # OK
+    current_app.config['SECRET'], _ = _, current_app.config['SECRET'] = 'secret', x # OK
+
+    current_app.config['SECRET'] = x  # OK
+    y, current_app.config['SECRET'] = 'some_random_string', x
+    current_app.config['SECRET'], y = z, current_app.config['SECRET'] = x, 'secret'
+
+    current_app.config['SECRET'], y = z, current_app.config['SECRET'] = x
```
joke1196, line 109: As I understand all of these tests are testing only the case where the KEY is `SECRET` instead of  `SECRET_KEY`. I think it would be a good idea to test if on the lhs we have the correct key and on the rhs we have a compliant solution. So right now the coverage is not really 100%, I know it is a bit deceiving but bear in mind that using `Optional` will make the coverage tool thinks there is only one path to visit.

joke1196:
So if we don't know about the value we should not raise an issue. If it is not certain the value is a hardcoded value we prefer a FN than an FP. A test to illustrate that would be something like:
```
secret = 'secret'

secret = get_secret_from_vault()

current_app.secret_key = secret # Compliant
```

python-checks/src/test/resources/checks/flaskHardCodedSecret.py
```diff
@@ -0,0 +1,110 @@
...
+
+    app.config['SECRET'] = 'secret'  # OK
+    app.config['SECRET'] = assigned_secret  # OK
+    x = app.config['SECRET'] = 'secret'  # OK
+    app.config['SECRET'] = x = 'secret'  # OK
+    app.config['SECRET'], _ = 'mysecret', x # OK
+    app.config['SECRET'], _ = _, app.config['SECRET_KEY'] = 'secret', x # OK
+
+    current_app.config.update(dict(
+        SECRET_KEY=x  # OK, but not sure if it should be.
```
joke1196, line 78: This is ok as we cannot for certain tell that x is an hardcode string

python-checks/src/test/resources/checks/flaskHardCodedSecret.py
```diff
@@ -0,0 +1,110 @@
...
+    def hello(SECRET_KEY):
+        ...
+
+    current_app.config.update(dict(
+        SECRET_KEY=hello('something')  # OK, but not sure if it should be.
+    ))
+
+
+    current_app.config.update(hello(
+        SECRET_KEY='SECRET'  # OK, but not sure if it should be.
```
joke1196, line 89: this is ok a well as we do not know what hello returns

python-checks/src/test/resources/checks/flaskHardCodedSecret.py
```diff
@@ -0,0 +1,110 @@
...
+    app.config['SECRET'], _ = _, app.config['SECRET_KEY'] = 'secret', x # OK
+
+    current_app.config.update(dict(
+        SECRET_KEY=x  # OK, but not sure if it should be.
+    ))
+    def hello(SECRET_KEY):
+        ...
+
+    current_app.config.update(dict(
+        SECRET_KEY=hello('something')  # OK, but not sure if it should be.
```
joke1196, line 84: this is ok as well, hello could fetch the secret. It is not a hardcoded value

python-checks/src/test/resources/checks/flaskHardCodedSecret.py
```diff
@@ -0,0 +1,110 @@
...
+    ))
+
+
+    current_app.config.update(hello(
+        SECRET_KEY='SECRET'  # OK, but not sure if it should be.
+    ))
+    current_app.config.update()
+    current_app.config.update('secret')
```
joke1196, line 92: Not sure what this does in fact. just curious

python-checks/src/test/resources/checks/flaskHardCodedSecret.py
joke1196: This is a good example of a missing test `dict(A_KEY='something')` inside the `current_app.config.update` function.

joke1196 (changes requested):
Looks good! A lot of cases where taken into account great job. I have left some comments mainly about tests. It would be best to go through them and check which ones are missing. As mentioned it could be that the coverage did not take into account some branches because of the functional API.

python-checks/src/main/java/org/sonar/python/checks/FlaskHardCodedSecretCheck.java
jdodinh: It seems that it's not quite working the way I want it with your suggestion, therefore I will stick to this. Thank you however for making me aware of this method!

python-checks/src/main/java/org/sonar/python/checks/FlaskHardCodedSecretCheck.java
jdodinh: Same as above.

python-checks/src/main/java/org/sonar/python/checks/FlaskHardCodedSecretCheck.java
jdodinh: Since I am working on a stream, I can't invoke `ifPresent`. I will still look for a neater option.

jdodinh: I changed the implementation slightly.

python-checks/src/test/resources/checks/flaskHardCodedSecret.py
jdodinh: Thanks for the tip. I made the key a little bit more explicit and added some tests.

python-checks/src/test/resources/checks/flaskHardCodedSecret.py
jdodinh: Done.

python-checks/src/test/resources/checks/flaskHardCodedSecret.py
```diff
@@ -0,0 +1,147 @@
...
+
+    current_app.config[x] = 'secret'  # OK
+    current_app.config['A_KEY'] = 'secret'  # OK
+    current_app.config['A_KEY', x] = 'secret'  # OK
+    current_app.config['A_KEY'] = assigned_secret  # OK
+    x = current_app.config['A_KEY'] = 'secret'  # OK
+    current_app.config['A_KEY'] = x = 'secret'  # OK
+    current_app.config['A_KEY'], _ = 'mysecret', x # OK
+    current_app.config['A_KEY'], _ = _, current_app.config['SECRET'] = 'secret', x # OK
```
joke1196, line 134: I think this is fine like this. But I don't think all the subsequent tests here are relevant. From what I get from the implementation only once the lhs contains a sensitive property, the rhs is checked. This makes all this code redundant. If you want we can discuss some of the approaches related to TDD to mitigate this issue.

joke1196 (approved):
LGTM! Still added a small comment about redundant tests, maybe we can take some time to discuss it.

PR #1637, opened 2023-11-07
SONARPY-1498 Rule S6786: Python GraphQL introspection should be disabled

joke1196:


python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
```diff
@@ -0,0 +1,123 @@
...
+
+  private static final String MESSAGE = "Disable introspection on this GraphQL server endpoint.";
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.CALL_EXPR, GraphQLIntrospectionCheck::checkGraphQLIntrospection);
+  }
+
+  private static void checkGraphQLIntrospection(SubscriptionContext ctx) {
+    CallExpression callExpression = (CallExpression) ctx.syntaxNode();
```
maksim-grebeniuk-sonarsource, line 61: maybe if you already using the `var` keyword - it could be here as well?

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
```diff
@@ -0,0 +1,123 @@
...
+    .map(String::toUpperCase)
+    .anyMatch(name -> name.contains("INTROSPECTION"));
+
+  private static Predicate<List<Expression>> expressionsContainsSensitiveRuleFQN = expressions -> expressions.stream()
+    .map(TreeUtils::getSymbolFromTree)
+    .filter(Optional::isPresent)
+    .map(Optional::get)
+    .map(Symbol::fullyQualifiedName)
+    .filter(Objects::nonNull)
+    .anyMatch(SENSITIVE_VALIDATION_RULE_FQNS::contains);
```
maksim-grebeniuk-sonarsource, line 113: I like functional API in Java, and I'm trying to use it as much as possible but even for me it took some time to figure out what's going on here. It is up to you but maybe it will be easier to read the code if you will have just two methods to check the expressions, like:
```java
private static boolean isExpressionsNameContainIntrospection(List<Expression> expressions) {
    return expressions.stream()
      .map(GraphQLIntrospectionCheck::nameFromIndentifierOrCallExpression)
      .filter(Optional::isPresent)
      .map(Optional::get)
      .map(String::toUpperCase)
      .anyMatch(name -> name.contains("INTROSPECTION"));
  }
  
  private static boolean isExpressionsContainsSensitiveRuleFQN(List<Expression> expressions) {
    return expressions.stream()
      .map(TreeUtils::getSymbolFromTree)
      .filter(Optional::isPresent)
      .map(Optional::get)
      .filter(Objects::nonNull)
      .anyMatch(SENSITIVE_VALIDATION_RULE_FQNS::contains);
  }
  
  private static boolean isExpressionsNameContainIntrospectionOrContainsSensitiveRuleFQN(List<Expression> expressions) {
    return isExpressionsNameContainIntrospection(expressions) || isExpressionsContainsSensitiveRuleFQN(expressions);
  }
```

joke1196: I see so remove the `or` on line 75. And I guess not use a predicate in the `areArgumentsStafe` method

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
joke1196: I guess I will change the `var` to `Argument`

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
joke1196: With the change I had a bit more duplicate code as I had to treat empty list and empty `Optional` differently but I think the intent is quite clear now.

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  private static Optional<List<Expression>> extractArgumentValues(RegularArgument argument) {
+    return Optional.of(argument)
+      .map(RegularArgument::expression)
+      .flatMap(GraphQLIntrospectionCheck::expressionsFromListOrTuple);
+  }
+
+  private static Optional<List<Expression>> expressionsFromListOrTuple(Expression expression) {
+    return TreeUtils.toOptionalInstanceOfMapper(ListLiteral.class)
```
maksim-grebeniuk-sonarsource, line 97: the `TreeUtils. toOptionalInstanceOf` method could be used there

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
```diff
@@ -0,0 +1,133 @@
...
+      .apply(expression)
+      .map(ListLiteral::elements)
+      .map(ExpressionList::expressions)
+      .or(() -> TreeUtils.toOptionalInstanceOfMapper(Tuple.class)
```
maksim-grebeniuk-sonarsource, line 101: the `TreeUtils. toOptionalInstanceOf` method could be used there

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
```diff
@@ -0,0 +1,133 @@
...
+      .filter(Optional::isPresent)
+      .map(Optional::get)
+      .map(Symbol::fullyQualifiedName)
+      .filter(Objects::nonNull)
+      .anyMatch(SENSITIVE_VALIDATION_RULE_FQNS::contains);
+  }
+
+  private static Optional<String> nameFromIndentifierOrCallExpression(Expression expression) {
+    return Optional.ofNullable(TreeUtils.nameFromExpression(expression))
+      .or(() -> TreeUtils.toOptionalInstanceOfMapper(CallExpression.class)
```
maksim-grebeniuk-sonarsource, line 127: the `TreeUtils. toOptionalInstanceOf` method could be used there

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
```diff
@@ -0,0 +1,133 @@
...
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.CALL_EXPR, GraphQLIntrospectionCheck::checkGraphQLIntrospection);
+  }
+
+  private static void checkGraphQLIntrospection(SubscriptionContext ctx) {
+    CallExpression callExpression = (CallExpression) ctx.syntaxNode();
+    Optional.ofNullable(callExpression.calleeSymbol())
+      .map(Symbol::fullyQualifiedName)
+      .filter(fqn -> GRAPHQL_VIEWS_FQNS.contains(fqn) &&
```
maksim-grebeniuk-sonarsource, line 63: I guess it could be split up by two `filter` method calls:
```java
.filter(GRAPHQL_VIEWS_FQNS::contains)
.filter(fqn -> hasSensitiveMiddlewares(callExpression.arguments()) || hasSensitiveValidationRules(callExpression.arguments())
```
or since you are working with the `callExpression` all the time - maybe we don't need to extract the symbol at all:
```java
Optional.of(callExpression)
  .filter(exp -> Optional.of(exp)
    .map(CallExpression::calleeSymbol)
    .stream()
    .allMatch(GRAPHQL_VIEWS_FQNS::contains)
  )
  .filter(exp -> hasSensitiveMiddlewares(callExpression.arguments()) || hasSensitiveValidationRules(callExpression.arguments()))
  .ifPresent(exp -> ctx.addIssue(exp, MESSAGE)
```

maksim-grebeniuk-sonarsource (approved):
LGTM, see a couple of minor small suggestions. They are mostly about code style, not logic, so feel free to ignore

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
joke1196: Yes good call thanks!

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
joke1196: The two filters make it clearer, thank you. I will keep the non-nested version though. I think the intent is still without the call expression being passed all the way down

python-checks/src/test/resources/checks/hotspots/graphQLIntrospection.py
```diff
@@ -0,0 +1,227 @@
+
```
guillaume-dequenne-sonarsource, line 1: _Nitpick:_ This seems unneeded.

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
```diff
@@ -0,0 +1,131 @@
...
+
+  private static final Set<String> GRAPHQL_VIEWS_FQNS = Set.of(
+    "flask_graphql.GraphQLView.as_view",
+    "graphql_server.flask.GraphQLView.as_view");
+
+  private static final Set<String> SENSITIVE_VALIDATION_RULE_FQNS = Set.of(
+    "graphene.validation.DisableIntrospection",
+    "graphql.validation.NoSchemaIntrospectionCustomRule");
+
+  private static final String MESSAGE = "Disable introspection on this GraphQL server endpoint.";
```
jdodinh, line 52: Maybe change this to: 
`private static final String MESSAGE = "Disable introspection on this \"GraphQL\" server endpoint."; `

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
```diff
@@ -0,0 +1,131 @@
...
+      .filter(Objects::nonNull)
+      .anyMatch(SENSITIVE_VALIDATION_RULE_FQNS::contains);
+  }
+
+  private static Optional<String> nameFromIndentifierOrCallExpression(Expression expression) {
+    return Optional.ofNullable(TreeUtils.nameFromExpression(expression))
+      .or(() -> TreeUtils.toOptionalInstanceOf(CallExpression.class, expression)
+        .map(CallExpression::callee)
+        .map(TreeUtils::nameFromExpression))
+      .filter(Objects::nonNull);
```
guillaume-dequenne-sonarsource, line 129: This doesn't seem to be needed.

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
```diff
@@ -0,0 +1,131 @@
...
+    return expressions.stream()
+      .map(TreeUtils::getSymbolFromTree)
+      .filter(Optional::isPresent)
+      .map(Optional::get)
+      .map(Symbol::fullyQualifiedName)
+      .filter(Objects::nonNull)
+      .anyMatch(SENSITIVE_VALIDATION_RULE_FQNS::contains);
+  }
+
+  private static Optional<String> nameFromIndentifierOrCallExpression(Expression expression) {
```
guillaume-dequenne-sonarsource, line 124: _Nitpick:_ typo in `Identifier` (extra n)

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
```diff
@@ -0,0 +1,131 @@
...
+    context.registerSyntaxNodeConsumer(Tree.Kind.CALL_EXPR, GraphQLIntrospectionCheck::checkGraphQLIntrospection);
+  }
+
+  private static void checkGraphQLIntrospection(SubscriptionContext ctx) {
+    CallExpression callExpression = (CallExpression) ctx.syntaxNode();
+    Optional.ofNullable(callExpression.calleeSymbol())
+      .map(Symbol::fullyQualifiedName)
+      .filter(GRAPHQL_VIEWS_FQNS::contains)
+      .filter(fqn -> hasSensitiveMiddlewares(callExpression.arguments()) ||
+        hasSensitiveValidationRules(callExpression.arguments()))
```
guillaume-dequenne-sonarsource, line 65: I think that the logical operator here should be `&&` rather than `||`. 
If either a middleware or a validation rule disables introspection, we should not raise an issue.

(Admittedly, the ticket says `When the .as_view(...) method is called on these classes, S6786 should be raised if there is no middleware or no validator that disables introspection.`, but looking at the RSPEC, it seems that the issue should be raised only if both are missing).

Note: Maybe instead of naming the methods `hasSensitiveMiddlewares`, we could consider having them named `doesNotHaveSafeMiddleware` and `doesNotHaveSafeValidationRule` to reflect more explicitly what the rule is looking for.

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
```diff
@@ -0,0 +1,131 @@
...
+import org.sonar.python.tree.TreeUtils;
+
+@Rule(key = "S6786")
+public class GraphQLIntrospectionCheck extends PythonSubscriptionCheck {
+
+  private static final Set<String> GRAPHQL_VIEWS_FQNS = Set.of(
+    "flask_graphql.GraphQLView.as_view",
+    "graphql_server.flask.GraphQLView.as_view");
+
+  private static final Set<String> SENSITIVE_VALIDATION_RULE_FQNS = Set.of(
```
guillaume-dequenne-sonarsource, line 48: I believe this should be `SAFE _VALIDATION_RULE_FQNS` as these are the validation rules that disable introspection.

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
```diff
@@ -0,0 +1,131 @@
...
+  private static boolean expressionsNameContainIntrospection(List<Expression> expressions) {
+    return expressions.stream()
+      .map(GraphQLIntrospectionCheck::nameFromIndentifierOrCallExpression)
+      .filter(Optional::isPresent)
+      .map(Optional::get)
+      .map(String::toUpperCase)
+      .anyMatch(name -> name.contains("INTROSPECTION"));
+  }
+
+  private static boolean expressionsContainsSensitiveRuleFQN(List<Expression> expressions) {
```
guillaume-dequenne-sonarsource, line 114: Here then this should probably be something like `expressionsContainsSafeRuleFQN`

python-checks/src/test/resources/checks/hotspots/graphQLIntrospection.py
```diff
@@ -0,0 +1,227 @@
...
+def flask_graphql_non_compliant(schema, some_middleware, introspection_middleware):
+    from flask_graphql import GraphQLView
+    from middleware import IntrospectionMiddleware, get_introspection_middleware
+
+    GraphQLView.as_view(  # Noncompliant {{Disable introspection on this GraphQL server endpoint.}}
+#   ^[el=+5;ec=5]
```
guillaume-dequenne-sonarsource, line 7: Are we sure we want to highlight the full call? This may be noisy (Not asking for a change, though - just asking).

jdodinh (changes requested):

guillaume-dequenne-sonarsource (changes requested):
Overall implementation looks good, but there seems to be some mismatch between the logic of the specification and the implementation.

python-checks/src/test/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheckTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.Collections;
+import org.junit.jupiter.api.Test;
+import org.sonar.python.checks.utils.PythonCheckVerifier;
+
+class GraphQLIntrospectionCheckTest {
+
+  @Test
+  void test() {
+    PythonCheckVerifier.verify(Collections.singletonList("src/test/resources/checks/hotspots/graphQLIntrospection.py"), new GraphQLIntrospectionCheck());
```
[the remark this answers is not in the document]
joke1196: None really

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
joke1196: This was the message in the RSPEC so not sure if I should change it...

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  private static boolean hasSensitiveMiddlewares(List<Argument> arguments) {
+    RegularArgument argument = TreeUtils.argumentByKeyword("middleware", arguments);
+    if (argument == null) {
+      return true;
+    }
+    return extractArgumentValues(argument)
+      .map(values -> values.isEmpty() || expressionsNameContainIntrospection(values))
+      .orElse(false);
```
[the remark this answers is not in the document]
joke1196: Yes you are right I did it the wrong way infact

python-checks/src/test/resources/checks/hotspots/graphQLIntrospection.py
joke1196: I was not sure the ticket said the method `as_view`. I guess it would be enough to highlight the name only

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
guillaume-dequenne-sonarsource: I think it's about the fact that we want to consistently highlight code in issue messages. That said, I can't (again) remember where this is supposed to be written, nor whether it should be backticks or double quotes...

joke1196: I would think here that GraphQL is just a brand. I am not sure I would consider it code. I.e. if it was Disable introspection on this Spring server endpoint. not sure I would highlight it. I can still make the change. The double quotes is the way to go.

python-checks/src/test/resources/checks/hotspots/graphQLIntrospection.py
```diff
@@ -0,0 +1,269 @@
...
+        schema=schema,
+        graphiql=True,
+        validation_rules=[SafeRule, some_rule]
+    )
+
+    GraphQLView.as_view( 
+        name="introspection",
+        schema=schema,
+        graphiql=True,
+        validation_rules=[OtherSensitiveRule, some_rule]
```
jdodinh, line 159: Maybe rename this to `OtherSafeRule`?

python-checks/src/test/resources/checks/hotspots/graphQLIntrospection.py
```diff
@@ -0,0 +1,269 @@
...
+        graphiql=True,
+        validation_rules=[SafeRule, some_rule]
+    )
+
+    GraphQLView.as_view( 
+        name="introspection",
+        schema=schema,
+        graphiql=True,
+        validation_rules=[introspection_rule]
+    )
```
jdodinh, line 269: Maybe it would also be useful to have a test case where we have both of the keyword arguments in the `as_view()` function call, and combine cases of these arguments containing compliant and non_compliant classes and methods?

jdodinh (approved):
LGTM! Well done on the fix of the logic, and the clarity of the code. Just left a few minor comments, otherwise it all looks good :)

python-checks/src/test/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheckTest.java
```diff
@@ -19,15 +19,16 @@
...
+
 import org.junit.jupiter.api.Test;
 import org.sonar.python.checks.utils.PythonCheckVerifier;
 
 class GraphQLIntrospectionCheckTest {
 
   @Test
   void test() {
-    PythonCheckVerifier.verify(Collections.singletonList("src/test/resources/checks/hotspots/graphQLIntrospection.py"), new GraphQLIntrospectionCheck());
+    PythonCheckVerifier.verify(List.of("src/test/resources/checks/hotspots/graphQLIntrospection.py"), new GraphQLIntrospectionCheck());
```
guillaume-dequenne-sonarsource, line 31: In fact I don't think you even need the list at all. This API works both with a single path an with a list of paths.
With a list, it will create a project symbol table for all the files, but this is unneeded for a single file.

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
```diff
@@ -61,31 +61,32 @@ private static void checkGraphQLIntrospection(SubscriptionContext ctx) {
...
     RegularArgument argument = TreeUtils.argumentByKeyword("middleware", arguments);
     if (argument == null) {
-      return true;
+      return false;
     }
+
     return extractArgumentValues(argument)
-      .map(values -> values.isEmpty() || expressionsNameContainIntrospection(values))
-      .orElse(false);
+      .map(values -> !values.isEmpty() && expressionsNameContainIntrospection(values))
```
guillaume-dequenne-sonarsource, line 76: Here we check for expression names (and I think it's good), but the ticket mentions class and method names.
So I guess something like:

```

    class IntrospectionCustomMiddleware:
        ...
    my_custom_middleware = IntrospectionCustomMiddleware()

    GraphQLView.as_view(
            name="introspection",
            schema=schema,
            graphiql=True,
            middleware=[my_custom_middleware]
        )
```
should not raise an issue, but it does.

(Same thing applies to validation rules)
Sorry this was missed in the first review...

guillaume-dequenne-sonarsource (commented):
Updated logic looks good!
I notice there is a safe case that may not be covered by the rule yet. Not sure if we want to address it in this PR, or merge this PR and address it in a later one. Let's discuss this.

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
joke1196: Yes, no worries do you think it is enough to check the runtime type for this one? I will do an implementation for this I guess some other cases would then miss and I will create a ticket. i.e. for the case where a method returns safe middleware.

guillaume-dequenne-sonarsource: I think checking runtime time is enough indeed!
The case where a method not named `introspection` returns a safe middleware would probably be too complex for a rule like this one, I don't think we need to cover it.

python-checks/src/test/resources/checks/hotspots/graphQLIntrospection.py
```diff
@@ -87,6 +87,28 @@ def compliant_import():
...
+            name="introspection",
+            schema=schema,
+            graphiql=True,
+            middleware=[my_custom_middleware]
+        )
+
+    def create_middleware():
+        return IntrospectionCustomMiddleware()
+
+    # FP this should be fixed with SONARPY-1554
```
guillaume-dequenne-sonarsource, line 104: I think we can just accept the FP without a ticket here: it doesn't look like something we'll really want to solve anytime soon.

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
```diff
@@ -104,8 +107,11 @@ private static Optional<List<Expression>> expressionsFromListOrTuple(Expression
   }
 
   private static boolean expressionsNameContainIntrospection(List<Expression> expressions) {
-    return expressions.stream()
-      .map(GraphQLIntrospectionCheck::nameFromIdentifierOrCallExpression)
+    Stream<Optional<String>> expressionsNameAndType = Stream.concat(expressions.stream()
+      .map(GraphQLIntrospectionCheck::nameFromIdentifierOrCallExpression), 
+      expressions.stream().map(GraphQLIntrospectionCheck::nameOfType) );
```
guillaume-dequenne-sonarsource, line 112: Extra space here before parenthesis?

guillaume-dequenne-sonarsource (approved):
LGTM!
Left very minor comments, no blocker for merging.

PR #1638, opened 2023-11-08
SONARPY-1434 Rule S5122: Add support for flask_cors 2

joke1196:


joke1196:
I am not sure if this is enough. As we do not have the stubs, we can resolve the decorator directly.

guillaume-dequenne-sonarsource (approved):
LGTM.
I think it's fine that we don't have stubs for this, since it's an older version of flask_cors anyway.

PR #1639, opened 2023-11-09
SONARPY-1474 Rule S6781: JWT secret keys should not be disclosed.

jdodinh:


python-checks/src/main/java/org/sonar/python/checks/FlaskHardCodedSecret.java
```diff
@@ -117,40 +115,40 @@ private static boolean isCallToDictConstructor(CallExpression callExpression) {
       .isPresent();
   }
 
-  private static boolean hasIllegalKeyValuePair(DictionaryLiteral dictionaryLiteral) {
+  private boolean hasIllegalKeyValuePair(DictionaryLiteral dictionaryLiteral) {
     return dictionaryLiteral.elements().stream()
       .filter(KeyValuePair.class::isInstance)
       .map(KeyValuePair.class::cast)
       .map(KeyValuePair::key)
```
maksim-grebeniuk-sonarsource, line 122: need to validate `KeyValuePair`'s value as well that it is string literal
Sample test case is:
```python
def foo(x):
  app.config.update({"SECRET_KEY": x}) # FP here
```

python-checks/src/main/java/org/sonar/python/checks/FlaskHardCodedSecret.java
```diff
@@ -117,40 +115,40 @@ private static boolean isCallToDictConstructor(CallExpression callExpression) {
...
-  private static void verifyAssignmentStatement(SubscriptionContext ctx) {
+  private void verifyAssignmentStatement(SubscriptionContext ctx) {
     AssignmentStatement assignmentStatementTree = (AssignmentStatement) ctx.syntaxNode();
     List<Expression> expressionList = assignmentStatementTree.lhsExpressions().stream()
       .map(ExpressionList::expressions)
       .filter(list -> list.size() == 1)
       .flatMap(List::stream)
-      .filter(FlaskHardCodedSecretCheck::isSensitiveProperty)
+      .filter(this::isSensitiveProperty)
       .filter(expression -> isStringLiteral(assignmentStatementTree.assignedValue()))
```
maksim-grebeniuk-sonarsource, line 143: move this check as an `if` condition in the method beginning:
```java
if (!isStringLiteral(assignmentStatementTree.assignedValue())) {
      return;
    }
```

python-checks/src/main/java/org/sonar/python/checks/FlaskHardCodedSecret.java
```diff
@@ -162,16 +160,16 @@ private static boolean isSensitiveProperty(Expression expression) {
         .map(ExpressionList::expressions)
         .filter(list -> list.size() == 1)
         .map(list -> list.get(0))
-        .map(FlaskHardCodedSecretCheck::getAssignedValue)
+        .map(FlaskHardCodedSecret::getAssignedValue)
         .flatMap(TreeUtils.toOptionalInstanceOfMapper(StringLiteral.class))
         .map(StringLiteral::trimmedQuotesValue)
-        .filter(SECRET_KEY_KEYWORD::equals)
+        .filter(key -> this.getSecretKeyKeyword().equals(key))
```
maksim-grebeniuk-sonarsource, line 166: replace with:
```java
.filter(this.getSecretKeyKeyword()::equals)
```

python-checks/src/main/java/org/sonar/python/checks/FlaskHardCodedSecret.java
```diff
@@ -162,16 +160,16 @@ private static boolean isSensitiveProperty(Expression expression) {
...
         .isPresent();
     } else if (expression.is(Tree.Kind.QUALIFIED_EXPR)) {
```
maksim-grebeniuk-sonarsource, line 168: move check of qualified expression to exact implementation from the abstract class since it is needed just for particular rule

python-checks/src/main/java/org/sonar/python/checks/FlaskHardCodedSecret.java
jdodinh: Thanks for the tip. I improved this in d194803430d05b7ae392a3a54adb07f5d392cbd5.

python-checks/src/main/java/org/sonar/python/checks/FlaskHardCodedSecret.java
jdodinh: Done in 73be1ac3b25eb4b2c9dc41cf27f4f93aa2251763.

python-checks/src/main/java/org/sonar/python/checks/FlaskHardCodedSecret.java
jdodinh: Fixed in 76ae07f32865c40b90d2654034fcbbf935f06188.

python-checks/src/main/java/org/sonar/python/checks/FlaskHardCodedSecret.java
jdodinh: This is a very good catch, thank you. I made the change in bd8ee36efa8204a5c0bd4c0895860bd5494092f0.

python-checks/src/main/java/org/sonar/python/checks/FlaskHardCodedSecret.java
```diff
@@ -117,74 +114,84 @@ private static boolean isCallToDictConstructor(CallExpression callExpression) {
...
+      .anyMatch(this::isIllegalKeyValuePair);
+  }
+
+  private boolean isIllegalKeyValuePair(KeyValuePair keyValuePair) {
+    return Optional.of(keyValuePair.key())
       .filter(StringLiteral.class::isInstance)
       .map(StringLiteral.class::cast)
       .map(StringLiteral::trimmedQuotesValue)
-      .anyMatch(SECRET_KEY_KEYWORD::equals);
+      .filter(this.getSecretKeyKeyword()::equals)
```
maksim-grebeniuk-sonarsource, line 129: is `this` keyword needed here?

maksim-grebeniuk-sonarsource (approved):
LGTM, just one small comment.

PR #1640, opened 2023-11-09
SONARPY-1552 Rule S5332: Raise an issue on `server_bind` calls.

joke1196:


python-checks/src/main/java/org/sonar/python/checks/hotspots/ClearTextProtocolsCheck.java
```diff
@@ -89,24 +88,26 @@ public void initialize(Context context) {
...
+    context.registerSyntaxNodeConsumer(Tree.Kind.CALL_EXPR, ClearTextProtocolsCheck::checkServerBindCalls);
+
     new ClearTextProtocolsCheckPart().initialize(context);
   }
 
   private static void checkServerCallFromSuper(SubscriptionContext ctx) {
     QualifiedExpression qualifiedExpression = (QualifiedExpression) ctx.syntaxNode();
     Optional.of(qualifiedExpression)
-      .filter(qe -> isName("serve_forever", qe.name()) && isCallToSensitiveSuperClass(qe))
+      .filter(qe -> (isName("serve_forever", qe.name()) || isName("server_bind", qe.name())) && isCallToSensitiveSuperClass(qe))
```
maksim-grebeniuk-sonarsource, line 99: create Set of names and check for `contains`

python-checks/src/main/java/org/sonar/python/checks/hotspots/ClearTextProtocolsCheck.java
```diff
@@ -57,10 +56,10 @@ public class ClearTextProtocolsCheck extends PythonSubscriptionCheck {
   private static final List<String> SENSITIVE_PROTOCOLS = Arrays.asList("http://", "ftp://", "telnet://");
   private static final Pattern LOOPBACK = Pattern.compile("localhost|127(?:\\.[0-9]+){0,2}\\.[0-9]+$|^(?:0*\\:)*?:?0*1", Pattern.CASE_INSENSITIVE);
   private static final Map<String, String> ALTERNATIVES = new HashMap<>();
```
maksim-grebeniuk-sonarsource, line 58: Use `Map.of` for initialization

python-checks/src/main/java/org/sonar/python/checks/hotspots/ClearTextProtocolsCheck.java
```diff
@@ -89,28 +85,26 @@ public void initialize(Context context) {
...
+    context.registerSyntaxNodeConsumer(Tree.Kind.CALL_EXPR, ClearTextProtocolsCheck::checkServerBindCalls);
+
     new ClearTextProtocolsCheckPart().initialize(context);
   }
 
   private static void checkServerCallFromSuper(SubscriptionContext ctx) {
     QualifiedExpression qualifiedExpression = (QualifiedExpression) ctx.syntaxNode();
     Optional.of(qualifiedExpression)
-      .filter(qe -> isName("serve_forever", qe.name()) && isCallToSensitiveSuperClass(qe))
+      .filter(qe -> SENSITIVE_HTTP_SERVER_METHOD_NAMES.contains(qe.name().name()) && isCallToSensitiveSuperClass(qe))
```
jdodinh, line 96: Maybe here it would be worth splitting this into two separate filters:
```java
.filter(qe -> SENSITIVE_HTTP_SERVER_METHOD_NAMES.contains(qe.name().name()))
  .filter(ClearTextProtocolsCheck::isCallToSensitiveSuperClass)
```

python-checks/src/main/java/org/sonar/python/checks/hotspots/ClearTextProtocolsCheck.java
```diff
@@ -89,28 +85,26 @@ public void initialize(Context context) {
...
-      .ifPresent(fd -> ctx.addIssue(fd.defKeyword(), fd.rightPar(), message("http")));
-  }
-
-  private static boolean isName(String nameToCheck, Name name) {
-    return nameToCheck.equals(name.name());
+  private static void checkServerBindCalls(SubscriptionContext ctx) {
+    CallExpression callExpression = (CallExpression) ctx.syntaxNode();
+    Optional.ofNullable(callExpression.calleeSymbol())
+      .map(Symbol::fullyQualifiedName)
+      .filter(fqn -> SENSITIVE_HTTP_SERVER_BIND_FQN.equals(fqn) && isParentClassExtendingSensitiveClass(callExpression))
```
jdodinh, line 106: Here I would suggest just having:
```java
.filter(SENSITIVE_HTTP_SERVER_BIND_FQN::equals)
```
and have an if statement at the beginning of the method such as:
```java
if (!isParentClassExtendingSensitiveClass(callExpression)) {
  return;
}
```

jdodinh (approved):
LGTM! Just added a few comments, but everything seems right to me.

python-checks/src/main/java/org/sonar/python/checks/hotspots/ClearTextProtocolsCheck.java
joke1196: I would rather check first if the FQN matches. The check of the parent class can be costly as it traverses the graph. But I could modify it to have two filters.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1646, opened 2023-11-14
SONARPY-1535 Rule S4830: Add support for HTTPX

maksim-grebeniuk-sonarsource:


python-checks/src/test/resources/checks/verifiedSslTlsCertificateCheck.py
```diff
@@ -236,3 +236,12 @@ def call_expression_in_arguments():
     import ssl
     conn = httplib.client.HTTPSConnection("123.123.21.21", context=ssl._create_unverified_context()) # Noncompliant {{Enable server certificate validation on this SSL/TLS connection.}}
 #                                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^ 0
+
+def httpx_verify():
+    import httpx
+
+    httpx.get("https://expired.badssl.com/", verify=False)  # Noncompliant
+    httpx.get("https://expired.badssl.com/")  # Compliant
```
jdodinh, line 244: Perhaps it's an overkill, but why don't we include test cases for all the other function names?

python-checks/src/main/java/org/sonar/python/checks/VerifiedSslTlsCertificateCheck.java
```diff
@@ -165,29 +162,43 @@ private IssueReport(String message, Token token) {
...
+    Map.entry("httpx.patch", VERIFY_ARG_NAME),
+    Map.entry("httpx.delete", VERIFY_ARG_NAME),
+    Map.entry("httpx.Client", VERIFY_ARG_NAME),
+    Map.entry("httpx.AsyncClient", VERIFY_ARG_NAME)
+  );
 
   private static void requestsCheck(SubscriptionContext subscriptionContext) {
-    CallExpression callExpr = (CallExpression) subscriptionContext.syntaxNode();
-    boolean isVulnerableMethod = ofNullable(callExpr.calleeSymbol())
+    var callExpr = (CallExpression) subscriptionContext.syntaxNode();
```
jdodinh, line 192: Could you maybe clarify why you choose to use `var` is this situation? I'm genuinely curious.

jdodinh (approved):
LGTM overall implementation seems correct. I noted some details regarding the naming, as I think in some cases it could be more clear. Some of it doesn't even apply to your implementation though.

python-checks/src/main/java/org/sonar/python/checks/VerifiedSslTlsCertificateCheck.java
maksim-grebeniuk-sonarsource: I generally prefer `var` since it allows you to write shorter lines.

python-checks/src/main/java/org/sonar/python/checks/VerifiedSslTlsCertificateCheck.java
```diff
@@ -209,12 +220,12 @@ private static void requestsCheck(SubscriptionContext subscriptionContext) {
    *
    * @return The <code>expr</code> part on the right hand side of the assignment.
    */
-  private static Optional<List<Expression>> searchVerifyAssignment(CallExpression callExpr) {
+  private static Optional<List<Expression>> searchVerifyAssignment(CallExpression callExpr, Set<String> argumentNames) {
     for (Argument a : callExpr.arguments()) {
       if (a.is(Tree.Kind.REGULAR_ARGUMENT)) {
```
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: good suggestion, I didn't try to fix existing code here but sure I will

PR #1648, opened 2023-11-15
SONARPY-1545 Rule S5659: Add support for python-jose

jdodinh:


python-checks/src/main/java/org/sonar/python/checks/JwtVerificationCheck.java
```diff
@@ -45,26 +55,107 @@ public class JwtVerificationCheck extends PythonSubscriptionCheck {
     "python_jwt.verify_jwt",
     "jwt.verify_jwt"));
 
+  private static final Set<String> FQNS_WHERE_VERIFY_KWARG_SHOULD_BE_TRUE = new HashSet<>(Arrays.asList(
```
maksim-grebeniuk-sonarsource, line 58: Use `Set.of(...)` method isnstead

python-checks/src/main/java/org/sonar/python/checks/JwtVerificationCheck.java
```diff
@@ -45,26 +55,107 @@ public class JwtVerificationCheck extends PythonSubscriptionCheck {
...
+    "jwt.decode",
+    "jose.jws.verify"));
+
+
+  private static final Set<String> UNVERIFIED_FQNS = new HashSet<>(Arrays.asList(
```
maksim-grebeniuk-sonarsource, line 63: Use `Set.of(...)` method isnstead

python-checks/src/main/java/org/sonar/python/checks/JwtVerificationCheck.java
maksim-grebeniuk-sonarsource, line 58: better to move `FQNS` to the end and rename other contacts as well do get rid of mix of `FQNS` and `NAMES

python-checks/src/main/java/org/sonar/python/checks/JwtVerificationCheck.java
```diff
@@ -45,26 +55,107 @@ public class JwtVerificationCheck extends PythonSubscriptionCheck {
...
-          ctx.addIssue(verifyArg, MESSAGE);
-        }
-      } else if (PROCESS_JWT_NAMES.contains(calleeSymbol.fullyQualifiedName())) {
-        Tree scriptOrFunction = TreeUtils.firstAncestorOfKind(call, Kind.FILE_INPUT, Kind.FUNCDEF);
-        if (!TreeUtils.hasDescendant(scriptOrFunction, JwtVerificationCheck::isCallToVerifyJwt)) {
-          ctx.addIssue(call, MESSAGE);
-        }
+    } else if (UNVERIFIED_FQNS.contains(calleeSymbol.fullyQualifiedName())) {
+      ctx.addIssue(call, MESSAGE);
+    } else if ("jose.jwt.decode".equals(calleeSymbol.fullyQualifiedName())) {
```
maksim-grebeniuk-sonarsource, line 97: extend to constant as for previous names

python-checks/src/main/java/org/sonar/python/checks/JwtVerificationCheck.java
```diff
@@ -38,39 +46,122 @@ public class JwtVerificationCheck extends PythonSubscriptionCheck {
...
       }
-    });
+    }
+  }
+
+  private static boolean isListOrDictWithSensitiveEntry(@Nullable Expression expression) {
+    if (expression == null) {
+      return false;
+    } else if (expression.is(Tree.Kind.NAME)) {
+      return isListOrDictWithSensitiveEntry(Expressions.singleAssignedValue((Name) expression));
```
guillaume-dequenne-sonarsource, line 109: Remembering the StackOverflowException we investigated recently, could this trigger one as well in presence of circular assignments? As far as I remember, we didn't fix the underlying cause in `Expression#singleAssignedValue`?

python-checks/src/main/java/org/sonar/python/checks/JwtVerificationCheck.java
```diff
@@ -38,39 +46,122 @@ public class JwtVerificationCheck extends PythonSubscriptionCheck {
...
+      .map(list -> list.get(1))
+      .anyMatch(Expressions::isFalsy);
+  }
+
+  private static boolean isSensitiveKey(Expression key) {
+    return key.is(Kind.STRING_LITERAL) && VERIFY_SIGNATURE_KEYWORD.equals(((StringLiteral) key).trimmedQuotesValue());
   }
 
   private static boolean isCallToVerifyJwt(Tree t) {
     if (t.is(Kind.CALL_EXPR)) {
```
maksim-grebeniuk-sonarsource, line 162: could be done like this:
```java
return TreeUtils.toOptionalInstanceOf(CallExpression.class, t)
      .map(CallExpression::calleeSymbol)
      .filter(VERIFY_JWT_FQNS::contains)
      .isPresent();
```

python-checks/src/main/java/org/sonar/python/checks/JwtVerificationCheck.java
jdodinh: Yes, indeed I am trying to resolve variables in the same way. I have added the test case that causes the infinite recursion in 46c635a1b0b8eefaef0a39bb623382e0ae0f577e. However, in this case, it doesn't seem to break, and I am not entirely sure why. We can discuss this maybe.

guillaume-dequenne-sonarsource: In your test, the recursion is not triggered because it's only done in the `jose.jwt.decode` case. The example uses `jose.jws.verify` for which a simple falsy check on the `verify` argument value is done.

guillaume-dequenne-sonarsource: This case will trigger an infinite recursion:
```
x = []
class TestInfiniteRecursion():
    x = x
    payload = jwt.decode(token, None, options=x)
```

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1649, opened 2023-11-16
SONARPY-1537 Rule S4830: Add support for aiohttp

maksim-grebeniuk-sonarsource:


python-checks/src/test/resources/checks/verifiedSslTlsCertificateCheck.py
```diff
@@ -270,3 +270,9 @@ def httpx_verify():
     insecure_client = httpx.AsyncClient(verify=False)  # Noncompliant
     secure_client = httpx.AsyncClient(verify=True)  # Compliant
 
+async def aiohttp_test():
+    import aiohttp
+    async with aiohttp.ClientSession() as session:
+        session.request('GET', "https://expired.badssl.com/", verify_ssl=False) # Noncompliant
+        session.request('GET', "https://expired.badssl.com/", ssl=False) # Noncompliant
+        session.request('GET', "https://expired.badssl.com/") # Compliant
```
jdodinh, line 278: I added the following test case:
```java
session.request('GET', "https://expired.badssl.com/", verify_ssl=True, ssl=False) # Noncompliant
```
Which seems to be failing. Can you maybe check if this is expected? From the ticket, it seems to me that the issue should be raised when `verify_ssl` *or* `ssl` is set to `False`, which I think is the case when at least one is `False`. I also checked, and it seems like both parameters can be specified. Let me know what you think.

jdodinh: I think perhaps having this set of test cases for each method would be nice. Maybe an overkill but probably worth is:
```java
session.request('GET', "https://expired.badssl.com/", verify_ssl=False) # Noncompliant 
  session.request('GET', "https://expired.badssl.com/", ssl=False) # Noncompliant
  session.request('GET', "https://expired.badssl.com/", verify_ssl=False, ssl=True) # Noncompliant
  session.request('GET', "https://expired.badssl.com/", verify_ssl=True, ssl=False) # Noncompliant
  session.request('GET', "https://expired.badssl.com/", verify_ssl=True, ssl=True) # OK
  session.request('GET', "https://expired.badssl.com/") # Compliant
```

jdodinh (approved):
LGTM! Thanks for adding these test cases.

PR #1650, opened 2023-11-17
SONARPY-1546 Rule S2092: Add support for FastAPI

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/hotspots/AbstractCookieFlagCheck.java
```diff
@@ -57,44 +64,66 @@ public void initialize(Context context) {
...
+        });
+    }
+  }
+
+  private Optional<MethodArgumentsToCheck> findMethodArgumentToCheck(CallExpression callExpression) {
+    return Optional.of(callExpression)
+      .map(AbstractCookieFlagCheck::getCallExpressionMethodFqn)
+      .map(methodArgumentsToCheckRegistry()::getByMethodFqn)
+      .or(() -> Optional.of(callExpression)
+          .map(AbstractCookieFlagCheck::getCallExpressionMethodName)
```
joke1196, line 96: Same as the previous comment

PR #1653, opened 2023-11-20
SONARPY-1588 Support declared types of class fields

guillaume-dequenne-sonarsource:


python-frontend/src/test/java/org/sonar/python/types/TypeInferenceTest.java
```diff
@@ -681,7 +681,19 @@ void user_defined_attributes() {
...
+  @Test
+  void user_defined_attributes_reassigned() {
+    assertThat(lastExpression(
+      "class Foo:",
+      "  attr: int",
+      "def f():",
+      "  e = Foo()",
+      "  e.attr = 'hello'",
+      "  e.attr"
+    ).type()).isEqualTo(DECL_INT);
```
guillaume-dequenne-sonarsource, line 696: We'll probably want to consider to support field sensitive type inference eventually, to correctly infer that the type is `STR` here.

However, this may come with its own set of challenges and is out of scope for this ticket. I feel like this case should be discussed in our next sprints about type inference.

python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -753,6 +754,26 @@ public void visitFunctionDef(FunctionDef functionDef) {
       super.visitFunctionDef(functionDef);
     }
 
+    @Override
+    public void visitAnnotatedAssignment(AnnotatedAssignment annotatedAssignment) {
```
joke1196, line 758: I am not sure I understand why we would need to do this visit in the ThirdPhase and not the Second one already? I am not sure I understand the purpose of the ThirdPhase could we have a quick chat about it?

joke1196 (approved):
LGTM! There are a few things I do not understand yet on the SymbolTable, but here everything looks good to me!

python-frontend/src/main/java/org/sonar/python/semantic/SymbolTableBuilder.java
```diff
@@ -753,6 +754,26 @@ public void visitFunctionDef(FunctionDef functionDef) {
...
+        Name variable = (Name) annotatedAssignment.variable();
+        TypeAnnotation annotation = annotatedAssignment.annotation();
+        Symbol symbol = variable.symbol();
+        Optional.ofNullable(symbol)
+          .ifPresent(s -> {
+            Expression expression = annotation.expression();
+            if (expression instanceof HasSymbol) {
+              Symbol typeSymbol = ((HasSymbol) expression).symbol();
+              if (typeSymbol != null) {
+                ((SymbolImpl) symbol).setInferredType(new DeclaredType(typeSymbol, Collections.emptyList()));
```
guillaume-dequenne-sonarsource, line 769: Note: after reviewing this [PR](https://github.com/SonarSource/sonar-python/pull/1654), I realize this is probably not the right way to create the `DeclaredType` and that I should probably use `InferredTypes#fromTypeAnnotation` instead.

joke1196 (approved):
LGTM!

jdodinh (approved):
LGTM!

PR #1654, opened 2023-11-21
SONARPY-1556 Parameters should properly support union type as declared type

joke1196:


python-frontend/src/test/java/org/sonar/python/types/InferredTypesTest.java
joke1196, line 1: Sorry about this it is unrelated to the PR. I will just add some more coverage.

python-frontend/src/test/java/org/sonar/python/types/InferredTypesTest.java
```diff
@@ -248,6 +250,28 @@ void test_none_annotation() {
...
+  @Test
+  void test_unknown_annotation() {
+    TypeAnnotation typeAnnotation = typeAnnotation(
+      "l : MyType"
+    );
+    assertThat(fromTypeshedTypeAnnotation(typeAnnotation)).isEqualTo(anyType());
+    InferredType declaredType = fromTypeAnnotation(typeAnnotation);
+    assertThat(declaredType).isInstanceOf(AnyType.class);
+  }
+
```
joke1196, line 274: Sorry about this. This is not related to the PR. It just raises the coverage.

python-frontend/src/test/java/org/sonar/python/types/InferredTypesTest.java
```diff
@@ -379,6 +459,10 @@ void test_type_from_protobuf() throws TextFormat.ParseException {
       "fully_qualified_name: \"mod.t\""
       ))
       .isEqualTo(InferredTypes.or(STR, INT));
+    assertThat(protobufType(
+      "kind: INSTANCE\n" +
+      "fully_qualified_name: \"typing._SpecialForm\""
+      )).isEqualTo(anyType());
```
joke1196, line 465: Sorry about this as well. This is not related to the PR. It just raises the coverage.

python-frontend/src/test/java/org/sonar/python/types/InferredTypesTest.java
```diff
@@ -282,6 +306,62 @@ void test_optional_type_annotations() {
...
+      "def f(x: int | A):",
+      "  x").type()).isEqualTo(anyType());
+
+    assertThat(lastExpression(
+      "def f(x: B | int):",
+      "  x").type()).isEqualTo(anyType());
+  }
+
+  @Test
+  void test_union_type_annotation_with_any_type() {
```
jdodinh, line 338: Maybe the title of this test is a little bit misleading, because we are not testing a uniton type here.

python-frontend/src/test/java/org/sonar/python/types/InferredTypesTest.java
```diff
@@ -282,6 +306,62 @@ void test_optional_type_annotations() {
...
+    assertThat(lastExpression(
+      "from typing import Any",
+      "def f(x: Any):",
+      "  x").type()).isEqualTo(anyType());
+  }
+
```
jdodinh, line 344: The following case seems to fail:
```java
  @Test
  void test_declared_type_outside_of_method() {
    DeclaredType union = new DeclaredType(new SymbolImpl("Union", "typing.Union"), List.of(new DeclaredType(BuiltinTypes.INT),
      new DeclaredType(BuiltinTypes.STR)));
    assertThat(lastExpression(
      "class Foo():",
      "  x: int | str",
      "  x").type()).isEqualTo(union);
  }
```
I am not entirely sure that it is within the scope of this ticket, but maybe it would be worth investigating?

jdodinh (changes requested):
LGTM, found a test case that fails, but not sure it needs to be addressed as part of this ticket.

python-frontend/src/test/java/org/sonar/python/types/InferredTypesTest.java
joke1196: Ok so I have taken a look how I could test this. It is a bit complicated as we cannot use `lastExpression`. Also as discussed this yet out of scope so for now we can leave it as is and come back to it later on if needed.

guillaume-dequenne-sonarsource: Probably the class attributes ticket would also be a blocker for this (although I'm not sure it would work in the current state of things once both tickets are merged).

A way to rewrite the test could make use of a separate function that instantiates the class and then accesses the attribute (similar to the tests in this [PR](https://github.com/SonarSource/sonar-python/pull/1653)) because currently the fact there is no function fails the test before it can actually test anything.

I think it makes sense to have such a test (either in this or in the other PR) to ensure we eventually cover this case though.

python-frontend/src/main/java/org/sonar/python/types/InferredTypes.java
```diff
@@ -235,9 +243,32 @@ private static DeclaredType declaredTypeFromTypeAnnotation(Expression expression
     if (expression.is(Kind.NONE)) {
       return new DeclaredType(builtinSymbols.get(BuiltinTypes.NONE_TYPE));
     }
+    if (expression.is(Kind.BITWISE_OR)) {
+      var binaryExpression = (BinaryExpression) expression;
+      var unionType = declaredUnionType(binaryExpression.leftOperand(), binaryExpression.rightOperand(), builtinSymbols);
+      if (unionType != null) {
+        return unionType;
+      }
```
guillaume-dequenne-sonarsource, line 251: You can probably just return unconditionally here, since the default is to return null.

guillaume-dequenne-sonarsource (approved):
Looks good!
Small comment about formatting.
About the test that combines class attributes and union types, please let me know if you decide to add it, otherwise I can do it from my PR.

python-frontend/src/main/java/org/sonar/python/types/InferredTypes.java
```diff
@@ -19,6 +19,8 @@
  */
 package org.sonar.python.types;
 
+import static org.sonar.plugins.python.api.symbols.Symbol.Kind.CLASS;
```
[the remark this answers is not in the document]
joke1196: Yes, I see I had to redo this already with the imports. I'll fix my configuration.

PR #1656, opened 2023-11-21
SONARPY-1286  Fix FP on S3457 when format specifier contains an equal sign.

jdodinh:


python-frontend/src/test/java/org/sonar/python/lexer/PythonLexerTest.java
```diff
@@ -326,6 +324,18 @@ void fstring_incorrect_walrus_operator() {
       hasToken("}", PythonPunctuator.RCURLYBRACE),
       hasToken("'", PythonTokenType.FSTRING_END)));
   }
+
+  @Test
+  void fstring_walrus_looking_format_specifier() {
```
guillaume-dequenne-sonarsource, line 329: I'm not sure this test was strictly needed, seeing it's quite similar to the one above it.

I think having tests ensuring the FP on rule S3457 is no longer raised (while the TP still is) might be more valuable: the underlying issue being already fixed and tested, we can just confirm that the rule works as intended and close the ticket.

python-checks/src/test/resources/checks/stringFormatCorrectness.py
```diff
@@ -1,5 +1,5 @@
 # Due to parsing errors with formatted expressions, the rule is disabled for f-strings: see SONARPY-726
```
guillaume-dequenne-sonarsource, line 1: I know this is not the scope of this PR, but could you take the opportunity to remove this comment?
It's been a while this is not relevant anymore.

guillaume-dequenne-sonarsource (approved):
LGTM.
If you can take the opportunity to remove the now misleading comment, that would be nice as well!

PR #1658, opened 2023-11-23
SONARPY-1010: Fix FP on S5607, S5864, S2159, S5644, S3862 when using mocks

joke1196:


joke1196:
Tell me if you think there is a better solution I could go for on this one. It seems to me that we can consider all the mocks to have any member, of course, this is a bit extreme. So if you see anything that I could break let me know!

python-checks/src/test/resources/checks/confusingTypeChecking/sillyEquality.py
```diff
@@ -24,3 +24,11 @@ def builtins(x: int, y: str):
 def foo(a: T):
   if a == 42: # FN
     ...
+
+
+from unittest.mock import Mock, MagicMock
+# We should not raise any issues on mocks as they could be monkey patched to fit the comparison type
+def mocks(mock: Mock, magic_mock:MagicMock):
```
guillaume-dequenne-sonarsource, line 31: _Nitpick:_ there are some formatting issues with the type annotation (missing whitespace), feel free to ignore.

python-frontend/src/main/java/org/sonar/python/types/DeclaredType.java
```diff
@@ -81,7 +81,10 @@ public boolean canHaveMember(String memberName) {
 
   @Override
   public boolean declaresMember(String memberName) {
-    if (hasUnresolvedHierarchy()) {
+    Symbol fetchedTypeClass = getTypeClass();
+    if (hasUnresolvedHierarchy() ||
+      (fetchedTypeClass.fullyQualifiedName() != null &&
+        EXCLUDED_FQNS.contains(fetchedTypeClass.fullyQualifiedName()))) {
```
guillaume-dequenne-sonarsource, line 87: I think the following
`EXCLUDED_FQNS.stream().anyMatch(this::mustBeOrExtend)`
would also include any subtypes of `Mock` and `MagicMock`, which could avoid other FPs for subtypes of mocks, such as:

```
class MyClassForMock(MagicMock):
    ...

def my_custom_mock(param: MyClassForMock):
    param + 3
    -param
```

python-frontend/src/main/java/org/sonar/python/types/RuntimeType.java
```diff
@@ -57,7 +57,12 @@ public boolean isIdentityComparableWith(InferredType other) {
 
   @Override
   public boolean canHaveMember(String memberName) {
-    return getTypeClass().canHaveMember(memberName);
+    ClassSymbol fetchedTypedClass = getTypeClass();
+    if (fetchedTypedClass.fullyQualifiedName() != null && 
+        EXCLUDED_FQNS.contains(fetchedTypedClass.fullyQualifiedName())){
```
guillaume-dequenne-sonarsource, line 62: Same as for `DeclaredType`, we can have tests such as:
```


class MyClassForMock(MagicMock):
    ...

def my_custom_mock():
    my_mock = MyClassForMock()
    my_mock + 3
    -my_mock
```

guillaume-dequenne-sonarsource (changes requested):
Looks good!

I think the approach makes sense as it clearly reflects our assumption that mocks can have any member. Also, I don't think the fact us making this assumption could lead to extra issues.
I appreciate the way commits are organized, this made reviewing the PR quite easy, thanks for that!

I left some suggestions, nothing major though.

python-frontend/src/main/java/org/sonar/python/types/DeclaredType.java
joke1196: Thank you for the suggestion this really cleans up the code!

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1659, opened 2023-11-23
SONARPY-1256 Fix FP on S5886 when Iterator is an ambiguous symbol

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/FunctionReturnTypeCheck.java
```diff
@@ -124,6 +128,22 @@ private static void raiseIssues(SubscriptionContext ctx, FunctionDef functionDef
     });
   }
 
+  private static boolean canBeOrExtendIterableType(InferredType inferredType) {
+    return Stream.concat(ITERABLE_TYPES.stream(), ASYNC_ITERABLE_TYPES.stream())
+      .anyMatch(typeName -> canBeOrExtend(inferredType, typeName));
+  }
+
+  private static Boolean canBeOrExtend(InferredType inferredType, String typeName) {
```
guillaume-dequenne-sonarsource, line 136: Do we need to use the boxed type here?

Also I'm a bit uncomfortable with the naming here, or at least with the fact that we still have a `DeclaredType#canBeOrExtend` method with a different implementation.

It would be nice to have a comment explaining that in this case, we consider the declared type of the type annotation to be as trustworthy as if it were a runtime type (which really is the reason for the different implementation).

Maybe a different name would be nice as well, but I know naming is all over the place for these methods and I don't have a good suggestion right now.

guillaume-dequenne-sonarsource (changes requested):
The fix looks good! Thanks for handling this!

I left some comments, mostly to improve tests, to make sure we cover all the known FPs that are related to this.

python-checks/src/test/resources/checks/functionReturnTypeWithConditionalImport.py
```diff
@@ -0,0 +1,8 @@
+import sys
```
guillaume-dequenne-sonarsource, line 1: I think all this could go into a function in the other test file, so we keep only one extra test file (if you decide to do this, it may make sense to rename the test file then to broaden it from just `collections.abc`.

guillaume-dequenne-sonarsource (approved):
LGTM.
Minor suggestion about test but feel free to ignore.

PR #1660, opened 2023-11-24
SONARPY-1565 [S6786] should be raised on GraphQLView.as_view() when the list of middlewares is passed through a variable

jdodinh:


python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
```diff
@@ -139,6 +149,13 @@ private static Optional<String> nameFromIdentifierOrCallExpression(Expression ex
     return Optional.ofNullable(TreeUtils.nameFromExpression(expression))
       .or(() -> TreeUtils.toOptionalInstanceOf(CallExpression.class, expression)
         .map(CallExpression::callee)
-        .map(TreeUtils::nameFromExpression));
+        .flatMap(GraphQLIntrospectionCheck::nameFromExpressionOrQualifiedExpression));
+  }
+
+  private static Optional<String> nameFromExpressionOrQualifiedExpression(Expression expression) {
+    return Optional.of(expression)
+      .flatMap(TreeUtils.toOptionalInstanceOfMapper(QualifiedExpression.class))
```
joke1196, line 157: Here you could use `TreeUtils.toOptionalInstanceOf()`

python-checks/src/test/resources/checks/hotspots/graphQLIntrospection.py
```diff
@@ -296,9 +296,51 @@ def graphql_server_validation_rules_compliant(schema, some_rule, introspection_r
...
+
+    from a_module import some_function
+    some_var = some_function()
+    GraphQLView.as_view( # OK
+        'graphiql',
+        schema=schema,
+        backend=CustomBackend(),
+        graphiql=True,
+        middleware=some_var,
+        format_error=format_custom_error
```
joke1196, line 345: I think this could be added to the other functions. This way we can keep the same grouping with compliant/noncompliant and stay a bit more consistant.

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
```diff
@@ -139,6 +149,13 @@ private static Optional<String> nameFromIdentifierOrCallExpression(Expression ex
     return Optional.ofNullable(TreeUtils.nameFromExpression(expression))
       .or(() -> TreeUtils.toOptionalInstanceOf(CallExpression.class, expression)
         .map(CallExpression::callee)
-        .map(TreeUtils::nameFromExpression));
+        .flatMap(GraphQLIntrospectionCheck::nameFromExpressionOrQualifiedExpression));
```
joke1196, line 152: I was wondering if this means that this would not work with 
```python
    gql_middlew = [
        middleware.IntrospectionMiddleware,
    ]

    GraphQLView.as_view( # OK
        'graphql',
        schema=schema,
        middleware=gql_middlew,
        backend=CustomBackend(),
        batch=True
    )
```
as the check for qualified expression is only done when encountering a CallExpression. Maybe I misunderstood this part of the code. Tell me if I am wrong.

Also this makes me think about the first implementation that could be missing the case where a Qualified expression is provided?

```python
    GraphQLView.as_view( # FN? 
        'graphql',
        schema=schema,
        middleware=(some_module.unsafemiddleware,),
        backend=CustomBackend(),
        batch=True
    )
```
Not sure about any of this but it could be nice to check. Also it is not part of this ticket and if it is too cumbersome to implement there is no need to do it.

python-checks/src/test/resources/checks/hotspots/graphQLIntrospection.py
jdodinh: Yes, I was thinking of doing this. I guess I just wanted the diff of this PR to be clean regarding the test file. I will make the change.

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
jdodinh: I will investigate this. Thanks for the tip. The first case at least should be possible without too much refactoring.

jdodinh: I pushed some changes to support qualified expressions as part of the elements of the list. The changes are in this commit: cf9f1a407d2afa9522d9ebc276a1fe59be296457.

python-checks/src/test/resources/checks/hotspots/graphQLIntrospection.py
jdodinh: I reorganized the test file in 985c67818094d9de1507b0021e12c89900ca2790.

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
jdodinh: Done in c3c9a320bf6afa8ea03d9bc9a8446be257743ef4.

joke1196 (approved):
LGTM! Thanks for updating the old logic!

PR #1663, opened 2023-11-27
SONARPY-1570 [S5332] Fix FN when calls are made to  from within inherited classes.

jdodinh:


python-checks/src/test/resources/checks/hotspots/clearTextProtocols.py
```diff
@@ -181,34 +181,9 @@ def python_web_server_compliant(ok_server):
...
-
-    my_server = MyServer()
-    my_server.serve_forever()
-    my_server.server_bind()
-
-    my_server = HTTPServer()
-    my_server.serve_forever()
-    my_server.server_bind()
+            self.socket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+            http.server.HTTPServer.server_bind(self) # Noncompliant
```
joke1196, line 189: I think we should keep the compliant examples here in the test cases. Also we could remove the line `self.socket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)` there is not much use for it.

python-checks/src/main/java/org/sonar/python/checks/hotspots/ClearTextProtocolsCheck.java
```diff
@@ -134,9 +134,9 @@ public static Stream<String> getClassFQNFromArgument(List<Argument> arguments) {
       .map(TreeUtils.toInstanceOfMapper(RegularArgument.class))
```
joke1196, line 134: Small detail as it is not part of this PR but as you mentioned the access modifier is incorrect on this method. Maybe could quickly correct it.

joke1196 (changes requested):
Looks good to me! I think we removed a few important test. We could maybe add the new test case not at the bottom of the file but maybe inside of `def python_web_server_noncompliant():` WDYT?

python-checks/src/test/resources/checks/hotspots/clearTextProtocols.py
```diff
@@ -202,8 +207,7 @@ class MyServer(HTTPServer):
         def run(self):
             super(self).serve_forever()  # Compliant
 
-        def server_bind(self):
-            HTTPServer.server_bind(self)
+    HTTPServer.server_bind(self)
```
joke1196, line 211: I think we can keep this test case as before.

joke1196 (approved):
LGTM! I have just added a comment on a test case that was not reverted properly.

PR #1664, opened 2023-11-27
SONARPY-1247 Fix inconsistencies in test rule results when test dirs are configured

guillaume-dequenne-sonarsource:
It is worth noting that failing to parse test files will now fail the analysis when `failFast=true`. I don't think this is a big issue, as this is supposed to be an internal property, but this might impact our Peach jobs if some projects contain invalid syntax in their test files. In which case we might want to explicitly exclude parse errors on test files from failing the analysis.

### DIFF
--- a/its/plugin/it-python-plugin-test/projects/test_code/tests/test_my_code.py
+++ b/its/plugin/it-python-plugin-test/projects/test_code/tests/test_my_code.py
@@ -12,3 +12,10 @@ from tests.my_other_test_file import MyTestClass
 class SomeCustomTestCase(MyTestClass):
         def test_addition(self):
             self.assertTrue(1 + 2 == 3)  # Issue here (S5906)
+
+
+from yest_another_test_file import MyOtherTestClass
+
+class SomeOtherCustomTestCase(MyTestClass):
+        def test_addition_2(self):
+            self.assertTrue(1 + 2 == 3)  # Issue here (S5906)
--- a/its/plugin/it-python-plugin-test/projects/test_code/tests/test_my_code.py
+++ b/its/plugin/it-python-plugin-test/projects/test_code/tests/test_my_code.py
@@ -14,8 +14,8 @@ class SomeCustomTestCase(MyTestClass):
             self.assertTrue(1 + 2 == 3)  # Issue here (S5906)
 
 
-from yest_another_test_file import MyOtherTestClass
+from yet_another_test_file import MyOtherTestClass
 
-class SomeOtherCustomTestCase(MyTestClass):
+class SomeOtherCustomTestCase(MyOtherTestClass):
         def test_addition_2(self):
-            self.assertTrue(1 + 2 == 3)  # Issue here (S5906)
+            self.assertTrue(1 + 2 == 3)  # FN here (S5906)
--- a/its/plugin/it-python-plugin-test/projects/test_code/tests/test_my_code.py
+++ b/its/plugin/it-python-plugin-test/projects/test_code/tests/test_my_code.py
@@ -5,3 +5,10 @@ def test_something():
       print("hello")
     else:
       print("hello")
+
+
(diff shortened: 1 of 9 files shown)


its/ruling/src/test/java/org/sonar/python/it/PythonExtendedRulingTest.java
```diff
@@ -72,7 +72,8 @@ void test_black() throws IOException {
     SonarScanner build = buildWithCommonProperties("black");
     build.setProperty("sonar.sources", "src");
     build.setProperty("sonar.tests", "tests");
-    build.setProperty("sonar.test.exclusions", "tests/data/async_as_identifier.py");
+    build.setProperty("sonar.test.exclusions", "tests/data/async_as_identifier.py, tests/data/tricky_unicode_symbols.py," +
+      "tests/data/pattern_matching_extras.py,tests/data/pep_572_py39.py,tests/data/pattern_matching_invalid.py");
```
guillaume-dequenne-sonarsource, line 76: At least some of these seem to indicate valid syntax that we fail to parse.
I'm currently investigating them and will create relevant tickets for the cases that make sense.

its/plugin/it-python-plugin-test/projects/test_code/tests/test_my_code.py
```diff
@@ -12,3 +12,10 @@ def test_something():
 class SomeCustomTestCase(MyTestClass):
         def test_addition(self):
             self.assertTrue(1 + 2 == 3)  # Issue here (S5906)
+
+
+from yest_another_test_file import MyOtherTestClass
```
joke1196, line 17: There is a small typo here for the imported module `yest` instead of `yet`

sonar-python-plugin/src/test/java/org/sonar/plugins/python/indexer/SonarQubePythonIndexerTest.java
```diff
@@ -351,11 +351,11 @@ void test_test_files_not_using_cache() throws IOException, NoSuchAlgorithmExcept
     pythonIndexer = new SonarQubePythonIndexer(inputFiles, cacheContext, context);
     pythonIndexer.buildOnce(context);
 
-    assertThat(pythonIndexer.canBePartiallyScannedWithoutParsing(file1)).isTrue();
+    assertThat(pythonIndexer.canBePartiallyScannedWithoutParsing(file1)).isFalse();
```
joke1196, line 354: I am not sure I understand this part. So now that we gather everything together, test and main code, the test files are already cached?

joke1196 (approved):
LGTM! There is just a small typo, so nothing major/ Thank you for taking care of this!

its/plugin/it-python-plugin-test/projects/test_code/tests/test_my_code.py
guillaume-dequenne-sonarsource: Good catch!
It's unfortunately more than just a typo as the issue actually isn't found when the import is correctly made (and the class below uses the correct parent).
The initial goal of this test was to document this possible limitation, and I was surprised to see it succeed. This makes more sense now.

its/ruling/src/test/java/org/sonar/python/it/PythonExtendedRulingTest.java
guillaume-dequenne-sonarsource: SONARPY-1572 and SONARPY-1573 have been created to handle some unicode characters we fail to parse, as well as to cover unparenthesized walrus in set literals/comprehensions.

I didn't manage to create a reproducer for the `pattern_matching_extras` case that is both valid Python and that we fail to parse (using the offending snippet verbatim raises an invalid syntax error in Python 3.11), so I didn't create a ticket for it.

joke1196 (approved):
LGTM!

PR #1665, opened 2023-11-28
SONARPY-1166 Avoid FPs when variables are used inside 'pandas.DataFrame.query' expression argument

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource (approved):
Looks good! Thanks for the refactoring and for the clean commits.
I left very minor comments/suggestions.

PR #1667, opened 2023-11-29
SONARPY-1251 Fix FP on S1940 Inverted boolean checks when comparing Sets.

jdodinh:


python-checks/src/test/resources/checks/booleanCheckNotInverted.py
```diff
@@ -68,3 +68,23 @@ def func2():
...
+def set_comparison_test():
+    from typing import Set
+    def validate(a: Set[str], b: Set[str]) -> None:
+        if not (a <= b): # OK
+            ...
+
+    ingredients: Set[str] = {'beef', 'potatos'}
+    consumable_by_vegans: Set[str] = {'nut', 'apple', 'potatos'}
+    # The example below is a FP. The type of both sets is inferred as ANY. Should be fixed as part of:
+    # https://sonarsource.atlassian.net/browse/SONARPY-1574
```
guillaume-dequenne-sonarsource, line 81: Generally, simply the ticket reference is enough, no need for the full link (no need to change though).

python-checks/src/test/resources/checks/booleanCheckNotInverted.py
```diff
@@ -68,3 +68,23 @@ def func2():
...
+    if not (ingredients <= consumable_by_vegans): # Noncompliant
+        ...
+
+    ingredients_not_annotated = {'beef', 'potato'}
+    consumable_by_vegans_not_annotated = {'nut', 'apple', 'potato'}
+    if not (ingredients_not_annotated < consumable_by_vegans_not_annotated):       # OK
+        ...
+
+    validate(ingredients, consumable_by_vegans)
```
guillaume-dequenne-sonarsource, line 90: Is this needed? I have the feeling that since the rule doesn't involve symbolic validation, this may be misleading to have this.

python-checks/src/test/resources/checks/booleanCheckNotInverted.py
```diff
@@ -68,3 +68,23 @@ def func2():
 #                 ^^^^^^^^^^
         b = 10
     return "item1" "item2"
+
+def set_comparison_test():
```
guillaume-dequenne-sonarsource, line 72: I think it would be nice to have some more tests highlighting the assumptions we make: 
* Comparing a known set with ANY type won't raise an issue
* Comparing two ANY types will raise an issue (I know the one related to SONARPY-1574 does that, but only by accident)
* Subtypes of `set` won't raise an issue as well.
* We can also document that we may still have FPs in cases such as (and maybe see if we can fix them without losing too many TPs):
```
def foo(a, cond):
  if cond:
    b = {'nut', 'apple', 'potatos'}
  else:
    b = "hello"
  if not (a <= b): ...
```
* And there may be some other cases like this that could be documented.

guillaume-dequenne-sonarsource (changes requested):
The approach looks good to avoid losing too many TPs while removing some obvious FPs.
I think we need some more tests to make our assumptions explicit though.

python-checks/src/test/resources/checks/booleanCheckNotInverted.py
jdodinh: I added a few test cases in 4f00f494229181d05dd8bb4a811b7777636c400d.

guillaume-dequenne-sonarsource (approved):
Looks good!
Left a few comments to ensure our tests are clear and intentional.

python-checks/src/test/resources/checks/booleanCheckNotInverted.py
```diff
@@ -51,20 +53,77 @@
...
+class CustomSet(set):
+    ...
+
+
+def check_set_as_custom_subclass(a: CustomSet, b: CustomSet) -> None:
+    if not (a <= b): # OK
+        ...
+
+
+def check_set_as_type_alias(a: A, b: A) -> None:
```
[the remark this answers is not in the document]
jdodinh: Yes, thanks. I forgot to move it after I put all the function definitions at the same level. Thanks for catching this!

python-checks/src/test/resources/checks/booleanCheckNotInverted.py
```diff
@@ -76,7 +76,7 @@ def func2():
 
 
 def check_set_as_declared_type(a: Set[str], b: Set[str]) -> None:
-    c: Set[str] = {'set_element'} # c will have type ANY
+    c: Any = "something" # c will have type ANY
```
guillaume-dequenne-sonarsource, line 79: I don't think this change is the best we could have, as the same bug will impact it again.
I'd suggest something like:
```
c = unknown_func() # c will have type ANY
```
Then it's really clear that we can't know anything about `c`. `unknown_func` doesn't even have to be defined in the scope.

PR #1668, opened 2023-11-29
SONARPY-1561: Infer types of variables used as decorators.

joke1196:


python-frontend/src/test/java/org/sonar/python/types/TypeInferenceTest.java
```diff
@@ -723,4 +729,36 @@ void user_defined_attributes_reassigned() {
...
+
+    decorator = lastDecorator(
+      "class A:",
+      "  def __call__():",
+      "    ...",
+      "@A",
+      "class OtherClass:",
+      "  ...");
+    Name name = (Name) decorator.expression();
+    assertThat(name.type()).isEqualTo(anyType());
```
joke1196, line 756: Not exactly sure which type `@A` should be here. I would think for now `Any` is fine.

joke1196:
Here the approach I took was to add the decorators to the CFG, this way the type inference will treat them as any other statement. As I am not 100% sure what kind of information the CFG should contain, I may be missing tests, so feel free to challenge.

python-frontend/src/test/java/org/sonar/plugins/python/api/cfg/ControlFlowGraphTest.java
```diff
@@ -706,6 +707,34 @@ void if_stmt_test_predecessors() {
       "after_if(succ = [END], pred = [before, if_body])");
   }
 
+  @Test
+  void decorators() {
+    ControlFlowGraph cfg = cfg("" +
+      "class Dec:",
```
jdodinh, line 713: I don't quite understand why we prepend the first element with the empty string. Is this intentional?

jdodinh (changes requested):
Looks good to me. Just added a few minor points.

python-frontend/src/test/java/org/sonar/python/types/TypeInferenceTest.java
```diff
@@ -723,4 +729,36 @@ void user_defined_attributes_reassigned() {
...
+    assertThat(name.symbol().fullyQualifiedName()).isEqualTo("package.mod.A");
+  }
+
+  private static Decorator lastDecorator(String... code) {
+    FileInput fileInput = parse(new SymbolTableBuilder("package", pythonFile("mod")), code);
```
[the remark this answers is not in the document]
joke1196: Sure so they don't influence the outcome of the test but they do influence the fully qualified name of the Symbols

python-frontend/src/main/java/org/sonar/python/cfg/ControlFlowGraphBuilder.java
```diff
@@ -186,13 +183,28 @@ private PythonCfgBlock build(Statement statement, PythonCfgBlock currentBlock) {
...
+      case FUNCDEF:
+        return buildFuncDefStatement((FunctionDef) statement, currentBlock);
       default:
         currentBlock.addElement(statement);
     }
 
     return currentBlock;
   }
 
+  private PythonCfgBlock buildClassDefStatement(ClassDef classDef, PythonCfgBlock successor) {
```
guillaume-dequenne-sonarsource, line 195: Isn't this a bit misleading to have the parameter called `successor`?
It seems it is the `currentBlock` that is passed as an argument, and this is indeed the one we want to add the elements to.

python-frontend/src/main/java/org/sonar/python/cfg/ControlFlowGraphBuilder.java
```diff
@@ -186,13 +183,28 @@ private PythonCfgBlock build(Statement statement, PythonCfgBlock currentBlock) {
...
+    PythonCfgBlock block = build(classDef.body().statements(), successor);
+    block.addElement(classDef.name()); // represents binding to class name
+    classDef.decorators().stream().forEach(decorator -> successor.addElement(decorator.expression()));
```
guillaume-dequenne-sonarsource, line 198: I have the feeling that if we wanted to be correct, we probably should add the `Decorator` tree as a distinct element (rather than just its expression), since the fact that it's part of a decorator has influence on the control flow (the underlying decorator function will be called here), which will not be represented by just having the `Name`.

However I see this approximation is already done for the class/function definitions themselves (although here the impact is more limited since it really is the creation of a symbol without other unexpected side effect).

Depending on how tricky this is to change without negative impact, I think we can either make the change directly, create a ticket to make the change later or just accept this approximation.

python-frontend/src/test/java/org/sonar/python/types/TypeInferenceTest.java
guillaume-dequenne-sonarsource: I think `Any` is fair here.

Currently, classes and functions are distinguished at symbol level and no type is defined for them. I think we eventually should infer types `type` and `Callable` (or something of the kind) for such symbols, but this is out of the scope of this ticket.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1672, opened 2023-11-30
SONARPY-1251 S6542 should not raise an issue on overrides and overloads.

jdodinh:


python-checks/src/main/java/org/sonar/python/checks/UseOfAnyAsTypeHintCheck.java
```diff
@@ -42,14 +52,52 @@ public void initialize(Context context) {
 
   private static void checkForAnyInTypeHint(SubscriptionContext ctx) {
     TypeAnnotation typeAnnotation = (TypeAnnotation) ctx.syntaxNode();
-    if (isTypeAny(typeAnnotation)) {
-      ctx.addIssue(typeAnnotation.expression(), MESSAGE);
-    }
+    Optional.of(typeAnnotation)
+      .filter(UseOfAnyAsTypeHintCheck::isTypeAny)
+      .filter(Predicate.not(UseOfAnyAsTypeHintCheck::isAnnotatingArgumentToChildMethod))
+      .filter(Predicate.not(UseOfAnyAsTypeHintCheck::isUnannotatedOverride))
```
jdodinh, line 58: Feel free to remove the negation of the predicate here. For some reason I just really didn't want to have `isNot...` at the beginning of the method name.

its/ruling/src/test/resources/expected/python-S6542.json
```diff
@@ -103,68 +111,69 @@
...
-'project:mypy-0.782/test-data/stdlib-samples/3.2/pprint.py':[
+"project:mypy-0.782/test-data/stdlib-samples/3.2/pprint.py": [
 89,
 92,
-92,
+92
 ],
-'project:mypy-0.782/test-data/stdlib-samples/3.2/random.py':[
+"project:mypy-0.782/test-data/stdlib-samples/3.2/random.py": [
+100,
```
joke1196, line 141: This is a new issue detected

its/ruling/src/test/resources/expected/python-S6542.json
```diff
@@ -1,64 +1,72 @@
...
 202,
 205,
 208,
 211,
-253,
+253
+],
+"project:mypy-0.782/mypy/main.py": [
+266,
+266
```
joke1196, line 42: These two are also new issues.

python-checks/src/main/java/org/sonar/python/checks/UseOfAnyAsTypeHintCheck.java
```diff
@@ -32,24 +45,81 @@
...
   }
+
+  private static boolean hasFunctionOverrideOrOverloadDecorator(FunctionDef currentFunctionDef) {
+    return currentFunctionDef.decorators().stream()
+      .map(Decorator::expression)
+      .map(TreeUtils::fullyQualifiedNameFromExpression)
+      .anyMatch(OVERRIDE_FQNS::contains);
+  }
+
+  private static boolean isParentFunctionAnOverrideWithoutDecorators(FunctionDef currentMethodDef) {
```
guillaume-dequenne-sonarsource, line 86: Could we achieve the same using the `FunctionSymbolImpl#owner` method? Though I know it's annoying as it's not part of the public API.

Also, maybe this is too lenient, but we have the helper `SymbolUtils#canBeAnOverridingMethod` which could also be used here. I assume we'd lose many issues, as it is much more lenient, but that might prevent quite a few FPs when overriding methods from other libraries.

guillaume-dequenne-sonarsource (changes requested):
Looks good overall.

I think we shouldn't be so strict when checking for a match between the child and parents' parameters.

I also left a suggestion that would make the rule more lenient (assuming an override whenever the class hierarchy is unresolved), happy to discuss this.

python-checks/src/main/java/org/sonar/python/checks/UseOfAnyAsTypeHintCheck.java
joke1196: Yes this is great. Thank you!

python-checks/src/main/java/org/sonar/python/checks/UseOfAnyAsTypeHintCheck.java
```diff
@@ -20,36 +20,65 @@
...
```
guillaume-dequenne-sonarsource, line 78: Probably now the name `isParentFunctionAnOverrideWithoutDecorators` doesn't make so much sense anymore and could be simplified to `canParentFunctionBeAnOverride` or something like that.

guillaume-dequenne-sonarsource (approved):
Looks good! Minor renaming suggestion.

PR #1674, opened 2023-12-01
SONARPY-1563 Fix FPs on S1172 when the parameter is intentionally unused

guillaume-dequenne-sonarsource:
I decided to go for a less aggressive approach than initially  discussed (excluding all methods from this rule). Instead, I decided to:
* Exclude parameter names `event` and `context` that may be required by AWS Lambda.
* Exclude any parameter that is referenced in a string literal or a comment (thus excluding any case of documented unused parameter for overriding purpose as well as anything that may be used through a DSL).

Seeing the results on ruling, this will make us lose some legitimate issues, although I believe the benefit of raising fewer FPs is worth the cost. Despite this change, this rule might still prove problematic and we might want to consider a more aggressive approach later.

### DIFF
--- a/its/ruling/src/test/resources/expected/python-S1172.json
+++ b/its/ruling/src/test/resources/expected/python-S1172.json
@@ -1,46 +1,39 @@
 {
-'project:biopython/Bio/Alphabet/__init__.py':[
-236,
+"project:biopython/Bio/Blast/ParseBlastTable.py": [
+122
 ],
-'project:biopython/Bio/Blast/ParseBlastTable.py':[
-122,
-],
-'project:biopython/Bio/ExPASy/__init__.py':[
-66,
-],
-'project:biopython/Bio/GenBank/__init__.py':[
+"project:biopython/Bio/GenBank/__init__.py": [
 937,
 1062,
 1352,
 1604,
-1701,
+1701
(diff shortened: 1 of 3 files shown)


its/ruling/src/test/resources/expected/python-S1172.json
```diff
@@ -1,46 +1,39 @@
 {
```
guillaume-dequenne-sonarsource, line 1: Here the actual changes are mixed with the format change of the output json, sorry about that.

joke1196 (approved):
LGTM! I was not sure about the string `'Issues are not raised when the variable is mentioned in a comment related to the function'`. If it was a comment then we would not raise an issue because of the word `a` in the string and the unused parameter `a`, is that correct?

guillaume-dequenne-sonarsource:
> LGTM! I was not sure about the string `'Issues are not raised when the variable is mentioned in a comment related to the function'`. If it was a comment then we would not raise an issue because of the word `a` in the string and the unused parameter `a`, is that correct?

Yes, without it, the comment above that string literal would be attached to the tokens of the function definition. As that comment refers to the parameter `"a"` in the asserted issue message, it would prevent the issue from being raised.

PR #1676, opened 2023-12-04
SONARPY-1580: Update build task to JDK 17

guillaume-dequenne-sonarsource:


joke1196 (approved):
LGTM! Is there anything we can/should do about the warning `task "promote" depends on task "ws_scan", but their only_if conditions are different`?

PR #1677, opened 2023-12-04
SONAPY-1553 Rule S4507: Add support for GraphQL endpoints with GraphiQL (interactive GraphQL) enabled

maksim-grebeniuk-sonarsource:


python-checks/src/test/resources/checks/hotspots/debugMode/graphql.py
```diff
@@ -0,0 +1,42 @@
+from flask import Flask
```
guillaume-dequenne-sonarsource, line 1: For the sake of completeness, can you add some of the usual tests (e.g. when the `GraphQLView.as_view` call is assigned to a variable before being used, or even when it's not used at all. We could also have one where `True` is assigned to a variable, which won't raise the issue) so that the behavior of the rule is well documented?

PR #1680, opened 2023-12-05
SONARPY-1569 [S6779] Highlight on the line that actually contains the Flask secret

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/FlaskHardCodedSecret.java
```diff
@@ -84,9 +84,13 @@ private void verifyUpdateCallArgument(SubscriptionContext ctx, CallExpression ca
       .flatMap(TreeUtils.toOptionalInstanceOfMapper(RegularArgument.class))
       .map(RegularArgument::expression)
       .map(FlaskHardCodedSecret::getAssignedValue)
-      .filter(this::isIllegalDictArgument)
-      .ifPresent(expr -> ctx.addIssue(callExpression, String.format(MESSAGE, getSecretKeyType())));
+      .flatMap(this::getIllegalDictArgument)
+      .ifPresent(illegalArgument -> ctx.addIssue(illegalArgument, getMessage())
+        .secondary(callExpression.callee(), getMessage()));
```
guillaume-dequenne-sonarsource, line 89: Having the primary message on the secondary location probably wouldn't make sense.
What about `The secret is used in this call.` as secondary message on the call site?

python-checks/src/test/resources/checks/flaskHardCodedJWTSecretKey.py
```diff
@@ -38,30 +38,39 @@ def test_non_compliant_call_expressions():
     app = Flask(__name__)
 
     # Tests for "flask.app.Flask.config[.update]"
-    app.config.update(dict( # Noncompliant
```
guillaume-dequenne-sonarsource, line 41: Given we are changing issues locations and messages, it makes sense to have assertions not only on primary locations, but also on secondaries and on messages.

guillaume-dequenne-sonarsource (changes requested):
Looks good, I suggest changing secondary location message and add assertions on secondary locations and messages.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1681, opened 2023-12-05
SONARPY-1567 S5659: Add support for PyJWT

maksim-grebeniuk-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/JwtVerificationCheck.java
```diff
@@ -87,6 +88,11 @@ private static void verifyCallExpression(SubscriptionContext ctx) {
       RegularArgument verifyArg = TreeUtils.argumentByKeyword("verify", call.arguments());
       if (verifyArg != null && Expressions.isFalsy(verifyArg.expression())) {
         ctx.addIssue(verifyArg, MESSAGE);
+      } else {
+        Optional.ofNullable(TreeUtils.argumentByKeyword("options", call.arguments()))
```
guillaume-dequenne-sonarsource, line 92: Is this intentional that this will also work for `"jose.jws.verify"`? I don't see it mentioned explicitly in the ticket.
If yes, I think it would make sense to add a test for it. If not, probably we should exclude it?

guillaume-dequenne-sonarsource (commented):
Looks good! Minor question on the fact this will also cover `jose.jws.verify`.

python-checks/src/main/java/org/sonar/python/checks/JwtVerificationCheck.java
maksim-grebeniuk-sonarsource: Good catch

python-checks/src/main/java/org/sonar/python/checks/JwtVerificationCheck.java
```diff
@@ -80,31 +80,30 @@ public void initialize(Context context) {
...
+    } else if (PROCESS_JWT_FQNS.contains(calleeFqn)) {
       Optional.ofNullable(TreeUtils.firstAncestorOfKind(call, Kind.FILE_INPUT, Kind.FUNCDEF))
         .filter(scriptOrFunction -> !TreeUtils.hasDescendant(scriptOrFunction, JwtVerificationCheck::isCallToVerifyJwt))
         .ifPresent(scriptOrFunction -> ctx.addIssue(call, MESSAGE));
-    } else if (UNVERIFIED_FQNS.contains(calleeSymbol.fullyQualifiedName())) {
+    } else if (UNVERIFIED_FQNS.contains(calleeFqn)) {
       Optional.ofNullable(TreeUtils.nthArgumentOrKeyword(0, "", call.arguments()))
         .flatMap(TreeUtils.toOptionalInstanceOfMapper(RegularArgument.class))
         .map(RegularArgument::expression)
         .ifPresent(argument -> ctx.addIssue(argument, MESSAGE));
```
guillaume-dequenne-sonarsource, line 104: _Nitpick:_ unrelated to the change, but there's an extra whitespace here that we could remove.

python-checks/src/main/java/org/sonar/python/checks/JwtVerificationCheck.java
```diff
@@ -79,26 +80,30 @@ public void initialize(Context context) {
...
     if (calleeSymbol == null || calleeSymbol.fullyQualifiedName() == null) {
       return;
     }
-    if (WHERE_VERIFY_KWARG_SHOULD_BE_TRUE_FQNS.contains(calleeSymbol.fullyQualifiedName())) {
+
+    String calleeFqn = calleeSymbol.fullyQualifiedName();
+    if (WHERE_VERIFY_KWARG_SHOULD_BE_TRUE_FQNS.contains(calleeFqn)) {
       RegularArgument verifyArg = TreeUtils.argumentByKeyword("verify", call.arguments());
       if (verifyArg != null && Expressions.isFalsy(verifyArg.expression())) {
         ctx.addIssue(verifyArg, MESSAGE);
```
guillaume-dequenne-sonarsource, line 93: Now, there is the possibility of raising 2 issues on `jwt.decode` calls that fail to meet both conditions. Given we raise on the argument, we could keep this as-is, but I think we'd need a test to show this is intentional.

Otherwise, maybe we could return here after raising the first issue, so that we don't raise twice?

guillaume-dequenne-sonarsource (changes requested):
Looks good. I feel we need to either have a test that now shows we can raise two issues on the same call, or prevent it from happening (I don't have a strong preference on the solution though).

python-checks/src/main/java/org/sonar/python/checks/JwtVerificationCheck.java
maksim-grebeniuk-sonarsource: return added

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1682, opened 2023-12-05
Refactor TreeUtils#fullyQualifiedNameFromExpression to return Optional

guillaume-dequenne-sonarsource:
Due to the fact this API is often used within stream and can be error prone due to the fact it's nullable, I believe it makes sense to have it return an optional instead.

### DIFF
--- a/python-checks/src/main/java/org/sonar/python/checks/UseOfAnyAsTypeHintCheck.java
+++ b/python-checks/src/main/java/org/sonar/python/checks/UseOfAnyAsTypeHintCheck.java
@@ -65,7 +65,8 @@ public class UseOfAnyAsTypeHintCheck extends PythonSubscriptionCheck {
 
   private static boolean isTypeAny(@Nullable TypeAnnotation typeAnnotation) {
     return Optional.ofNullable(typeAnnotation)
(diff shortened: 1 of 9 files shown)


python-checks/src/main/java/org/sonar/python/checks/UseOfAnyAsTypeHintCheck.java
```diff
@@ -65,15 +65,16 @@ private static void checkForAnyInReturnTypeAndParameters(SubscriptionContext ctx
 
   private static boolean isTypeAny(@Nullable TypeAnnotation typeAnnotation) {
     return Optional.ofNullable(typeAnnotation)
-      .map(annotation -> "typing.Any".equals(TreeUtils.fullyQualifiedNameFromExpression(annotation.expression())))
+      .flatMap(a -> TreeUtils.fullyQualifiedNameFromExpression(a.expression()))
```
maksim-grebeniuk-sonarsource, line 68: could be
```java
.map(TypeAnnotation::expression)
.flatMap(TreeUtils:: fullyQualifiedNameFromExpression)
```

PR #1685, opened 2023-12-05
SONARPY-1583 Add sonarProduct() API to PythonInputFileContext

andreaguarino:


python-frontend/src/main/java/org/sonar/plugins/python/api/PythonInputFileContext.java
```diff
@@ -34,10 +35,20 @@ public class PythonInputFileContext {
...
+    this.workingDirectory = workingDirectory;
+    this.cacheContext = cacheContext;
+    this.sonarProduct = sonarProduct;
+  }
+
   public PythonInputFileContext(PythonFile pythonFile, @Nullable File workingDirectory, CacheContext cacheContext) {
     this.pythonFile = pythonFile;
     this.workingDirectory = workingDirectory;
     this.cacheContext = cacheContext;
+    this.sonarProduct = SonarProduct.SONARQUBE;
```
anton-haubner-sonarsource, line 51: Minor: As far as I can tell, this constructor is mostly just used in test contexts. However, in general, is it safe to just silently assume SonarQube as the product for a public constructor?

Maybe this property should always be supplied explicitly.

andreaguarino: Unfortunately `PythonInputFileContext` and `PythonVisitorContext` are part of the public API and adding an explicit parameter to all constructors would break back-compatibility. Policies for changing APIs compatibility are pretty conservative (at least one SQ LTS cycle). Given that the impact is pretty small, I think it's fine to have `SonarProduct.SONARQUBE` by default. Wdyt?

anton-haubner-sonarsource: I see. Yes, it was mostly just a concern about accidental use of the wrong constructor, so for me, it is also fine to keep it like this. Especially, if it is required for compatibility.

PR #1687, opened 2023-12-06
SONARPY-1555 Rule S6785: GraphQL queries should not be vulnerable to Denial of Service attacks

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/GraphQLDenialOfServiceCheck.java
```diff
@@ -0,0 +1,146 @@
...
+
+@Rule(key="S6785")
+public class GraphQLDenialOfServiceCheck extends PythonSubscriptionCheck {
+
+  private static final Set<String> SAFE_VALIDATION_RULE_FQNS = Set.of("graphene.validation.DepthLimitValidator");
+
+  private static final List<String> VALID_MIDDLEWARE_NAMES = List.of("DEPTH", "COST");
+
+  private static final Predicate<String> VALID_MIDDLEWARE_PREDICATE = name ->
+    VALID_MIDDLEWARE_NAMES.stream().anyMatch(mwName -> name.toUpperCase(Locale.ROOT).contains(mwName));
```
maksim-grebeniuk-sonarsource, line 51: I guess `VALID_MIDDLEWARE_NAMES` could be `Set` and then the predicate could be like:
```java
name -> VALID_MIDDLEWARE_NAMES.contains(name.toUpperCase(Locale.ROOT))
```

python-checks/src/main/java/org/sonar/python/checks/GraphQLDenialOfServiceCheck.java
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  private static boolean hasSafeMiddlewares(List<Argument> arguments) {
+    RegularArgument argument = TreeUtils.argumentByKeyword("middleware", arguments);
+    if (argument == null) {
+      return false;
+    }
+
+    return GraphQLUtils.extractListOrTupleArgumentValues(argument)
+      .map(values -> GraphQLUtils.expressionsNameMatchPredicate(values, VALID_MIDDLEWARE_PREDICATE))
```
maksim-grebeniuk-sonarsource, line 101: maybe it is better to have
```java
.filter(values -> ...)
.isPresent()
```

python-checks/src/main/java/org/sonar/python/checks/GraphQLDenialOfServiceCheck.java
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  private static boolean hasSafeValidationRules(List<Argument> arguments) {
+    RegularArgument argument = TreeUtils.argumentByKeyword("validation_rules", arguments);
+    if (argument == null) {
+      return false;
+    }
+
+    return GraphQLUtils.extractListOrTupleArgumentValues(argument)
+      .map(values -> (GraphQLUtils.expressionsNameMatchPredicate(values, VALID_MIDDLEWARE_PREDICATE)
```
maksim-grebeniuk-sonarsource, line 112: what if instead of having the predicate extract `GraphQLUtils.expressionsNameMatchPredicate(values, VALID_MIDDLEWARE_PREDICATE)` to method accepting just values? like
```java
private static boolean isValidMiddlewareNames(values) {
  return GraphQLUtils.expressionsNameMatchPredicate(values, name -> VALID_MIDDLEWARE_NAMES.contains(name.toUpperCase(Locale.ROOT))))
}
```
and then `VALID_MIDDLEWARE_PREDICATE` is not needed

python-checks/src/main/java/org/sonar/python/checks/GraphQLDenialOfServiceCheck.java
```diff
@@ -0,0 +1,146 @@
...
+
+    final Set<Tree> circularReferences;
+
+    public CircularRelationshipVisitor(Set<Tree> circularReferences) {
+      this.circularReferences = circularReferences;
+    }
+
+    @Override
+    public void visitCallExpression(CallExpression callExpression) {
+      boolean createsCircularDependency = Optional.of(callExpression.callee())
```
maksim-grebeniuk-sonarsource, line 127: instead of `Optional.of` could be `TreeUtils.toOptionalInstanceOf(QualifiedExpression.class, callExpression.callee())` and it will next line not needed

python-checks/src/main/java/org/sonar/python/checks/GraphQLDenialOfServiceCheck.java
```diff
@@ -0,0 +1,146 @@
...
+        .filter(qualifiedExpression -> "relationship".equals(qualifiedExpression.name().name()))
+        .map(QualifiedExpression::qualifier)
+        .flatMap(TreeUtils.toOptionalInstanceOfMapper(Name.class))
+        .map(Expressions::singleAssignedNonNameValue)
+        .flatMap(TreeUtils.toOptionalInstanceOfMapper(CallExpression.class))
+        .map(CallExpression::callee)
+        .filter(HasSymbol.class::isInstance)
+        .map(HasSymbol.class::cast)
+        .map(HasSymbol::symbol)
+        .filter(s -> "flask_sqlalchemy.SQLAlchemy".equals(s.fullyQualifiedName()))
```
maksim-grebeniuk-sonarsource, line 138:
```java
.filter(s -> "flask_sqlalchemy.SQLAlchemy".equals(s.fullyQualifiedName()))
```
could be replaced with
```java
.filter("flask_sqlalchemy.SQLAlchemy"::equals)
```

python-checks/src/main/java/org/sonar/python/checks/GraphQLDenialOfServiceCheck.java
```diff
@@ -0,0 +1,146 @@
...
+        .isPresent();
+      if (createsCircularDependency) {
```
maksim-grebeniuk-sonarsource, line 140: why not `ifPresent(() -> circularReferences.add(callExpression))` instead?

python-checks/src/main/java/org/sonar/python/checks/hotspots/GraphQLIntrospectionCheck.java
```diff
@@ -20,46 +20,33 @@
...
-
-  private static final String AS_VIEW_CALLEE_NAME = "as_view";
-
   private static final Set<String> SAFE_VALIDATION_RULE_FQNS = Set.of(
     "graphene.validation.DisableIntrospection",
     "graphql.validation.NoSchemaIntrospectionCustomRule");
 
   private static final String MESSAGE = "Disable introspection on this \"GraphQL\" server endpoint.";
 
+  private static final Predicate<String> CONTAINS_INTROSPECTION = s -> s.toUpperCase(Locale.ROOT).contains("INTROSPECTION");
```
maksim-grebeniuk-sonarsource, line 48: why not method?

python-checks/src/main/java/org/sonar/python/checks/utils/GraphQLUtils.java
```diff
@@ -0,0 +1,94 @@
...
+        .map(TreeUtils::nameFromQualifiedOrCallExpression),
+      expressions.stream().map(Expression::type).map(t -> Optional.ofNullable(InferredTypes.typeName(t))));
+
+    return expressionsNameAndType
+      .filter(Optional::isPresent)
+      .map(Optional::get)
+      .anyMatch(predicate);
+  }
+
+  public static boolean expressionsContainsSafeRuleFQN(List<Expression> expressions, Predicate<String> predicate) {
```
maksim-grebeniuk-sonarsource, line 85: public static Optional<String> expressionsContainsSafeRuleFQN(List<Expression> expressions, Predicate<String> predicate) {
  return TreeUtils.getSymbolFromTree(expression)
}

maksim-grebeniuk-sonarsource (changes requested):
Looks good, after discussion several changes requested

python-checks/src/main/java/org/sonar/python/checks/GraphQLDenialOfServiceCheck.java
guillaume-dequenne-sonarsource: Actually the logic will not be the same because we want to check that the name contains either `depth` or `cost`, not that it actually is one of them.

python-checks/src/main/java/org/sonar/python/checks/GraphQLDenialOfServiceCheck.java
guillaume-dequenne-sonarsource: I think it does look nicer but here the peculiarity of the logic is that if the optional is initially empty, it is because the argument could not be resolved to a list or tuple (which means there's something, but it's entirely unknown), in which case we don't raise an issue.

If the list/tuple is present but its contained expressions fail to match the predicate, then we'll raise an issue (even if it's empty).

maksim-grebeniuk-sonarsource (approved):
LGTM! Thanks

PR #1688, opened 2023-12-07
DBD-691: Add support for shared checks.

anton-haubner-sonarsource:
Shared checks are responsible for multiple rule keys at the same time. They are executed if at least one of the rules registered for them is active.
A similar concept has for example also been implemented for sonar-java. See org.sonar.plugins.java.api.CheckRegistrar::registerMainSharedCheck.

In practice, this feature is needed by DBD.
DBD needs to supply a custom check to sonar-python for generating IR.

Previously, this was achieved by supplying a check for every DBD rule and only the first one that is executed would actually generate the IR. This approach required static fields that keep track whether IR has already been generated or not.
Since static fields like this don't work well with SonarLint we now need proper support for checks that cover multiple rules.

### DIFF
--- a/python-frontend/src/main/java/org/sonar/plugins/python/api/PythonSharedCheck.java
+++ b/python-frontend/src/main/java/org/sonar/plugins/python/api/PythonSharedCheck.java
@@ -1,0 +1,27 @@
+/*
+ * SonarQube Python Plugin
+ * Copyright (C) 2011-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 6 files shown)


anton-haubner-sonarsource:
This Ticket / PR has been put on hold, since the DBD sprint on SonarLint support has ended and this ticket did not contain a required feature.

Some more notes on the purpose of this PR and what steps are required to complete it have been added to the ticket: https://sonarsource.atlassian.net/browse/DBD-691?atlOrigin=eyJpIjoiNDhiM2JiZTJlYWQ4NDdhYjliNTMzNTc4NzgyNzlhNjEiLCJwIjoiaiJ9

Thus, I am closing this PR for now.

PR #1689, opened 2023-12-08
SONARPY-1579 Fix FN on S6437 when library stubs are missing

maksim-grebeniuk-sonarsource:


python-frontend/src/main/java/org/sonar/python/types/InferredTypes.java
```diff
@@ -399,6 +417,9 @@ static boolean isTypeClassCompatibleWith(Symbol typeClass, InferredType other) {
     if (other instanceof UnionType) {
       return ((UnionType) other).types().stream().anyMatch(t -> InferredTypes.isTypeClassCompatibleWith(typeClass, t));
     }
+    if (other instanceof UnknownClassType) {
+      return InferredTypes.areSymbolsCompatible(typeClass, ((UnknownClassType) other).typeClass());
```
guillaume-dequenne-sonarsource, line 421: Do we need this?
Given the type is unknown, I feel we should be conservative here and return true like with `AnyType`.
Otherwise we'd risk some false positives I guess?

python-frontend/src/main/java/org/sonar/python/types/UnknownClassType.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import org.sonar.plugins.python.api.symbols.Symbol;
+import org.sonar.plugins.python.api.types.InferredType;
+import org.sonar.python.semantic.SymbolImpl;
+
+public class UnknownClassType implements InferredType {
+
+  private final Symbol typeClass;
```
guillaume-dequenne-sonarsource, line 31: Does it make sense to call this `typeClass`?
We assume this to be a class but really we don't know. 
Probably `typeSymbol` or `assumedClassSymbol` may make sense to highlight the uncertainty?

python-frontend/src/main/java/org/sonar/python/types/UnknownClassType.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  @Override
+  public boolean canBeOrExtend(String typeName) {
+    return true;
+  }
+
+  @Override
+  public boolean isCompatibleWith(InferredType other) {
+    return InferredTypes.isTypeClassCompatibleWith(typeClass, other);
```
guillaume-dequenne-sonarsource, line 80: Probably here we could return true by default as well? Since we know nothing of the implementation?

python-frontend/src/main/java/org/sonar/python/types/UnknownClassType.java
```diff
@@ -0,0 +1,91 @@
...
+package org.sonar.python.types;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import org.sonar.plugins.python.api.symbols.Symbol;
+import org.sonar.plugins.python.api.types.InferredType;
+import org.sonar.python.semantic.SymbolImpl;
+
+public class UnknownClassType implements InferredType {
```
guillaume-dequenne-sonarsource, line 29: We should probably have a comment that explains that this type is used for symbols that we assume to be class due to how they are used, so that type inference correctly tracks objects that were created from this.

python-frontend/src/main/java/org/sonar/python/types/InferredTypes.java
```diff
@@ -153,6 +154,23 @@ public static InferredType runtimeType(@Nullable Symbol typeClass) {
     return anyType();
   }
 
+  public static InferredType anyOrUnknownClassType(Symbol symbol) {
```
guillaume-dequenne-sonarsource, line 157: Probably here as well we need a comment as well explaining that we assume unknown imported symbols whose name start with a capital letter to be classes.

guillaume-dequenne-sonarsource (changes requested):
This looks good!

I'm still not sure this is the right direction to go, but I appreciate how simple you made the change looks like, as this is way less impactful and hacky looking than the previous iteration.

I left some small comments, nothing major though.

python-frontend/src/main/java/org/sonar/python/types/UnknownClassType.java
```diff
@@ -0,0 +1,95 @@
...
+    var member = members.computeIfAbsent(memberName, n -> Optional.of(typeSymbol)
+      .map(Symbol::fullyQualifiedName)
+      .map(fqn -> new SymbolImpl(memberName, fqn + "." + memberName))
+      .orElse(null));
+    return Optional.ofNullable(member);
+  }
+
+  @Override
+  public Optional<Symbol> resolveDeclaredMember(String memberName) {
+    return Optional.empty();
```
guillaume-dequenne-sonarsource, line 69: One last thing I missed during first review: I'm wondering whether we should return `Optional.empty()` here or just defer to `resolveMember` and return a fake symbol.

In practical terms I believe this wouldn't change much as this method is seldom used, but probably returning a fake symbol would make sense (since we consider this unknown type to be capable of declaring any member).

guillaume-dequenne-sonarsource (commented):
Looks good!
Left a question about how `resolveDeclaredMember` should actually behave. I have the feeling it should defer to `resolveMember`, but happy to discuss it.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1694, opened 2023-12-13
SONARPY-1582b: Independent Indexer and Cache

anton-haubner-sonarsource:
This PR is meant to address [review comments](https://github.com/SonarSource/sonar-python/pull/1690#pullrequestreview-1777960787) on MMF-3482.

`SonarLintCache` and `PythonIndexer` are now more independent.
I.e. the `SonarLintCache` extension is added independently of whether `SonarLintPythonIndexer` can be added or not.
Also, Sensors can be instantiated if no `SonarLintCache` is present, but a `PythonIndexer` is available.

### DIFF
--- a/python-frontend/src/main/java/org/sonar/python/caching/CacheContextImpl.java
+++ b/python-frontend/src/main/java/org/sonar/python/caching/CacheContextImpl.java
@@ -19,11 +19,13 @@
  */
 package org.sonar.python.caching;
 
+import javax.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.sonar.api.SonarProduct;
 import org.sonar.api.batch.sensor.SensorContext;
 import org.sonar.api.utils.Version;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import org.sonar.plugins.python.api.SonarLintCache;
 import org.sonar.plugins.python.api.caching.CacheContext;
 import org.sonar.plugins.python.api.caching.PythonReadCache;
 import org.sonar.plugins.python.api.caching.PythonWriteCache;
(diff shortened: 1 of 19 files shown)


python-frontend/src/main/java/org/sonar/python/caching/CacheContextImpl.java
```diff
@@ -58,19 +60,24 @@ public PythonWriteCache getWriteCache() {
...
-      && context.runtime().getApiVersion().isGreaterThanOrEqual(MINIMUM_RUNTIME_VERSION)
-      && !isUsingSonarModules
-    ) {
-      return new CacheContextImpl(context.isCacheEnabled(), new PythonWriteCacheImpl(context.nextCache()), new PythonReadCacheImpl(context.previousCache()));
+
+    if (context.runtime().getApiVersion().isGreaterThanOrEqual(MINIMUM_RUNTIME_VERSION) &&
+      !isUsingSonarModules) {
+      if (context.runtime().getProduct().equals(SonarProduct.SONARLINT)) {
+        if (sonarLintCache != null) {
+          return new CacheContextImpl(true, new PythonWriteCacheImpl(sonarLintCache), new PythonReadCacheImpl(sonarLintCache));
```
guillaume-dequenne-sonarsource, line 74: Until now, the `CacheContext#isCacheEnabled` was understood to mean that the SQ/SC PR analysis cache was enabled.

This information is then used by the `SonarQubePythonIndexer` to try to retrieve/store information from the PR analysis cache. I believe it is used similarly by the SonarSecurity frontend to retrieve and save UCFGs (not 100% sure about that - didn't double check).

By chance, none of these components are involved in the SonarLint context, but I'm afraid this is a bit implicit and could lead to nasty bugs should the underlying assumption change.

Even if we were setting `false` here, I feel the meaning of this field may no longer be clear, now that this component handles 2 distinct caches for 2 quite different use cases.

guillaume-dequenne-sonarsource (commented):
I like the idea that we can get independence between the `PythonIndexer` and the `SonarLintCache`. 

However, I have the feeling that this is at the cost of muddying the waters between the PR analysis cache and the SonarLint cache, which in my opinion should be well separated due to how easy it would be to confuse those two components whose goals are quite different. I could imagine the possibility of exposing ourselves to some bugs due to this.

All in all, I have the feeling that the initial implementation, while not ideal, carries less risk: all recent versions of SonarLint have the `ModuleFileListener` and, in practice, both the `SonarLintPythonIndexer` and the `SonarLintCache` will be injected. Versions that do not have it predate DBD itself and may not be able to make use of it even with successful injection of the `SonarLintCache`.

It's probably best that we discuss it live to decide the next course of action.

anton-haubner-sonarsource:
I recorded this open problem as a Jira ticket that could perhaps be addressed by a future DBD sprint on SonarLint support: https://sonarsource.atlassian.net/browse/SONARPY-1592?atlOrigin=eyJpIjoiYjZmOTc4ODZlMjdiNDgyY2EzYjlkMmYzOGNjMmQyYjYiLCJwIjoiaiJ9

Thus, I'm closing this PR for now.

PR #1707, opened 2024-02-14
SONARPY-1609 Fix FP on S5886 when returning optional unions of unknown symbols

guillaume-dequenne-sonarsource:


joke1196 (approved):
LGTM! It makes sense to do this to remove FPs

PR #1708, opened 2024-02-14
SONARPY-1628 S1172: should not raise on methods of abstract classes

ghislainpiot:


python-checks/src/main/java/org/sonar/python/checks/UnusedFunctionParameterCheck.java
```diff
@@ -102,7 +104,18 @@ private static boolean isException(SubscriptionContext ctx, FunctionDef function
...
       isSpecialMethod(functionDef) ||
       hasNonCallUsages(functionSymbol) ||
-      isTestFunction(ctx, functionDef);
+      isTestFunction(ctx, functionDef) ||
+      isAbstractClass(functionDef);
+  }
+
+  private static boolean isAbstractClass(FunctionDef functionDef) {
+    FunctionSymbol functionSymbol = ((FunctionDefImpl) functionDef).functionSymbol();
+    if (functionSymbol == null) {
```
guillaume-dequenne-sonarsource, line 113: To be noted that you have to do this defensive check because `FunctionDefImpl::functionSymbol` can, in theory, be `null`, but I don't think it can actually happen in practice. 
This is why you won't be able to have 100% code coverage here. One trick would be to wrap this into an `Optional` similarly to what `hasNonCallUsages` is doing, but the implications of that have to be understood and shouldn't be abused (e.g, the `null` case still won't be tested despite 100% coverage).

(No action required, just an FYI)

python-checks/src/test/resources/checks/unusedFunctionParameter/unusedFunctionParameter.py
```diff
@@ -161,3 +161,35 @@ def param_referenced_in_comment_no_issue_3(unused_param_2):
...
+    def execute_test(self, name):
+        print("Test")
+
+from abc import ABC as XYZ
+class Test2(XYZ):
+    def execute_test(self, name):
+        print("Test")
+
+# source :  https://stackoverflow.com/a/392767
+class FakeMetaclass(type):
```
guillaume-dequenne-sonarsource, line 188: While it's interesting to have an actual example of a metaclass, it's not really needed to have a minimal test (since the current implementation will avoid raising issues to any class that defines a metaclass).

Since the goal of the unit tests is not necessarily to have exhaustive code, defining a metaclass such as:
`class FakeMetaClass(type): ...` would be enough. Having a test case where the metaclass is a symbol that is not even in the scope is acceptable as well, such as:
```
class ClassWithUnknownMetaClass(metaclass=UnknownMeta):
  def execute(self, name):
    print("Execute")
```

python-checks/src/test/resources/checks/unusedFunctionParameter/unusedFunctionParameter.py
```diff
@@ -161,3 +161,35 @@ def param_referenced_in_comment_no_issue_3(unused_param_2):
     def param_accessed_through_pandas_no_issue(sample_df, area_of_interest): # OK
         sample_df.query('area == @area_of_interest').population
         print("hello")
+
+import abc
+class Test(abc.ABC):
```
guillaume-dequenne-sonarsource, line 166: Generally, we try to name the test classes/methods with names that represent what is tested. I'd also recommend avoiding calling stuff `Test` whenever possible, since some rules define heuristics to avoid raising issues on test code ([SONARPY-1630](https://sonarsource.atlassian.net/browse/SONARPY-1630) being the latest example) and you wouldn't want to trigger one accidentally (this is not the case here, just a general principle).

I'd suggest names such as:

```
class SomeAbstractClass(abc.ABC):
  def no_issue_on_abstract_class_methods(self, name):
    print("Hello")
```

python-checks/src/test/resources/checks/unusedFunctionParameter/unusedFunctionParameter.py
```diff
@@ -161,3 +161,35 @@ def param_referenced_in_comment_no_issue_3(unused_param_2):
...
+    def execute_test(self, name):
+        print("Test")
+
```
guillaume-dequenne-sonarsource, line 169: As we discussed, I'd like to see an additional test where the inheritance from `abc.ABC` is not direct (having another class that inherits from the current `Test` for example).
I expect that an issue will be raised in that case and I think it's important to have that behavior documented.

guillaume-dequenne-sonarsource (changes requested):
Looks good!
I left some comments on the tests that you should see more as good practices that I'd like to share rather than real change requests.
The only think I think we really need is a test of indirect `abc.ABC` inheritance.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1709, opened 2024-02-15
 SONARPY-1612 S905: should ignore assigned and unused Airflow operators.

ghislainpiot:


python-checks/src/test/resources/checks/uselessStatement.py
```diff
@@ -136,3 +136,19 @@ def no_issue_within_contextlib_suppress(a):
   with suppress(TypeError):
     a + ''  # OK
     return a
+
```
guillaume-dequenne-sonarsource, line 139: Those tests look good.
What about the case you mentioned with decorators? I think it was a good catch and would be relevant to have as a dedicated ticket and possibly as a documenting test here (just showing the FP might occur in this case, without actually addressing it).

python-checks/src/main/java/org/sonar/python/checks/UselessStatementCheck.java
```diff
@@ -122,13 +120,13 @@ private static void checkNode(SubscriptionContext ctx) {
     if (parent == null || !parent.is(Kind.EXPRESSION_STMT)) {
       return;
     }
-    if (isWithinContextlibSuppress(tree)) {
+    if (isWithinSuppressionContext(tree)) {
```
guillaume-dequenne-sonarsource, line 123: Super minor thing but we tend to use `ignored` (like `isWithinIgnoredContext`) or possibly `isException` (though here it may be confusing with actual exceptions) for silenced issues - to have consistent terminology.

The prior `suppress` in the method name was related to the fact we were targeting the `contextlib.suppress` context manager only. Keeping the same name to refer to suppressed issues may make things confusing.

guillaume-dequenne-sonarsource (approved):
Looks good!
I left some pretty minor comments.
If you decide to add new tests for the decorator case, please ask for another review.

PR #1711, opened 2024-02-15
SONARPY-1624 Fix FP on S5886 when the expected return type is a tuple subclass

guillaume-dequenne-sonarsource:


ghislainpiot (approved):
Looks good to me !

PR #1714, opened 2024-02-19
SONARPY-1633 S1481: Add a quick fix to rename loop indices

ghislainpiot:


python-checks/src/main/java/org/sonar/python/checks/UnusedLocalVariableCheck.java
```diff
@@ -53,7 +55,8 @@ public class UnusedLocalVariableCheck extends PythonSubscriptionCheck {
   private static final String DEFAULT = "(_[a-zA-Z0-9_]*|dummy|unused|ignored)";
   private static final String MESSAGE = "Remove the unused local variable \"%s\".";
   private static final String SEQUENCE_UNPACKING_MESSAGE = "Replace unused local variable \"%s\" with \"_\".";
-  private static final String SEQUENCE_UNPACKING_QUICK_FIX_MESSAGE = "Replace with \"_\"";
+  private static final String LOOP_INDEX_MESSAGE = "Replace unused loop index \"%s\" with \"_\".";
```
guillaume-dequenne-sonarsource, line 58: I see you stayed consistent with `SEQUENCE_UNPACKING_MESSAGE `, which is good. However I feel like there is a `the` article missing from both messages.
I'd suggest taking the opportunity to replace both messages with `Replace the unused ...`.

python-checks/src/main/java/org/sonar/python/checks/UnusedLocalVariableCheck.java
```diff
@@ -120,17 +123,36 @@ private static boolean isVariableAccessedInStringTemplate(Symbol symbol, StringL
...
   public PreciseIssue createIssue(SubscriptionContext ctx, Symbol symbol, Usage usage) {
     if (isSequenceUnpacking(usage)) {
-      var quickFix = PythonQuickFix.newQuickFix(SEQUENCE_UNPACKING_QUICK_FIX_MESSAGE, TextEditUtils.replace(usage.tree(), "_"));
+      var quickFix = PythonQuickFix.newQuickFix(RENAME_QUICK_FIX_MESSAGE, TextEditUtils.replace(usage.tree(), "_"));
       var issue = ctx.addIssue(usage.tree(), String.format(SEQUENCE_UNPACKING_MESSAGE, symbol.name()));
       issue.addQuickFix(quickFix);
       return issue;
+    } else if (isLoopIndex(usage, symbol)) {
+      PreciseIssue issue = ctx.addIssue(usage.tree(), String.format(LOOP_INDEX_MESSAGE, symbol.name()));
+      PythonQuickFix quickFix = PythonQuickFix.newQuickFix(RENAME_QUICK_FIX_MESSAGE, TextEditUtils.replace(usage.tree(), "_"));
```
guillaume-dequenne-sonarsource, line 132: It probably makes more sense to instantiate the quick fix only if we're going to apply it (i.e in the `foundUnderscoreSymbol == null` branch).

python-checks/src/main/java/org/sonar/python/checks/UnusedLocalVariableCheck.java
```diff
@@ -120,17 +123,36 @@ private static boolean isVariableAccessedInStringTemplate(Symbol symbol, StringL
...
+      Symbol foundUnderscoreSymbol = null;
+      Tree searchTree = usage.kind().equals(Usage.Kind.LOOP_DECLARATION) ? ctx.syntaxNode() : null;
+      while (foundUnderscoreSymbol == null && searchTree != null) {
+        foundUnderscoreSymbol = ((FunctionDef) searchTree).localVariables().stream().filter(symbol1 -> "_".equals(symbol1.name())).findAny().orElse(null);
+        searchTree = TreeUtils.firstAncestorOfKind(searchTree, Kind.FUNCDEF);
+      }
```
guillaume-dequenne-sonarsource, line 138: I'd suggest extracting this in a dedicated method whose purpose is to check whether the `_` symbol exists in another scope.

python-checks/src/main/java/org/sonar/python/checks/UnusedLocalVariableCheck.java
```diff
@@ -120,17 +123,36 @@ private static boolean isVariableAccessedInStringTemplate(Symbol symbol, StringL
...
       issue.addQuickFix(quickFix);
       return issue;
+    } else if (isLoopIndex(usage, symbol)) {
+      PreciseIssue issue = ctx.addIssue(usage.tree(), String.format(LOOP_INDEX_MESSAGE, symbol.name()));
+      PythonQuickFix quickFix = PythonQuickFix.newQuickFix(RENAME_QUICK_FIX_MESSAGE, TextEditUtils.replace(usage.tree(), "_"));
+      Symbol foundUnderscoreSymbol = null;
+      Tree searchTree = usage.kind().equals(Usage.Kind.LOOP_DECLARATION) ? ctx.syntaxNode() : null;
+      while (foundUnderscoreSymbol == null && searchTree != null) {
+        foundUnderscoreSymbol = ((FunctionDef) searchTree).localVariables().stream().filter(symbol1 -> "_".equals(symbol1.name())).findAny().orElse(null);
+        searchTree = TreeUtils.firstAncestorOfKind(searchTree, Kind.FUNCDEF);
```
guillaume-dequenne-sonarsource, line 137: What if `_` is defined at the file input level or at the class level?

```
_ = 42
def foo():
  for i in range(5):  # quick fixing i into _ would change the behavior of the program
    print("hello")
  print(_)
foo()
```

I believe we wouldn't have this problem for a class, but we might want to have a test showing that this is the expected behavior.

python-checks/src/test/resources/checks/unusedLocalVariable.py
```diff
@@ -144,3 +144,17 @@ def foo[T](a: T) -> T:
         ...
 
     return foo
+
+def additional_test_for_loop_comprehension():
+    return ["a" for i in range(5)] # Noncompliant
```
guillaume-dequenne-sonarsource, line 149: Since we changed the issue message for comprehension indices, I think it would make sense to have an assertion on the message as well.

python-checks/src/test/java/org/sonar/python/checks/UnusedLocalVariableCheckTest.java
```diff
@@ -87,4 +87,40 @@ void exceptClauseQuickFixTest() {
     PythonQuickFixVerifier.verify(check, before, after);
     PythonQuickFixVerifier.verifyQuickFixMessages(check, before, "Remove the unused local variable");
   }
+
```
guillaume-dequenne-sonarsource, line 90: Related to the previous comment, I think having a test where `_` is defined at the top level (thus we don't offer a quick fix) and one where `_` is defined at a class level (where we do offer a quick fix, assuming this indeed doesn't change the behavior of the program) would be nice.

guillaume-dequenne-sonarsource (changes requested):
I left a few comments for minor improvements and to cover some edge cases, but this looks good otherwise!

python-checks/src/main/java/org/sonar/python/checks/UnusedLocalVariableCheck.java
```diff
@@ -147,6 +143,21 @@ public PreciseIssue createIssue(SubscriptionContext ctx, Symbol symbol, Usage us
     }
   }
 
+  @Nullable
+  private static Symbol isUnderscoreSymbolAlreadyAssigned(SubscriptionContext ctx, Usage usage) {
```
guillaume-dequenne-sonarsource, line 147: Since we only perform a null check on the return value, I'd suggest making that null check inside this method and returning a boolean instead.
That would also be more consistent with the method name.

guillaume-dequenne-sonarsource (approved):
Looks good to me.
A minor last comment before merging: I think `isUnderscoreSymbolAlreadyAssigned` should return a boolean directly rather than performing a null check on its return value.

PR #1716, opened 2024-02-20
SONARPY-1482 Modify PythonQuickfixVerify to parse the code after the quickfix

ghislainpiot:
…quickfix


python-checks/src/test/java/org/sonar/python/checks/quickfix/PythonQuickFixVerifier.java
```diff
@@ -96,6 +97,8 @@ public static void verify(Function<String, PythonVisitorContext> createVisitorCo
       .as("The code with the quickfix applied is not the expected result.\n" +
         "\"Applied QuickFixes are:\n%s\nExpected result:\n%s", appliedQuickFix, Arrays.asList(codesFixed))
       .isEqualTo(Arrays.asList(codesFixed));
+
+    assertThatCode( () -> PythonParser.createIPythonParser().parse(String.join("\n", codesFixed))).doesNotThrowAnyException(); // We assume that any IPython code is valid Python code
```
guillaume-dequenne-sonarsource, line 101: We have `PythonQuickFixVerifierTest` to test the behavior of the verifier.

I think it would make sense to have a test where the expected code after quick fix is syntactically invalid to ensure this catches it.

You might want to move this assertion before the ones checking for the quick fix number/result so that you can trigger this condition more easily. 

Additionally, you can add a description with `as` and override the thrown exception message with `overridingErrorMessage` (similar to what is done in other assertions of the verifier) to ensure the message is informative enough (applying the quick fix results in a parse error) - this is something you can test as well.

guillaume-dequenne-sonarsource, line 101: On another note, I do have the feeling that we could refactor a bit the behavior for the `createIPythonVisitorContext` and `createPythonVisitorContext` methods so that we propagate the parser they use, which would mean we can use the correct one here as well.

I don't have a very strong opinion that we should do this, although I think it may avoid mistakes where we quick fix Python code into IPython code - however unlikely that is to happen.

guillaume-dequenne-sonarsource (changes requested):
This looks good! I think we need to add a test for this though.

I also suggest a broader refactoring to use the correct parser each time, but I'm not sure we should actually do it as it may be cumbersome.

Happy to discuss this further.

python-checks/src/test/java/org/sonar/python/checks/quickfix/PythonQuickFixVerifierTest.java
```diff
@@ -109,7 +121,22 @@ public void initialize(Context context) {
       });
     }
   }
+  private class SimpleCheckInvalidSyntax extends PythonSubscriptionCheck {
```
guillaume-dequenne-sonarsource, line 124: I don't think you need this special made quick fix to create the syntax error, since the mere fact that the expected fixed code contains a syntax error will fail the test, without checking what was actually applied.

python-checks/src/test/java/org/sonar/python/checks/quickfix/PythonQuickFixVerifier.java
```diff
@@ -92,10 +93,16 @@ public static void verify(Function<String, PythonVisitorContext> createVisitorCo
...
+    PythonParser pythonParser = isIPython ? PythonParser.createIPythonParser() : PythonParser.create();
+    assertThatCode(() -> pythonParser.parse(String.join("\n", codesFixed))).as("Correction of quick fixes")
+      .overridingErrorMessage("The code generated by the quickfix is not valid (I)Python code.\nResults is :\n%s", Arrays.asList(codesFixed))
+      .doesNotThrowAnyException(); // We assume that any IPython code is valid Python code
+
     assertThat(appliedQuickFix)
       .as("The code with the quickfix applied is not the expected result.\n" +
         "\"Applied QuickFixes are:\n%s\nExpected result:\n%s", appliedQuickFix, Arrays.asList(codesFixed))
       .isEqualTo(Arrays.asList(codesFixed));
+
```
guillaume-dequenne-sonarsource, line 105: Is this newline needed?

guillaume-dequenne-sonarsource (changes requested):
Looks good!
While it may not seem like the cleanest solution, I like the fact you didn't pursue a possibly long refactoring and went for a simple boolean to instantiate the right parser.
I left some minor comments to fix but this is good otherwise.

PR #1718, opened 2024-02-21
SONARPY-1634 S1481: Add a quick fix to remove assignment target

ghislainpiot:


python-checks/src/test/java/org/sonar/python/checks/UnusedLocalVariableCheckTest.java
```diff
@@ -157,4 +157,36 @@ void loopIndexComprehensionClassQuickFixTest() {
     PythonQuickFixVerifier.verify(check, before, after);
     PythonQuickFixVerifier.verifyQuickFixMessages(check, before, "Replace with \"_\"");
   }
+
+  @Test
+  void assignementQuickFixTest() {
```
guillaume-dequenne-sonarsource, line 162: Super minor typo: extra `e` in `assignment` (same on line 179)

python-checks/src/main/java/org/sonar/python/checks/UnusedLocalVariableCheck.java
```diff
@@ -135,6 +137,16 @@ public PreciseIssue createIssue(SubscriptionContext ctx, Symbol symbol, Usage us
         issue.addQuickFix(quickFix);
       }
       return issue;
+    } else if (usage.kind().equals(Usage.Kind.ASSIGNMENT_LHS)) {
+      PreciseIssue issue = ctx.addIssue(usage.tree(), String.format(MESSAGE, symbol.name()));
+      AssignmentStatement assignmentStatement = ((AssignmentStatement) TreeUtils.firstAncestorOfKind(usage.tree(), Kind.ASSIGNMENT_STMT));
+      // This can only trigger if there is 1 lhs expression, the other cases are handled by isSequenceUnpacking
+      if (assignmentStatement != null) {
```
guillaume-dequenne-sonarsource, line 144: From my understanding, this can happen when we have an annotated assignment, such as (in the tests):

```
def no_fn_type_annotation_with_assignment():
    value: str = "hello"  # Noncompliant
    return [int(value) for value in something()]
```

Maybe we could explicitly have a test showing no quick fix is offered in such case (and maybe we actually want to eventually cover that case as well - although you can technically trigger side effects through the evaluation of the type annotation...).

guillaume-dequenne-sonarsource (approved):
Looks good!
Left a small comment about annotated assignment. Not sure we really want to do something about this, maybe a test documenting the current behavior (no quick fix) would be nice thoughl.

python-checks/src/main/java/org/sonar/python/checks/UnusedLocalVariableCheck.java
ghislainpiot: As discussed, we will generate the quickfix even when there is a type annotation. We also allow the quick fix have this behaviour :  `value: str = "hello`  ->`"hello"`

python-checks/src/main/java/org/sonar/python/checks/UnusedLocalVariableCheck.java
```diff
@@ -135,13 +139,37 @@ public PreciseIssue createIssue(SubscriptionContext ctx, Symbol symbol, Usage us
...
     }
   }
 
+  private static void createAssignmentQuickFix(Usage usage, Statement assignmentStatement, PreciseIssue issue) {
+    Token lastQuickFixToken;
+    if (assignmentStatement.is(Kind.ASSIGNMENT_STMT)) {
+      lastQuickFixToken = ((AssignmentStatement) assignmentStatement).assignedValue().firstToken();
+    } else {
+      var assignedValue = ((AnnotatedAssignment) assignmentStatement).assignedValue();
+      // This is an assignment, so it can't be null
```
guillaume-dequenne-sonarsource, line 164: Actually, the following:
```
x: str
```
is considered a valid annotated assignment with a null assigned value.

The only reason this is never null in the current test suite is because `hasOnlyBindingUsages` explicitly filters out symbols which are only used as type annotations (through `isOnlyTypeAnnotation`).

In fact modifying one of the existing tests such as:
```
def no_fp_type_annotation():
    value: str  # assigned value is null here
    value = "hello"
```
will make the assert statement fail. 

As a sidenote, I think we should never use `assert` statements to test our assumptions. As these statements may not be executed in production, it is unsafe to rely on them (cf. my comment on the Jira ticket).

Besides, this case is a good example why it is better to let code go uncovered through defensive programming than relying on assumptions. With an uncovered `null` check, we'd miss providing a quick fix in some cases. With an `assert` statement or an unsafe dereference, we'd have a crash in production because the assumption was ultimately incorrect.

(Note: I think we probably need an extra test case that reflects the example here)

python-checks/src/main/java/org/sonar/python/checks/UnusedLocalVariableCheck.java
```diff
@@ -135,13 +139,37 @@ public PreciseIssue createIssue(SubscriptionContext ctx, Symbol symbol, Usage us
         issue.addQuickFix(quickFix);
       }
       return issue;
+    } else if (usage.kind().equals(Usage.Kind.ASSIGNMENT_LHS)) {
+      PreciseIssue issue = ctx.addIssue(usage.tree(), String.format(MESSAGE, symbol.name()));
+
+      Statement assignmentStatement = ((Statement) TreeUtils.firstAncestorOfKind(usage.tree(), Kind.ASSIGNMENT_STMT, Kind.ANNOTATED_ASSIGNMENT));
+
+      // This can only trigger if there is 1 assigned variable in the same statement, the other cases are handled by isSequenceUnpacking
+
```
guillaume-dequenne-sonarsource, line 148: _Nitpick:_ I don't really think we need those white lines here. I don't really mind them though, no strong opinion.

python-checks/src/main/java/org/sonar/python/checks/UnusedLocalVariableCheck.java
```diff
@@ -138,10 +141,30 @@ public PreciseIssue createIssue(SubscriptionContext ctx, Symbol symbol, Usage us
...
+      Optional.ofNullable(assignmentStatement).filter(stmt -> stmt.is(Kind.ASSIGNMENT_STMT)).map(AssignmentStatement.class::cast).ifPresent(stmt -> {
+        PythonQuickFix quickFix = PythonQuickFix.newQuickFix(ASSIGNMENT_QUICK_FIX_MESSAGE,
+          TextEditUtils.removeUntil(usage.tree(), stmt.assignedValue().firstToken()));
+        issue.addQuickFix(quickFix);
+      });
+
+      Optional.ofNullable(assignmentStatement).filter(stmt -> stmt.is(Kind.ANNOTATED_ASSIGNMENT)).map(AnnotatedAssignment.class::cast)
+        .map(AnnotatedAssignment::assignedValue).ifPresent(stmt -> {
+          PythonQuickFix quickFix = PythonQuickFix.newQuickFix(ASSIGNMENT_QUICK_FIX_MESSAGE,
+            TextEditUtils.removeUntil(usage.tree(), stmt.firstToken()));
```
guillaume-dequenne-sonarsource, line 162: _Nitpick:_ I guess here `stmt` is no longer the statement but the assigned value (if present), I think it'd be better if the variable name reflected that.

python-checks/src/main/java/org/sonar/python/checks/UnusedLocalVariableCheck.java
```diff
@@ -138,10 +141,30 @@ public PreciseIssue createIssue(SubscriptionContext ctx, Symbol symbol, Usage us
...
+      Statement assignmentStatement = ((Statement) TreeUtils.firstAncestorOfKind(usage.tree(), Kind.ASSIGNMENT_STMT, Kind.ANNOTATED_ASSIGNMENT));
+
+      Optional.ofNullable(assignmentStatement).filter(stmt -> stmt.is(Kind.ASSIGNMENT_STMT)).map(AssignmentStatement.class::cast).ifPresent(stmt -> {
+        PythonQuickFix quickFix = PythonQuickFix.newQuickFix(ASSIGNMENT_QUICK_FIX_MESSAGE,
+          TextEditUtils.removeUntil(usage.tree(), stmt.assignedValue().firstToken()));
+        issue.addQuickFix(quickFix);
+      });
+
+      Optional.ofNullable(assignmentStatement).filter(stmt -> stmt.is(Kind.ANNOTATED_ASSIGNMENT)).map(AnnotatedAssignment.class::cast)
+        .map(AnnotatedAssignment::assignedValue).ifPresent(stmt -> {
```
guillaume-dequenne-sonarsource, line 160: As mentioned in the previous review, I think that having a test case similar to:

```
def annotated_assignment_null_value():
    value: str
    value = "hello"
```

is important.

The lack of this test case led you to assume `AnnotatedAssignment::assignedValue` could never return `null` when reaching this point (although it can). Here, the `Optional` chain is hiding the fact we are missing some attainable coverage (I'm not saying we shouldn't use an `Optional` here, just that we shouldn't use it for the wrong reasons).

Namely, in this case, an issue would be raised but no quick fix will be offered (we can have a test in the regular check test as well as one in the quick fix test to show this).

guillaume-dequenne-sonarsource (changes requested):
I admit this is a bit nitpicky to request further changes for this, but I think the underlying principle is important enough to insist on this.

The actual implementation looks good though.

PR #1719, opened 2024-02-21
SONARPY-1660 Fix FP on S905 when using a decorator to declare an Airflow DAG

ghislainpiot:


python-checks/src/main/java/org/sonar/python/checks/UselessStatementCheck.java
```diff
@@ -123,9 +125,22 @@ private static void checkNode(SubscriptionContext ctx) {
...
     }
+    // Safe cast because the rule only subscribes to expressions
+    if (isAnAirflowException((Expression) tree)) {
+      return;
+    }
     ctx.addIssue(tree, MESSAGE);
   }
 
+  private static boolean isAnAirflowException(Expression expression) {
+    if (isWithinAirflowContext(expression) && expression.type().canBeOrExtend("airflow.models.baseoperator.BaseOperator")) {
```
guillaume-dequenne-sonarsource, line 136: I think we can be lenient here and not actually check whether the expression type is an Airflow operator.
I think focusing the exception on the last statement of the block is already strict enough (I was personally tempted to just keep it disabled in all Airflow context).

Our main goal is not to avoid FNs at all cost, it is to make sure the rule is not noisy when dealing with edge cases/frameworks that may alter the runtime behavior in a way we don't understand.

I also feel we are missing a test where the expression actually isn't a `BaseOperator` despite meeting all other conditions to raise the issue (while coverage is 100%, removing this particular condition doesn't break any test).

guillaume-dequenne-sonarsource (changes requested):
While the implementation looks good and the change is not actually big, I think this PR does too much at once conceptually.

The goal of the ticket is to remove FPs by not reporting issues on methods decorated with `@dag`, yet in the implementation you change the behavior of the `with DAG` case as well to remove some FNs (as well as FPs in the case of nested `with`).

I believe it's important to separate concerns and have a clear purpose when implementing a ticket. In this case, the goal is to make the rule less noisy by removing FPs, not fixing FNs. Especially as we deemed those FNs acceptable when discussing the first ticket.

If we decide to fix those FNs/if it appears convenient to do so in a single PR, I'd expect separate commits and possibly even separate tickets to clearly separate the concerns.

To remain pragmatic, I think we can keep the implementation as it is (aside from a small change request, cf my comment) as it is still a net improvement over what we initially had. That said, I think this is important to keep in mind, especially when we'll deal with more complex tasks.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1720, opened 2024-02-21
SONARPY-1662 Fix FP on S117 when a class is assigned to a variable

guillaume-dequenne-sonarsource:


python-checks/src/test/resources/checks/itemOperationsTypeCheck/itemOperations_delitem.py
```diff
@@ -180,3 +180,12 @@ class ExtendedMock(MagicMock):
...
     def custom_mock():
         del ExtendedMock()[42]
+
+
+class SomeClass:
+    ...
+
+def get_state_fp():
+    some_class = SomeClass()
+    state = SomeClass.__getstate__(some_class)  # __getstate__ usually returns a dict but is annotated to return 'object' in Typeshed
```
guillaume-dequenne-sonarsource, line 190: Unfortunately, I believe we'd need to update the annotation in Typeshed to fix this. I think we can accept the FP for now.

its/ruling/src/test/resources/expected/python-S5644.json
```diff
@@ -1,49 +1,53 @@
 {
```
guillaume-dequenne-sonarsource, line 1: Unfortunately changes here are FPs (documented in the tests for S5644) due to a wrong stub annotation for `type.__getstate__`

its/ruling/src/test/resources/expected/python-S3699.json
```diff
@@ -1,19 +1,22 @@
 {
```
guillaume-dequenne-sonarsource, line 1: I believe the change here is a TP.

ghislainpiot (approved):
Looks good to me!

PR #1721, opened 2024-02-22
SONARPY-1600 Fix FP on S5655: boolean parameter incompatible with float type annotation.

ghislainpiot:


python-checks/src/test/resources/checks/argumentType.py
```diff
@@ -243,3 +243,12 @@ def win32_no_support(machine, object, instance, inum, counter):
 def fn_len():
     len(len([1,2,3]))  # Noncompliant
     len(42)  # Noncompliant
+
+def boolean_compatible_with_float() -> None:
```
guillaume-dequenne-sonarsource, line 247: In the same vein, I think we should also have a test with a function that expects an `int` and receives a `bool`, just to make sure this case works as well.

guillaume-dequenne-sonarsource (changes requested):
Some minor comments.

python-frontend/src/main/java/org/sonar/python/types/InferredTypes.java
```diff
@@ -102,8 +102,9 @@ public class InferredTypes {
   private static final Set<Set<String>> HARDCODED_INCOMPATIBLE_TYPES = Set.of(Set.of(BuiltinTypes.TUPLE, BuiltinTypes.LIST));
 
   static {
-    HARDCODED_COMPATIBLE_TYPES.put(BuiltinTypes.INT, new HashSet<>(Arrays.asList(BuiltinTypes.FLOAT, BuiltinTypes.COMPLEX)));
-    HARDCODED_COMPATIBLE_TYPES.put(BuiltinTypes.FLOAT, new HashSet<>(Collections.singletonList(BuiltinTypes.COMPLEX)));
+    HARDCODED_COMPATIBLE_TYPES.put(BuiltinTypes.INT, new HashSet<>(Arrays.asList(BuiltinTypes.FLOAT, BuiltinTypes.COMPLEX, BuiltinTypes.BOOL)));
```
guillaume-dequenne-sonarsource, line 105: I think `BuiltinTypes.BOOL` is not needed here. Similarly to the fact we raise an issue when `float` is provided to a function expecting `bool`, it would make sense to raise an issue when passing `int` to a function expecting `bool` (Mypy raises the issue).

In this case, for some unrelated reason, the issue is not raised (actually because `int` defines all the members that `bool` defines and so our generic duck-type compatibility check returns true - and it's okay), but there is no need to hard-code the fact that `int` can be provided where `bool` is expected. (It would probably make sense to have another test highlighting the behavior though: `takes_boolean(42) # FN`).

guillaume-dequenne-sonarsource (approved):
Looks good, left a last comment.

PR #1722, opened 2024-02-23
SONARPY-1662 Fix FPs related to 'type' type

guillaume-dequenne-sonarsource:
It's a bit heavy handed to assume `type` can have any member, but until we know how to properly resolve these types and their members, I think the assumption makes sense to avoid FPs.


python-checks/src/main/java/org/sonar/python/checks/CaughtExceptionsCheck.java
```diff
@@ -121,7 +121,7 @@ private static boolean canBeOrExtendBaseException(InferredType type) {
       // avoid FP on variables holding a tuple: SONARPY-713
       return true;
     }
-    if (type.canOnlyBe("type")) {
+    if (type.canBeOrExtend("type")) {
```
ghislainpiot, line 124: I see that you changed this test for the check, but the same thing happens in the checks from SONARPY-1662. It could be worth investigating

ghislainpiot (commented):
These changes looks good, but there might be some false positives with SONARPY-1662

python-checks/src/main/java/org/sonar/python/checks/CaughtExceptionsCheck.java
guillaume-dequenne-sonarsource: Indeed! I only focused on rules for which I saw FPs through differential validation, but the same FP pattern could be triggered on some other rules.
I pushed 9f525ca6e87a3ab2d3f6e09a87ddcfe5b943211e to address the remaining FPs and add tests to document the behavior for unimpacted rules.

ghislainpiot (approved):

PR #1723, opened 2024-02-23
SONARPY-1598 Fix FP on S6542 with re-exported/imported override

ghislainpiot:


python-checks/src/test/resources/checks/useOfOverrideOrOverloadDecorator.py
```diff
@@ -0,0 +1,21 @@
...
+    def addItem(self, data: Any = ...) -> None:
+        ...
+
+    @override
+    def addItem(self, data: Any = ...) -> None:
+        ...
+
+from reexport_typing_overload_override import reexported_override, reexported_overload
+
+class ClassWithReExports(UnknownParent):
```
guillaume-dequenne-sonarsource, line 14: I think it would be good to express that these are false positive. Possibly through the name of the class, or through some comment.
Also, if you created some ticket to handle this case, you could consider having a small comment to reference it.

python-checks/src/test/resources/checks/useOfOverrideOrOverloadDecorator.py
```diff
@@ -0,0 +1,21 @@
+from typing import Any
+
+class Class(UnknownParent):
```
guillaume-dequenne-sonarsource, line 3: Probably here we could name this class `ClassUsingUnknownOverloadSymbols` or something similar to clearly express what the test is doing.

guillaume-dequenne-sonarsource (approved):
Looks good!
Some very minor test improvement suggestions.

PR #1730, opened 2024-02-27
SONARPY-1645 Rule S6887: pytz.timezone should not be passed to the datetime.datetime constructor

ghislainpiot:


python-checks/src/main/java/org/sonar/python/checks/DateTimeNoPyTzTimeZoneInConstructor.java
```diff
@@ -0,0 +1,82 @@
...
+
+  private void checkContext(SubscriptionContext context) {
+    CallExpression callExpression = (CallExpression) context.syntaxNode();
+    Symbol calleeSymbol = callExpression.calleeSymbol();
+
+    if (calleeSymbol != null && "datetime.datetime".equals(calleeSymbol.fullyQualifiedName())) {
+      ArgList argList = callExpression.argumentList();
+      if (argList != null) {
+        List<Argument> argumentList = argList.arguments();
+        argumentList.stream().filter(this::checkArgument)
```
guillaume-dequenne-sonarsource, line 58: Maybe this is enough to just search for any argument (probably `pytz.timezone` should never be passed no matter which argument it is about). 

However, you'd probably be interested to know there is a method called `TreeUtils#nthArgumentOrKeyword` that allows you to search for a specific argument (passed either by name or position). It would make sense to search specifically for the timezone argument (assuming its name and position are consistent throughout the different versions of the library - which I guess should be the case).

python-checks/src/main/java/org/sonar/python/checks/DateTimeNoPyTzTimeZoneInConstructor.java
```diff
@@ -0,0 +1,82 @@
...
+          .findAny().ifPresent(argument -> context.addIssue(argument, MESSAGE));
+      }
+    }
+  }
+
+  private boolean checkArgument(Argument argument) {
+    if (argument.is(Tree.Kind.REGULAR_ARGUMENT)) {
```
guillaume-dequenne-sonarsource, line 65: Here I think early returns would actually make the code a bit more readable.

python-checks/src/main/java/org/sonar/python/checks/DateTimeNoPyTzTimeZoneInConstructor.java
```diff
@@ -0,0 +1,82 @@
...
+        CallExpression callExpression1 = (CallExpression) regularArgument.expression();
+        Symbol calleeSymbol = callExpression1.calleeSymbol();
+        return calleeSymbol != null && "pytz.timezone".equals(calleeSymbol.fullyQualifiedName());
+      } else if (regularArgument.expression().is(Tree.Kind.NAME)) {
+        return reachingDefinitionsAnalysis.valuesAtLocation((Name) regularArgument.expression()).stream()
+          .filter(expression -> expression.is(Tree.Kind.CALL_EXPR))
+          .map(CallExpression.class::cast)
+          .map(CallExpression::calleeSymbol)
+          .filter(Objects::nonNull)
+          .anyMatch(symbol -> "pytz.timezone".equals(symbol.fullyQualifiedName()));
```
guillaume-dequenne-sonarsource, line 77: The detection logic is good.
However we need to keep in mind what's going to happen from the user perspective.

In this case, if the timezone object is created far from the call site of `datetime.datetime`, we're only going to highlight the latter, and no context will be given to the user (who will wonder "how is this a `pytz.timezone` object?").

We should have a way to add a secondary location to the creation of the `pytz.timezone` object to provide that context.

python-checks/src/main/java/org/sonar/python/checks/DateTimeNoPyTzTimeZoneInConstructor.java
```diff
@@ -0,0 +1,82 @@
...
+  private void checkContext(SubscriptionContext context) {
+    CallExpression callExpression = (CallExpression) context.syntaxNode();
+    Symbol calleeSymbol = callExpression.calleeSymbol();
+
+    if (calleeSymbol != null && "datetime.datetime".equals(calleeSymbol.fullyQualifiedName())) {
+      ArgList argList = callExpression.argumentList();
+      if (argList != null) {
+        List<Argument> argumentList = argList.arguments();
+        argumentList.stream().filter(this::checkArgument)
+          .findAny().ifPresent(argument -> context.addIssue(argument, MESSAGE));
```
guillaume-dequenne-sonarsource, line 59: Related to my comment about secondary location, the issue message would probably look like: `context.addIssue(argument, MESSAGE).secondary(callExpression, SECONDARY_MESSAGE)`
So that the secondary location is added.

The secondary location message could be `"The pytz.timezone is created here"` or `"The pytz.timezone is assigned here"` if we decide to make the secondary location on the assignment, which may be even clearer.

python-checks/src/main/java/org/sonar/python/checks/DateTimeNoPyTzTimeZoneInConstructor.java
```diff
@@ -0,0 +1,82 @@
...
+import org.sonar.plugins.python.api.tree.Name;
+import org.sonar.plugins.python.api.tree.RegularArgument;
+import org.sonar.plugins.python.api.tree.Tree;
+import org.sonar.python.cfg.fixpoint.ReachingDefinitionsAnalysis;
+
+@Rule(key = "S6887")
+public class DateTimeNoPyTzTimeZoneInConstructor extends PythonSubscriptionCheck {
+
+  private ReachingDefinitionsAnalysis reachingDefinitionsAnalysis;
+  private static final String MESSAGE = "pytz.timezone should not be passed to the datetime.datetime constructor";
```
guillaume-dequenne-sonarsource, line 40: This is my bad, as issue messages should be part of the rule specification and I forgot to add it for this one.
In any case, we have some [guidelines](https://docs.sonarsource.com/sonarqube/latest/extension-guide/adding-coding-rules/) to write issue messages. One of them is that the issue message should generally be imperative.

I'd suggest something like `Don't pass a "pytz.timezone" here.` or `Don't pass a "pytz.timezone" to the "datetime.datetime" constructor.`

python-checks/src/test/resources/checks/datetime_constructor_pytz_timezone.py
```diff
@@ -0,0 +1,37 @@
...
+
+dt = datetime.datetime(2022, 1, 1, tzinfo=pytz.timezone('US/Eastern'))  # Noncompliant
+
+def timezone_in_a_variable():
+    some_timezone = pytz.timezone('US/Eastern')
+    dt = datetime.datetime(2022, 1, 1, tzinfo=some_timezone)  # Noncompliant
+    return dt
+
+some_timezone_1 = pytz.timezone('US/Eastern')
+dt1 = datetime.datetime(2022, 1, 1, tzinfo=some_timezone_1)  # FN because of ReachingDefinitionsAnalysis limitations
```
guillaume-dequenne-sonarsource, line 12: What limitation are you referring to? (no need to add it in the test - this is a legitimate question)

I believe the limitation is that it doesn't actually run on module level code and only does so within functions?

If yes, it's totally fair to document it. However, I feel it would then make sense to place the other calls to `datetime.datetime` below within a function, so that it's clear that this limitation cannot impact those tests (even though it doesn't anyway since no intermediate variable is involved for those tests).

guillaume-dequenne-sonarsource (changes requested):
This looks pretty good!

I left some comments as some things are missing. Though you couldn't possibly know them since this is the first rule you implement 😄

Another thing that is missing is the rules metadata which needs to be generated through `rule-api`. The process is very similar to the update. You need to run:

`java -jar ../rule-api-{your-version}.jar generate -rule S5886`.

Still, good job on your first rule implementation!

python-checks/src/main/java/org/sonar/python/checks/PytzTimeZoneInDatetimeConstructorCheck.java
```diff
@@ -19,64 +19,65 @@
...
+        .filter(expression -> expression.is(Tree.Kind.CALL_EXPR))
+        .map(CallExpression.class::cast)
+        .findFirst();
+      var lastAssignmentSymbol = lastAssignmentExpression.map(CallExpression::calleeSymbol)
+        .filter(symbol -> "pytz.timezone".equals(symbol.fullyQualifiedName()));
+
+      if (lastAssignmentExpression.isEmpty()) {
+        return;
+      }
+      lastAssignmentSymbol.ifPresent(assignment -> context.addIssue(argument, MESSAGE).secondary(lastAssignmentExpression.get(), SECONDARY_MESSAGE));
```
guillaume-dequenne-sonarsource, line 80: I find the logic here a bit imprecise.

`ReachingDefinitionsAnalysis#valuesAtLocation` can actually return multiple values as it is not path sensitive:
```
def foo(cond):
  if cond:
    my_tz = get_some_tz()
  else:
    my_tz = pytz.timezone(...)
  datetime.datetime(..., tzinfo=my_tz) # Reaching definitions are both get_some_tz() and pyt.timezone(...)
```

By calling `findFirst()` you'll keep the first one of those two definitions you encounter. In the best case, if the ordering was based on syntax, it would be a bit imprecise and could simply lead to some FPs/FNs depending on how the code is written.
It's actually worse, as `ReachingDefinitionsAnalysis#valuesAtLocation` returns a `Set` which means that the order of iteration is not guaranteed and the element you retrieve may be non-deterministic (which would in turn make whether we raise the issue or not non-deterministic).

Rather than looking for the first element we can find, I think we need to make a conscious choice: do all of the reaching definitions need to be noncompliant for us to report the issue? Or is it enough that one of the reaching definition is noncompliant? Depending on that choice, I think this snippet should be refactored to be a bit more intentional.

Also, some additional tests will be needed to document the behavior when the object has multiple definitions.

python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S6887.json
```diff
@@ -0,0 +1,23 @@
...
+  "defaultSeverity": "Major",
+  "ruleSpecification": "RSPEC-6887",
+  "sqKey": "S6887",
+  "scope": "All",
+  "quickfix": "unknown",
+  "code": {
+    "impacts": {
+      "MAINTAINABILITY": "HIGH",
+      "RELIABILITY": "HIGH",
+      "SECURITY": "LOW"
```
guillaume-dequenne-sonarsource, line 19: This is an action for me but I have to update the rules metadata before we merge. Namely, there shouldn't be a security impact, and we may or may not want to keep both maintainability and reliability impacts.

python-checks/src/main/java/org/sonar/python/checks/PytzTimeZoneInDatetimeConstructorCheck.java
```diff
@@ -19,64 +19,65 @@
...
+import org.sonar.python.tree.TreeUtils;
 
 @Rule(key = "S6887")
-public class DateTimeNoPyTzTimeZoneInConstructor extends PythonSubscriptionCheck {
+public class PytzTimeZoneInDatetimeConstructorCheck extends PythonSubscriptionCheck {
 
   private ReachingDefinitionsAnalysis reachingDefinitionsAnalysis;
-  private static final String MESSAGE = "pytz.timezone should not be passed to the datetime.datetime constructor";
+  private static final String MESSAGE = "Don't pass a \"pytz.timezone\" to the \"datetime.datetime\" constructor.";
+  private static final String SECONDARY_MESSAGE = "The pytz.timezone is created here";
```
guillaume-dequenne-sonarsource, line 38: I'm not sure where we can find guidelines for secondary messages and I don't even remember what they are, but it seems we have a tendency to have dots at the end of them (it is quick fix messages that don't have them).

guillaume-dequenne-sonarsource (changes requested):
Good improvements!
I left some small comments, as well as a bigger one regarding how we handle the results of `ReachingDefinitions`. Happy to discuss it.

python-checks/src/main/java/org/sonar/python/checks/PytzTimeZoneInDatetimeConstructorCheck.java
```diff
@@ -0,0 +1,88 @@
...
+
+    context.registerSyntaxNodeConsumer(Tree.Kind.CALL_EXPR, this::checkCallExpression);
+  }
+
+  private void checkCallExpression(SubscriptionContext context) {
+    CallExpression callExpression = (CallExpression) context.syntaxNode();
+    Symbol calleeSymbol = callExpression.calleeSymbol();
+
+    if (calleeSymbol != null && "datetime.datetime".equals(calleeSymbol.fullyQualifiedName())) {
+      RegularArgument argument = TreeUtils.nthArgumentOrKeyword(-1, "tzinfo", callExpression.arguments());
```
guillaume-dequenne-sonarsource, line 57: Unfortunately `nthArgumentOrKeyword` doesn't support using `-1` to express the last parameter (not sure if that's what you were aiming for?).

Anyway, the `tzinfo` parameter can actually be passed positionally (you can check that by looking at the `.pyi` stubs for `datetime`, looking at the documentation or simply trying to run it locally) and while I don't suspect it's a common thing to do, I feel it would make sense to cover this case by providing the correct parameter index (and have a test for it).

python-checks/src/test/resources/checks/datetime_constructor_pytz_timezone.py
```diff
@@ -0,0 +1,69 @@
...
+        some_timezone = pytz.timezone('US/Eastern')
+                       #^^^^^^^^^^^^^^^^^^^^^^^^^^^> 1 {{The pytz.timezone is created here.}}
+    if cond: # This is technically wrong because we will never use the pytz.timezone
+        dt11 = datetime.datetime(2022, 1, 1, tzinfo=some_timezone) # Noncompliant {{Don't pass a "pytz.timezone" to the "datetime.datetime" constructor.}}
+                                            #^^^^^^^^^^^^^^^^^^^^
+
+    another_timezone = ValidTimeZone()
+    aliased_timezone = another_timezone
+    another_timezone = pytz.timezone('US/Eastern')
+    dt12 = datetime.datetime(2022, 1, 1, tzinfo=aliased_timezone)
```
guillaume-dequenne-sonarsource, line 69: I'm not sure I understand what you're aiming for with this test.
Are you missing another call to `datetime` that takes `another_timezone` as an argument?

guillaume-dequenne-sonarsource (changes requested):
Looking good!
Sorry about the comment on `nthArgumentOrKeyword`, I should've caught that in the previous review.
Left another comment about test intentionality, but overall this is pretty good!

python-checks/src/test/resources/checks/datetime_constructor_pytz_timezone.py
ghislainpiot: This test was mostly to make sure that the ReachingDefinitionsAnalysis was returning the correct symbol, even if the value is overwritten.

python-checks/src/test/resources/checks/datetime_constructor_pytz_timezone.py
```diff
@@ -36,7 +38,9 @@ def with_tuple_unpacking():
     date_tuple = [2024, 2, 28]
     some_timezone_2 = pytz.timezone('US/Eastern')
     dt9 = datetime.datetime(*date_tuple, tzinfo=some_timezone_2)  # Noncompliant
-    dt9 = datetime.datetime(*date_tuple, tzinfo=pytz.timezone('US/Eastern'))  # Noncompliant
+    dt9_1 = datetime.datetime(*date_tuple, tzinfo=pytz.timezone('US/Eastern'))  # Noncompliant
+    full_date_tuple = [2024, 2, 28, 5, 5, 5, 1]
+    dt9_2 = datetime.datetime(*full_date_tuple, some_timezone_2)  # FN because nthArgumentOrKeyword does not take into account the size of the tuple
```
guillaume-dequenne-sonarsource, line 43: Good one!

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1732, opened 2024-02-29
SONARPY-1685 Fix FP on S6742 when DataFrame class is not a pandas's l ibrary class

maksim-grebeniuk-sonarsource:


joke1196 (approved):
LGTM! Thank you for taking care of this!

PR #1733, opened 2024-02-29
SONARPY-1636 S6903: Using timezone aware "datetime" objects should be preferred over using "datetime.datetime.utcnow" and "datetime.datetime.utcfromtimestamp"

ghislainpiot:


python-checks/src/main/java/org/sonar/python/checks/UseTimeZoneAwareDateTimeConstructorsCheck.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Set;
+import org.sonar.check.Rule;
+import org.sonar.plugins.python.api.PythonSubscriptionCheck;
+import org.sonar.plugins.python.api.SubscriptionContext;
+import org.sonar.plugins.python.api.symbols.Symbol;
+import org.sonar.plugins.python.api.tree.CallExpression;
+import org.sonar.plugins.python.api.tree.Tree;
+
+@Rule(key = "S6903")
+public class UseTimeZoneAwareDateTimeConstructorsCheck extends PythonSubscriptionCheck {
```
guillaume-dequenne-sonarsource, line 31: Similar to previous review, I'd suggest updating the name to highlight the pattern the rule is trying to detect rather than the one it's promoting.
`TimezoneNaiveDatetimeConstructorsCheck` or `TimezoneNaiveDatetimeFunctionsCheck` (I'm not sure whether these functions strictly are constructor, though it might still be the most understandable name).

python-checks/src/main/java/org/sonar/python/checks/UseTimeZoneAwareDateTimeConstructorsCheck.java
```diff
@@ -0,0 +1,55 @@
...
+  private static final String MESSAGE = "Using timezone aware \"datetime\"s should be preferred over using \"datetime.datetime.utcnow\" and \"datetime.datetime.utcfromtimestamp\"";
```
guillaume-dequenne-sonarsource, line 32: The message should be imperative (sorry it's not part of the specification, it really should be...).
I'd suggest:
Don't use `{FQN}` to create this `datetime` object. 

Also now that I think about it a bit more, I have the feeling that the real message we're trying to convey is that `datetime.datetime.utcnow` and `datetime.datetime.utcfromtimestamp` create timezone-naive objects despite the fact their name imply timezone-aware objects. 

It could very well be that it's perfectly intentional to create a timezone-naive object (and doing otherwise could lead to bugs if compared with other timezone-naive objects). The problem is that the intention is unclear from the method name. Maybe we should discuss that and consider changing the rule title accordingly.

python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S6903.json
```diff
@@ -0,0 +1,22 @@
+{
+  "title": "Using timezone aware \"datetime\"s should be preferred over using \"datetime.datetime.utcnow\" and \"datetime.datetime.utcfromtimestamp\"",
```
guillaume-dequenne-sonarsource, line 2: There was an unaddressed comment on the RSPEC. I made some minor changes. You should probably regenerate the metadata to reflect that.

python-checks/src/test/resources/checks/datetime_constructor_use_timezone_aware.py
```diff
@@ -0,0 +1,40 @@
...
+    from datetime import datetime as dt
+    dt.utcnow() # Noncompliant
+    dt.utcfromtimestamp(timestamp) # Noncompliant
+
+def unknown_symbols():
+
+    unrelated_call()
+
+    datetime.datetime.utcnow() # FN because there is no symbol
+    datetime.datetime.utcfromtimestamp() # FN because there is no symbol
```
guillaume-dequenne-sonarsource, line 24: If the symbols are unknown because not imported, I think it's fair to not report any issue: they could be anything at this point. So it's not FN (FN kind of implies we'd like to raise an issue here - which we don't).

python-checks/src/test/resources/checks/datetime_constructor_use_timezone_aware.py
```diff
@@ -0,0 +1,40 @@
...
+
+    datetime.now(timezone.utc)
+    timestamp = 1571595618.0
+    datetime.fromtimestamp(timestamp, timezone.utc)
+
+def aliased_utcnow():
+    from datetime import datetime
+    aliased = datetime.utcnow
+    # FN because we lose the FQN in the assignment
+    aliased()
```
guillaume-dequenne-sonarsource, line 37: This is quite nitpicky but to me, aliased would be something like:
```
from datetime.datetime import utcnow as aliased
aliased() # Noncompliant
```
Which we'd most likely raise correctly.
Here the symbol is reassigned (and it's indeed FN in that case, good thing to document it!)

guillaume-dequenne-sonarsource (changes requested):
This looks good!
Requesting changes as there are some very minor changes that need to be applied.
There are other small comments I'd like to discuss (e.g the rule title one).

python-checks/src/main/java/org/sonar/python/checks/TimezoneNaiveDatetimeConstructorsCheck.java
```diff
@@ -0,0 +1,53 @@
...
+import org.sonar.check.Rule;
+import org.sonar.plugins.python.api.PythonSubscriptionCheck;
+import org.sonar.plugins.python.api.SubscriptionContext;
+import org.sonar.plugins.python.api.symbols.Symbol;
+import org.sonar.plugins.python.api.tree.CallExpression;
+import org.sonar.plugins.python.api.tree.Tree;
+
+@Rule(key = "S6903")
+public class TimezoneNaiveDatetimeConstructorsCheck extends PythonSubscriptionCheck {
+  private static final String MESSAGE = "Don't use %s to create this datetime object.";
```
guillaume-dequenne-sonarsource, line 32: Sorry I forgot it in my previous comment, we generally use backticks around code snippets, especially if highlighting the code that the developer wrote.
So I'd suggest: "Don't use \`%s\` to create this datetime object."

guillaume-dequenne-sonarsource (approved):
Looks good!
One last comment about message formatting, sorry about that.

PR #1734, opened 2024-03-04
SONARPY-1639 S6882: Constructor attributes of date and time objects should be in the range of possible values

ghislainpiot:


python-checks/src/test/resources/checks/datetime_constructor_values.py
```diff
@@ -0,0 +1,85 @@
...
+
+    datetime.datetime(2024, 3, 1, 12, 30, 0, True)
+    datetime.datetime(2024, 3, 1, 12, 30, True, 0)
+    datetime.datetime(2024, 3, 1, 12, True, 0, 0)
+    datetime.datetime(2024, 3, 1, True, 30, 0, 0)
+    datetime.datetime(2024, 3, True, 12, 30, 0, 0)
+    datetime.datetime(2024, True, 1, 12, 30, 0, 0)
+    datetime.datetime(True, 3, 1, 12, 30, 0, 0)
+
+    datetime.time(-random(), 12, 5, 0)
```
guillaume-dequenne-sonarsource, line 85: _Nitpick_ Newline at the end of file.

python-checks/src/main/java/org/sonar/python/checks/ProvideCorrectParameterDateTimeConstructorsCheck.java
```diff
@@ -0,0 +1,117 @@
...
+    if (secondArgument != null) {
+      checkArgument(secondArgument, 0, 59, context, "second");
+    }
+    if (microsecondArgument != null) {
+      checkArgument(microsecondArgument, 0, 999_999, context, "microsecond");
+    }
+  }
+
+  private static void checkArgument(RegularArgument argument, long min, long max, SubscriptionContext context, String name) {
+    if (!argument.expression().is(Tree.Kind.NUMERIC_LITERAL) && !argument.expression().is(Tree.Kind.UNARY_MINUS)) {
```
guillaume-dequenne-sonarsource, line 85: You can probably cover some more cases by calling `Expressions#singleAssignedNonNameValue` or `Expressions#singleAssignedValue` without too much overhead when the argument expression is a `NAME`.
If you decide to cover this case, it may make sense to have a primary location on the argument itself and a secondary on the value with message such as e.g `"An invalid value is assigned here."`

python-checks/src/main/java/org/sonar/python/checks/ProvideCorrectParameterDateTimeConstructorsCheck.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+    if (secondArgument != null) {
+      checkArgument(secondArgument, 0, 59, context, "second");
+    }
+    if (microsecondArgument != null) {
+      checkArgument(microsecondArgument, 0, 999_999, context, "microsecond");
+    }
+  }
+
+  private static void checkArgument(RegularArgument argument, long min, long max, SubscriptionContext context, String name) {
```
guillaume-dequenne-sonarsource, line 84: _Super nitpick:_ For consistency it's good to have the `SubscriptionContext` first in the parameter list (it's not always the case though and I wouldn't ask to change this - this is just so you know of some common practices).

guillaume-dequenne-sonarsource (changes requested):
Looks good!
I left a few comments/suggestions but only the issue message is really important one.

I think it would also be nice to use `Expressions#singleAssignedValue` to cover some more dataflow case, but we can discuss this.
In any case, I think it would also make sense to have some tests to document the behaviour when the argument is not a literal.

maksim-grebeniuk-sonarsource (approved):
LGTM

PR #1735, opened 2024-03-05
SONARPY-1637 tmp

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/PandasToDatetimeFormatCheck.java
```diff
@@ -0,0 +1,267 @@
...
+
+  private static ParseResult parseResult(String normalizedDateString, ParseResult expected) {
+    boolean parsedOnce = false;
+    for (Map.Entry<String, ParseResult> entry : DATE_FORMATS.entrySet()) {
+      if (expected.areAllConditionsMet()) {
+        return expected;
+      }
+      String dateFormat = entry.getKey();
+      ParseResult parseResult = entry.getValue();
+      DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern(dateFormat);
```
ghislainpiot, line 204: I don't know about the performance impact, but it might be interesting to create all the DateTimeFormatter only once.

python-checks/src/main/java/org/sonar/python/checks/PandasToDatetimeFormatCheck.java
```diff
@@ -0,0 +1,267 @@
...
+
+  static final String PANDAS_TO_DATETIME_FQN = "pandas.core.tools.datetimes.to_datetime";
+  static final String MESSAGE = "Remove this `%s=%s` parameter or make sure the provided date(s) can be parsed accordingly.";
+  static final String SECONDARY_MESSAGE = "Invalid date.";
+  static final String DAYFIRST = "dayfirst";
+  static final String YEARFIRST = "yearfirst";
+
+  static final Map<String, ParseResult> DATE_FORMATS = new HashMap<>();
+
+  static {
```
ghislainpiot, line 59: I think it would be helpful to add a comment saying that the formats are normalized to this form, and therefore the forms with / as separators are included

python-checks/src/test/resources/checks/pandasToDatetimeFormatCheck.py
```diff
@@ -0,0 +1,155 @@
...
+    pd.to_datetime("02.03.2022", yearfirst=True) # Noncompliant
+    pd.to_datetime("02;03;2022", yearfirst=True) # Noncompliant
+    pd.to_datetime("02_03_2022", yearfirst=True) # Noncompliant
+
+
+def list_argument_noncompliant():
+    pd.to_datetime(["02-03-2022"], yearfirst=True) # Noncompliant
+    pd.to_datetime(
+        ["02-03-2022", "03-03-2022"],
+    #   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^> {{This contains invalid date(s).}}
```
ghislainpiot, line 23: I think it would be better to assert that there is only 1 secondary location, instead of the 2 the user might expect

ghislainpiot (commented):
It looks pretty good to me, I think my concerns about performance might only be valid in edge cases.

python-checks/src/test/resources/checks/pandasToDatetimeFormatCheck.py
guillaume-dequenne-sonarsource: There are actually 2 secondary locations (the assertion for the second one is on the line below with a `@-1` specifier). I figure if the list is long, showing an example of the invalid literal is still interesting.
I added some assertions on the order of secondary locations to make things more explicit.

ghislainpiot (approved):
Looks good to me !

PR #1738, opened 2024-03-07
SONARPY-1644 Rule S6890: zoneinfo should be preferred to pytz when using Python 3.9 and later

ghislainpiot:


python-checks/src/main/java/org/sonar/python/checks/PytzDontUseAfterPy39Check.java
```diff
@@ -0,0 +1,79 @@
...
+    Stream<DottedName> dottedNames = context.syntaxNode().is(Tree.Kind.IMPORT_FROM) ? ((ImportFrom) context.syntaxNode()).importedNames().stream().map(AliasedName::dottedName)
+      : ((ImportName) context.syntaxNode()).modules().stream().map(AliasedName::dottedName);
+    checkDottedNamesGetLocations(context, dottedNames);
+  }
+
+  private static void checkDottedNamesGetLocations(SubscriptionContext context, Stream<DottedName> dottedNames) {
+    dottedNames
+      .flatMap(dottedName -> dottedName.names().stream())
+      .filter(dottedName -> "pytz".equals(dottedName
+        .name()) || Optional.ofNullable(dottedName.symbol())
```
maksim-grebeniuk-sonarsource, line 65: I think it is better to have `.name())` on the previous line

python-checks/src/main/java/org/sonar/python/checks/PytzDontUseAfterPy39Check.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    Stream<DottedName> dottedNames = context.syntaxNode().is(Tree.Kind.IMPORT_FROM) ? ((ImportFrom) context.syntaxNode()).importedNames().stream().map(AliasedName::dottedName)
+      : ((ImportName) context.syntaxNode()).modules().stream().map(AliasedName::dottedName);
+    checkDottedNamesGetLocations(context, dottedNames);
+  }
+
+  private static void checkDottedNamesGetLocations(SubscriptionContext context, Stream<DottedName> dottedNames) {
+    dottedNames
+      .flatMap(dottedName -> dottedName.names().stream())
```
maksim-grebeniuk-sonarsource, line 63: this solution has FP when there is `pytz` package in another library like:
```python
import something.else.pytz as p
```

python-checks/src/main/java/org/sonar/python/checks/PytzDontUseAfterPy39Check.java
```diff
@@ -0,0 +1,79 @@
...
+    return context.sourcePythonVersions().stream()
+      .allMatch(version -> version.compare(REQUIRED_VERSION.major(), REQUIRED_VERSION.minor()) >= 0);
+  }
+
+  private static void checkImport(SubscriptionContext context) {
+    if (!isRelevantPythonVersion(context)) {
+      return;
+    }
+
+    Stream<DottedName> dottedNames = context.syntaxNode().is(Tree.Kind.IMPORT_FROM) ? ((ImportFrom) context.syntaxNode()).importedNames().stream().map(AliasedName::dottedName)
```
maksim-grebeniuk-sonarsource, line 56: I'd suggest to have just single stream chain to get the tree elements for validation like:
```java
Stream.of(
      Optional.of(context.syntaxNode())
        .flatMap(TreeUtils.toOptionalInstanceOfMapper(ImportFrom.class))
        .map(ImportFrom::importedNames),
      Optional.of(context.syntaxNode())
        .flatMap(TreeUtils.toOptionalInstanceOfMapper(ImportName.class))
        .map(ImportName::modules)
    )
      .filter(Optional::isPresent)
      .map(Optional::get)
      .flatMap(Collection::stream)
      .map(AliasedName::dottedName)
      .map(DottedName::names)
      .filter(Predicate.not(List::isEmpty))
      .map(names -> names.get(0))
      .filter(name -> "pytz".equals(name.name()) 
        || Optional.ofNullable(name.symbol())
        .filter(fqn -> fqn.startsWith("pytz")).isPresent())
      .forEach(name -> raiseIssue(context, name));
```

maksim-grebeniuk-sonarsource (changes requested):
Generally looks good, one critical issue found, see the comments

python-checks/src/main/java/org/sonar/python/checks/PytzDontUseAfterPy39Check.java
```diff
@@ -0,0 +1,79 @@
...
+import org.sonar.plugins.python.api.SubscriptionContext;
+import org.sonar.plugins.python.api.symbols.Symbol;
+import org.sonar.plugins.python.api.tree.AliasedName;
+import org.sonar.plugins.python.api.tree.DottedName;
+import org.sonar.plugins.python.api.tree.ImportFrom;
+import org.sonar.plugins.python.api.tree.ImportName;
+import org.sonar.plugins.python.api.tree.Tree;
+
+@Rule(key = "S6890")
+public class PytzDontUseAfterPy39Check extends PythonSubscriptionCheck {
```
maksim-grebeniuk-sonarsource, line 36: Maybe the check name could be just like `PytzUsageCheck`

maksim-grebeniuk-sonarsource (approved):
LGTM

PR #1741, opened 2024-03-11
SONARPY-1649 S6908: tf.function should not be recursive

ghislainpiot:


python-checks/src/main/java/org/sonar/python/checks/TfFunctionRecursivityCheck.java
```diff
@@ -0,0 +1,114 @@
...
+    collector.expressionList.forEach(call -> issue.secondary(call, SECONDARY_MESSAGE));
+  }
+
+  private static boolean isTfFunction(FunctionDef functionDefinition) {
+    return functionDefinition.decorators().stream()
+      .map(Decorator::expression).map(TreeUtils::getSymbolFromTree)
+      .filter(Optional::isPresent).map(Optional::get)
+      .map(Symbol::fullyQualifiedName)
+      .filter(Objects::nonNull)
+      .anyMatch("tensorflow.function"::equals);
```
guillaume-dequenne-sonarsource, line 77: I must say I don't really like this construct, as this kind of construct has caused some NPEs in production in the past.

In particular, I believe the `Objects::nonNull` filter is not actually covered in the tests (no `null` object ever gets through it, although I believe it can be triggered). If you forgot the filter and performed an unsafe operation afterward, you could get a NPE that may go unseen until we reach production. I believe this code is actually safe, but it is quite implicit and untested.

I believe extracting the `isTfFunctionDecorator` bit in a dedicated method (as well as adding the relevant tests for other kind of decorators) may help achieving safer code. It would also avoid the awkward `Optional::isPresent / Optional::get` part. Something that would look like:

```
private static boolean isTfFunctionDecorator(Decorator decorator) {
    return Optional.of(decorator.expression())
      .flatMap(TreeUtils::getSymbolFromTree)
      .filter("tensorflow.function"::equals)
      .isPresent();
  }
```

(Note that even with this implementation, we have blind spots as to which of these steps can actually return `null` and we shouldn't just rely on the 100% coverage to leave these cases uncovered)

python-checks/src/main/java/org/sonar/python/checks/TfFunctionRecursivityCheck.java
```diff
@@ -0,0 +1,114 @@
...
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.FUNCDEF, TfFunctionRecursivityCheck::checkFunctionDef);
+  }
+
+  private static void checkFunctionDef(SubscriptionContext context) {
+    FunctionDef functionDef = (FunctionDef) context.syntaxNode();
+    if (!isTfFunction(functionDef)) {
+      return;
+    }
+    var selfSymbol = TreeUtils.getFunctionSymbolFromDef(functionDef);
```
guillaume-dequenne-sonarsource, line 58: `selfSymbol` sort of implies the symbol behind the `self` keyword. I'd suggest something more straightforward such as `functionSymbol` to avoid ambiguity.

python-checks/src/main/java/org/sonar/python/checks/TfFunctionRecursivityCheck.java
```diff
@@ -0,0 +1,114 @@
...
+    if (selfSymbol == null) {
+      return;
+    }
+    var collector = new CallCollector(selfSymbol);
+    context.syntaxNode().accept(collector);
```
guillaume-dequenne-sonarsource, line 63: You can use the `functionDef` already defined here.

python-checks/src/main/java/org/sonar/python/checks/TfFunctionRecursivityCheck.java
```diff
@@ -0,0 +1,114 @@
...
+    @Override
+    public void visitCallExpression(CallExpression callExpression) {
+      Symbol symbol = TreeUtils.getSymbolFromTree(callExpression.callee()).orElse(null);
+      if (symbol == null) {
+        return;
+      }
+      if (symbol.equals(originalSymbol)) {
+        expressionList.add(callExpression.callee());
+      }
+      if (symbol.is(Symbol.Kind.FUNCTION)) {
```
guillaume-dequenne-sonarsource, line 99: I believe having a `else if` here fixes the `StackOverflowError` you've been having in the current tests.

Some can still be trigger if two functions that do not match the original symbol call each other. Still, I assume it is not the intention to visit the same function declaration again.

python-checks/src/main/java/org/sonar/python/checks/TfFunctionRecursivityCheck.java
```diff
@@ -0,0 +1,114 @@
...
+        symbol.usages().stream().filter(usage -> usage.kind() == (Usage.Kind.FUNC_DECLARATION)).forEach(usage -> usage.tree().parent().accept(this));
```
guillaume-dequenne-sonarsource, line 100: I guess this (in combination with the `visited` list) works though I'm not 100% sure we care that much about indirect recursion (I think keeping it as an accepted FN would have been a good baby step) that we're going to visit all called functions for which we can find a definition. 

I'm not against it though as I guess the performance impact will be limited, given how narrow the scope of the rule is. Still, I'm a bit wondering whether this could cause FPs in some ways (with direct recursion, we can always assume a path is either dead code or can lead to the actual recursion - with indirect recursion, maybe the path that leads to the recursive code is neither dead code nor an actually possible path).

python-checks/src/test/resources/checks/tfFunctionRecursivity.py
```diff
@@ -0,0 +1,75 @@
...
+
+@tf.function
+def indirect_rec1(n): # Noncompliant {{Remove this recursive call.}}
+   #^^^^^^^^^^^^^
+    return indirect_rec2(n)
+          #^^^^^^^^^^^^^> {{Recursive call is here.}}
+@tf.function
+def indirect_rec2(n): # Noncompliant {{Remove this recursive call.}}
+   #^^^^^^^^^^^^^
+    return indirect_rec1(n)
```
guillaume-dequenne-sonarsource, line 33: I think this should stay out of scope given the complexity it would add but another secondary such as `Indirect recursive call.` would be really nice here if we're going to support indirect recursion.

guillaume-dequenne-sonarsource (changes requested):
Looking good!

I left some improvements comment, as well as some more generic comments that may not need addressing (e.g comments on the support of indirect recursion).
Happy to discuss them.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1747, opened 2024-03-14
SONARPY-1652 S6918: "tf.Variable" objects should be singletons when c…

guillaume-dequenne-sonarsource:
…reated inside of a "tf.function"


python-checks/src/main/java/org/sonar/python/checks/NonSingletonTfVariableCheck.java
```diff
@@ -0,0 +1,59 @@
...
+      return;
+    }
+    if (isUnconditional(callExpression)) {
+      ctx.addIssue(callExpression, "Refactor this variable declaration to be a singleton.");
+    }
+  }
+
+  private static boolean isUnconditional(CallExpression callExpression) {
+    Tree firstAncestor = TreeUtils.firstAncestorOfKind(callExpression, Tree.Kind.FUNCDEF, Tree.Kind.IF_STMT, Tree.Kind.CONDITIONAL_EXPR);
+    return firstAncestor != null && firstAncestor.is(Tree.Kind.FUNCDEF);
```
guillaume-dequenne-sonarsource, line 53: I believe `firstAncestor` can't ever be null here, as we already ensured we are within a `tf.function`. It is purely defensive.

PR #1748, opened 2024-03-15
SONARPY-1654 S6919: The input_shape parameter should not be specified for tf.keras.Model subclasses

ghislainpiot:


guillaume-dequenne-sonarsource (changes requested):
My comments are mostly stylistic but I still think it is worth considering/discussing, as I believe this can be made more readable and maintainable.

Implementation looks good otherwise. I just want to promote good habits to avoid issues when dealing with more complex cases. Happy to be challenged though.

guillaume-dequenne-sonarsource (approved):
LGTM.

PR #1749, opened 2024-03-19
SONARPY-1650 S6911: tf.function should not depend on global or free python var

ghislainpiot:


python-checks/src/main/java/org/sonar/python/checks/TfFunctionDependOnOutsideVariableCheck.java
```diff
@@ -0,0 +1,83 @@
...
+import org.sonar.plugins.python.api.symbols.Usage;
+import org.sonar.plugins.python.api.tree.BaseTreeVisitor;
+import org.sonar.plugins.python.api.tree.FunctionDef;
+import org.sonar.plugins.python.api.tree.Name;
+import org.sonar.plugins.python.api.tree.Tree;
+import org.sonar.python.tree.TreeUtils;
+
+@Rule(key = "S6911")
+public class TfFunctionDependOnOutsideVariableCheck extends PythonSubscriptionCheck {
+  private static final String MESSAGE = "This function should not depend implicitly on a global or free variable.";
```
guillaume-dequenne-sonarsource, line 42: To make the message imperative I'd suggest:
`Make sure this function does not depend on a global or free variable.`

python-checks/src/main/java/org/sonar/python/checks/TfFunctionDependOnOutsideVariableCheck.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Override
+    public void visitName(Name pyNameTree) {
+      if (pyNameTree.isVariable()) {
+        Optional.ofNullable(pyNameTree.symbol())
+          .filter(symbol -> !symbol.is(Symbol.Kind.FUNCTION, Symbol.Kind.CLASS))
+          .filter(symbol -> symbol.usages().stream().map(Usage::kind).noneMatch(usageKind -> usageKind == Usage.Kind.IMPORT))
+          .ifPresent(symbol -> {
+            symbolToNames.putIfAbsent(symbol, new ArrayList<>());
+            symbolToNames.get(symbol).add(pyNameTree);
```
guillaume-dequenne-sonarsource, line 78: I believe you can use `computeIfAbsent` to achieve this in a single line.

python-checks/src/test/resources/checks/tfFunctionDependOnOutsideVariable.py
```diff
@@ -0,0 +1,64 @@
...
+def non_compliant2(): # Noncompliant {{This function should not depend implicitly on a global or free variable.}}
+    something_else = no_symbol
+    return something_else + 1
+
+@tf.function
+def non_compliant3(): # Noncompliant
+   #^^^^^^^^^^^^^^
+    something = foo
+               #^^^< {{Variable used here.}}
+    smth = no_symbol
```
guillaume-dequenne-sonarsource, line 23: Since `no_symbol` may be a function, do we really want it to raise an issue/be a secondary location?
I feel we shouldn't raise when we are uncertain of what the symbol represents (ideally only raising on "pure" Python variables).

python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S6911.html
```diff
@@ -0,0 +1,78 @@
...
+
+foo = 10
+addition() # tf.Tensor(5, shape=(), dtype=int32): unexpected result of 5 instead of 11
+</pre>
+<p>As we can see in the example above the second time <code>addition</code> is called, we obtain the same result as the first call. This is due to the
+fact that between the 2 calls of <code>addition</code> the value of the argument passed to the function did not change. This result in the creation of
+a single <code>ConcreteFunction</code> during the first call of <code>addition</code>, with the value of foo set to 4.</p>
+<p>This is why it is a good practice to not use and mutate global variables or nonlocal variables inside of a <code>tensorflow.function</code>.</p>
+<h3>Exceptions</h3>
+<p>This rule will not raise an issue if the global or nonlocal variable is a <code>tensorflow.Variable</code>.</p>
```
guillaume-dequenne-sonarsource, line 24: I believe this point of the RSPEC is not covered. We should make sure no issue is raised in such case (probably by checking if a `tf.Variable` is explicitly assigned to the variable). It's likely we'll still have to accept FPs though.

its/ruling/src/test/resources/expected/python-S6911.json
```diff
@@ -0,0 +1,10 @@
+{
```
guillaume-dequenne-sonarsource, line 1: Are these TPs in your opinion?
Acceptable FPs? Non acceptable FPs?

If the issues are not TPs I'd expect a comment mentioning that you've reviewed them and that you think they're acceptable. If they're not, we should either reconsider the rule or at least its implementation.

I'm saying this because I have the impression these are FPs since these all are Tensorflow / Keras objects that are involved. Maybe the issue is still warranted as using parameters would be cleaner, though. But since the rule is about reliability of the execution, it's still not as bad as the rule suggests.

python-checks/src/main/resources/org/sonar/l10n/py/rules/python/S6911.json
```diff
@@ -0,0 +1,21 @@
+{
+  "title": "\"tf.function\"s should not depend on global or free Python variables",
```
guillaume-dequenne-sonarsource, line 2: Before merging the RSPEC and the rule, I think we should rephrase the title to avoid the awkward pluralization of `tf.function`, consistently with the other rules.

python-checks/src/test/java/org/sonar/python/checks/TfFunctionDependOnOutsideVariableCheckTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.junit.jupiter.api.Test;
+import org.sonar.python.checks.utils.PythonCheckVerifier;
+
+class TfFunctionDependOnOutsideVariableCheckTest {
+  @Test
+  void test() {
+    PythonCheckVerifier.verify("src/test/resources/checks/tfFunctionDependOnOutsideVariable.py", new TfFunctionDependOnOutsideVariableCheck());
+  }
+}
```
guillaume-dequenne-sonarsource, line 30: _Nitpick:_ newline.

python-checks/src/main/java/org/sonar/python/checks/TfFunctionDependOnOutsideVariableCheck.java
```diff
@@ -0,0 +1,83 @@
...
+
+  private static class NameCollector extends BaseTreeVisitor {
+    Map<Symbol, List<Name>> symbolToNames = new HashMap<>();
+
+    @Override
+    public void visitName(Name pyNameTree) {
+      if (pyNameTree.isVariable()) {
+        Optional.ofNullable(pyNameTree.symbol())
+          .filter(symbol -> !symbol.is(Symbol.Kind.FUNCTION, Symbol.Kind.CLASS))
+          .filter(symbol -> symbol.usages().stream().map(Usage::kind).noneMatch(usageKind -> usageKind == Usage.Kind.IMPORT))
```
guillaume-dequenne-sonarsource, line 75: Related to my other comments,  I feel we should be a bit more lenient here.
As we add conditions, maybe it will make sense to extract this in a dedicated method to better convey the goal of this filtering.

guillaume-dequenne-sonarsource (changes requested):
This looks good overall.
I do have questions on the ruling results which I'm not sure about. This relates also to the rule specification and the fact we probably want to be a bit more lenient about what we consider to be a free variable use,

its/ruling/src/test/resources/expected/python-S6911.json
```diff
@@ -1,10 +1,6 @@
 {
 "project:tensorflow/python/keras/integration_test/function_test.py": [
 168,
-191,
-221
-],
-"project:tensorflow/python/keras/integration_test/saved_model_test.py": [
-164
+191
```
guillaume-dequenne-sonarsource, line 4: What about those remaining issues? Again, if it's not TP, we should address or acknowledge it somehow. For example with a unit test, even if it's a documented FP we don't address (or just a bordeline case we actually want to raise an issue on - but the intention should be clear).

python-checks/src/main/java/org/sonar/python/checks/TfFunctionDependOnOutsideVariableCheck.java
```diff
@@ -55,29 +58,54 @@ private static void checkFunction(SubscriptionContext context) {
...
+        .filter(Objects::nonNull)
+        .map(AssignmentStatement.class::cast)
+        .map(AssignmentStatement::assignedValue)
+        .filter(assignedValue -> assignedValue.is(Tree.Kind.CALL_EXPR))
+        .map(CallExpression.class::cast)
+        .map(CallExpression::calleeSymbol)
+        .filter(Objects::nonNull)
+        .map(Symbol::fullyQualifiedName)
+        .filter(Objects::nonNull)
+        .anyMatch(fqn -> fqn.startsWith("tensorflow."));
```
guillaume-dequenne-sonarsource, line 108: I think we generally should avoid those long streams that do different things at once (here, there are clearly two parts - searching for assigned values and checking whether they are Tensorflow values - this could be split).

For instance, I think the highlighted bit could be factored into a `isTensorflowValue`, which would then be easier to reason about in terms of what we consider to be a tensorflow object.

With a dedicated method, it might then be extended to consider more expressions (the one from the ruling could be handled there, and we even have an `Expressions::removeParentheses` helper that could help here).

python-checks/src/test/resources/checks/tfFunctionDependOnOutsideVariable.py
```diff
@@ -62,3 +62,12 @@ def indirect_tf_function():
...
     return indirect_tf_function()
+
+def containing():
+    foo_tf = tf.Variable(3)
+    @tf.function
+    def compliant2():
+        tenf = tf.Variable(10)
+        return foo_tf + tenf
+
+    foo_tf = True
```
guillaume-dequenne-sonarsource, line 73: _Nitpick:_ newline.

guillaume-dequenne-sonarsource (changes requested):
Good job on this difficult rule.
I still have a few comments seeing there are still issues on the ruling that may not be clearly intentional. I left a small refactoring suggestion that may help in that sense.

its/ruling/src/test/resources/expected/python-S6911.json
ghislainpiot: I have added some extra logic to handle some cases, but I think the implementation is open to discussion

python-checks/src/main/java/org/sonar/python/checks/TfFunctionDependOnOutsideVariableCheck.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+    var issue = context.addIssue(functionDef.name(), MESSAGE);
+    allNames.forEach(symbol -> collector.symbolToNames.get(symbol).forEach(name -> issue.secondary(name, MESSAGE_SECONDARY)));
+  }
+
+  private static class NameCollector extends BaseTreeVisitor {
+    Map<Symbol, List<Name>> symbolToNames = new HashMap<>();
+
+    @Override
+    public void visitName(Name pyNameTree) {
```
guillaume-dequenne-sonarsource, line 77: Would it make sense for this method to have access of the local variables of the enclosing function and to early return if the current symbol belongs to the function?
Then we'd only look at usages for symbols that were actually defined outside the function.

guillaume-dequenne-sonarsource (approved):
Looks good! The implementation seems a bit heavy but I don't really think you have a better choice here.
I left a small suggestion: I think we can optimize this a bit by checking whether the name symbol is a local variable of the function before checking if it's a tensorflow object.

PR #1750, opened 2024-03-19
SONARPY-1730 Fix warning with sonar.python.version

ghislainpiot:


python-frontend/src/main/java/org/sonar/plugins/python/api/PythonVersionUtils.java
```diff
@@ -106,7 +106,9 @@ private PythonVersionUtils() {
   }
 
   public static Set<Version> fromString(String propertyValue) {
-    String[] versions = propertyValue.split(",");
+    return fromString(propertyValue.split(","));
+  }
+  public static Set<Version> fromString(String[] versions) {
```
guillaume-dequenne-sonarsource, line 111: Would it make sense to rename this to `fromArray`, `fromVersionArray`, `fromStringArray` or anything of the kind that would be more accurate? I find it weird that a method called `fromString` actually takes an array as parameter.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPynbSensor.java
```diff
@@ -63,9 +63,10 @@ public void describe(SensorDescriptor descriptor) {
   @Override
   public void execute(SensorContext context) {
     List<InputFile> pythonFiles = getInputFiles(context);
-    context.config().get(PYTHON_VERSION_KEY)
-      .map(PythonVersionUtils::fromString)
-      .ifPresent(ProjectPythonVersion::setCurrentVersions);
+    var pythonVersions = context.config().getStringArray(PYTHON_VERSION_KEY);
+    if (pythonVersions.length != 0) {
+      ProjectPythonVersion.setCurrentVersions(PythonVersionUtils.fromString(pythonVersions));
```
guillaume-dequenne-sonarsource, line 68: This goes untested, yet I think we can have a (very simple) test in `IPynbSensorTest` that does something similar to what is done in `PythonSensorTest#test_python_version_parameter_no_warning` (setting the property in the context before executing the sensor test) so that this would be well tested. (It would not exactly be the same test though, as there is no warning on SonarLint for missing version property).

Note that the missing test probably simply went unnoticed due to the optional chain that was used before.

guillaume-dequenne-sonarsource (approved):
Looks good! Thanks for handling this.
About the missing test - I think it would be nice to add it but I won't block the PR for it since it predates this PR.
If you end up adding it, please put back in review though.

guillaume-dequenne-sonarsource (approved):
LGTM, thanks for the additional test!

PR #1751, opened 2024-03-20
SONARPY-1732 Fix non-determinism on S2638 when positional-only parameters have no name available

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/SymbolUtils.java
```diff
@@ -306,12 +306,23 @@ public static List<FunctionSymbol> getFunctionSymbols(@Nullable Symbol symbol) {
 
   public static Optional<FunctionSymbol> getFirstAlternativeIfEqualArgumentNames(List<FunctionSymbol> alternatives) {
     return Optional.of(alternatives)
+      .filter(SymbolUtils::haveSameParameterCount)
```
ghislainpiot, line 309: I think it would be interesting to see if it is possible to test the behavior of this function in `SymbolUtilsTest` instead of in `ChangeMethodContractCheckTest`.

ghislainpiot (approved):
The fix looks good to me! Maybe it is possible to make a new test so it is easier to see the impact of any other change to SymbolUtils.

python-frontend/src/main/java/org/sonar/python/semantic/SymbolUtils.java
guillaume-dequenne-sonarsource: Good point! And in fact, coming back to this, I realize we could also have possible unwanted behavior if the functions define the same parameters in a different order (in which case we'd want to be conservative instead of just picking one alternative randomly).

I've updated the implementation to compare lists of parameter names instead of sets, which will make sure we take ordering and duplicates into account.

I've added tests accordingly. Note that it is not really possible to create a case with duplicate parameter names (as it is not valid Python). This only happens when we get the information from Typeshed and the parameter is positional only. I therefore reused `dict` inheritance to test this.

ghislainpiot (approved):
Looks good to me !

PR #1754, opened 2024-03-22
SONARPY 1745b

ghislainpiot:
- **SONARPY-1745 Update instance casts**
- **Update some Collectors#toList into toList**
- **Fix unused imports**
- **Update switch statements to new syntax**
- **Update .collect(Collectors.toList()) to .toList()**

### DIFF
--- a/python-checks/src/main/java/org/sonar/python/checks/ChildAndParentExceptionCaughtCheck.java
+++ b/python-checks/src/main/java/org/sonar/python/checks/ChildAndParentExceptionCaughtCheck.java
@@ -104,7 +104,7 @@ public class ChildAndParentExceptionCaughtCheck extends PythonSubscriptionCheck 
         .stream()
         .map(ChildAndParentExceptionCaughtCheck::collectNames)
         .flatMap(Collection::stream)
-        .toList();
+        .collect(Collectors.toList());
     }
     throw new IllegalArgumentException("Unsupported kind of tree element: " + expression.getKind().name());
   }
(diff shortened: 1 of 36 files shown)


python-checks/src/main/java/org/sonar/python/checks/ChildAndParentExceptionCaughtCheck.java
```diff
@@ -132,7 +132,7 @@ private static PythonQuickFix createQuickFix(Expression currentException) {
         .map(ExceptClause.class::cast)
         .map(ExceptClause::exception)
         .map(exceptions -> {
-          List<String> names = collectNamesFromTuple(exceptions);
+          List<String> names = new ArrayList<>(collectNamesFromTuple(exceptions));
```
guillaume-dequenne-sonarsource, line 135: In this case, wouldn't it be better to keep `Collectors.toList()` and flag the issue as FP?
It feels a bit convoluted to create an immutable list which is only be used to re-create a mutable one.

guillaume-dequenne-sonarsource (approved):
LGTM, thanks for handling this.
Left one comment about the case where the list is modified after. I feel previous behavior was okay in this case. No strong feeling about it though.

python-checks/src/main/java/org/sonar/python/checks/ChildAndParentExceptionCaughtCheck.java
ghislainpiot: Yes, let's do that

PR #1762, opened 2024-04-08
SONARPY-1747 Fix incorrect type equality when classes vary in metaclass status

guillaume-dequenne-sonarsource:
It seems `RuntimeType#equals` relies on a somewhat loose mechanism to determine that classes are equal, while `equals` is not implemented on `Symbol` level.

I made sure that `metaclass`, `decorator` and type hierarchy resolution status are accounted for when performing this check. 
I think it would be better to have proper `equals` implemented on `ClassSymbol`. However, it seems we actually rely on the loose comparison to implicitly merge almost-identical symbols.  Given we are in the process of refactoring our type inference model and engine, I think the quick-fix makes sense.


maksim-grebeniuk-sonarsource (approved):
LGTM

PR #1765, opened 2024-04-17
SONARPY-1760 Infer types of imported serialized module

maksim-grebeniuk-sonarsource:


python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -57,6 +60,63 @@ void test() {
     System.out.println("hello");
   }
 
+  @Test
+  void testTypeshedImports() {
```
guillaume-dequenne-sonarsource, line 64: _Nitpick_ to make our tests more robust I think it would be better to avoid relying too much on statement indexes when parsing a different file.
I think we can also use `PythonTestUtils#parse` or even `parseWithoutSymbols` with code provided as a text block to create the `FileInput` and visit it rather than needing an actual file for the simpler cases.

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -57,6 +60,63 @@ void test() {
...
+
+    {
+      var statement = fileInput.statements().statements().get(0);
+      var importedNames = statement.children().get(1).children().get(0).children();
+      {
+        var item = importedNames.get(0);
+        Assertions.assertInstanceOf(Name.class, item);
+        var name = (Name) item;
+        var type = name.typeV2();
+        Assertions.assertInstanceOf(ModuleType.class, type);
```
guillaume-dequenne-sonarsource, line 82: I guess we can also assert by name of the module here on top of the fact it's a `ModuleType`

python-frontend/src/test/java/org/sonar/python/semantic/v2/SymbolTableBuilderV2Test.java
```diff
@@ -37,13 +37,13 @@ class SymbolTableBuilderV2Test {
 
   @BeforeAll
   static void init() {
-    var context = TestPythonVisitorRunner.createContext(new File("src/test/resources/semantic/v2/script.py"));
+    var context = TestPythonVisitorRunner.createContext(new File("src/test/resources/semantic/v2/symbol_table.py"));
```
guillaume-dequenne-sonarsource, line 40: I think your branch and mine have a conflict on how we dealt with this issue. I decided to inline the code in a text-block, which I feel makes it clearer the snippet should not be changed. Wdyt of this approach?

python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeInferenceV2.java
```diff
@@ -64,7 +65,7 @@ public TypeInferenceV2(ProjectLevelTypeTable projectLevelTypeTable) {
 
   @Override
   public void visitFileInput(FileInput fileInput) {
-    var type = new ModuleType("somehow get its name", new HashMap<>());
+    var type = new ModuleType("somehow get its name");
```
guillaume-dequenne-sonarsource, line 68: Can you maybe create a small ticket for this?
I think basically we need to get the `PythonFile` here to fix this.

python-frontend/src/main/java/org/sonar/python/semantic/v2/ProjectLevelTypeTable.java
```diff
@@ -42,48 +42,76 @@
...
   }
 
-  private PythonType convertToType(Symbol symbol) {
-    return switch (symbol.kind()) {
-      case CLASS -> converToClassType((ClassSymbol) symbol);
-      case FUNCTION -> convertToFunctionType((FunctionSymbol) symbol);
-      case AMBIGUOUS -> convertToUnionType((AmbiguousSymbol) symbol);
-      case OTHER -> converToObjectType(symbol);
-    };
+  private void initBuiltinModule() {
```
guillaume-dequenne-sonarsource, line 92: I do have some doubts about how we eventually merge stub symbols from project symbols in the project table.

We used to have a somewhat clear order of priority in how we resolve symbols (project symbols > custom typeshed > stdlib typeshed) that I feel we may lose. I'd be tempted to have a project `ModuleType` and a stub `ModuleType` instead, wdyt?

Still, I think it's a good idea to centralize those symbols/types in the project table (as opposed to the previous implementation where we had project symbols in the project table and Typeshed symbols in the Typeshed class, sequentially looking in each location at the import resolution).

python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeInferenceV2.java
maksim-grebeniuk-sonarsource: Good catch, I've created the [SONARPY-1778](https://sonarsource.atlassian.net/browse/SONARPY-1778)

python-frontend/src/main/java/org/sonar/python/semantic/v2/ProjectLevelTypeTable.java
maksim-grebeniuk-sonarsource: As discussed the priority should be handled not as different storages but on type population stage. 
Maybe it will make sense to introduce a type source field for PythonType but I suggest adding it only once it is really used

guillaume-dequenne-sonarsource: That makes sense. I must say I'm still partial to having clearly separated storages, but at the moment I can't really express a compelling "why" (it's just a gut feeling).

Let's keep things as they are now and keep this option as a possibility if we face limitations.

As I'm typing this, one thing we may want to double-check is how PR analysis caching works with project symbols vs. stub symbols. I tend to think we don't want to cache stub symbols (due to the network cost and them being embedded in the analyzer anyway) however I know we had problems in the past with this (because stub symbols depend on which files are actually parsed). We don't need to act on this as of now, this is just a consideration I wanted to share.

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -55,6 +68,116 @@ void test() {
...
 
+  @Test
+  void testTypeshedImports() {
+    FileInput root = inferTypes("""
+      import cryptography.hazmat.primitives.asymmetric
+      import datetime
+      a = datetime.date(year=2023, month=12, day=1)
+      """);
+
+    var importName = (ImportName) root.statements().statements().get(0);
```
guillaume-dequenne-sonarsource, line 79: Nice job on the test refactoring, this looks much cleaner and maintainable.

I think we should remove the base `test` that relies on the `script.py` file though (assuming we've migrated all of its use cases to other tests). 
In this PR we could leave it there as a reference if not all cases have dedicated tests, but we shouldn't forget to remove it later.

python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeInferenceV2.java
```diff
@@ -141,8 +145,44 @@ private FunctionType buildFunctionType(FunctionDef functionDef) {
...
+          setTypeToName(names.get(i), module);
+          module = Optional.ofNullable(module)
+            .map(ModuleType::parent)
+            .orElse(null);
+        }
+      });
+  }
+
+  @Override
+  public void visitImportFrom(ImportFrom importFrom) {
```
guillaume-dequenne-sonarsource, line 167: We should probably have a ticket about wildcard imports and make sure we don't have issues with them in the current implementation.

python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeInferenceV2.java
```diff
@@ -141,8 +145,44 @@ private FunctionType buildFunctionType(FunctionDef functionDef) {
...
+
+    var module = projectLevelTypeTable.getModule(fqn);
+    importFrom.importedNames().forEach(aliasedName -> aliasedName
+      .dottedName()
+      .names()
+      .stream()
+      .findFirst()
+      .ifPresent(name -> {
+        var type = module.resolveMember(name.name());
+        setTypeToName(name, type);
```
guillaume-dequenne-sonarsource, line 181: If the name actually has an alias, is the type also bound the the original name?
If not we should probably avoid conflicts here (in case the name has an alias specifically not to clash with the same name already defined in the namespace).

guillaume-dequenne-sonarsource (changes requested):
Looks good overall, left some comments.

python-frontend/src/main/java/org/sonar/python/types/v2/ModuleType.java
```diff
@@ -38,4 +48,12 @@ public PythonType resolveMember(String memberName) {
     // FIXME: handle case where type is missing
     return members.getOrDefault(memberName, PythonType.UNKNOWN);
   }
+
+  @Override
+  public String toString() {
```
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: cause it is record and by default it includes `parent` field, but since member and parent have links to each other it leads to stack overflow

python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeInferenceV2.java
maksim-grebeniuk-sonarsource: agreed, [SONARPY-1781](https://sonarsource.atlassian.net/browse/SONARPY-1781)

python-frontend/src/main/java/org/sonar/python/types/v2/ModuleType.java
```diff
@@ -19,10 +19,20 @@
...
 package org.sonar.python.types.v2;
 
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nullable;
 
-public record ModuleType(String name, Map<String, PythonType> members) implements PythonType {
+public record ModuleType(@Nullable @CheckForNull String name, @Nullable @CheckForNull ModuleType parent, Map<String, PythonType> members) implements PythonType {
```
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: Optional vs null is not about records but it is about just personal preferences. Personally, I prefer to use Optionals for business logic but for a model keep it as basic as possible

python-frontend/src/test/java/org/sonar/python/types/v2/ModuleTypeTest.java
```diff
@@ -0,0 +1,46 @@
...
+  @Test
+  void toStringTest() {
+    var root = new ModuleType(null);
+    var module = new ModuleType("pkg", root);
+    var moduleString = module.toString();
+    Assertions.assertThat(moduleString).isEqualTo("ModuleType{name='pkg', members={}}");
+  }
+
+  @Test
+  void isCompatibleWithTest() {
```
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: `isCompatibleWith` mainly designed to check for compatibility between different classes objects and classes, for modules it is just a comparison by name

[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: removed

python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeInferenceV2.java
maksim-grebeniuk-sonarsource: tests added

PR #1766, opened 2024-04-22
SONARPY-1777 S1128: Do not raise an unused import on from sklearn.experimental

ghislainpiot:


python-checks/src/test/resources/checks/unusedImport/unusedImport.py
```diff
@@ -38,3 +38,7 @@ def import_inside_function():
     from my_module import if_import # Compliant
 
 if_import()
+
+from sklearn.experimental import enable_iterative_imputer
```
joke1196, line 42: it there a use case where we could import experimental itself? `from sklearn import experimental` ? This would trigger a FP no? But not sure it is relevant.

joke1196 (approved):
LGTM! I am not a fan of the nested Optional, but I think it is the cleanest way of doing this. Really good job! Just added a small question but not sure it should be dealt with.

python-checks/src/test/resources/checks/unusedImport/unusedImport.py
ghislainpiot: The [documentation](https://scikit-learn.org/stable/modules/classes.html#module-sklearn.experimental) specifies that it only enables features, so it should be ok !

PR #1767, opened 2024-04-22
SONARPY-1786 Add basic information to locally defined ClassType (parents and metaclass)

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeBuilder.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.python.semantic.v2;
+
+import org.sonar.python.types.v2.PythonType;
+
+public interface TypeBuilder {
```
maksim-grebeniuk-sonarsource, line 24: maybe what we can do here is to make the `TypeBuilder` generic:
```java
public interface TypeBuilder<T extends PythonType> {
  T build()
}
```

python-frontend/src/main/java/org/sonar/python/semantic/v2/FunctionTypeBuilder.java
```diff
@@ -37,7 +37,7 @@
 
 import static org.sonar.python.tree.TreeUtils.locationInFile;
 
-public class FunctionTypeBuilder {
+public class FunctionTypeBuilder implements TypeBuilder {
```
maksim-grebeniuk-sonarsource, line 40: I still think we can simply use the [record builder library](https://github.com/Randgalt/record-builder) which will generate builders for us cause it does completely the same

python-frontend/src/main/java/org/sonar/python/semantic/v2/SymbolsModuleTypeProvider.java
```diff
@@ -99,26 +99,48 @@ private static PythonType convertToObjectType(Symbol symbol) {
     return PythonType.UNKNOWN;
   }
 
-  private static PythonType convertToFunctionType(FunctionSymbol symbol) {
-    return new FunctionType(symbol.name(), List.of(), List.of(), PythonType.UNKNOWN, false, false, false, false, null);
+  private static PythonType convertToFunctionType(FunctionSymbol symbol, Map<Symbol, TypeBuilder> createdTypesBySymbol) {
```
maksim-grebeniuk-sonarsource, line 102: I think it will be easier to operate with PythonTypes instead of TypeBuilders map here

python-frontend/src/main/java/org/sonar/python/semantic/v2/SymbolsModuleTypeProvider.java
```diff
@@ -99,26 +99,48 @@ private static PythonType convertToObjectType(Symbol symbol) {
...
+    // TODO: ensure we don't build twice
+    if (createdTypesBySymbol.containsKey(symbol)) {
+      return createdTypesBySymbol.get(symbol).build();
```
maksim-grebeniuk-sonarsource, line 105: then it will allow just to get type from map here

python-frontend/src/main/java/org/sonar/python/semantic/v2/SymbolsModuleTypeProvider.java
```diff
@@ -99,26 +99,48 @@ private static PythonType convertToObjectType(Symbol symbol) {
...
+      new FunctionTypeBuilder(symbol.name())
+      .withAttributes(List.of())
+        .withParameters(List.of())
+        .withReturnType(PythonType.UNKNOWN)
+        .withAsynchronous(false)
+        .withHasDecorators(false)
+        .withInstanceMethod(false)
+        .withHasVariadicParameter(false)
+        .withOwner(null);
+    createdTypesBySymbol.put(symbol, functionTypeBuilder);
```
maksim-grebeniuk-sonarsource, line 117: here we'll be able to put the built `functiontype`

python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeInferenceV2.java
```diff
@@ -100,12 +114,49 @@ public void visitListLiteral(ListLiteral listLiteral) {
...
     inTypeScope(type, () -> scan(classDef.body()));
   }
 
+  static void resolveTypeHierarchy(ClassDef classDef, ClassTypeBuilder classTypeBuilder) {
+    ArgList argList = classDef.args();
+    if (argList == null) {
+      return;
+    }
+    for (Argument argument : argList.arguments()) {
+      if (!argument.is(Tree.Kind.REGULAR_ARGUMENT)) {
```
maksim-grebeniuk-sonarsource, line 131: could be done like:
```java
if (argument instanceof RegularArgument regularArgument) {
  addParentClass(classTypeBuilder, regularArgument);
} else {
  classTypeBuilder.superClasses().add(PythonType.UNKNOWN);
}
```

python-frontend/src/main/java/org/sonar/python/types/v2/PythonType.java
```diff
@@ -45,4 +45,13 @@ default PythonType resolveMember(String memberName) {
     return PythonType.UNKNOWN;
   }
 
+  default TriBool hasMember(String memberName) {
+    return TriBool.UNKNOWN;
+  }
+
+  enum TriBool {
```
maksim-grebeniuk-sonarsource, line 52: I guess it is better to make it a top level class (not nested)

python-frontend/src/main/java/org/sonar/python/semantic/v2/FunctionTypeBuilder.java
guillaume-dequenne-sonarsource: As discussed, I think we can introduce this later if we actually stick with records.
Even then, I'm not sure how important this is for us given I don't expect we will create that many more records.

PR #1768, opened 2024-04-23
SONARPY-1780 Rule S6794: Subclasses of Scikit-Learn's "BaseEstimator" should not set attributes ending with "_" in the "__init__" method

ghislainpiot:


python-checks/src/main/java/org/sonar/python/checks/SkLearnEstimatorDontInitializeEstimatedValuesCheck.java
```diff
@@ -0,0 +1,157 @@
...
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.FUNCDEF, SkLearnEstimatorDontInitializeEstimatedValuesCheck::checkFunction);
+  }
+
+  private static boolean inheritsMixin(ClassSymbol classSymbol) {
+    boolean directlyInheritsMixin = classSymbol.superClasses().stream().filter(superClass -> superClass.fullyQualifiedName() != null)
+      .anyMatch(superClass -> MIXINS_FULLY_QUALIFIED_NAME.contains(superClass.fullyQualifiedName()));
+    return directlyInheritsMixin || classSymbol.superClasses().stream().filter(symbol -> symbol.is(Symbol.Kind.CLASS)).map(ClassSymbol.class::cast)
+      .anyMatch(SkLearnEstimatorDontInitializeEstimatedValuesCheck::inheritsMixin);
```
joke1196, line 71: I think there is an easier way with the use of `ClassSymbol.isOrExtend(String fqn)` method.
```suggestion
    return MIXINS_FULLY_QUALIFIED_NAME.stream().anyMatch(mixin -> classSymbol.isOrExtend(mixin)); 
```
I did not test this so take it with a grain of salt :D

joke1196 (changes requested):
This looks really good. I have a few suggestions. Also I am not 100% convinced about the quickfix as removing the statement would not fix the issue per se. Because one use case would be: the variable was badly named (incorrect convention) and is used for some other purpose (but should still be initialized in the __init__ method).

python-checks/src/main/java/org/sonar/python/checks/SkLearnEstimatorDontInitializeEstimatedValuesCheck.java
ghislainpiot: It looks like it is working

python-checks/src/main/java/org/sonar/python/checks/SkLearnEstimatorDontInitializeEstimatedValuesCheck.java
```diff
@@ -0,0 +1,153 @@
...
+        .flatMap(tuple -> tuple.elements().stream())
+        .filter(expression -> expression.is(Tree.Kind.QUALIFIED_EXPR))
+        .map(QualifiedExpression.class::cast);
+
+      var qualifiedExpressionsStream = Stream.concat(
+        offendingQualifiedExpressions, offendingTuples)
+        .filter(VariableDeclarationEndingWithUnderscoreVisitor::isOffendingQualifiedExpression);
+
+      qualifiedExpressionsStream.forEach(qualifiedExpression -> qualifiedExpressions.put(qualifiedExpression, pyAssignmentStatementTree));
+      super.visitAssignmentStatement(pyAssignmentStatementTree);
```
joke1196, line 150: I did not see that in the previous iteration, but we could drop the assignment to qualifiedExpressionsStream
```suggestion
      Stream.concat(
        offendingQualifiedExpressions, offendingTuples)
        .filter(VariableDeclarationEndingWithUnderscoreVisitor::isOffendingQualifiedExpression)
        .forEach(qualifiedExpression -> qualifiedExpressions.put(qualifiedExpression, pyAssignmentStatementTree));
      super.visitAssignmentStatement(pyAssignmentStatementTree);
```

joke1196 (approved):
Looks really good just one small change with an assignment that we do not need. Otherwise GTG

python-checks/src/test/java/org/sonar/python/checks/SkLearnEstimatorDontInitializeEstimatedValuesCheckTest.java
```diff
@@ -0,0 +1,65 @@
...
+      """
+        from sklearn.base import ClassifierMixin
+        class InheritingEstimator(ClassifierMixin):
+            def __init__(self) -> None:
+                self.a_ = None""",
+      """
+        from sklearn.base import ClassifierMixin
+        class InheritingEstimator(ClassifierMixin):
+            def __init__(self) -> None:
+                pass"""
```
joke1196, line 62: Maybe we could have a variation of the quickfix. One version would be remove and the other could be rename. The rename would remove the "_" I guess.

joke1196 (approved):
I think this good still [one comment was not addressed](https://github.com/SonarSource/sonar-python/pull/1768/files#r1585939489). Small fix but after that you can merge. Also the quickfix looks really good. Thanks for that!

PR #1769, opened 2024-04-24
SONARPY-1783 Migrate ArgumentNumberCheck to the new type model

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource:
@maksim-grebeniuk-sonarsource I feel this PR is outdated after the work we did for the past month. The changes introduced here are probably no longer relevant for the rule migration when we actually tackle it.
Do you agree? Is it okay if we close it?

guillaume-dequenne-sonarsource:
Outdated.

PR #1773, opened 2024-04-25
SONARPY-1796 Infer types for set, dict and tuple literals

guillaume-dequenne-sonarsource:
Tuple content type is inferred immediately if (and unique) available as tuple are immutables.
This should be seen as a baby step/low-hanging fruit. I'd expect a dedicated effort to tackle any more complex case.

Also, this PR does not deal with display name/representation of those generic types.

Not 100% related to the ticket, but I also took the opportunity to add some missing test for some already covered literal types (str, numeric literals and None).

### DIFF
--- a/python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeInferenceV2.java
+++ b/python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeInferenceV2.java
@@ -30,6 +30,7 @@ import org.sonar.plugins.python.api.tree.ArgList;
 import org.sonar.plugins.python.api.tree.AssignmentStatement;
 import org.sonar.plugins.python.api.tree.BaseTreeVisitor;
 import org.sonar.plugins.python.api.tree.ClassDef;
+import org.sonar.plugins.python.api.tree.DictionaryLiteral;
 import org.sonar.plugins.python.api.tree.DottedName;
 import org.sonar.plugins.python.api.tree.Expression;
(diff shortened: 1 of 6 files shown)


python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeInferenceV2.java
```diff
@@ -85,6 +91,35 @@ public void visitStringLiteral(StringLiteral stringLiteral) {
...
+  public void visitTuple(Tuple tuple) {
+    super.visitTuple(tuple);
+    List<PythonType> contentTypes = tuple.elements().stream().map(Expression::typeV2).distinct().toList();
+    List<PythonType> attributes = List.of();
+    if (contentTypes.size() == 1 && !contentTypes.get(0).equals(PythonType.UNKNOWN)) {
+      attributes = contentTypes;
+    }
+    ModuleType builtins = this.projectLevelTypeTable.getModule(BUILTINS);
+    PythonType tupleType = builtins.resolveMember("tuple").orElse(PythonType.UNKNOWN);
+    ((TupleImpl) tuple).typeV2(new ObjectType(tupleType,  attributes, List.of()));
```
maksim-grebeniuk-sonarsource, line 104: using `List.of()` could be dangerous since it is immutable. I'd prefer to use array lists instead

maksim-grebeniuk-sonarsource (approved):
LGTM

PR #1774, opened 2024-04-25
SONARPY-1798 Try to resolve built-in types for names which have no symbol

maksim-grebeniuk-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeInferenceV2.java
```diff
@@ -279,15 +277,20 @@ public void visitQualifiedExpression(QualifiedExpression qualifiedExpression) {
...
-      return;
-    }
-    List<PythonType> types = new ArrayList<>();
-    for (UsageV2 usage : symbolV2.usages()) {
-      if (usage.kind().equals(UsageV2.Kind.GLOBAL_DECLARATION)) {
+    Optional.of(name)
+      .map(Name::symbolV2)
+      .map(SymbolV2::usages)
+      .map(TypeInferenceV2::getUsagesType)
+      .or(() -> projectLevelTypeTable.getModule().resolveMember(name.name()))
```
guillaume-dequenne-sonarsource, line 284: I think this is a subtle yet impactful change from the previous implementation.

In the previous implementation, any variable that either has multiple assignments or a global declaration usage would not be assigned a type.

Now we fall back to `builtins` instead, which means that in the following (admittedly convoluted) code:

```
global int
int = 42
int = "hello"
int
```

The final `int` would be considered to be `builtins.int` despite the multiple reassignments.

I think the fallback to builtins should only happen if a symbol has no binding usages.

I'm a bit unhappy about the focus of our discussion regarding the imperative loop vs the optional chain. I think the main focus should be on functional behavior rather than technical implementation and I believe this case was an example where we can improve.

Taking the previous implementation as a reference, I believe that a simple

```
if (types.size().isEmpty() {
  // the underlying assumption is that each binding usage will have a type and add to the list
  // if it's not the case, a simple boolean "hasNoBindingUsage" computed during the iteration would do the trick
  setTypeToName(name, projectLevelTypeTable.getModule().resolveMember(name.name()));
}
```
at the end of the method would have achieved what you're trying to do, without requiring an additional iteration over `usages` and without risking an undesired functional change.

I also think we are missing tests for this (like variations of the convoluted case I showed), which is hidden due to the optional chain.

Again, I have nothing against optional chains, but they do have limitations that need to be acknowledged and I don't think we should aim to refactor everything to these just for the sake of it.

python-frontend/src/main/java/org/sonar/python/semantic/v2/FunctionTypeBuilder.java
```diff
@@ -148,8 +148,16 @@ private void addParameter(org.sonar.plugins.python.api.tree.Parameter parameter,
     Token starToken = parameter.starToken();
     if (parameterName != null) {
       ParameterType parameterType = getParameterType(parameter);
-      this.parameters.add(new ParameterV2(parameterName.name(), parameterType.pythonType(), parameter.defaultValue() != null,
-        parameterState.keywordOnly, parameterState.positionalOnly, parameterType.isKeywordVariadic(), parameterType.isPositionalVariadic(), locationInFile(parameter, fileId)));
+      var parameterV2 = new ParameterV2(parameterName.name(),
```
guillaume-dequenne-sonarsource, line 151: Do we need the changes regarding `FunctionType` building for this PR?
I feel we can focus on the builtins fallback which would then be clearly separated from this.

python-frontend/src/main/java/org/sonar/python/semantic/v2/SymbolsModuleTypeProvider.java
```diff
@@ -47,9 +49,8 @@ public SymbolsModuleTypeProvider(ProjectLevelSymbolTable projectLevelSymbolTable
     this.projectLevelSymbolTable = projectLevelSymbolTable;
   }
 
-  public void createBuiltinModule(ModuleType parent) {
-    var name = "builtins";
-    createModuleFromSymbols(name, parent, TypeShed.builtinSymbols().values());
+  public ModuleType createBuiltinModule() {
```
guillaume-dequenne-sonarsource, line 52: I'm not sure yet that I see the value of removing the `builtins` prefix.

I believe we could still have a `builtins` dedicated module with a helper method that would help retrieve it so that we don't have to duplicate the `"builtins"` literal everywhere?

guillaume-dequenne-sonarsource (changes requested):
This PR is a good example of what I'm trying to promote and why.

I believe that if this PR had been focused on fixing the following problem:

"If a name does not have any binding usage, let's search in the builtins module to allocate a type for it"

It could have been much smaller and focused, easier to review and to make progress (carving it out of the `ArgumentNumberCheck` was a great change, IMO - it would have been even better to do it from the start).

I also have a question regarding names whose type is a builtin: what does it mean for them in terms of symbol? I believe for now, they would have a type but no symbol (because no assignment). Is this intentional? I don't have a strong opinion on the behavior at this point, but I think we should probably test/document this. It may have impacts on rules that aim to detect undefined symbols.

python-frontend/src/main/java/org/sonar/python/semantic/v2/SymbolsModuleTypeProvider.java
maksim-grebeniuk-sonarsource: it is mainly needed to convert symbols (old ones) to types, the builtin types don't have `builtins` in their FQN

guillaume-dequenne-sonarsource: I see! I think it was a mistake on our end in the original implementation, though. Since other tools (including the Python interpreter itself) will map builtin types to the `builtins` module.

Ideally, I think I'd like to follow that convention as well, however I think for now the least disruptive approach is okay (i.e, if this PR still achieves its goals without removing `builtins`, I'd keep the prefix - and if it breaks a lot of other stuff, we can consider adding it back later instead).

python-frontend/src/main/java/org/sonar/python/tree/SubscriptionExpressionImpl.java
```diff
@@ -77,4 +79,13 @@ public List<Tree> computeChildren() {
   public Kind getKind() {
     return Kind.SUBSCRIPTION;
   }
+
+  @Override
+  public PythonType typeV2() {
```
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: excluded from PR

python-frontend/src/main/java/org/sonar/python/types/v2/ClassType.java
```diff
@@ -171,4 +171,9 @@ public boolean equals(Object o) {
   public int hashCode() {
     return Objects.hash(name, members, attributes, superClasses);
   }
+
+  public boolean isOrExtends(String s) {
+//    TODO: implement and maybe it should accept PythonType instead of string
+    return false;
+  }
```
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: excluded from PR

python-frontend/src/main/java/org/sonar/python/semantic/v2/FunctionTypeBuilder.java
maksim-grebeniuk-sonarsource: excluded

guillaume-dequenne-sonarsource (commented):
I still have some questions before we proceed with merging this PR.

python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeInferenceV2.java
```diff
@@ -312,29 +310,39 @@ public void visitQualifiedExpression(QualifiedExpression qualifiedExpression) {
     }
   }
 
+
   @Override
   public void visitName(Name name) {
     SymbolV2 symbolV2 = name.symbolV2();
     if (symbolV2 == null) {
+      projectLevelTypeTable.getModule().resolveMember(name.name())
```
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: in my implementation, there was no difference between no symbol and symbol doesn't have a binding usage, that's why it has been working.
About symbol for builtin types - agree but it is out of this ticket scope

python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeInferenceV2.java
```diff
@@ -75,8 +75,6 @@ public TypeInferenceV2(ProjectLevelTypeTable projectLevelTypeTable) {
     this.projectLevelTypeTable = projectLevelTypeTable;
   }
 
-  private static final String BUILTINS = "builtins";
```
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: cause when you resolve types e.g. for serialized symbols or project level symbols the FQN of type str is just str, it doesn't operate with builtins, and I want to have common behavior for it

python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeInferenceV2.java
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: TODO: add comment with ticket

PR #1775, opened 2024-04-25
SONARPY-1783 Migrate ArgumentNumberCheck to the new type model

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource (commented):
To be honest, I didn't do a thorough review of this PR yet.

However given the many FPs this introduces and the shift in our approach, I suggest we don't touch this PR for now and we come back to it when we feel ready to deliver the updated `NonCallableCalledCheck` to production. Then we can use this rule to track which feature we're still missing so that we can prioritize them.

guillaume-dequenne-sonarsource:
@maksim-grebeniuk-sonarsource I feel this PR is outdated after the work we did for the past month. The changes introduced here are probably no longer relevant for the rule migration when we actually tackle it.
Do you agree? Is it okay if we close it?

guillaume-dequenne-sonarsource:
Outdated.

PR #1776, opened 2024-04-26
SONARPY-1801 infer type of subscription expression when it is a class with generic type

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource (changes requested):
I'm not sure about keeping this logic in the `Tree` itself despite the fact we did that in the previous implementation. I think it would make sense to do it during the visit of `TypeInferenceV2` to centralize the places where types are set (I realize it's actually quite difficult to know where a type is set in the current implementation, as it can happen in many places...).

I'm also unsure about the edge cases we could face with this. Ideally it could be nice to have an extra test for a possible edge case that is actually not generic (if it exists0>

PR #1777, opened 2024-04-26
SONARPY-1800 NonCallableCallCheck migration: Use the display name of the new type model

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/NonCallableCalledCheck.java
```diff
@@ -55,22 +54,20 @@ public void initialize(Context context) {
...
 
   public boolean isNonCallableType(PythonType type) {
     return type.hasMember("__call__") == TriBool.FALSE;
   }
 
-  public String message(InferredType calleeType, @Nullable String name) {
+  public String message(PythonType typeV2, @Nullable String name) {
     if (name != null) {
-      return String.format("Fix this call; \"%s\"%s is not callable.", name, addTypeName(calleeType));
+      return String.format("Fix this call; \"%s\"%s is not callable.", name, addTypeName(typeV2));
```
maksim-grebeniuk-sonarsource, line 69: In currently used Java version `String.format` could be simplified by the `String.formatted` method:
```java
"Fix this call; \"%s\"%s is not callable.".formatted(name, addTypeName(typeV2))
```

maksim-grebeniuk-sonarsource (approved):
LGTM, just one minor suggestion related to strings formatting (no need to review again, feel free to ignore)

PR #1778, opened 2024-04-26
SONARPY-1791: Typeshed add Scikit-learn stubs

joke1196:


README.md
```diff
@@ -32,7 +32,13 @@ The Python interpreter is not required in that case.
 - Maven 3.0.0 or newer
 - Python 3.9 or newer
 - [tox](https://tox.readthedocs.io/en/latest/) - `pip install tox`
-- Run `git submodule update --init` to retrieve [Typeshed](https://github.com/python/typeshed) as a Git submodule
+- Run `git submodule update --init` to retrieve [Typeshed](https://github.com/python/typeshed) and [SKlearn stubs](https://github.com/microsoft/python-type-stubs/tree/main/stubs/sklearn) as a Git submodules.
+- Run the following commands to extract only the needed files for the SKlearn stubs
+```
+cd python-frontend/typeshed_serializer/resources/python-type-stubs
+git sparse-checkout set --no-cone stubs/sklearn 
+git checkout
```
ghislainpiot, line 40: It works, but it is quite convoluted. If you do the steps incorrectly, you can get stuck in a sparse checkout and it is not obvious how to fix it.

python-frontend/src/main/resources/org/sonar/python/types/third_party_protobuf/Xlib.ext.record.protobuf
```diff
@@ -39,17 +39,17 @@ GetVersionXlib.ext.record.GetVersion"
...
 UnregisterClients!Xlib.ext.record.UnregisterClients"Xlib.protocol.rq.Requestj36j37j38j39j310j311a
 
-GetContextXlib.ext.record.GetContext"
Xlib.protocol.rq.ReplyRequestj36j37j38j39j310j311�
-
EnableContext
Xlib.ext.record.EnableContext"
Xlib.protocol.rq.ReplyRequest*�
+GetContextXlib.ext.record.GetContext"
Xlib.protocol.rq.ReplyRequestj36j37j38j39j310j311�
```
ghislainpiot, line 42: If I build the stubs on my machine, I get instability on the protobuf. We can discuss it offline

its/ruling/src/test/resources/expected_extended/scikit-learn/python-S1226.json
```diff
@@ -1,8 +1,8 @@
 {
-'scikit-learn:sklearn/ensemble/_forest.py':[
-2678,
-],
-'scikit-learn:sklearn/ensemble/_iforest.py':[
-236,
+"scikit-learn:sklearn/ensemble/_forest.py": [
```
ghislainpiot, line 2: We should probably avoid reformatting files that didn't change

python-frontend/typeshed_serializer/checksum
```diff
@@ -1,2 +1,2 @@
-b0f9d8a6ebbbbe82ab815781dbff772d320ba6f67915c2cef80308dc6d209251
-6c15956136fdf02bde39d93ead39411a3c7908e21b1a8250ed3600e2bde1430e
\ No newline at end of file
+51ede4a5b5795ebd7cf3a0201643b16bef8ba617136121af51bcab09746d12a5
+40539c4a5c33308396640492f6bb790585ad37fb444d77404018dc838bef47e9
```
ghislainpiot, line 2: Checksums don't match on my environment, I will retry it anyway with a fresh one

python-frontend/typeshed_serializer/serializer/serializers.py
```diff
@@ -105,8 +121,12 @@ def serialize(self, output_dir_name="output") -> None:
                 continue
             current_file = build_result.files.get(file)
             module_symbol = symbols.ModuleSymbol(current_file)
-            symbols.save_module(module_symbol, self.output_folder, is_debug=self.is_debug,
-                                debug_dir=output_dir_name)
+            symbols.save_module(
```
ghislainpiot, line 124: Not sure if the formatting changes are really necessary, but it does look better

python-frontend/typeshed_serializer/serializer/serializers.py
```diff
@@ -149,27 +173,39 @@ def build_for_every_python_version(self):
...
             else walk_typeshed_stdlib(opt)
+        )
         return build_result, source_paths
 
     def is_exception(self, file, build_result, source_paths):
         file_path = build_result.files[file].path
-        return self.is_third_parties and file_path not in source_paths
+        return (
+            self.is_third_parties and file_path not in source_paths
+        ) or "/numpy/" in file_path
```
ghislainpiot, line 208: Is that intentional or a remnant ?

python-frontend/typeshed_serializer/tests/conftest.py
```diff
@@ -47,6 +47,11 @@ def typeshed_custom_stubs():
     assert len(build_result.errors) == 0
     return build_result
 
+@pytest.fixture(scope="session")
+def microsoft_stubs():
+    build_result, _ = MicrosoftStubsSerializer().get_build_result()
+    assert len(build_result.errors) == 323
```
ghislainpiot, line 53: Nice, should we have a ticket or something in place to solve that ?

ghislainpiot (changes requested):
Good work on a pretty sensitive part of the project. Other than the instability, I think it is mostly good !

python-frontend/typeshed_serializer/serializer/serializers.py
joke1196: No good catch this is an error during merge!

python-frontend/typeshed_serializer/tests/conftest.py
joke1196: I don't think we need to fix this as they are all import  error as mypy tries to solve everything, even non relevant imports.

its/ruling/src/test/resources/expected_extended/scikit-learn/python-S1226.json
joke1196: Good call! Thanks

guillaume-dequenne-sonarsource: I think reformating files is okay, as we're in a somewhat inconsistent state right now (for a long time the produce json had a special format and it's hasn't been that long since we product valid json), and I personally prefer seeing formatting-only changes rather than to see mixed formatting and functional changes when this file is actually impacted.

However, I suggest having explicit formatting-only commits when this happens so that actual ruling changes can be reviewed in a focused way.

README.md
joke1196: I know it is not perfect sadly there is no easier way. I think we could maybe automate the process, in a maven task or small script. But I think this could be a separate ticket/PR

python-frontend/typeshed_serializer/serializer/serializers.py
```diff
@@ -94,6 +102,15 @@ class Serializer(ABC):
     save_location: str
     output_folder: str
 
+    logger = logging.getLogger(__name__)
```
guillaume-dequenne-sonarsource, line 105: Is the logging needed as part of the `scikit-learn` stubs change?

I'm not against having it, but I think this (as well as the formatting change) would be nice to have in a separate commit to clearly separate the QoL improvements for the serializer from the functional changes.

Also, I think if we have logs, we should have assertions for those logs in our tests (it's actually a point I like to make when referring to our string formatting rules, people tend not to make assertions on logs and we did find a few interesting issues in some big projects due to that).

python-frontend/typeshed_serializer/tests/conftest.py
guillaume-dequenne-sonarsource, line 53: Does it even make sense to have an assert of this if it's not relevant?
I guess some kind of assert on the build result content to ensure it contains what we want makes sense, but why this one in particular?

its/ruling/src/test/resources/expected_extended/scikit-learn/python-S5644.json
```diff
@@ -1,6 +1,14 @@
 {
```
guillaume-dequenne-sonarsource, line 1: When encountering issue differences on ruling, can you clarify whether it's TP/FP and what is your overall evaluation of the differences?

Given these are on sklearn itself, I would assume these are FPs due to some mismatch between the sklearn version analyzed and the version of the stubs we use? 
I know we have [SONARPY-1518](https://sonarsource.atlassian.net/browse/SONARPY-1518) to eventually deal with this, but another reviewer may not know that.

guillaume-dequenne-sonarsource (approved):
I left mostly feedback about how I'd prefer the PR to be prepared for review (cleaning commit history and splitting functional from non functional changes) but the changes themselves look good to me.
Just a few points:
* I'd recommend adding some tests for the logger.
* I'd suggest considering using a fork of the Microsoft stubs rather than the `sparse-checkout`, but this could be done later.
* I'm wondering about the priority/way we split stubs from different sources, but no strong opinion here, I'm okay with the current implementation.

python-frontend/typeshed_serializer/tests/conftest.py
joke1196: Yes this was mainly to acknowledge that there are errors we do not have to fix. I would add some assertions that's a good point

README.md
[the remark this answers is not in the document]
joke1196: Yes `no-cone` is not relevant here. I forgot to remove it after some tests. The fork would be a good idea, I will create a ticket for it

python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
```diff
@@ -302,6 +303,12 @@ private static Map<String, Symbol> searchTypeShedForModule(String moduleName) {
       return thirdPartySymbolsMypy;
     }
 
+    Map<String, Symbol> thirdPartySymbolsMicrosoft = getSymbolsFromProtobufModule(moduleName, PROTOBUF_THIRD_PARTY_MICROSOFT);
```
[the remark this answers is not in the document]
joke1196: Yes I think we could merge it. It would be less code to maintain

ghislainpiot (approved):
LGTM!

python-frontend/typeshed_serializer/serializer/serializers.py
joke1196: No, you are right this should have been in a different commit. For the sake of time, I will merge this PR. But I will better organize the commit history in the following PRs

PR #1779, opened 2024-04-26
SONARPY-1770 Rule S6971 : Transformers should not be accessed directly when a Scikit-Learn Pipeline uses caching

ghislainpiot:


python-checks/src/main/java/org/sonar/python/checks/SklearnCachedPipelineDontAccessTransformersCheck.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+    if (expression.is(Tree.Kind.QUALIFIED_EXPR)) {
+      return getAssignedName(((QualifiedExpression) expression).name());
+    }
+    var assignment = (AssignmentStatement) TreeUtils.firstAncestorOfKind(expression, Tree.Kind.ASSIGNMENT_STMT);
+    if (assignment == null) {
+      return Optional.empty();
+    }
+    var expressions = SymbolUtils.assignmentsLhs(assignment);
+    if (expressions.size() != 1) {
```
joke1196, line 183: could it be possible for the expressions to have no elements?

python-checks/src/main/java/org/sonar/python/checks/SklearnCachedPipelineDontAccessTransformersCheck.java
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  private static Optional<Name> getAssignedName(Expression expression) {
+    if (expression.is(Tree.Kind.NAME)) {
+      return Optional.of((Name) expression);
+    }
+    if (expression.is(Tree.Kind.QUALIFIED_EXPR)) {
+      return getAssignedName(((QualifiedExpression) expression).name());
+    }
+    var assignment = (AssignmentStatement) TreeUtils.firstAncestorOfKind(expression, Tree.Kind.ASSIGNMENT_STMT);
```
joke1196, line 178: This could also make use of the Optional.ofNullable

joke1196 (changes requested):
Looks really good. Nice job as it is not straighforward. I added a few suggestions to maybe simplify the code when possible.

python-checks/src/main/java/org/sonar/python/checks/SklearnCachedPipelineDontAccessTransformersCheck.java
ghislainpiot: I think it would impact the readability

python-checks/src/main/java/org/sonar/python/checks/SklearnCachedPipelineDontAccessTransformersCheck.java
ghislainpiot: I have thought about it, and I don't know how to make one

python-checks/src/main/java/org/sonar/python/checks/SklearnCachedPipelineDontAccessTransformersCheck.java
```diff
@@ -0,0 +1,209 @@
...
+      .filter(u -> u.tree().parent().is(Tree.Kind.QUALIFIED_EXPR))
+      .forEach(u -> qualifiedExpressions.put(((QualifiedExpression) u.tree().parent()).qualifier(), ((QualifiedExpression) u.tree().parent())));
+
+    return qualifiedExpressions;
+  }
+
+  private enum PipelineCreation {
+    PIPELINE,
+    MAKE_PIPELINE,
+    NOT_PIPELINE
```
joke1196, line 155: I think we can get rid of the `NOT_PIPELINE`

python-checks/src/main/java/org/sonar/python/checks/SklearnCachedPipelineDontAccessTransformersCheck.java
```diff
@@ -0,0 +1,209 @@
...
+      .map(RegularArgument::expression);
+
+    var stepNames = stepArgumentExpression.map(
+      e -> pipelineCreation == PipelineCreation.PIPELINE ? extractFromPipeline(e, nameToStepName) : extractFromMakePipeline(e))
+      .orElse(Stream.empty());
+
+    var qualifiedUses = stepNames
+      .map(name -> Map.entry(name, symbolIsUsedInQualifiedExpression(name)));
+
+    qualifiedUses.forEach(entry -> {
```
joke1196, line 88: In hindsight maybe we could split this method into small methods. It would make it easier to read. We can also drop the assignment to qualifiedUses. 
```suggestion
      stepNames
      .map(name -> Map.entry(name, symbolIsUsedInQualifiedExpression(name)));
      .forEach(...)
```

joke1196 (changes requested):
Great job, just a few small details here and there.

joke1196 (approved):
LGTM!

PR #1780, opened 2024-04-29
SONARPY-1803 Store type definition location in the new type model

guillaume-dequenne-sonarsource:
To be honest I'm not entirely convinced of the proposed solution and I think you made good points in our discussion, so my goal here is to make sure we have a first baby step that can be used in the `NonCallableCalledCheck` and can be revisited later.

One thing to note is that both `ObjectType` and `ClassType` will point to the definition of the class. To be consistent with the `displayName` implementation, `ClassType#definitionLocation` should probably refer to the definition of `type`, however, I didn't think that makes sense (no such definition could be found anyway) so they both return the location of the class definition instead.


maksim-grebeniuk-sonarsource (approved):
LGTM

PR #1781, opened 2024-04-29
SONARPY-1804: Typeshed: Updating futures and PyYAML dependencies

joke1196:


its/ruling/src/test/resources/expected/python-S930.json
```diff
@@ -1,40 +1,69 @@
 {
```
guillaume-dequenne-sonarsource, line 1: As a general habit, I think it would be good to leave a small explanation about ruling changes in the PR description, especially when it's about accepted FPs.
If I understand correctly, all changes we're seeing here are due to projects using the outdated PyYAML 5.4 version?

guillaume-dequenne-sonarsource (approved):
LGTM.
I left some questions but nothing that would be a blocker to merge.

python-frontend/typeshed_serializer/requirements.txt
```diff
@@ -38,8 +37,8 @@ pandas==2.1.1
 pandas-stubs==2.0.3.230814
 pathlib==1.0.1
 PyJWT==2.7.0
-PyYAML==5.4.1
-types-PyYAML==5.4.1
+PyYAML==6.0.1
+types-PyYAML==6.0.12.20240311
```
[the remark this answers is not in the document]
joke1196: That was my impression as well, but apparently not :). So this version of the types is effectively linked to 6.0.1 but with some bug fixed apparently. So I do not think there are really convention. maybe minor versions should match.

its/ruling/src/test/resources/expected/python-S930.json
joke1196: Oh yes sorry. Exactly so those are "FPs" as we are analysing the outdated version of PyYAML. I have created a ticket as well to update these sources. Because for example salt, made the change to the new version of PyYAML but we are still analysing an old version of salt that uses PyYAML 5.4

PR #1782, opened 2024-04-30
SONARPY-1807 Populate symbol table out of SymbolTableBuilder and make it accessible from TypeInferenceV2

maksim-grebeniuk-sonarsource:


python-frontend/src/main/java/org/sonar/plugins/python/api/PythonVisitorContext.java
```diff
@@ -49,7 +49,8 @@ public PythonVisitorContext(FileInput rootTree, PythonFile pythonFile, @Nullable
     symbolTableBuilder.visitFileInput(rootTree);
     SymbolTableBuilderV2 symbolTableBuilderV2 = new SymbolTableBuilderV2();
     symbolTableBuilderV2.visitFileInput(rootTree);
-    rootTree.accept(new TypeInferenceV2(new ProjectLevelTypeTable(ProjectLevelSymbolTable.empty()), pythonFile));
+    var symbolTable = symbolTableBuilderV2.build();
```
guillaume-dequenne-sonarsource, line 52: I think there is something weird in this change about whether `build` or `visitFileInput` has to be called to trigger the visit.
Here we are calling `build` without setting the `FileInput` associated to the `SymbolTableBuilderV2` which should do nothing, and only works because `symbolTableBuilderV2.visitFileInput(rootTree); ` was called explicitly.

The duplication between `withFileInput` and the public `visitFileInput` doesn't help here I believe.

Would it make sense to instead have `FileInput` as a constructor parameter for `SymbolTableBuilder` then have `build` as its unique public method which calls the visit and then returns the `SymbolTable`?

python-frontend/src/test/java/org/sonar/python/semantic/v2/SymbolTableBuilderV2Test.java
```diff
@@ -49,89 +106,69 @@ def script_do_something(param):
...
+      assertNameSymbol(functionCallName, "script_do_something", 2);
     }
 
   }
+
+  private static void assertNameSymbol(Name functionCallName, String script_do_something, int expected) {
+    var functionCallNameSymbol = functionCallName.symbolV2();
+    Assertions.assertThat(functionCallNameSymbol).isNotNull();
+    Assertions.assertThat(functionCallNameSymbol.name()).isEqualTo(script_do_something);
+    Assertions.assertThat(functionCallNameSymbol.usages()).hasSize(expected);
```
guillaume-dequenne-sonarsource, line 172: I guess you forgot to update variable names here?
If I understand correctly, the first parameter can be any name (as in tree)?
Second parameter is the expected symbol name and third parameter is expectedUsageCount?

python-frontend/src/test/java/org/sonar/python/semantic/v2/SymbolTableBuilderV2Test.java
```diff
@@ -19,21 +19,78 @@
...
 import org.sonar.python.PythonTestUtils;
+import org.sonar.python.tree.TreeUtils;
 
 class SymbolTableBuilderV2Test {
 
 
   @Test
-  void test() {
-    var builder = new SymbolTableBuilderV2();
+  void testSymbolTableGlobalSymbols() {
```
guillaume-dequenne-sonarsource, line 37: `GlobalSymbols` is a misleading wording (I assume you mean `ModuleLevel`) since it could refer to symbols declared with the `global` keyword.

We probably actually need some tests around those global / nonlocal symbols as well to assert they are not part of any local symbol table (this would probably deserve a separate and isolated unit test though).

guillaume-dequenne-sonarsource (changes requested):
Looking good!
I like this change and how it can lead us to decouple a bit more our symbol table from the AST itself.

I left some polishing comments around the new API for `SymbolTableBuilderV2` as well as for tests, but nothing major.

guillaume-dequenne-sonarsource (approved):
LGTM.
I still think it would be good to have an additional test for `global` symbols, but it's no blocker to merge.

PR #1785, opened 2024-05-01
SONARPY-1769: Scikit-Learn random_state check

joke1196:


python-checks/src/main/java/org/sonar/python/checks/RandomSeedCheck.java
```diff
@@ -67,14 +74,47 @@ public void initialize(Context context) {
...
+  private static boolean hasRandomStateParameter(Symbol calleeSymbol) {
+    if (calleeSymbol.is(Kind.CLASS)) {
+      List<Symbol> constructorSymbols = ((ClassSymbolImpl) calleeSymbol).declaredMembers()
+        .stream()
+        .filter(s -> "__init__".equals(s.name()))
+        .toList();
+      if (constructorSymbols.size() == 1) {
+        return hasRandomStateParameter(constructorSymbols.get(0));
+      }
+      return false;
```
ghislainpiot, line 102: Removing this line should improve test coverage slightly

python-checks/src/main/java/org/sonar/python/checks/RandomSeedCheck.java
```diff
@@ -67,14 +74,47 @@ public void initialize(Context context) {
...
+    }
+    if (calleeSymbol.is(Kind.FUNCTION)) {
+      List<FunctionSymbol> symbols = SymbolUtils.getFunctionSymbols(calleeSymbol);
```
ghislainpiot, line 105: Is there a reason not to just cast the current symbol ?

python-checks/src/main/java/org/sonar/python/checks/RandomSeedCheck.java
```diff
@@ -67,14 +74,47 @@ public void initialize(Context context) {
...
+          .map(FunctionSymbol.Parameter::name)
+          .anyMatch(SKLEARN_ARG_NAME::equals);
+      }
+    }
+    return false;
   }
 
-  private boolean isSeedArgumentAbsentOrNone(String argName, List<Argument> args) {
-    RegularArgument arg = TreeUtils.nthArgumentOrKeyword(0, argName, args);
+  private boolean isArgumentAbsentOrNone(@Nullable RegularArgument arg) {
```
ghislainpiot, line 117: This is a good candidate to be put in the utils, if it is not there already

python-checks/src/test/resources/checks/randomSeedSKlearn.py
```diff
@@ -0,0 +1,28 @@
...
+    sgd = SGDClassifier(random_state=foo()) # Compliant
+    
+    def sklearn_seed(rng):
+        svc = SVC(random_state=rng) # Compliant
+
+    def foo(random_state=None):
+        pass
+
+    foo() # Compliant
+
```
ghislainpiot, line 28: We might be able to document that ambiguous symbols are not handled

```
def ambiguous():
    from sklearn.svm import SVC as something
    from sklearn.datasets import make_blobs as something

    something = something()
```

python-checks/src/test/resources/checks/randomSeedSKlearn.py
```diff
@@ -0,0 +1,28 @@
+from sklearn.model_selection import train_test_split
+from sklearn.svm import SVC
+from sklearn.datasets import load_iris, make_blobs
+
+def failure():
+    X, y = load_iris(return_X_y=True)
+    X_train, X_test, y_train, y_test = train_test_split(X, y) # Noncompliant {{Provide a seed for the random_state parameter.}}
+    #                                  ^^^^^^^^^^^^^^^^^^^^^^
```
ghislainpiot, line 8: Only having the issue on the function name might be clearer, some scikit-learn functions have a lot of parameters. It would make a really large red blob

python-checks/src/main/java/org/sonar/python/checks/RandomSeedCheck.java
```diff
@@ -67,14 +74,47 @@ public void initialize(Context context) {
...
+      if (symbols.size() == 1) {
+        FunctionSymbol symbol = symbols.get(0);
+        return symbol.parameters()
+          .stream()
+          .map(FunctionSymbol.Parameter::name)
+          .anyMatch(SKLEARN_ARG_NAME::equals);
+      }
+    }
+    return false;
   }
```
ghislainpiot, line 115: Using the good old Optional chain might side step the coverage.

```  private static boolean hasRandomStateParameter(Symbol calleeSymbol) {
    return Optional.of(calleeSymbol)
      .filter(s -> s.is(Kind.CLASS))
      .map(ClassSymbolImpl.class::cast)
      .flatMap(s -> s.declaredMembers()
        .stream()
        .filter(member -> "__init__".equals(member.name()))
        .findFirst()
        .map(RandomSeedCheck::hasRandomStateParameter))
      .or(
        () -> Optional.of(calleeSymbol)
          .filter(s1 -> s1.is(Kind.FUNCTION))
          .map(SymbolUtils::getFunctionSymbols)
          .filter(symbols -> symbols.size() == 1)
          .map(symbols -> symbols.get(0))
          .map(symbol -> symbol.parameters()
            .stream()
            .map(FunctionSymbol.Parameter::name)
            .anyMatch(SKLEARN_ARG_NAME::equals)))
      .orElse(false);
  
```

python-checks/src/main/java/org/sonar/python/checks/CheckList.java
```diff
@@ -282,7 +282,7 @@ public static Iterable<Class> getChecks() {
       NumpyWeekMaskValidationCheck.class,
       NumpyIsNanCheck.class,
       NumpyListOverGeneratorCheck.class,
-      NumpyRandomSeedCheck.class,
+      RandomSeedCheck.class,
```
ghislainpiot, line 285: nitpick : alphabetical order

ghislainpiot (changes requested):
Really solid PR, the only issue is the test coverage. The use of defensive programming does make it almost impossible to fully test correctly

python-checks/src/main/java/org/sonar/python/checks/RandomSeedCheck.java
joke1196: No not really I think using the optional API will in fact make things clearer as well

python-checks/src/main/java/org/sonar/python/checks/RandomSeedCheck.java
joke1196: I will create a ticket for this.

ghislainpiot (approved):
Looks good to me!

PR #1786, opened 2024-05-02
SONARPY-1810 Track types in case of multiple assignments in module scope

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/v2/types/FlowSensitiveTypeInference.java
```diff
@@ -59,40 +61,39 @@ public FlowSensitiveTypeInference(Set<SymbolV2> trackedVars,
...
       // update rhs
       updateTree(assignment.assignedValue(), state);
       handleAssignment(assignment, state);
       // update lhs
       assignment.lhsExpressions().forEach(lhs -> updateTree(lhs, state));
-    } else if (element.is(COMPOUND_ASSIGNMENT)) {
+    } else if (element instanceof CompoundAssignmentStatement) {
       // Assumption: compound assignments don't change types
       updateTree(element, state);
-    } else if (element.is(ANNOTATED_ASSIGNMENT)) {
```
guillaume-dequenne-sonarsource: I wasn't actually sure about removing this bit, because I suspect it actually handles the case where we know the RHS type:
```
x: int = "hello" # We may not support type annotations in this iteration but we can still infer this is actually a string
```

maksim-grebeniuk-sonarsource (approved):
LGTM

PR #1788, opened 2024-05-03
SONARPY-1775: Important hyperparameters should be specified for Scikit-Learn estimators.

joke1196:


python-checks/src/main/java/org/sonar/python/checks/SklearnEstimatorHyperparametersCheck.java
```diff
@@ -0,0 +1,178 @@
...
+        .anyMatch(SklearnEstimatorHyperparametersCheck::isArgumentPartOfGridSearch))
+      .orElse(false);
+  }
+
+  private static boolean isArgumentPartOfGridSearch(RegularArgument arg) {
+    return Optional.ofNullable(TreeUtils.firstAncestorOfKind(arg, CALL_EXPR))
+      .flatMap(toOptionalInstanceOfMapper(CallExpressionImpl.class))
+      .map(CallExpressionImpl::calleeSymbol)
+      .map(Symbol::fullyQualifiedName)
+      .map("sklearn.model_selection._search.GridSearchCV"::equals)
```
ghislainpiot, line 155: There are 3 other FQNs that we should check to avoid FPs

python-checks/src/main/java/org/sonar/python/checks/SklearnEstimatorHyperparametersCheck.java
```diff
@@ -0,0 +1,178 @@
...
+      .ifPresent(parameters -> ctx.addIssue(callExpression, MESSAGE));
+  }
+
+  private static boolean isMissingAHyperparameter(CallExpression callExpression, List<Param> parametersToCheck) {
+    return parametersToCheck.stream()
+      .map(param -> {
+        if (param.position.isPresent()) {
+          return TreeUtils.nthArgumentOrKeyword(param.position.get(), param.name, callExpression.arguments());
+        }
+        return TreeUtils.argumentByKeyword(param.name, callExpression.arguments());
```
ghislainpiot, line 116: If you want to refactor, there should be a way with Optional to do that ( `orElseGet` or something like that). But it is not important at all

ghislainpiot (approved):
Looks good to me! There are just a few FPs that could be easily avoided.

python-checks/src/main/java/org/sonar/python/checks/SklearnEstimatorHyperparametersCheck.java
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  private static void checkEstimator(SubscriptionContext ctx) {
+    CallExpression callExpression = (CallExpression) ctx.syntaxNode();
+
+    Symbol calleeSymbol = callExpression.calleeSymbol();
+
+    Optional.ofNullable(calleeSymbol)
+      .filter(callee -> callee.is(Kind.CLASS))
+      .map(ClassSymbol.class::cast)
```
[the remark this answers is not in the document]
joke1196: Sadly not as ClassSymbol is not a `Tree`

python-checks/src/main/java/org/sonar/python/checks/SklearnEstimatorHyperparametersCheck.java
joke1196: You are talking about the RandomSearch and the HalvingGrid and HalvingRandom search right?

ghislainpiot: Yes, they are the ones

PR #1789, opened 2024-05-03
SONARPY-1771 Rule S6972: Nested estimator parameters adjustment in a Pipeline should refer to valid parameters

ghislainpiot:


python-checks/src/main/java/org/sonar/python/checks/SklearnPipelineParameterAreCorrectCheck.java
```diff
@@ -0,0 +1,182 @@
...
+      .filter(Objects::nonNull)
+      .map(SklearnPipelineParameterAreCorrectCheck::getStepAndParameterFromName)
+      .filter(Optional::isPresent)
+      .map(Optional::get)
+      .collect(Collectors.toMap(StepAndParameter::step, stepAndParameter -> Set.of(new StringAndTree(stepAndParameter.parameter(), stepAndParameter.location())),
+        (set1, set2) -> {
+          var set = new HashSet<>(set1);
+          set.addAll(set2);
+          return set;
+        }));
```
joke1196, line 165: Maybe here it would be a good idea to move the collector to another function. it would help a bit with the comprehension of the code.

python-checks/src/main/java/org/sonar/python/checks/SklearnPipelineParameterAreCorrectCheck.java
```diff
@@ -0,0 +1,182 @@
...
+      if (classifier == null) {
+        // Maybe quickfix ?
+        continue;
+      }
+      var possibleParameters = getInitFunctionSymbol(classifier).map(FunctionSymbol::parameters).orElse(List.of());
+
+      parameters.forEach(parameter -> {
+        if (possibleParameters.stream().noneMatch(symbol -> Objects.equals(symbol.name(), parameter))) {
+          stringAndTree.stream().filter(stringAndTree1 -> stringAndTree1.string().equals(parameter)).findFirst()
+            .ifPresent(location -> subscriptionContext.addIssue(location.tree, MESSAGE));
```
joke1196, line 99: This is not straightforward to understand as well. We could also maybe create small functions just to name things.

python-checks/src/main/java/org/sonar/python/checks/SklearnPipelineParameterAreCorrectCheck.java
```diff
@@ -0,0 +1,182 @@
...
+
+    var parsedFunction = getStepAndParametersFromArguments(callExpression);
+    var pipelineName = Optional.of(callExpression).map(CallExpression::callee).flatMap(TreeUtils.toOptionalInstanceOfMapper(QualifiedExpression.class))
+      .map(QualifiedExpression::qualifier).flatMap(TreeUtils.toOptionalInstanceOfMapper(Name.class));
+
+    pipelineName.flatMap(Expressions::singleAssignedNonNameValue)
+      .ifPresent(c1 -> pipelineName.ifPresent(pipelineName1 -> {
+        var parsedPipeline = parsePipeline(((CallExpression) c1));
+        findProblems(parsedFunction, parsedPipeline, subscriptionContext);
+      }));
```
joke1196, line 80: I don't think you need to check twice if the pipelineName isPresent?

joke1196 (changes requested):
The logic look good, but I think the formatting is a bit messy. Maybe when using the stream or optional API try to format it with each transformation or filter on the next this would help for the review.

python-checks/src/main/java/org/sonar/python/checks/SklearnPipelineParameterAreCorrectCheck.java
```diff
@@ -0,0 +1,320 @@
...
+      .or(() -> Optional.ofNullable(callExpression.calleeSymbol())
+        .map(Symbol::fullyQualifiedName)
+        .filter("sklearn.pipeline.Pipeline.set_params"::equals)
+        .map(callExpr -> getStepAndParametersFromArguments(callExpression))
+        .flatMap(parsedParameters -> getPipelineNameAndParsedParametersFromPipelineSetParamsFunction(parsedParameters, callExpression)));
+
+    parsedFunctionOptional.ifPresent(
+      pipelineNameAndParsedParameters -> {
+        var parsedFunction = pipelineNameAndParsedParameters.parsedParameters();
+        var pipelineName = pipelineNameAndParsedParameters.pipelineName();
```
joke1196, line 90: Here just in case you did not know. You can access Record field directly (`.parsedParameters;`).

python-checks/src/main/java/org/sonar/python/checks/SklearnPipelineParameterAreCorrectCheck.java
```diff
@@ -0,0 +1,320 @@
...
+          createIssue(subscriptionContext, parameter, stringAndTree);
+        }
+      });
+    }
+  }
+
+  private static void createIssue(SubscriptionContext subscriptionContext, String parameter, Set<StringAndTree> stringAndTree) {
+    stringAndTree.stream().filter(stringAndTree1 -> stringAndTree1.string().equals(parameter)).findFirst()
+      .ifPresent(location -> subscriptionContext.addIssue(location.tree, MESSAGE));
+  }
```
joke1196, line 149: same here with the formatting

python-checks/src/main/java/org/sonar/python/checks/SklearnPipelineParameterAreCorrectCheck.java
```diff
@@ -0,0 +1,320 @@
...
+    return expressionAndPrefix -> {
+      var tuple = expressionAndPrefix.tuple();
+      var step = tuple.get(0);
+      var classifier = tuple.get(1);
+
+      if (!step.is(Tree.Kind.STRING_LITERAL)) {
+        return Stream.of(expressionAndPrefix);
+      }
+
+      if (classifier.is(Tree.Kind.NAME)) {
```
joke1196, line 209: I guess we could merge these ifs together. As the code is quite complex, having less branching to go through make it easier to reason about.

python-checks/src/main/java/org/sonar/python/checks/SklearnPipelineParameterAreCorrectCheck.java
```diff
@@ -0,0 +1,320 @@
...
+  private static Optional<Result> getResult(List<Expression> tuple) {
+    var step = tuple.get(0);
+    var classifier = tuple.get(1);
+    var stepName = Optional.ofNullable(step)
+      .flatMap(TreeUtils.toOptionalInstanceOfMapper(StringLiteral.class))
+      .map(StringLiteral::trimmedQuotesValue);
+    var classifierName = Optional.ofNullable(classifier)
+      .flatMap(TreeUtils.toOptionalInstanceOfMapper(CallExpression.class))
+      .map(CallExpression::calleeSymbol)
+      .filter(symbol -> symbol.is(Symbol.Kind.CLASS) && !Objects.equals(symbol.fullyQualifiedName(), "sklearn.pipeline.Pipeline"))
```
joke1196, line 241: Here you can use the equals on the String literal directly. `"sklearn.pipeline.Pipeline".equals(symbol.fQN)`. No need to change anything just a suggestion. I just find it more readable than referring to Objects.

python-checks/src/main/java/org/sonar/python/checks/SklearnPipelineParameterAreCorrectCheck.java
```diff
@@ -0,0 +1,320 @@
...
+        var set = new HashSet<>(set1);
+        set.addAll(set2);
+        return set;
+      });
+  }
+
+  private record StringAndTree(String string, Tree tree) {
+  }
+
+  private static Optional<StepAndParameter> getStepAndParameterFromName(Name name) {
```
joke1196, line 299: Maybe here the usage of a regex could help in just splitting the last `__`. This could also help in the creation of StepAndParameter as it could be reused for both, the string and the name.

python-checks/src/main/java/org/sonar/python/checks/SklearnPipelineParameterAreCorrectCheck.java
```diff
@@ -0,0 +1,320 @@
...
+    var classifierName = Optional.ofNullable(classifier)
+      .flatMap(TreeUtils.toOptionalInstanceOfMapper(CallExpression.class))
+      .map(CallExpression::calleeSymbol)
+      .filter(symbol -> symbol.is(Symbol.Kind.CLASS) && !Objects.equals(symbol.fullyQualifiedName(), "sklearn.pipeline.Pipeline"))
+      .map(ClassSymbol.class::cast);
+
+    return stepName.flatMap(stepName1 -> classifierName.map(classifierName1 -> new Result(stepName1, classifierName1)));
+  }
+
+  private record Result(String stepName, ClassSymbol classifierName) {
```
joke1196, line 247: Not convinced about the naming here. Maybe something more explicit on what should be expected in the tuple would be great. Maybe if you are facing issues with some naming, sometimes it could help namespacing them in classes that could hold some data and logic.

python-checks/src/main/java/org/sonar/python/checks/SklearnPipelineParameterAreCorrectCheck.java
```diff
@@ -0,0 +1,320 @@
...
+  private static Collector<StepAndParameter, ?, Map<String, Set<StringAndTree>>> mergeStringAndTreeToMapCollector() {
+    return Collectors.toMap(StepAndParameter::step, stepAndParameter -> Set.of(new StringAndTree(stepAndParameter.parameter, stepAndParameter.location)),
+      (set1, set2) -> {
+        var set = new HashSet<>(set1);
+        set.addAll(set2);
+        return set;
+      });
+  }
+
+  private record StringAndTree(String string, Tree tree) {
```
joke1196, line 296: Here the naming does not give more information than the types in the tuples. `ParameterNameAndLocation` could be a candidate.

python-checks/src/main/java/org/sonar/python/checks/SklearnPipelineParameterAreCorrectCheck.java
```diff
@@ -0,0 +1,320 @@
...
+  }
+
+  private record ExpressionAndPrefix(List<Expression> tuple, String prefix) {
+  }
+
+  private static Map<String, ClassSymbol> parsePipeline(CallExpression callExpression) {
+    var stepsArgument = TreeUtils.nthArgumentOrKeyword(0, "steps", callExpression.arguments());
+    var out = new HashMap<String, ClassSymbol>();
+
+    var a = getExpressionsFromArgument(stepsArgument)
```
joke1196, line 177: Here I guess the `a` could be removed (and the `forEach` appended to the stream) as it does not help with the comprehension of the code.

joke1196 (changes requested):
This is quite a complex implementation. I think you did very well. It is a bit difficult to review as the code is quite long and interleaved, I would encourage you to keep splitting functions when possible (to a degree that makes sense of course) to make it easier to navigate the code.

python-checks/src/main/java/org/sonar/python/checks/SklearnPipelineParameterAreCorrectCheck.java
ghislainpiot: Good catch, it was a remnant from the debugging phase

python-checks/src/main/java/org/sonar/python/checks/SklearnPipelineParameterAreCorrectCheck.java
```diff
@@ -0,0 +1,330 @@
...
+
+@Rule(key = "S6972")
+public class SklearnPipelineParameterAreCorrectCheck extends PythonSubscriptionCheck {
+
+  public static final String MESSAGE = "Provide valid parameters to the estimator.";
+  private static final Set<String> SKLEARN_SEARCH_FQNS = Set.of(
+    "sklearn.model_selection._search.GridSearchCV",
+    "sklearn.model_selection._search.HalvingGridSearchCV",
+    "sklearn.model_selection._search.RandomizedSearchCV",
+    "sklearn.model_selection._search.HalvingRandomSearchCV");
```
joke1196, line 66: Those are not the correct FQNs. You could try adding a test to confirm, but it should be under `_search_successive_halving` for the 2 halving search.

python-checks/src/main/java/org/sonar/python/checks/SklearnPipelineParameterAreCorrectCheck.java
```diff
@@ -0,0 +1,330 @@
...
+        .map(elements -> new ExpressionAndPrefix(elements,
+          expressionAndPrefix.prefix() + ((StringLiteral) step).trimmedQuotesValue() + "__", expressionAndPrefix.depth + 1))
+        .flatMap(expandRecursivePipelines());
+    };
+  }
+
+  private static Optional<CallExpression> classifierIsANestedPipeline(Name classifier) {
+    return Expressions.singleAssignedNonNameValue(classifier)
+      .flatMap(TreeUtils.toOptionalInstanceOfMapper(CallExpression.class))
+      .filter(callExpression1 -> Optional.of(callExpression1)
```
joke1196, line 242: __Nitpick__: here we could keep the `callExpression` name as it is the only one

joke1196 (changes requested):
Looks good! Just a small test to add and the correction of the FQNS, but otherwise good job! It was not an easy one.

joke1196 (approved):
LGTM! Thank you!

PR #1792, opened 2024-05-06
SONARPY-1815 Enable AST-based type inference for functions/module containing try/catch blocks

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/v2/types/Assignment.java
```diff
@@ -19,9 +19,86 @@
...
+    this.rhs = rhs;
+    this.assignmentsByLhs = assignmentsByLhs;
+  }
+
+  void computeDependencies(Expression expression, Set<SymbolV2> trackedVars) {
+    Deque<Expression> workList = new ArrayDeque<>();
+    workList.push(expression);
+    while (!workList.isEmpty()) {
+      Expression e = workList.pop();
+      if (e.is(Tree.Kind.NAME)) {
```
maksim-grebeniuk-sonarsource, line 58: it is possible to do type check and type cast in single operation here:
```java
if (e instanceof Name name) {
...
}
```

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/Assignment.java
```diff
@@ -19,9 +19,86 @@
...
+import org.sonar.plugins.python.api.tree.Tree;
 import org.sonar.python.semantic.v2.SymbolV2;
+import org.sonar.python.semantic.v2.UsageV2;
+import org.sonar.python.tree.NameImpl;
+import org.sonar.python.types.HasTypeDependencies;
+import org.sonar.python.types.v2.PythonType;
+import org.sonar.python.types.v2.UnionType;
 
-public record Assignment(SymbolV2 lhsSymbol, Name lhsName, Expression rhs) {
+public class Assignment {
```
maksim-grebeniuk-sonarsource, line 37: do we really need to have it as a class instead of a record?

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/Assignment.java
```diff
@@ -19,9 +19,86 @@
...
+
+  final SymbolV2 lhsSymbol;
+  Name lhsName;
+  Expression rhs;
+  Set<SymbolV2> variableDependencies = new HashSet<>();
```
maksim-grebeniuk-sonarsource, line 42: I'd prefer to have a single place to initialize everything. If you have a constructor - may be it makes sense to initialize these fields there as well

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/Assignment.java
```diff
@@ -19,9 +19,86 @@
...
+    this.lhsSymbol = lhsSymbol;
+    this.lhsName = lhsName;
+    this.rhs = rhs;
+    this.assignmentsByLhs = assignmentsByLhs;
+  }
+
+  void computeDependencies(Expression expression, Set<SymbolV2> trackedVars) {
+    Deque<Expression> workList = new ArrayDeque<>();
+    workList.push(expression);
+    while (!workList.isEmpty()) {
```
maksim-grebeniuk-sonarsource, line 56: Add a comment here to rework this part later. Add tests showing type tracking of collection item types

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/PropagationVisitor.java
```diff
@@ -83,9 +84,39 @@ public void visitAnnotatedAssignment(AnnotatedAssignment annotatedAssignment){
...
+
+    applyPropagations(propagations, initializedVars, true);
+    applyPropagations(propagations, initializedVars, false);
+  }
+
+  private static void applyPropagations(Set<Assignment> propagations, Set<SymbolV2> initializedVars, boolean checkDependenciesReadiness) {
+    Set<Assignment> workSet = new HashSet<>(propagations);
+    while (!workSet.isEmpty()) {
+      Iterator<Assignment> iterator = workSet.iterator();
+      Assignment propagation = iterator.next();
```
maksim-grebeniuk-sonarsource, line 112: rename to `assignment` and `propagations` parameter as well

maksim-grebeniuk-sonarsource (approved):
LGTM, just couple of small changes requested

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/Assignment.java
guillaume-dequenne-sonarsource: I don't have a strong opinion on what it should be, but if you don't mind, I'd rather keep this as it is here, since SONARPY-1826 refactors this again. Please have a look at that PR and I'm happy to further refactor it there into a record if that still makes sense (I didn't consider this tbh).

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/Assignment.java
guillaume-dequenne-sonarsource: If you don't mind I'll address this as part of SONARPY-1826.

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/PropagationVisitor.java
guillaume-dequenne-sonarsource: This will be updated in SONARPY-1826 (introducing `Propagation` class again).

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/Assignment.java
guillaume-dequenne-sonarsource: I didn't add the comment, as I think we already have some tickets that relate to this anyway. However, I added a test that shows that collection item types actually work (for a trivial case).

PR #1796, opened 2024-05-07
SONARPY-1826 Enable flow sensitive type inference for function types

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/v2/types/Propagation.java
```diff
@@ -0,0 +1,100 @@
...
+import org.sonar.python.semantic.v2.SymbolV2Utils;
+import org.sonar.python.semantic.v2.UsageV2;
+import org.sonar.python.tree.NameImpl;
+import org.sonar.python.types.v2.PythonType;
+import org.sonar.python.types.v2.UnionType;
+
+public abstract class Propagation {
+
+  final Set<SymbolV2> variableDependencies = new HashSet<>();
+  final Set<Propagation> dependents = new HashSet<>();
```
maksim-grebeniuk-sonarsource, line 36: initialize in the constructor?

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/Propagation.java
```diff
@@ -0,0 +1,100 @@
...
+   */
+  boolean propagate(Set<SymbolV2> initializedVars) {
+    PythonType rhsType = rhsType();
+    if (initializedVars.add(lhsSymbol)) {
+      lhsSymbol.usages().stream()
+        .filter(u -> !SymbolV2Utils.isFunctionOrClassDeclaration(u))
+        .map(UsageV2::tree)
+        .filter(NameImpl.class::isInstance)
+        .map(NameImpl.class::cast)
+        .forEach(n -> n.typeV2(rhsType));
```
maksim-grebeniuk-sonarsource, line 58: this part is the same as usages stream below, so maybe it makes sense to extend this common part?

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/Propagation.java
```diff
@@ -0,0 +1,100 @@
...
+
+  public abstract Name lhsName();
+
+  public abstract PythonType rhsType();
+
+  static PythonType currentType(Name lhsName) {
+    return lhsName.symbolV2().usages()
+      .stream()
+      .filter(u -> !SymbolV2Utils.isFunctionOrClassDeclaration(u))
+      .map(UsageV2::tree)
```
maksim-grebeniuk-sonarsource, line 93: maybe we can simply move this part to SymbolV2Utils:
```java
public static Stream<Tree> getSymbolNonDeclarationUsageTrees(SymbolV2 symbol) {
  
  return symbol.usages()
      .stream()
      .filter(u -> !SymbolV2Utils.isFunctionOrClassDeclaration(u))
      .map(UsageV2::tree)
}
```

maksim-grebeniuk-sonarsource (approved):
LGTM, see a couple of small comments

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/Propagation.java
guillaume-dequenne-sonarsource: Good point! I extracted this into a dedicated method.
However, I kept it in `Propagation`, as I want the filtering out of definitions to be super explicit (I think it only makes sense in this specific case). I added a comment to clarify this in the now mutualized method.

PR #1802, opened 2024-05-14
SONARPY-1817 Ensure member access types are resolved correctly

maksim-grebeniuk-sonarsource:


python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -993,6 +994,84 @@ void try_except_list_attributes() {
...
+      .map(QualifiedExpression.class::cast)
+      .get();
+
+    Assertions.assertThat(qualifiedExpression)
+      .isNotNull()
+      .extracting(QualifiedExpression::typeV2)
+      .isNotNull();
+
+    var qualifiedExpressionType = qualifiedExpression.typeV2();
+    Assertions.assertThat(qualifiedExpressionType)
```
guillaume-dequenne-sonarsource, line 1042: Here also I think we could have some more meaningful assertions.
the fact that the method is called `append` is one thing. What we want to assert is that this is the `append` of `list`.
Since we don't have FQNs yet in the type model, we can't assert on that. However, maybe we can assert on owner, parameters, etc... Maybe we can have another `list.append` call and assert these are the same types.

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -993,6 +994,84 @@ void try_except_list_attributes() {
...
+      .isInstanceOf(ObjectType.class)
+      .extracting(ObjectType.class::cast)
+      .extracting(ObjectType::type)
+      .extracting(PythonType::name)
+      .isEqualTo("int");
+  }
+
```
guillaume-dequenne-sonarsource, line 1074: Another case that would be interesting to test is when qualified expressions are assigned to a variable.
```
class A:
  def foo(self): ...

a = A()
b = a.foo
b() # is the type of b the same as the type of a.foo?
```

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -993,6 +994,84 @@ void try_except_list_attributes() {
...
+    var qualifiedExpression = TreeUtils.firstChild(root.statements().statements().get(0), QualifiedExpression.class::isInstance)
+      .map(QualifiedExpression.class::cast)
+      .get();
+
+    Assertions.assertThat(qualifiedExpression)
+      .isNotNull()
+      .extracting(QualifiedExpression::typeV2)
+      .isNotNull();
+
```
guillaume-dequenne-sonarsource, line 1065: Another test that we may want to have is when the attribute is actually overridden:

```
class A:
  def foo(self): ...

a = A()
a.foo = 42
a.foo() # type of a.foo() ?
```

I don't think this would be supported, and I don't think it was supported in the previous model anyway, but it would be good to document (and compare to what we had before if it's different). I see this could be a source of FPs on which we might want to keep an eye.

python-checks/src/test/resources/checks/nonCallableCalled.py
```diff
@@ -255,3 +255,8 @@ def my_rec(x):
         my_rec(True)
     finally:
         my_rec = None
+
+def call_non_callable_property():
+    e = OSError()
+    e.errno()  # FN
+
```
guillaume-dequenne-sonarsource, line 262: I assume we should detect some new issues as well thanks member type resolutions, could we have some tests highlighting that?

I see we have a test already that highlights that reassignments are not tracked (see snippet below), but I feel we can have cases where we actually detect issues.
```
def member_access():
  my_callable = MyCallable()
  my_callable.non_callable = 42
  my_callable.non_callable() # FN
```

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/ProgramStateTypeInferenceVisitor.java
```diff
@@ -60,9 +61,9 @@ public void visitFunctionDef(FunctionDef pyFunctionDefTree) {
   public void visitQualifiedExpression(QualifiedExpression qualifiedExpression) {
     scan(qualifiedExpression.qualifier());
     if (qualifiedExpression.name() instanceof NameImpl name) {
-      qualifiedExpression.qualifier()
```
guillaume-dequenne-sonarsource: If I understand correctly, this used to crash with an NPE if the qualifier type is `null`.

Do you have a reproducer when this can happen?
I think it would be good to have such case documented and tested (at least to see whether it is an edge case or something currently unsupported that we eventually need to look at).

python-checks/src/test/resources/checks/nonCallableCalled.py
```diff
@@ -255,3 +255,13 @@ def my_rec(x):
...
+def member_access():
```
guillaume-dequenne-sonarsource, line 263: I think you misread my comment. We already had that test showing a FN. I was asking for another test showing a new TP instead.

guillaume-dequenne-sonarsource (changes requested):
Thanks for the additional test cases!
The one from `nonCallableCalled.py` is now duplicated however, that should be fixed.
Ideally we should also have a test case that reproduces the NPE we had (currently the Optional hides the missing coverage for this).

python-checks/src/test/resources/checks/nonCallableCalled.py
maksim-grebeniuk-sonarsource: removed

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1805, opened 2024-05-16
SONARPY-1836 Infer item type for basic loop over list iteration

maksim-grebeniuk-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
```diff
@@ -153,11 +157,31 @@ public void visitNone(NoneExpression noneExpression) {
...
+        if (Stream.of(c1, c2).anyMatch(UnknownType.class::isInstance)) {
+          return PythonType.UNKNOWN;
+        } else {
+          ((UnionType) c1).candidates().add(c2);
+          return c1;
+        }
+      });
+
+    var attributes = new ArrayList<PythonType>();
+    attributes.add(elementsType);
```
guillaume-dequenne-sonarsource, line 182: Could this be a one-liner? Or, you don't want to use something like `List#of` to keep the list mutable? Genuine question.

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -1158,6 +1160,176 @@ def f():
...
+      .flatMap(expressionStatement -> TreeUtils.firstChild(expressionStatement, Name.class::isInstance))
+      .map(Name.class::cast)
+      .map(Expression::typeV2)
+      .get();
+
+    Assertions.assertThat(iType).isInstanceOf(ObjectType.class)
+      .extracting(PythonType::unwrappedType)
+      .isInstanceOf(ClassType.class)
+      .extracting(PythonType::name)
+      .isEqualTo("int");
```
guillaume-dequenne-sonarsource, line 1183: You can have a direct type comparison to `INT_TYPE` instead of this indirect check (same for the other tests).

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -1158,6 +1160,176 @@ def f():
...
+      .map(Name.class::cast)
+      .map(Expression::typeV2)
+      .get();
+
+    Assertions.assertThat(iType).isInstanceOf(UnionType.class);
+    var candidates = ((UnionType) iType).candidates();
+    Assertions.assertThat(candidates)
+      .allMatch(FunctionType.class::isInstance)
+      .extracting(PythonType::name)
+      .contains("foo", "bar");
```
guillaume-dequenne-sonarsource, line 1235: Here also you can probably retrieve the function types of `foo` and `bar` directly.

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -1158,6 +1160,176 @@ def f():
...
+
+    Assertions.assertThat(iType).isInstanceOf(UnknownType.class)
+      .isSameAs(PythonType.UNKNOWN);
+  }
+
+  @Test
+  void hasUnknownListItemType() {
+    var root = inferTypes("""
+      def f(i):
+        l = [1,2,l]
```
guillaume-dequenne-sonarsource, line 1305: Did you mean for the last item to be `i` or `l`? Both could make sense but as I read the test I have the feeling you were going for `i` here.

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
maksim-grebeniuk-sonarsource: I wanted to keep it mutable

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -1231,8 +1243,99 @@ def f():
...
+    var candidates = ((UnionType) iType).candidates();
+    Assertions.assertThat(candidates)
+      .allMatch(ObjectType.class::isInstance)
+      .extracting(PythonType::unwrappedType)
+      .contains(INT_TYPE, STR_TYPE);
+  }
+
+  @Test
+  @Disabled("The item type for loop iteration should be taken from __iter__ method return type")
+  void inferLoopOverCustomIterableVarType() {
```
guillaume-dequenne-sonarsource, line 1311: I'm not 100% sure about this, as it involves declared types on top of custom iterable, and we don't know yet how much trust we want to give to declared types.

I'd suggest saying "We should consider the declared return type of the `__iter__` method here" to be less categoric.

guillaume-dequenne-sonarsource (approved):
Looks good.
Some minor changes requested, please have a look at the comments.

PR #1811, opened 2024-05-23
SONARPY-1887 Resolve the fact that a class has a metaclass across its type hierarchy

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource (approved):
LGTM

PR #1814, opened 2024-05-27
SONARPY-1889 Fix non-deterministic stubFilesSymbols API for sonar-security

maksim-grebeniuk-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/v2/SymbolsModuleTypeProvider.java
```diff
@@ -71,7 +71,7 @@ private Optional<ModuleType> createModuleTypeFromProjectLevelSymbolTable(String
   }
 
   private Optional<ModuleType> createModuleTypeFromTypeShed(String moduleName, String moduleFqn, ModuleType parent) {
-    return Optional.ofNullable(TypeShed.symbolsForModule(moduleFqn))
+    return Optional.ofNullable(TypeShed.searchTypeShedForModule(moduleFqn))
```
guillaume-dequenne-sonarsource, line 74: While it makes sense to avoid caching the new results we retrieve here, I think it would make sense to read from what is already cached. Here it seems we will deserialize again everything, which could have a big and unnecessary performance impact.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1816, opened 2024-05-28
SONARPY-1894 Mark new type inference API as Beta

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource (approved):
LGTM.
We should look into the nondeterminism issue before merging though (even though it's obvious that it is not due to this PR).

PR #1817, opened 2024-05-28
SONARPY-1889 Fix non-deterministic stubFilesSymbols API for sonar-security Pt. 2

maksim-grebeniuk-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/ClassSymbolImpl.java
```diff
@@ -198,11 +198,21 @@ public ClassSymbolImpl copyWithoutUsages() {
...
   }
 
+  @Override
+  public List<String> superClassesFqn() {
+    return superClassesFqns;
+  }
+
+  @Override
+  public boolean isStub() {
+    return !hasAlreadyReadSuperClasses && superClasses.isEmpty() && !superClassesFqns.isEmpty();
```
guillaume-dequenne-sonarsource, line 208: When I said this condition was a proxy, I meant that it was the condition used to check in `Typeshed` for the class, but I think it's a bit of a stretch to have it in a dedicated method like this.

If we go for this, maybe we can actually set the `isStub` property when we create any `ClassSymbol` from Typeshed (like we do for `FunctionSymbol`). Or, to keep things simple, we could have a method on the `ClassSymbolImpl` that returns this boolean, with a name such as `shouldSearchHierarchyInTypeshed`.

python-frontend/src/main/java/org/sonar/python/types/TypeShed.java
```diff
@@ -110,6 +109,13 @@ public static Map<String, Symbol> builtinSymbols() {
     return builtins;
   }
 
+  public static Set<String> supportedPythonVersions() {
```
guillaume-dequenne-sonarsource, line 112: This is probably unneeded?

guillaume-dequenne-sonarsource (commented):
The duplication of `Typeshed` logic is admittedly messy. However, kudos for this, as it seems to solve our nondeterminism issues.

As discussed, I'm a bit concerned about the performance impact of this change. Let's see if we can improve it a bit before merging.

guillaume-dequenne-sonarsource (approved):
This is not ideal, but I believe we need this to move forward.
Let's keep improving performances and try to de-duplicate code in next PRs.
Before merging, can you close the SQ issues (mark them as accepted, most likely)?

PR #1820, opened 2024-06-03
SONARPY-1897 Get rid of recursive calls during PythonType hashCode calculations

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource (approved):
LGTM.

PR #1821, opened 2024-06-03
SONARPY-1896 Fix performance of new type inference hashCode computations

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource (approved):
LGTM

PR #1825, opened 2024-06-05
SONARPY-1891 Represent decorators in ClassType

guillaume-dequenne-sonarsource:


maksim-grebeniuk-sonarsource (approved):
LGTM, maybe it makes sense to have a test where a decorator is not just a name but a call expression or subscription

PR #1830, opened 2024-06-10
SONARPY-1905 Populate the TypeShed v2 type model bypassing proto-models to symbols conversion

maksim-grebeniuk-sonarsource:


guillaume-dequenne-sonarsource:
No longer relevant.

PR #1833, opened 2024-06-24
SONARPY-1944 Investigate TypeShed failures on CI

ghislainpiot:


joke1196 (commented):
Looks good. But I am not sure I understand why we would have different stubs if we fix the numpy version

ghislainpiot:
I don't think the differences in the protobufs come from the modification, but instead from the Linux/Mac differences. 
I needed to regenerate them because the checksum changed from the `requirements.txt` modifications

joke1196 (approved):
Ok some of the differences will disappear with the changes from SONARPY-1943

PR #1834, opened 2024-06-24
SONARPY-1480 Optimization of plugin download

ghislainpiot:


joke1196 (approved):
LGTM. Still we can wait for Guillaume's input, as I remember he was not 100% this would not break something

guillaume-dequenne-sonarsource:
> LGTM. Still we can wait for Guillaume's input, as I remember he was not 100% this would not break something

If I remember correctly, there was a concern about the interaction of this setting with sonar-security / DBD frontend generation.
However, I see that sonar-java and sonar-security have already enabled this, so I don't think there is a risk for us to break something here.

PR #1843, opened 2024-06-28
SONARPY-1943 Add integration test for sonar.python.version setting

ghislainpiot:


sonar-python-plugin/src/test/java/org/sonar/plugins/python/PythonSensorTest.java
```diff
@@ -693,6 +695,33 @@ void test_python_version_parameter_no_warning() {
...
+    assertThat(context.allIssues()).isEmpty();
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {"3.10", "3.11", "3.12"})
+  void test_python_version_post_310(String version) {
+    setup_typing_concise_rule(version);
+
+    assertThat(context.allIssues()).hasSize(1);
+  }
```
joke1196, line 723: Looks good! Maybe we could add a test when the version is not in our list for example `3.111` we should default to `3` I think.

joke1196 (changes requested):
Looks good. Thank you for taking care of this. I added a suggestion for an extra test.

PR #1846, opened 2024-07-09
POC: Analyze notebook location offset

joke1196:
This PR gives a rough idea of how we could work with the escaped characters to properly report locations.
Still, a few issues are not addressed in this PR:
- this does not work if the JSON file is a single line. I would still have to integrate Maskim's work.
- not all cases are taken into account only the escaping of quotes.
- tests are missing for the highlighting. symbol reporting.
- tests are sparse.

image

### DIFF
--- a/sonar-python-plugin/src/main/java/org/sonar/plugins/python/GeneratedIPythonFile.java
+++ b/sonar-python-plugin/src/main/java/org/sonar/plugins/python/GeneratedIPythonFile.java
@@ -19,6 +19,7 @@
  */
 package org.sonar.plugins.python;
 
+import com.google.thirdparty.publicsuffix.PublicSuffixPatterns;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
@@ -26,6 +27,8 @@ import java.net.URI;
 import java.nio.charset.Charset;
 import java.nio.file.Path;
 import java.util.Map;
+import java.util.ArrayList;
+import java.util.List;
 import javax.annotation.CheckForNull;
 import org.sonar.api.batch.fs.InputFile;
 import org.sonar.api.batch.fs.TextPointer;
@@ -36,15 +39,23 @@ public class GeneratedIPythonFile implements InputFile {
   InputFile originalFile;
   String pythonContent;
   Map<Integer, Offset> offsetMap;
+  // A map from the line to a Map of column location of escaped chars: 
+  // from the position in the Python code to the position in the original file
+  Map<Integer, Map<Integer, Integer>> colOffSet;
 
-  public GeneratedIPythonFile(InputFile originalFile, String pythonContent, Map<Integer, Offset> locationMap) {
+  public GeneratedIPythonFile(InputFile originalFile, String pythonContent, Map<Integer, Offset> locationMap, Map<Integer, Map<Integer, Integer>> colOffSet) {
     this.originalFile = originalFile;
     this.pythonContent = pythonContent;
     this.offsetMap = locationMap;
+    this.colOffSet = colOffSet;
   }
 
   public Map<Integer, Offset> offsetMap() {
     return offsetMap;
+  }
+
+  public Map<Integer, Map<Integer, Integer>> colOffSet(){
+    return colOffSet;
   }
 
   public record Offset(int line, int column) {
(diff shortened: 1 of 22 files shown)


joke1196:
Currently, this does not work if the JSON file is a single line. I  would still have to integrate maskim's work.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPythonNotebookParser.java
```diff
@@ -78,4 +82,23 @@ private static void processCodeCell(JsonParser parser) throws IOException {
       }
     }
   }
+   private static Map<Integer, Integer> countEscapeCharacters(String sourceLine, Map<Integer, Integer> colMap, int colOffSet) {
+    int count = 0;
+    var arr = sourceLine.toCharArray();
+    for (int i = 1; i < sourceLine.length() - 1; ++i) {
+      char c = arr[i];
+      switch (c) {
+        case '"','\'', '/':
```
guillaume-dequenne-sonarsource, line 91: Is `/` an escaped character in JSON string? I'm wondering why this one appears here.
In any case, I guess this category concerns characters that are escaped when converting the JSON string to a Java string, right?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPythonNotebookParser.java
```diff
@@ -78,4 +82,23 @@ private static void processCodeCell(JsonParser parser) throws IOException {
...
+    int count = 0;
+    var arr = sourceLine.toCharArray();
+    for (int i = 1; i < sourceLine.length() - 1; ++i) {
+      char c = arr[i];
+      switch (c) {
+        case '"','\'', '/': 
+          // + 1 as we do have to count the open quote.
+          colMap.put(i ,i + colOffSet + count + 1);
+          break;
+        case '\\', '\b', '\f', '\n', '\r', '\t':
```
guillaume-dequenne-sonarsource, line 95: Do we actually need to increase the count for `\n` when integrating the changes from Maksim?
I guess we'd make that into 2 lines of Python code, with their own column offset?

Also, I guess this category concerns characters that are not escaped when converting the JSON string to Java, but are when converted to Python, right? 

We'd probably need to document explicitly all characters that may be escaped in Python and JSON strings, so that we're sure we're not missing one.

guillaume-dequenne-sonarsource (commented):
It seems like this approach is simple and could work pretty well.
However I feel like we'll need to document and test more thoroughly when moving to an actual implementation.
Left some small comments/questions.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPythonNotebookParser.java
joke1196: Yes exactly. This list is not exhaustive at all and ha to be confirmed.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/GeneratedIPythonFile.java
```diff
@@ -131,7 +142,21 @@ public TextRange newRange(TextPointer textPointer, TextPointer textPointer1) {
   public TextRange newRange(int i, int i1, int i2, int i3) {
     Offset offsetFrom = offsetMap.get(i);
     Offset offsetTo = offsetMap.get(i2);
-    return originalFile.newRange(offsetFrom.line(), i1 + offsetFrom.column(), offsetTo.line(), i3 + offsetTo.column());
+
+    Map<Integer, Integer> escapes = colOffSet.get(i);
+    // the column location is an escape char we directly get the position from the column offset.
```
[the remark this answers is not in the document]
joke1196: Sure, that would work as well! If we try to raise an issue, for example, on a string, we will land on an escaped character, and I will need to start the location on the backslash. But I think you are right those will occur less.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPythonNotebookParser.java
joke1196: You are correct the `\n` and `\r` should not update the counter. And it is correct these are characters that are escaped in the Python code

PR #1847, opened 2024-07-15
SONARPY-1977 Change CI Releasability checks to ignore failures and allow manual triggering

ghislainpiot:


.github/workflows/releasability.yml
```diff
@@ -3,6 +3,8 @@ name: Releasability status
   check_suite:
     types:
       - completed
+  workflow_dispatch:
```
joke1196, line 6: I am not sure does this workflow_dispatch does.

joke1196 (approved):
LGTM. Just a small question.

.github/workflows/releasability.yml
ghislainpiot: It allows running the action manually in the menu, like when re-deploying Peach.

PR #1851, opened 2024-07-16
SONARPY-1978 Change CI Releasability checks to always be green allowing us to see the status by looking at the text

ghislainpiot:


joke1196 (approved):
Looks good. When running the workflow manually would we also rewrite the status to success? I guess it does not really matter as if we run it manually we would go and check the actual status.

ghislainpiot:
Yes, it would. It is unfortunate, but it is only temporary until a new version of the action comes out.

PR #1854, opened 2024-07-18
SONARPY-1984 Infer type of function parameters based on its type hints

maksim-grebeniuk-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
```diff
@@ -368,13 +402,13 @@ public void visitName(Name name) {
...
-      .map(Expression::typeV2)
-      // TODO: classes (SONARPY-1829) and functions should be propagated like other types
-      .filter(t -> (t instanceof ClassType) || (t instanceof FunctionType))
+      .flatMap(usage -> Optional.of(usage)
+        .map(UsageV2::tree)
+        .filter(Expression.class::isInstance)
+        .map(Expression.class::cast)
+        .map(Expression::typeV2)
+        // TODO: classes (SONARPY-1829) and functions should be propagated like other types
+        .filter(t -> (usage.kind() == UsageV2.Kind.PARAMETER) || (t instanceof ClassType) || (t instanceof FunctionType)))
```
guillaume-dequenne-sonarsource, line 411: Should this be here? I have the feeling this piece of code is here to assign types for names referencing classes and functions that are assigned only once, because we cannot propagate them properly.

To me for type hints, the initial propagation should be when the parameter is declared with its type hint, then it should propagate like any other name.

Here, if the parameter is reassigned, such as:
```
def foo(param: int):
    bar(param)
    param = "hello"
    qix(param)
```

We'll not consider that the call to `bar` is made with a declared type `int` because of the check on the number of binding usages above.

Note: I also think we'd need such reassignment scenarios in our tests.

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
```diff
@@ -329,6 +333,36 @@ public void visitAssignmentStatement(AssignmentStatement assignmentStatement) {
...
+    scan(parameter.defaultValue());
+    Optional.ofNullable(parameter.typeAnnotation())
+      .map(TrivialTypeInferenceVisitor::resolveTypeAnnotationType)
+      .ifPresent(type -> setTypeToName(parameter.name(), type));
+    scan(parameter.name());
+  }
+
+  private static PythonType resolveTypeAnnotationType(TypeAnnotation typeAnnotation) {
+    if (typeAnnotation.expression() instanceof Name name && name.typeV2() != PythonType.UNKNOWN) {
+      return new ObjectType(name.typeV2(), TypeSource.TYPE_HINT);
```
guillaume-dequenne-sonarsource, line 348: Should we also guard against having nested `ObjectType` here somehow?
What will we have if the annotation is a type alias? (This case, I feel, can be left out of scope, but it could be interesting to have a test for it nevertheless).

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
```diff
@@ -329,6 +333,36 @@ public void visitAssignmentStatement(AssignmentStatement assignmentStatement) {
...
+    } else if (typeAnnotation.expression() instanceof SubscriptionExpression subscriptionExpression && subscriptionExpression.object().typeV2() != PythonType.UNKNOWN) {
```
guillaume-dequenne-sonarsource, line 349: This is certainly out of scope for this PR, but we could have a test for syntax such as `X | Y` to represent union types as well.

guillaume-dequenne-sonarsource (changes requested):
Looking good, left some comments/questions.

There is also a very small piece of coverage missing, about the default `typeSource` implementation, which should be simple to cover.

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeCheckerBuilderTest.java
```diff
@@ -0,0 +1,41 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.python.semantic.v2;
+
+import org.assertj.core.api.Assertions;
```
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: Personally, I don't like static imports at all and try not to use them anywhere, even in tests

python-frontend/src/main/java/org/sonar/python/types/v2/TypeCheckBuilder.java
```diff
@@ -42,6 +42,11 @@ public TypeCheckBuilder instancesHaveMember(String memberName) {
     return this;
   }
 
+  public TypeCheckBuilder isTypeHintTypeSource() {
```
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: I think one method doesn't prevent from another

python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeInferenceV2.java
```diff
@@ -130,11 +133,13 @@ private static void inferTypesAndMemberAccessSymbols(Tree scopeTree,
...
   private static void flowSensitiveTypeInference(ControlFlowGraph cfg, Set<SymbolV2> trackedVars, PropagationVisitor propagationVisitor) {
-    // TODO: infer parameter type based on type hint or default value assignement
+    // TODO: infer parameter type based on default value assignement
     var parameterTypes = trackedVars
       .stream()
-      .filter(s -> s.usages().stream().map(UsageV2::kind).anyMatch(UsageV2.Kind.PARAMETER::equals))
-      .collect(Collectors.toMap(SymbolV2::name, v -> PythonType.UNKNOWN));
+      .map(TypeInferenceV2::getParameterType)
+      .filter(Optional::isPresent)
+      .map(Optional::get)
```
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: good catch

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
maksim-grebeniuk-sonarsource: I didn't find a Python code that could represent it

guillaume-dequenne-sonarsource: Fair enough, and it most likely makes sense not to have such check everywhere anyway.
I'm still concerned that I feel it's theoretically possible to create such nested structures, but we can handle that in another PR/ticket.

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
```diff
@@ -359,6 +361,10 @@ private static PythonType resolveTypeAnnotationType(TypeAnnotation typeAnnotatio
       var attributes = new ArrayList<PythonType>();
       attributes.add(new ObjectType(elementsType, TypeSource.TYPE_HINT));
       return new ObjectType(subscriptionExpression.object().typeV2(), attributes, new ArrayList<>(), TypeSource.TYPE_HINT);
+    } else if (expression instanceof BinaryExpression binaryExpression) {
+      var left = resolveTypeAnnotationExpressionType(binaryExpression.leftOperand());
+      var right = resolveTypeAnnotationExpressionType(binaryExpression.rightOperand());
+      return UnionType.or(left, right);
```
guillaume-dequenne-sonarsource, line 367: Should it be `UnionType` or `ObjectType[UnionType[...]]`?

I don't remember exactly what we decided, however I have the feeling it should be the latter (that way any `value` related stuff can go in the `ObjectType` itself).

python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeInferenceV2.java
```diff
@@ -130,11 +131,10 @@ private static void inferTypesAndMemberAccessSymbols(Tree scopeTree,
   }
 
   private static void flowSensitiveTypeInference(ControlFlowGraph cfg, Set<SymbolV2> trackedVars, PropagationVisitor propagationVisitor) {
-    // TODO: infer parameter type based on type hint or default value assignement
+    // TODO: infer parameter type based on default value assignement
     var parameterTypes = trackedVars
       .stream()
-      .filter(s -> s.usages().stream().map(UsageV2::kind).anyMatch(UsageV2.Kind.PARAMETER::equals))
-      .collect(Collectors.toMap(SymbolV2::name, v -> PythonType.UNKNOWN));
+      .collect(Collectors.toMap(SymbolV2::name, TypeInferenceV2::getParameterType));
```
guillaume-dequenne-sonarsource, line 137: Does that mean that non-parameters will now be set to `PythonType.UNKNOWN`?
Could this lead to any possible problem?

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -393,11 +394,82 @@ def foo(param: int):
...
+
+    var functionDef = (FunctionDef) root.statements().statements().get(0);
+    var lastExpressionStatement = (ExpressionStatement) functionDef.body().statements().get(functionDef.body().statements().size() -1);
+    var type = (UnionType) lastExpressionStatement.expressions().get(0).typeV2();
+    
+    Assertions.assertThat(type.candidates())
+      .extracting(PythonType::unwrappedType)
+      .containsOnly(INT_TYPE, STR_TYPE);
+  }
+
```
guillaume-dequenne-sonarsource, line 472: Reassignment scenario is still missing from the tests:

```
def foo(param: int):
    bar(param)
    param = "hello"
    qix(param)
```

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -393,6 +394,44 @@ def foo(param: int):
...
+      def foo(param: list[int]):
+        param
+      """);
+
+    var functionDef = (FunctionDef) root.statements().statements().get(0);
+    var lastExpressionStatement = (ExpressionStatement) functionDef.body().statements().get(functionDef.body().statements().size() -1);
+    Assertions.assertThat(lastExpressionStatement.expressions().get(0).typeV2().unwrappedType()).isEqualTo(LIST_TYPE);
+    Assertions.assertThat(lastExpressionStatement.expressions().get(0).typeV2().typeSource()).isEqualTo(TypeSource.TYPE_HINT);
+  }
+
```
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: This test case is still missing I believe.

guillaume-dequenne-sonarsource (changes requested):
I still have a question on how we set the initial parameter types, and I believe some additional tests would be worth it as well.

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
maksim-grebeniuk-sonarsource: I think it is not that clear cause potentially you can have the:
```python
def foo4(a: list[str] | set[int]):
```

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1857, opened 2024-07-19
SONARPY-1997 Propagate function return types to the result of call expressions

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeShed.java
```diff
@@ -91,7 +91,7 @@ public class TypeShed {
   public TypeShed(ProjectLevelSymbolTable projectLevelSymbolTable) {
     // workaround to initialize supported python versions used in ClassSymbolImpl
     // TODO: remove once v2 types model will be populated from TypeShed bypassing conversion to symbols
-    org.sonar.python.types.TypeShed.builtinSymbols();
+    builtins = org.sonar.python.types.TypeShed.builtinSymbols();
```
guillaume-dequenne-sonarsource, line 94: This is needed to ensure we convert builtins to type from the same set of `Symbol` objects.

python-frontend/src/main/java/org/sonar/python/tree/CallExpressionImpl.java
```diff
@@ -188,6 +190,21 @@ public PythonType typeV2() {
...
+      return functionType.returnType();
+    }
+    if (callee().typeV2() instanceof UnionType unionType) {
+      PythonType result = null;
+      for (PythonType candidate : unionType.candidates()) {
+        if (candidate instanceof ClassType classType) {
+          result = result == null ? classType : UnionType.or(result, classType);
+        }
+        if (candidate instanceof FunctionType functionType) {
+          result = result == null ? functionType.returnType() : UnionType.or(result, functionType.returnType());
```
maksim-grebeniuk-sonarsource, line 203: maybe it makes sense to have this logic as a part of the `UnionType.or` method if the first argument is null - return the second one

python-frontend/src/main/java/org/sonar/python/semantic/v2/SymbolsModuleTypeProvider.java
```diff
@@ -47,14 +49,16 @@
 public class SymbolsModuleTypeProvider {
   private final ProjectLevelSymbolTable projectLevelSymbolTable;
   private final TypeShed typeShed;
+  private ModuleType rootModule;
 
   public SymbolsModuleTypeProvider(ProjectLevelSymbolTable projectLevelSymbolTable, TypeShed typeShed) {
     this.projectLevelSymbolTable = projectLevelSymbolTable;
     this.typeShed = typeShed;
+    this.rootModule = createModuleFromSymbols(null, null, typeShed.builtinSymbols().values());
```
maksim-grebeniuk-sonarsource, line 57: why do we need to have the type here?

maksim-grebeniuk-sonarsource (approved):
LGTM, just couple of small comments, discussed on call

PR #1858, opened 2024-07-20
SONARPY-1974: fix false positive for unused variable (assignment expression in generator expression)

peteraisher:
- Add test case for scope of assignment expression target.
- Hoist target of assignment expression in generator expression to surrounding scope — See [PEP 572](https://peps.python.org/pep-0572/#scope-of-the-target).
- Add final modifiers to fields where appropriate
- Add test case for quick fix in assignment expression
- Add quick fix to remove assignment target from assignment expression
- Add true positive test case for unused target in assignment expression

(I wasn't able to verify these commits against the SonarQube Next instance without an access token.)


joke1196:
Hi @peteraisher,

Thank you so much for this contribution. We will try and review it as soon as possible!

Cheers,

David

PR #1860, opened 2024-07-22
SONARPY 2002

ghislainpiot:
- **SONARPY-1988: Create an interface to wrap InputFile (#1855)**
- **SONARPY-2002 Implement basic notebook parser with support for array code cells**


joke1196 (approved):
Looks really good, thank you for splitting the PR! Just one question do we need the added required file size?

ghislainpiot:
Unfortunately, we do need the increased size because of Jackson. I didn't realize before, but I think the gson library was already one of our dependency. Maybe we should have used it

PR #1861, opened 2024-07-22
SONARPY-2003 Implement basic notebook parser with support for multiline string code cells

ghislainpiot:


sonar-python-plugin/src/main/java/org/sonar/plugins/python/IpynbNotebookParser.java
```diff
@@ -103,6 +103,25 @@ private boolean parseSourceArray(JsonParser jParser, JsonToken jsonToken) throws
...
 
+  private boolean parseSourceMultilineString(JsonParser jParser, JsonToken jsonToken) throws IOException {
+    if (jsonToken != JsonToken.VALUE_STRING) {
+      return false;
+    }
+    String sourceLine = jParser.getValueAsString();
+    JsonLocation tokenLocation = jParser.currentTokenLocation();
+
+    for (String line : sourceLine.lines().toList()) {
+      aggregatedSource.append(line);
```
joke1196, line 114: I think this part could be refactored, with the lines 96 to 98.

joke1196 (changes requested):
Just a small refactor could be done, otherwise it is really clean thank you!

joke1196 (approved):
This looks good! Don't forget to rebase, before merging.

PR #1865, opened 2024-07-22
SONARPY-1934

guillaume-dequenne-sonarsource:


its/ruling/src/test/resources/expected/python-S3699.json
```diff
@@ -1,7 +1,4 @@
 {
-"project:biopython/Bio/Restriction/Restriction.py": [
-2432
```
guillaume-dequenne-sonarsource, line 3: This is a lost TP, for which SONARPY-2007 has been created

python-frontend/src/main/java/org/sonar/python/semantic/v2/SymbolsModuleTypeProvider.java
```diff
@@ -137,15 +136,21 @@ private PythonType convertToFunctionType(FunctionSymbol symbol, Map<Symbol, Pyth
...
+    var returnType = PythonType.UNKNOWN;
+    if (classSymbol == null) {
+      return returnType;
+    }
+    if (rootModule == null) {
+      return convertToType(classSymbol, createdTypesBySymbol);
+    }
+    // If the symbol is a built-in symbol and the root module already exists, we search for it instead of recreating the symbol
+    // This is necessary to avoid duplicated symbols in the original ProjectSymbolTable to translate into duplicated PythonType
+    returnType = Optional.ofNullable(classSymbol.fullyQualifiedName())
```
guillaume-dequenne-sonarsource, line 149: Note: I think this approach is a bit hacky, and very specific to the way we currently handle builtins (which are at the root of the project, whereas I think we should move them to a `builtins` package).

Somehow I feel like conversion from all symbols to types should first look into the `ProjectLevelTypeTable` to see if that symbol has already been created  (not only builtins), especially in the use case of converting a return type, which is not a place where a type is actually defined (i.e. it probably doesn't make sense to ever create new `PythonType` from here). Maybe implementing the `LazyType` resolution would make sense.

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
```diff
@@ -440,7 +440,7 @@ public void visitName(Name name) {
       .map(Expression.class::cast)
       .map(Expression::typeV2)
       // TODO: classes (SONARPY-1829) and functions should be propagated like other types
-      .filter(t -> (t instanceof ClassType) || (t instanceof FunctionType))
+      .filter(t -> (t instanceof ClassType) || (t instanceof FunctionType) || (t instanceof ModuleType))
```
guillaume-dequenne-sonarsource, line 443: This is needed to ensure we propagate imported names to branches they shouldn't otherwise reach when there is only one of them.
This avoids false negatives.

python-frontend/src/main/java/org/sonar/python/semantic/v2/SymbolsModuleTypeProvider.java
```diff
@@ -137,15 +136,21 @@ private PythonType convertToFunctionType(FunctionSymbol symbol, Map<Symbol, Pyth
...
-    return new ParameterV2(parameter.name(),
-      PythonType.UNKNOWN,
-      parameter.hasDefaultValue(),
-      parameter.isKeywordOnly(),
-      parameter.isPositionalOnly(),
-      parameter.isKeywordVariadic(),
-      parameter.isPositionalVariadic(),
-      null);
+  private PythonType resolveReturnType(Map<Symbol, PythonType> createdTypesBySymbol, @Nullable ClassSymbol classSymbol) {
+    var returnType = PythonType.UNKNOWN;
```
maksim-grebeniuk-sonarsource, line 140: do you need to define the variable here? seems that in the condition below you can simply return `UNKNOWN` and define the `returnType` on the line 149 or simply return the optional chain

python-frontend/src/main/java/org/sonar/python/semantic/v2/SymbolsModuleTypeProvider.java
maksim-grebeniuk-sonarsource, line 149: it makes sense to leave a TODO comment here that this logic should be cleaned up once lazy type will be implemented

maksim-grebeniuk-sonarsource (approved):
LGTM

PR #1867, opened 2024-07-23
SONARPY-1936 Migrate rule S5864 NonCallableCalled to typeV2

maksim-grebeniuk-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/v2/SymbolsModuleTypeProvider.java
```diff
@@ -49,6 +49,12 @@
 public class SymbolsModuleTypeProvider {
   private final ProjectLevelSymbolTable projectLevelSymbolTable;
   private final TypeShed typeShed;
+  private final Map<String, Map<String, PythonType>> syntheticModuleMembers = Map.ofEntries(
```
guillaume-dequenne-sonarsource, line 52: I understand we need to do something given that `TypeShed` doesn't clearly define some `typing` protocols.
I'm not sure how we should deal with that, though. We actually have several options:
- Process known protocols directly in the `typeshed_serializer`, and serialize them in a more expected format
- Have a list such as this one that could be used when converting symbols to types without storing them in a separate`syntheticModuleMembers`; potentially, we could alter the `TYPE_SOURCE` as a result.
- Maintain 2 hierarchies such as these ones

I don't really have a preference, and I think the current approach is fine for now; I just think it's good to highlight these options.

That being said, I have some comments:
- If I remove this, unit tests (outside of those specifically testing this synthetic look-up feature) still seem to work. In particular, unit tests for `confusingTypeChecking/nonCallableCalled.py` still pass. I would suspect this is needed to avoid FPs, but then I would expect a reproducer for those FPs to be tested in the rule tests.

- Following on the first point, I would expect that missing symbol information (especially here where `typing.Callable` is defined in Typeshed as a simple variable) would not result in FPs (they should at most lead to FNs). Therefore, why do we need this?

- If we do need these synthetic types, why not give them the expected `__call__` member, so that we don't have to special-case them in the relevant check later?

EDIT: As I tested this further, it seems that the behavior when removing these synthetic types is actually nondeterministic. I suspect that what is inferred as `typing.Coroutine` and what is in the predicate may be different objects, therefore failing the check. It's true that the `typing.Coroutine`we retrieve lacks the `__call__` method either way, which is annoying and might warrant the special case. However, I'm not sure synthetic types are an appropriate fix for that.

python-frontend/src/test/java/org/sonar/python/semantic/v2/ProjectLevelTypeTableTest.java
```diff
@@ -0,0 +1,45 @@
...
+import org.junit.jupiter.api.Test;
+import org.sonar.python.semantic.ProjectLevelSymbolTable;
+import org.sonar.python.types.v2.ClassType;
+import org.sonar.python.types.v2.ModuleType;
+import org.sonar.python.types.v2.PythonType;
+
+class ProjectLevelTypeTableTest {
+
+  @Test
+  void getTypeTest() {
```
guillaume-dequenne-sonarsource, line 32: Should we only focus on synthetic types here?
While the method is created with this use case in mind, this is probably not the only one for it.

guillaume-dequenne-sonarsource (changes requested):
I think this PR goes a bit deep in the direction of synthetic types while it may not actually address the underlying root cause of problems encountered without it (nondeterminism).

Some of the new tests assume a bit too much the context of synthetic types while I believe they should be more generic. In particular `isInstanceOfTest` is focused on that single use case and is misleading in its current form.

I'd expect you to do a more thorough pre-review of such PR and possibly break it in various tickets, seeing that there are many different things that may need to be fixed (the cases triggering the FP/FNs, the `InstanceOf` check and the nondeterminism/unreliable comparison of `typing.Coroutine` types).

python-checks/src/main/java/org/sonar/python/checks/ConfusingTypeCheckingCheck.java
```diff
@@ -218,6 +204,20 @@ String message(String result) {
...
 
+  private static class NonCallableCalledCheck extends NonCallableCalled {
+
+    @Override
+    protected boolean isTypeSourceMatches(SubscriptionContext ctx, PythonType calleeType) {
+      return ctx.typeChecker().typeCheckBuilder().isTypeHintTypeSource().check(calleeType) == TriBool.TRUE;
+    }
+
+    @Override
+    protected boolean isNonCallableCall(SubscriptionContext ctx, PythonType calleeType) {
```
guillaume-dequenne-sonarsource, line 215: `isNonCallableType` would make more sense here, as you're checking a type and not a call.

python-checks/src/main/java/org/sonar/python/checks/NonCallableCalled.java
```diff
@@ -20,45 +20,51 @@
...
-    return "";
+  protected boolean isNonCallableCall(SubscriptionContext ctx, PythonType calleeType) {
+    var typeSourceMatches = isTypeSourceMatches(ctx, calleeType);
+    var isCallMemberMissed = ctx.typeChecker().typeCheckBuilder().hasMember("__call__").check(calleeType) == TriBool.FALSE;
+    return typeSourceMatches
+           && isCallMemberMissed;
   }
 
-  public abstract boolean isNonCallableType(InferredType type);
+  protected abstract boolean isTypeSourceMatches(SubscriptionContext ctx, PythonType calleeType);
```
guillaume-dequenne-sonarsource, line 55: I'd suggest `isExpectedTypeSource` since `isTypeSourceMatches` sort of implies that the expected source is provided as an argument as well.

python-frontend/src/main/java/org/sonar/python/types/v2/TypeCheckBuilder.java
```diff
@@ -125,4 +136,15 @@ public TriBool test(PythonType pythonType) {
...
+  record IsInstanceOfPredicate(PythonType pythonType)  implements TypePredicate {
+
+    @Override
+    public TriBool test(PythonType pythonType) {
+      if (pythonType instanceof ObjectType objectType) {
+        return this.pythonType.equals(objectType.unwrappedType()) ? TriBool.TRUE : TriBool.FALSE;
+      }
+      return TriBool.UNKNOWN;
+    }
+  }
```
guillaume-dequenne-sonarsource, line 149: As stated in my previous comment, this is not correct.

```
class Parent: ...
class Child(Parent): ...
c = Child()
isinstance(c, Parent) # True
```

This predicate, based on type equality, would incorrectly return false in such a case.

I understand you want to test that the type is an actual instance (`ObjectType`) and not something like a `ClassType`; however, the method naming is misleading. If you really need to make that test, you may combine a predicate that ensures a type is a `ObjectType` with the one for `IsSameTypeAs` (which currently simply unwraps `ObjectType` so that it would return true for an object or for the type itself).

Note that, `ClassType` being an instance of `type` is also an edge case to take into account (simply checking `ObjectType` vs `ClassType` vs `FunctionType` may not be enough depending on context, since everything is ultimately an object in Python)

python-checks/src/main/java/org/sonar/python/checks/NonCallableCalled.java
```diff
@@ -20,45 +20,51 @@
...
 
-  protected static String addTypeName(InferredType type) {
-    String typeName = InferredTypes.typeName(type);
-    if (typeName != null) {
-      return " has type " + typeName + " and it";
-    }
-    return "";
+  protected boolean isNonCallableCall(SubscriptionContext ctx, PythonType calleeType) {
+    var typeSourceMatches = isTypeSourceMatches(ctx, calleeType);
+    var isCallMemberMissed = ctx.typeChecker().typeCheckBuilder().hasMember("__call__").check(calleeType) == TriBool.FALSE;
```
guillaume-dequenne-sonarsource, line 50: Probably `isCallMemberMissing` or `doesNotHaveCallMember` would be more correct.

python-frontend/src/main/java/org/sonar/python/semantic/v2/ProjectLevelTypeTable.java
```diff
@@ -61,4 +62,30 @@ public ModuleType getModule(List<String> moduleNameParts) {
...
+    return getType(typeFqn.split("\\."));
+  }
+
+  public PythonType getType(String... typeFqnParts) {
+    return getType(List.of(typeFqnParts));
+  }
+
+  public PythonType getType(List<String> typeFqnParts) {
+    var parent = (PythonType) rootModule;
+    for (int i = 0; i < typeFqnParts.size(); i++) {
```
guillaume-dequenne-sonarsource, line 76: I feel we're repeating these loops quite a lot (I also have a few variations of them in my PR). I'm wondering if we could somehow generalize these FQN traversal somehow (no action required on this PR, just sharing my thoughts).

python-frontend/src/main/java/org/sonar/python/semantic/v2/ProjectLevelTypeTable.java
```diff
@@ -61,4 +62,30 @@ public ModuleType getModule(List<String> moduleNameParts) {
     }
     return parent;
   }
+
+  public PythonType getType(String typeFqn) {
```
guillaume-dequenne-sonarsource, line 66: I think the approach makes sense, although I think we'll need to be careful about using this notion of a fully qualified name since we haven't finished defining it yet.
In this form, it probably only make sense while we still rely on the previous `ProjectLevelSymbolTable` conversion.

python-checks/src/test/resources/checks/confusingTypeChecking/nonCallableCalled.py
```diff
@@ -4,32 +4,38 @@
...
-  z() # Noncompliant
+  z() # FN Value calculation with involved parameters support is needed (SONARPY-2023)
 
 class Base: ...
 class CallableBase:
   def __call__(self, *args, **kwargs): ...
 def with_isinstance(x: Base):
   if isinstance(x, CallableBase):
-    x()
+    # FP type check guard need to be supported in V2 type inference (SONARPY-2017)
```
guillaume-dequenne-sonarsource, line 37: As discussed, I'd probably like to see some of the FP/FNs tickets addressed before merging the PR for the rule itself (if only to know which ones we can actually tackle simply). Open to discussion though.

guillaume-dequenne-sonarsource (changes requested):
Left some minor comments, and a bigger one related to the `IsInstance` predicate. Please have a look.

guillaume-dequenne-sonarsource (changes requested):
The rule itself looks good overall.
However, I still have some issues with the implementation of `IsInstanceOf`, I believe it relies too much on existing code that was introduced prematurely and under-tested (`isCompatibleWith`) and I think this could lead to subtle bugs in its current state.

python-checks/src/test/resources/checks/nonCallableCalledNoIssues.py
```diff
@@ -0,0 +1,3 @@
+
```
guillaume-dequenne-sonarsource, line 1: I'm not sure this needs a new test and test file. Why not simply have this case in `src/test/resources/checks/nonCallableCalled.py`?

python-checks/src/test/resources/checks/confusingTypeChecking/nonCallableCalled.py
```diff
@@ -4,25 +4,30 @@
...
   x()
 
-def foo(param1: int, param2: Set[int], param3: FrozenSet[int]):
+def foo(param1: int, param2: Set[int], param3: FrozenSet[int], param4: list[str], param5: set[int]):
   param1() # Noncompliant {{Fix this call; Previous type checks suggest that "param1" has type int and it is not callable.}}
   x = 42
   x() # OK, raised by S5756
 
-  param2() # Noncompliant
+  param2() # FN SONARPY-2022
```
guillaume-dequenne-sonarsource: I would expect these FNs to be fixed after rebasing on the corresponding ticket.
Can you double check that this is fixed and update the test accordingly?

python-checks/src/main/java/org/sonar/python/checks/ConfusingTypeCheckingCheck.java
```diff
@@ -218,6 +204,20 @@ String message(String result) {
...
+  private static class NonCallableCalledCheck extends NonCallableCalled {
+
+    @Override
+    protected boolean isExpectedTypeSource(SubscriptionContext ctx, PythonType calleeType) {
+      return ctx.typeChecker().typeCheckBuilder().isTypeHintTypeSource().check(calleeType) == TriBool.TRUE;
+    }
+
+    @Override
+    protected boolean isNonCallableType(SubscriptionContext ctx, PythonType calleeType) {
+      var isNotTypingCoroutine = ctx.typeChecker().typeCheckBuilder().isInstanceOf("typing.Coroutine").check(calleeType) == TriBool.FALSE;
```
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: This comment is still unaddressed.
While I don't mind keeping the code as-is (as mentioned originally, it's a minor comment after all), I would still expect you to acknowledge it (even if only by answering "I'd prefer to keep the code as it is" to the original comment).
(To be super explicit I think we can just leave the code as-is here)

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeCheckerBuilderTest.java
```diff
@@ -38,4 +40,68 @@ void typeSourceTest() {
...
+    var strObjectType = new ObjectType(strClassType);
+    var aObject = new ObjectType(aClassType);
+    var bObject = new ObjectType(bClassType);
+    var cObject = new ObjectType(cClassType);
+    var dObject = new ObjectType(dClassType);
+    var eObject = new ObjectType(UnionType.or(intClassType, strClassType));
+    var fObject = UnionType.or(new ObjectType(intClassType), new ObjectType(strClassType));
+    var gObject = new ObjectType(UnionType.or(intClassType, aClassType));
+    var hObject = UnionType.or(new ObjectType(intClassType), new ObjectType(aClassType));
+    var iObject = new ObjectType(eClassType);
```
guillaume-dequenne-sonarsource, line 87: It is confusing to have `iObject` of type `eClassType` while `eObject` is an object of another type. It would be good to make it consistent.

python-frontend/src/main/java/org/sonar/python/types/v2/TypeCheckBuilder.java
```diff
@@ -125,4 +138,94 @@ public TriBool test(PythonType pythonType) {
...
+
+    private static Set<PythonType> collectTypes(PythonType type) {
+      var result = new HashSet<PythonType>();
+      collectTypes(type, result);
+      return result;
+    }
+
+    private static void collectTypes(PythonType type, Set<PythonType> result) {
+      result.add(type);
+      if (type instanceof ClassType classType) {
```
guillaume-dequenne-sonarsource, line 215: Isn't it possible to have a `UnionType` as parent somewhere? Then shouldn't we consider them?
Also, if anything unexpected is in the type hierarchy (i.e not a `ClassType`), how do we make sure we don't miss it?


```
class A: ...
if x:
    class B(A):
        def foo(self): ...
else:
    class B(A):
        def bar(self): ...

class C(B): ...  # checking B instance of A?
```

I think the only reason why it would currently work is because we may not correctly create a `UnionType` for `B`. But if we do, this code would probably miss it.

python-frontend/src/main/java/org/sonar/python/types/v2/TypeCheckBuilder.java
```diff
@@ -125,4 +137,57 @@ public TriBool test(PythonType pythonType) {
       return pythonType.equals(expectedType) ? TriBool.TRUE : TriBool.FALSE;
     }
   }
+
+  record IsInstanceOfPredicate(PythonType expectedType)  implements TypePredicate {
+
+    @Override
+    public TriBool test(PythonType pythonType) {
+      if (expectedType instanceof ClassType expectedClassType) {
```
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: This comment was not addressed.

guillaume-dequenne-sonarsource (changes requested):
Looks good overall, but I have still doubts about potential edge cases for the `IsInstanceOf` predicate.
Also there remains some small unaddressed comments that would be good to address before merging.

python-frontend/src/main/java/org/sonar/python/types/v2/TypeCheckBuilder.java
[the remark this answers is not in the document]
maksim-grebeniuk-sonarsource: then the variable instantiation used outside of if which for me super unclear to read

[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: You could also have an explicit variable instantiation with an old school cast after the `if`, but fair enough.

guillaume-dequenne-sonarsource (approved):
LGTM.

PR #1868, opened 2024-07-23
SONARPY-1935 Migrate rule S112 GenericExceptionRaised to typeV2

guillaume-dequenne-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/GenericExceptionRaisedCheck.java
```diff
@@ -49,19 +43,12 @@ public void initialize(Context context) {
         return;
       }
       Expression expression = expressions.get(0);
-      InferredType type = expression.type();
-      if (GENERIC_EXCEPTION_NAMES.stream().anyMatch(type::canOnlyBe) || isGenericExceptionClass(expression)) {
+      PythonType pythonType = expression.typeV2();
+      TriBool isException = ctx.typeChecker().typeCheckBuilder().isBuiltinWithName(EXCEPTION).check(pythonType);
+      TriBool isBaseException = ctx.typeChecker().typeCheckBuilder().isBuiltinWithName(BASE_EXCEPTION).check(pythonType);
```
guillaume-dequenne-sonarsource, line 48: Ideally this should be merged in a single builder, either through some way to OR predicates (currently predicates are implicitly ANDed), or by making some predicates variadic.
I'm not yet sure what this API should look like and I believe we have other priorities at the moment, so I suggest fixing that in SONARPY-2013.

maksim-grebeniuk-sonarsource (approved):
LGTM

PR #1869, opened 2024-07-23
SONARPY-2004 SONARPY-2004 Track escaped characters correctly in notebook parser

ghislainpiot:


joke1196 (changes requested):
This looks really good. I think the only change I would is to have the colOffSet being part of the IPythonLocation. So IPythonLocation would hold the line number, the column number for the start of a line and the map between the column in Python and the Escaped char. This way, we do not have to change the signature of functions by passing a new argument. Everything is abstracted in the IPythonLocation object.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IpynbNotebookParser.java
```diff
@@ -91,7 +93,9 @@ private boolean parseSourceArray(JsonParser jParser, JsonToken jsonToken) throws
     }
     while (jParser.nextToken() != JsonToken.END_ARRAY) {
       String sourceLine = jParser.getValueAsString();
-      addLineToSource(sourceLine,  jParser.currentTokenLocation());
+      var tokenLocation = jParser.currentTokenLocation();
+      colOffSet.put(aggregatedSourceLine, countEscapeCharacters(sourceLine, new LinkedHashMap<>(), tokenLocation.getColumnNr()));
+      addLineToSource(sourceLine, tokenLocation, colOffSet.get(aggregatedSourceLine));
```
joke1196, line 98: I don't think we need the colOffSet anymore. We could just do 
```suggestion
   var countEscapedChar =  countEscapeCharacters(sourceLine, new LinkedHashMap<>(), tokenLocation.getColumnNr()));
   addLineToSource(sourceLine, tokenLocation, c);
```

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IpynbNotebookParser.java
```diff
@@ -104,19 +108,29 @@ private boolean parseSourceMultilineString(JsonParser jParser, JsonToken jsonTok
...
+    var previousLen = 0;
+    var previousExtraChars = 0;
 
     for (String line : sourceLine.lines().toList()) {
-      aggregatedSource.append(line);
-      addLineToSource("\n", tokenLocation);
+      colOffSet.put(aggregatedSourceLine, countEscapeCharacters(line, new LinkedHashMap<>(), tokenLocation.getColumnNr()));
+      var currentCount = colOffSet.get(aggregatedSourceLine).get(-1);
+      addLineToSource(line, new IPythonLocation(tokenLocation.getLineNr(),
+        tokenLocation.getColumnNr() + previousLen + previousExtraChars + 1, colOffSet.get(aggregatedSourceLine)));
```
joke1196, line 118: Same as above

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IpynbNotebookParser.java
```diff
@@ -68,7 +70,7 @@ public ParseResult parseNotebook() throws IOException {
       }
     }
 
-    return new ParseResult(inputFile, aggregatedSource.toString(), locationMap);
+    return new ParseResult(inputFile, aggregatedSource.toString(), locationMap, colOffSet);
```
joke1196, line 73: I guess we could remove the colOffSet here, as all the data is in the IPythonLocation

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IpynbNotebookParser.java
```diff
@@ -125,10 +139,36 @@ private void addDelimiterToSource() {
...
+          break;
+        default:
+          break;
+      }
+    }
+    colMap.put(-1, numberOfExtraChars);
+    return colMap;
+  }
+
+  public record ParseResult(PythonInputFile inputFile, String aggregatedSource, Map<Integer, IPythonLocation> locationMap, Map<Integer, Map<Integer, Integer>> colOffSet) {
```
joke1196, line 168: Here as well we could remove the colOffSet

joke1196 (changes requested):
Just a bit of clean up; we could remove the colOffset as we already have the information we need in the IPythonLocation.

joke1196 (approved):
LGTM! Thank you!

PR #1870, opened 2024-07-24
SONARPY-1986: The IPythonSensor should register to .ipynb files

joke1196:


sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPynbSensor.java
```diff
@@ -67,10 +77,38 @@ public void execute(SensorContext context) {
...
+      scanner.execute(pythonFiles, context);
+    } else {
+      processNotebooksFiles(pythonFiles, context);
+    }
+  }
+
+  private void processNotebooksFiles(List<PythonInputFile> pythonFiles, SensorContext context) {
+    pythonFiles = parseNotebooks(pythonFiles);
+    // Disable caching for IPynb files for now
+    CacheContext cacheContext = CacheContextImpl.dummyCache();
```
ghislainpiot, line 91: We should have that documented

sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonPlugin.java
```diff
@@ -91,6 +93,22 @@ public void define(Context context) {
...
+        .multiValues(true)
+        .category(PYTHON_CATEGORY)
+        .subCategory(GENERAL)
+        .onQualifiers(Qualifiers.PROJECT)
+        .defaultValue("ipynb")
+        .build(),
+      IPynb.class,
+      IPynbProfile.class,
+      IPynbSensor.class,
+      IPynbRuleRepository.class);
```
ghislainpiot, line 110: I don't know if we should have that property, is the goal to have other extensions or to allow disabling the notebooks ?

ghislainpiot (approved):
LGTM! Is the CI failure expected since the other ticket is not merged in that branch ?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonPlugin.java
joke1196: No, you are right I don't think we need this. I think it was to mirror the Python file properties.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPynbSensor.java
joke1196: Done in https://sonarsource.atlassian.net/browse/SONARPY-2020

PR #1872, opened 2024-07-24
SONARPY-1989 Extend Token to hold original and virtual locations

ghislainpiot:


python-frontend/src/main/java/org/sonar/python/tree/TokenImpl.java
```diff
@@ -33,23 +33,49 @@ public class TokenImpl extends PyTree implements Token {
...
+    return column != null ? column : physicalColumn();
+  }
+
+  @Override
+  public int physicalLine() {
+    return token.getLine();
+  }
+
+  @Override
+  public int includedEscapeChars() {
```
joke1196, line 73: Here we should add a `@CheckForNull` or maybe set a default value to `0`

joke1196 (approved):
Looks good! Just left a comment. It's a bit of a shame for the test coverage, but I am not sure it makes sense to test anything here.

python-frontend/src/main/java/org/sonar/python/tree/TokenImpl.java
ghislainpiot: The type of the variable is int, so it is 0 by default and it can't be set to `null` manually

PR #1873, opened 2024-07-24
SONARPY-1987: IPynbSensor should not crash when an error occurs during the parsing of notebooks

joke1196:


ghislainpiot (approved):
LGTM!

PR #1876, opened 2024-07-25
SONARPY-2019 Fix NPE for workDir in IPynbSensorTest

ghislainpiot:
- **SONARPY-1987: IPynbSensor should not crash when an error occurs during the parsing of notebooks**
- **SONARPY-2019 Fix NPE for workDir in IPynbSensorTest**


joke1196 (approved):
LGTM!

PR #1878, opened 2024-07-25
SONARPY-2027 IPynbParser doesn't create new lines when parsing multiline strings code cells

ghislainpiot:


sonar-python-plugin/src/main/java/org/sonar/plugins/python/IpynbNotebookParser.java
```diff
@@ -30,7 +30,7 @@
 
 public class IpynbNotebookParser {
 
-  public static final String SONAR_PYTHON_NOTEBOOK_CELL_DELIMITER = "\n#SONAR_PYTHON_NOTEBOOK_CELL_DELIMITER\n";
+  public static final String SONAR_PYTHON_NOTEBOOK_CELL_DELIMITER = "#SONAR_PYTHON_NOTEBOOK_CELL_DELIMITER\n";
```
joke1196, line 33: Do we still need this `\n` even with the append on each line of the aggregated source.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IpynbNotebookParser.java
```diff
@@ -113,7 +114,7 @@ private boolean parseSourceMultilineString(JsonParser jParser, JsonToken jsonTok
     for (String line : sourceLine.lines().toList()) {
       var countEscapedChar = countEscapeCharacters(line, new LinkedHashMap<>(), tokenLocation.getColumnNr());
       var currentCount = countEscapedChar.get(-1);
-      addLineToSource(line, new IPythonLocation(tokenLocation.getLineNr(),
+      addLineToSource(line + "\n", new IPythonLocation(tokenLocation.getLineNr(),
```
joke1196, line 117: Maybe here it would be cleaner to have: 
```suggestion
      addLineToSource(line + "\n", new IPythonLocation(tokenLocation.getLineNr(), ...)
      aggregatedSource.append("\n");
```
For consistency.

joke1196 (changes requested):
Thanks for taking care of this. Just a small question and a suggestion, to make it cleaner.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IpynbNotebookParser.java
ghislainpiot: Yes, we need to append `\n` to the source because the `lines()` iterator removes them. But for the delimiter, we also need to have it.

joke1196 (approved):
LGTM!

PR #1880, opened 2024-07-26
SONARPY-2029 Implement API to get a module type's member type from project level types table

maksim-grebeniuk-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/v2/ProjectLevelTypeTable.java
```diff
@@ -61,4 +62,30 @@ public ModuleType getModule(List<String> moduleNameParts) {
...
+    var parent = (PythonType) rootModule;
+    for (int i = 0; i < typeFqnParts.size(); i++) {
+      var part = typeFqnParts.get(i);
+      if (parent.hasMember(part) == TriBool.TRUE) {
+        parent = parent.resolveMember(part).orElse(PythonType.UNKNOWN);
+      } else if (parent instanceof ModuleType module) {
+        var moduleFqn = IntStream.rangeClosed(0, i)
+          .mapToObj(typeFqnParts::get)
+          .toList();
+        parent = symbolsModuleTypeProvider.createModuleType(moduleFqn, module);
```
guillaume-dequenne-sonarsource, line 84: I think we could use one more test, as I couldn't trigger this case when running the dedicated tests in `ProjectLevelTypeTableTest` (some other tests in `TypeInferenceV2Test` does trigger it though).

I think one thing that is important to acknowledge is that querying for a type will convert it an modify the `ProjectLevelTypeTable` if it is absent. I'm not sure if it could have some unintended consequences? (My gut feeling tells me that should be okay - still I think it's important to consider).

guillaume-dequenne-sonarsource (approved):
Looks good overall, left just one small comment. Happy to discuss it.

python-frontend/src/main/java/org/sonar/python/semantic/v2/ProjectLevelTypeTable.java
maksim-grebeniuk-sonarsource: tests added

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1882, opened 2024-07-26
SONARPY-2010 Avoid creating duplicate types when the corresponding symbols are duplicated

guillaume-dequenne-sonarsource:


python-checks/src/test/resources/checks/datetime_constructor_use_timezone_aware.py
```diff
@@ -36,8 +36,5 @@ def aliased_utcnow():
     # FN because we lose the FQN in the assignment
     reassigned()
 
-    from datetime.datetime import utcnow as aliased
-    aliased() # Noncompliant
-
-    from datetime.datetime import utcnow
-    utcnow() # Noncompliant
+    from datetime import datetime as aliased
+    aliased.utcnow() # Noncompliant
```
guillaume-dequenne-sonarsource, line 40: This triggered an edge case when replacing members of `ModuleType` because we'd end up replacing the `datetime.datetime` `ClassType` with an empty `ModuleType` (after unsuccessfully trying to resolve it as a module).

python-frontend/src/main/java/org/sonar/python/semantic/v2/SymbolsModuleTypeProvider.java
```diff
@@ -69,6 +75,52 @@ public ModuleType createModuleType(List<String> moduleFqn, ModuleType parent) {
...
+      }
+      // We have resolved an enclosing module, continue searching
+      currentModule = moduleType;
+    }
+    return PythonType.UNKNOWN;
+  }
+
+  public PythonType resolveLazyType(LazyType lazyType) {
+    if (lazyTypesContext.resolvedLazyTypes().containsKey(lazyType)) {
+      return lazyTypesContext.resolvedLazyTypes().get(lazyType);
```
guillaume-dequenne-sonarsource, line 104: This line is uncovered, and I think because `FunctionType#returnType` now automatically resolves lazy types when called, I don't think it can be triggered anymore. Still, I prefer to keep this defensively, as it will be possible to trigger it when we resolve more types as lazy types.

python-frontend/src/main/java/org/sonar/python/types/v2/FunctionType.java
```diff
@@ -98,6 +98,9 @@ public List<ParameterV2> parameters() {
   }
 
   public PythonType returnType() {
+    if (returnType instanceof LazyType lazyType) {
```
maksim-grebeniuk-sonarsource, line 101: wrap it to some utils, since we'll needed in a lot of places once we'll introduce lazy types for other places like superclasses

python-frontend/src/main/java/org/sonar/python/types/v2/FunctionType.java
```diff
@@ -121,4 +124,11 @@ public boolean hasVariadicParameter() {
   public PythonType owner() {
     return owner;
   }
+
+  public void resolveLazyReturnType(PythonType pythonType) {
+    if (!(returnType instanceof LazyType)) {
```
maksim-grebeniuk-sonarsource, line 129: same here, possibly we may want to move this check to utils

python-frontend/src/main/java/org/sonar/python/types/v2/LazyType.java
```diff
@@ -0,0 +1,112 @@
...
+  public PythonType resolve() {
+    PythonType resolvedType = symbolsModuleTypeProvider.resolveLazyType(this);
+    consumers.forEach(c -> c.accept(resolvedType));
+    consumers.clear();
+    return resolvedType;
+  }
+
+  @Override
+  public PythonType unwrappedType() {
+    return this.resolve().unwrappedType();
```
maksim-grebeniuk-sonarsource, line 65: since the current idea is that we never operate with lazy type outside of TypeInference, possibly we don't need to override this methods or need to throw an exception here

python-frontend/src/main/java/org/sonar/python/types/v2/ModuleType.java
```diff
@@ -36,6 +36,14 @@ public ModuleType(@Nullable String name, @Nullable ModuleType parent, Map<String
     this.name = name;
     this.parent = parent;
     this.members = members;
+    if (this.parent == null) {
+      return;
```
maksim-grebeniuk-sonarsource, line 40: having an early return in a simple constructor confused me a bit from understanding what's going on here. Maybe you can extend this logic to a private method like `registerToParent` or revert the condition

python-frontend/src/main/java/org/sonar/python/types/v2/TypeCheckBuilder.java
```diff
@@ -116,6 +116,9 @@ public IsSameAsTypePredicate(PythonType expectedType) {
 
     @Override
     public TriBool test(PythonType pythonType) {
+      if (pythonType instanceof LazyType lazyType) {
+        pythonType = lazyType.resolve();
```
maksim-grebeniuk-sonarsource, line 120: not needed since lazy type should not be accessible outside of type inference

python-frontend/src/main/java/org/sonar/python/semantic/v2/LazyTypesContext.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Queue;
+import org.sonar.python.types.v2.LazyType;
+import org.sonar.python.types.v2.PythonType;
+
+public class LazyTypesContext {
+  private final Map<String, LazyType> lazyTypes;
+  private final Map<LazyType, Queue<String>> lazyTypeRequirements;
```
maksim-grebeniuk-sonarsource, line 30: I guess this information could be inferred from LazyType itself

python-frontend/src/main/java/org/sonar/python/types/v2/LazyType.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.Queue;
+import java.util.function.Consumer;
+import org.sonar.plugins.python.api.LocationInFile;
+import org.sonar.python.semantic.v2.SymbolsModuleTypeProvider;
+
+public class LazyType implements PythonType {
+
+  String fullyQualifiedName;
+  private final Queue<Consumer<PythonType>> consumers;
+  private final SymbolsModuleTypeProvider symbolsModuleTypeProvider;
```
maksim-grebeniuk-sonarsource, line 33: I think it is better to operate with ProjectLevelTypesTable cause once you trigger lazy type resolution - it's type could be already resolved

python-frontend/src/main/java/org/sonar/python/semantic/v2/SymbolsModuleTypeProvider.java
```diff
@@ -140,16 +195,63 @@ private PythonType resolveReturnType(Map<Symbol, PythonType> createdTypesBySymbo
...
-      .orElseGet(() -> convertToType(classSymbol, createdTypesBySymbol));
+    return resolvePossibleLazyType(createdTypesBySymbol, classSymbol, fullyQualifiedName);
+  }
+
+  PythonType resolvePossibleLazyType(Map<Symbol, PythonType> createdTypesBySymbol, ClassSymbol classSymbol, String fullyQualifiedName) {
+    PythonType currentModule = rootModule;
+    String[] split = fullyQualifiedName.split("\\.");
+    StringBuilder sb = new StringBuilder();
+    for (int i = 0; i < split.length; i++) {
+      String currentElement = split[i];
```
maksim-grebeniuk-sonarsource, line 213: I think this logic is somehow implemented in `ProjectLevelTypeTable`, so maybe what we just need is to have a link to the types table in lazy type and call `getType` or `getModule` from for unresolved module and then call resolution consumers. WDYT?

python-frontend/src/main/java/org/sonar/python/semantic/v2/SymbolsModuleTypeProvider.java
maksim-grebeniuk-sonarsource: as discussed this is not needed anymore

python-frontend/src/main/java/org/sonar/python/semantic/v2/SymbolsModuleTypeProvider.java
```diff
@@ -69,6 +75,52 @@ public ModuleType createModuleType(List<String> moduleFqn, ModuleType parent) {
...
+      currentModule = moduleType;
+    }
+    return PythonType.UNKNOWN;
+  }
+
+  public PythonType resolveLazyType(LazyType lazyType) {
+    if (lazyTypesContext.resolvedLazyTypes().containsKey(lazyType)) {
+      return lazyTypesContext.resolvedLazyTypes().get(lazyType);
+    }
+    Queue<String> modules = lazyTypesContext.lazyTypeRequirements().get(lazyType);
```
maksim-grebeniuk-sonarsource, line 106: same here, if LazyType will have a link to project level type table then this  entire method is not needed

maksim-grebeniuk-sonarsource (changes requested):
Looks good! I think it requires some changes (as discussed and mentioned in the comments) to get rid of duplicated logic for the modules resolution and unused `LazyType`'s methods

python-frontend/src/main/java/org/sonar/python/types/v2/FunctionType.java
guillaume-dequenne-sonarsource: I guess this case is a bit different since `resolveLazyReturnType` is essentially a setter that can be triggered only if the current return type is a lazy type. I still want to prevent using it if the current return type is not a lazy type.

python-frontend/src/main/java/org/sonar/python/types/v2/LazyType.java
guillaume-dequenne-sonarsource: I'm still not sure on that one, because `LazyType` themselves are created in `SymbolsModuleTypeProvider` during some actual conversions from symbols to types.

While the current architecture is already a bit odd, I feel it would be even weirder if `SymbolsModuleTypeProvider` held a reference to `ProjectLevelTypeTable` simply for the sake of passing it to `LazyType`. I don't really have a strong opinion here though; and.I feel this is temporary code anyway since it will only exist while we are doing symbol to type conversions.

python-frontend/src/main/java/org/sonar/python/semantic/v2/SymbolsModuleTypeProvider.java
guillaume-dequenne-sonarsource: I moved the `getType` implementation in `SymbolsModuleTypeProvider` (for the reason highlighted in my other comment: that I'm not sure it's a good idea to pass `ProjectLevelTypeTable` to `SymbolsModuleTypeProvider` then to `LazyType`). We can discuss this further though.

maksim-grebeniuk-sonarsource (changes requested):
From the current implementation perspective the code looks good.
As we discussed I guess the implementation will keep proper responsibilities if the implementation will look like:
## getModule lazy type workflow
1. `ProjectLevelTypeTable.getModule()`
2. Create module type with members
2.2. For a return type create a lazy type without any checks for existing types, add them to LazyContext
2.3 return module type
3. check if newly created lazy types already have types in the type table, resolve for existing ones
4. check if there are previously created lazy types exist for just created module type, resolve it
5. if there is no requested module  - then set a type of all lazy types assigned to this module to `UNKNOWN`
6. if there is a module but no member for lazy type - set its type to unknown

## LazyType resolving when accessed
1. call projectLevelTypeTable.getModule()
2. if a type is still not resolved, resolve it to UNKNOWN

As result: 
- SymbolModuleTypeProvider will be still just a convertor from Symbol to Type model
- Logic of LazyType creation will be simple without any checks and conditions
- LazyTypes resolution will operate with types API independent on symbols layer

maksim-grebeniuk-sonarsource (approved):
LGTM, great job, thanks!
As discussed, the only change needed is to rewrite and clean up a bit the `SymbolsModuleTypeProvider.resolvePossibleLazyType` method

PR #1883, opened 2024-07-26
SONARPY-2032 Adapt (executable) lines of code metric for Jupyter Notebooks

ghislainpiot:


sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonScanner.java
```diff
@@ -302,7 +302,7 @@ private static NewIssueLocation newLocation(PythonInputFile inputFile, NewIssue
   }
 
   private void saveMeasures(PythonInputFile inputFile, PythonVisitorContext visitorContext) {
-    FileMetrics fileMetrics = new FileMetrics(visitorContext);
+    FileMetrics fileMetrics = new FileMetrics(visitorContext, inputFile instanceof GeneratedIPythonFile);
```
joke1196, line 305: Here you could use the `kind()` method on the inputFile instead of `instanceof`.

joke1196 (changes requested):
LGTM! Just a small change to use the PythonInputFile API

joke1196 (approved):
LGTM. It is a small detail but you could add a test in PythonSensor to cover the line in the PythonScanner, when passing a GeneratedIpython file or a default input file.

PR #1884, opened 2024-07-26
SONARPY-1990: Map JSON locations to python locations in IPythonTreeMaker

joke1196:


python-frontend/src/main/java/org/sonar/python/tree/TokenEnricher.java
```diff
@@ -0,0 +1,61 @@
...
+  private TokenEnricher() {
+  }
+
+  public static List<TokenImpl> enrichTokens(List<Token> tokens, Map<Integer, IPythonLocation> offsetMap) {
+    return tokens.stream().map(token -> enrichToken(token, offsetMap)).toList();
+  }
+
+  public static TokenImpl enrichToken(Token token, Map<Integer, IPythonLocation> offsetMap) {
+    if (!offsetMap.isEmpty()) {
+      IPythonLocation location = offsetMap.get(token.getLine());
```
ghislainpiot, line 40: While testing, I had to add 
```
if (token.getType() == PythonTokenType.DEDENT || token.getType() == GenericTokenType.EOF){
        return new TokenImpl(token);
      }
```

but it might be premature

python-frontend/src/main/java/org/sonar/python/tree/PythonTreeMaker.java
```diff
@@ -433,17 +433,17 @@ public ImportStatement importStatement(StatementWithSeparator statementWithSepar
...
       .getFirstChild(PythonGrammar.DOTTED_AS_NAMES)
       .getChildren(PythonGrammar.DOTTED_AS_NAME).stream()
-      .map(PythonTreeMaker::aliasedName)
+      .map(this::aliasedName)
       .toList();
     return new ImportNameImpl(importKeyword, aliasedNames, separators);
   }
 
-  private static ImportFrom importFromStatement(AstNode astNode, Separators separators) {
+  private  ImportFrom importFromStatement(AstNode astNode, Separators separators) {
```
ghislainpiot, line 446: nitpick : There is an extra space, maybe the file should be reformatted

ghislainpiot (changes requested):
It looks very good. I only have concerns with formatting and two files that became executable. 

I quickly tried plugging it in the Scanner, and found quite a few problems, but they should probably be tackled in the next ticket

python-frontend/src/main/java/org/sonar/python/tree/TokenEnricher.java
joke1196: No, I think you are right, I should test that as we are adding these ourselves, we need to take care of these tokens. I will add some tests, and correct the behavior. And as I decided to throw rapidly this should be taken into account.

joke1196: I will create a ticket for this as it would be a good idea to enrich them. For example just not to have EOF tokens before other tokens

PR #1885, opened 2024-07-29
SONARPY-2017 Avoid type-related FPs in case of isinstance checks

maksim-grebeniuk-sonarsource:


python-checks/src/test/resources/checks/nonCallableCalled.py
```diff
@@ -269,9 +269,9 @@ def foo():
 def using_isinstance_with_runtime_type():
     my_non_callable = MyNonCallable()
     if isinstance(my_non_callable, whatever):
-        my_non_callable() # Noncompliant
+        my_non_callable()
```
guillaume-dequenne-sonarsource, line 272: This was not a FP, because `my_non_callable` does not come from a type hint (it is `EXACT` source).
Our previous implementation only removed type information for declared types, not for runtime type. I think we should keep it that way.

If it's not possible I'd expect at least a ticket to fix this now FN.

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/FlowSensitiveTypeInference.java
```diff
@@ -96,6 +99,8 @@ public void updateProgramState(Tree element, ProgramState programState) {
       handleLoopAssignment(element, state);
     } else {
       // Here we should run "isinstance" visitor when we handle declared types, to avoid FPs when type guard checks are made
```
guillaume-dequenne-sonarsource, line 101: This comment is probably not needed anymore once this is implemented.

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -1843,6 +1843,58 @@ void basic_imported_symbols() {
...
+        x = 10
+        if isinstance(x.b, Foo):
+          ...
+        x
+      """).typeV2();
+
+    Assertions.assertThat(xType.unwrappedType()).isSameAs(INT_TYPE);
+
+    xType = lastExpression("""
+      def foo():
```
guillaume-dequenne-sonarsource, line 1879: In the test you took this from, `x: int` is a parameter of `foo`. Without it, I don't think the test makes sense since `x` is going to be unknown no matter what.

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -1843,6 +1843,58 @@ void basic_imported_symbols() {
...
+  void isInstanceTests() {
+    var xType = lastExpression("""
+      def foo():
+        x = 10
+        if isinstance(x, Foo):
+          ...
+        x
+      """).typeV2();
+
+    Assertions.assertThat(xType).isSameAs(PythonType.UNKNOWN);
```
guillaume-dequenne-sonarsource, line 1856: It's good to document the behavior anyway, but this is not what we'd expect (c.f my other comment).

guillaume-dequenne-sonarsource (changes requested):
The goal of this ticket is to avoid FPs for type-hinted types that are further type-guarded.
For types we know to be exact, we don't want to take the `isinstance` check into account (because it doesn't provide us new information, since we already know the exact type).

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1887, opened 2024-07-29
SONARPY-2018 Only analyze notebooks with a Python kernel

ghislainpiot:


joke1196 (changes requested):
As discussed it would be a good idea to split the parsing of the language and the processing of cells. For now we can live without this optimization, so no need to prioritize this change request.

PR #1889, opened 2024-07-30
SONARPY-2040 Add the sonar.ipynb.file.suffixes property

ghislainpiot:


ghislainpiot:
The ruling crashes are to be expected, since before the IPython sensor was not running

joke1196 (approved):
This looks good to me! Let's rebase on MMF-3813, we should get a better idea of what is left to fix.

PR #1890, opened 2024-07-30
SONARPY-2024 Ensure call expressions return types have same type source as their callee

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/tree/CallExpressionImpl.java
```diff
@@ -203,8 +209,15 @@ public PythonType typeV2() {
...
-      return result == null ? PythonType.UNKNOWN : new ObjectType(result);
+      return result == null ? PythonType.UNKNOWN : new ObjectType(result, typeSource);
     }
     return PythonType.UNKNOWN;
   }
+
+  TypeSource computeTypeSource() {
+    if (callee instanceof QualifiedExpression qualifiedExpression) {
+      return qualifiedExpression.qualifier().typeV2().typeSource();
+    }
```
maksim-grebeniuk-sonarsource, line 220: I guess in some cases we can compute the type source out of `Name` like in the example below:
```python
class A:
    def __call__(self) -> int:
        return 42
    
def foo(a: A):
    a.__call__() # the type source is coming from qualified expression and it is TYPE_HINT
    a() # the type source is coming from name and it is the same - TYPE_HINT
```

maksim-grebeniuk-sonarsource (changes requested):
Looks good, the only thing missing is to resolve a callee type source when called is a `Name`

python-frontend/src/main/java/org/sonar/python/tree/CallExpressionImpl.java
```diff
@@ -187,13 +188,23 @@ public List<Expression> typeDependencies() {
...
-    if (callee().typeV2() instanceof ClassType classType) {
-      return new ObjectType(classType);
+    TypeSource typeSource = computeTypeSource();
+    PythonType pythonType = returnTypeOfCall(callee().typeV2());
+    return pythonType != PythonType.UNKNOWN ? new ObjectType(pythonType, typeSource) : PythonType.UNKNOWN;
+  }
+
+  static PythonType returnTypeOfCall(PythonType calleeType) {
+    if (calleeType instanceof ClassType classType) {
+      return classType;
```
maksim-grebeniuk-sonarsource, line 198: should it be an ObjectType of a ClassType?

python-frontend/src/main/java/org/sonar/python/tree/CallExpressionImpl.java
```diff
@@ -187,13 +188,23 @@ public List<Expression> typeDependencies() {
...
-      return functionType.returnType();
+    if (calleeType instanceof FunctionType functionType) {
+      PythonType returnType = functionType.returnType();
+      if (returnType.equals(PythonType.UNKNOWN)) {
+        return PythonType.UNKNOWN;
+      }
+      return returnType;
     }
-    if (callee().typeV2() instanceof UnionType unionType) {
+    if (calleeType instanceof UnionType unionType) {
```
maksim-grebeniuk-sonarsource, line 207: I guess it makes sense to make union-type case processing recursively - get the return type of every candidate and join them over the `UnionType.or` method

maksim-grebeniuk-sonarsource (changes requested):
Overall looks good, I just have two comments, ready to discuss

python-frontend/src/main/java/org/sonar/python/tree/CallExpressionImpl.java
guillaume-dequenne-sonarsource: The actual `ObjectType` is only created when its `TypeSource` is known. In `typeV2()`, I get both the type and the source from different methods, then instantiate the `ObjectType` with that information.

python-frontend/src/main/java/org/sonar/python/tree/CallExpressionImpl.java
guillaume-dequenne-sonarsource: `UnionType.or` itself should already be recursive. However, I made some slight changes to make this piece of code more intentional.
In particular, if the original `UnionType` contains anything unexpected, I think we should return `PythonType.UNKNOWN`.

python-frontend/src/main/java/org/sonar/python/tree/CallExpressionImpl.java
```diff
@@ -187,24 +189,46 @@ public List<Expression> typeDependencies() {
...
+  static PythonType returnTypeOfCall(PythonType calleeType) {
+    if (calleeType instanceof ClassType classType) {
+      return classType;
     }
-    if (callee().typeV2() instanceof FunctionType functionType) {
-      return functionType.returnType();
+    if (calleeType instanceof FunctionType functionType) {
+      PythonType returnType = functionType.returnType();
+      if (returnType.equals(PythonType.UNKNOWN)) {
+        return PythonType.UNKNOWN;
```
maksim-grebeniuk-sonarsource, line 204: do we need this condition since UNKNOWN is singleton?

python-frontend/src/main/java/org/sonar/python/tree/CallExpressionImpl.java
```diff
@@ -187,24 +189,46 @@ public List<Expression> typeDependencies() {
...
+      }
+      return returnType;
     }
-    if (callee().typeV2() instanceof UnionType unionType) {
-      PythonType result = null;
+    if (calleeType instanceof UnionType unionType) {
+      Set<PythonType> types = new HashSet<>();
       for (PythonType candidate : unionType.candidates()) {
         if (candidate instanceof ClassType classType) {
```
maksim-grebeniuk-sonarsource, line 211: why can't you instead this conditional checking for instanceof call:
```java
for (var candidate : unionType.candidates()) {
  var candidateReturnType = returnTypeOfCall(candidate)
  ...
}
```

guillaume-dequenne-sonarsource: Yes indeed, thanks! I now realize I missed your initial point about recursion, sorry about that.

PR #1891, opened 2024-07-30
SONARPY-2036: Enrich EOF tokens

joke1196:


ghislainpiot (approved):
Thanks for taking care of this, it looks good to me !

PR #1894, opened 2024-07-31
SONARPY-2043: Deactivate CPD for Notebooks

joke1196:


sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonScanner.java
```diff
@@ -308,7 +308,9 @@ private void saveMeasures(PythonInputFile inputFile, PythonVisitorContext visito
     noSonarFilter.noSonarInFile(inputFile.wrappedFile(), fileLinesVisitor.getLinesWithNoSonar());
 
     if (!isInSonarLint(context)) {
-      cpdAnalyzer.pushCpdTokens(inputFile.wrappedFile(), visitorContext);
+      if (inputFile.kind() != PythonInputFile.Kind.IPYTHON) {
```
ghislainpiot, line 311: I think the condition should be inverted to make it more explicit.

ghislainpiot (changes requested):

ghislainpiot (approved):
LGTM!

PR #1895, opened 2024-07-31
SONARPY-2034 Refactor IPythonLocation to hold an ordered list rather than a linkedHashMap

ghislainpiot:


joke1196 (changes requested):
This looks good! Thanks for taking care of this. The only thing I would suggest would be to extract the number of escaped char from the list. When we will merge SONARPY-2045 you will see that the number of escaped chars would be the list size. So I would wait for the merge of 2045 and rebase then, you will be able to simplify this PR a bit.

PR #1897, opened 2024-07-31
SONARPY-2045: Fix incorrect count of escaped chars

joke1196:


ghislainpiot (approved):
Nice catch! It looks good to me!

PR #1899, opened 2024-08-02
SONARPY-2033: Adapt Highlighter for Jupyter Notebooks

joke1196:


python-frontend/src/main/java/org/sonar/python/TestPythonVisitorRunner.java
```diff
@@ -81,9 +97,16 @@ public static ProjectLevelSymbolTable globalSymbols(List<File> files, File baseD
...
+    var parser = PythonParser.createIPythonParser();
+    var treeMaker = new IPythonTreeMaker(locations);
+    var astNode = parser.parse(content);
+    return treeMaker.fileInput(astNode);
+  }
+
   private static FileInput parseFile(TestPythonFile file) {
     var parser = file.isIPython() ? PythonParser.createIPythonParser() : PythonParser.create();
-    var treeMaker = file.isIPython() ? new IPythonTreeMaker() : new PythonTreeMaker();
+    var treeMaker = file.isIPython() ? new IPythonTreeMaker(Map.of()) : new PythonTreeMaker();
```
ghislainpiot, line 109: The behavior might be surprising, but I don't know how to fix it

ghislainpiot (approved):
Thanks for taking care of this, looks good to me!

python-frontend/src/main/java/org/sonar/python/TestPythonVisitorRunner.java
joke1196: Yes I thought making it explicit would simplify things. but we could still make it better during the hardening sprint

PR #1900, opened 2024-08-02
SONARPY-1985 Infer a function return type based on its type hint for locally defined functions

guillaume-dequenne-sonarsource:
It would probably make sense to review this only after SONARPY-2024 is merged, and this PR is rebased.

I have my doubts on the introduction of `TypeOrigin` and would be happy to discuss it. I feel it actually should be included in `TypeSource`, but I felt this would complexify the propagation of type sources, so this solution felt like a simpler step, that we can revisit later.


maksim-grebeniuk-sonarsource (changes requested):
As discussed on call:
In addition to functions defined in the currently analyzing file, the `TypeOrigin.LOCAL` should be set to functions defined in another project files

maksim-grebeniuk-sonarsource (approved):
LGTM

PR #1901, opened 2024-08-02
SONARPY-2049 Remove extra line break when the last line of a source array ends with \n

ghislainpiot:


joke1196 (approved):
LGTM! That was a good catch on the empty cell!

PR #1904, opened 2024-08-05
Fix microsoft stub tests

joke1196:


ghislainpiot (approved):
One day, we should really fix the instability

PR #1910, opened 2024-08-07
SONARPY-2056 IPython line magics starting with numbers or special characters should be parsed correctly

ghislainpiot:


joke1196 (approved):
LGTM! I am not sure I understand what the `name` was used for before.

ghislainpiot:
There was no conflict when I made the change, so it was probably for future proofing

PR #1912, opened 2024-08-07
SONARPY-2059 PythonScanner should log the parsing exception on the correct line for notebooks

ghislainpiot:
- **SONARPY-2059 PythonScanner should log the parsing exception on the correct line for notebooks**
- **Add the notebook**


joke1196 (approved):
Looks good. Thank you for taking care of this. I was wondering though if we could streamline it by either subclassing the exception with our own or if  the exception is only thrown by the TreeMaker (which I don't think it is) we could override `recognitionException` in the IPythonTreeMaker. Did you look into such cases? For me it is good like as we know we will catch all of them.

ghislainpiot:
The `RecognitionException` comes straight from SSLR and the `PythonParser`, which don't have knowledge of the remapping. I don't think they should.

PR #1920, opened 2024-08-09
SONARPY-2067 First line of the first cell of notebooks should end at the correct location

ghislainpiot:


sonar-python-plugin/src/test/java/org/sonar/plugins/python/IpynbNotebookParserTest.java
```diff
@@ -138,6 +138,10 @@ void testParseNotebookSingleLine() throws IOException {
     assertThat(result.locationMap()).hasSize(5);
     assertThat(result.contents()).hasLineCount(5);
     // position of variable t
-    assertThat(result.locationMap().get(4).column()).isEqualTo(451);
+    assertThat(result.locationMap().get(4).column()).isEqualTo(452);
```
joke1196, line 141: Here the column count should start at 0
as stated in sonar-python-plugin/src/test/java/org/sonar/plugins/python/PythonHighlighterTest.java line 225. The t should be the 451st character no? Why did you need to change the count? Was the issue only related to the first line? in which case?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IpynbNotebookParser.java
```diff
@@ -171,6 +172,10 @@ private boolean parseSourceMultilineString(JsonParser jParser, JsonToken jsonTok
       aggregatedSource.append("\n");
       previousLen = previousLen + line.length() + 2;
```
joke1196, line 173: I do not understand why the first line of the multiline cell should have 1 more character

joke1196 (changes requested):
I am a bit confused with this PR as we are shifting by one all column count on the single line source. How did you find out about this issue. I have a few doubts on why the length of first line of the string should be different than length + 2 characters for the `\n`

sonar-python-plugin/src/test/java/org/sonar/plugins/python/IpynbNotebookParserTest.java
ghislainpiot: I checked which character it was manually, and at 451 the location is between the `\` and the `n`, which was incorrect.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IpynbNotebookParser.java
ghislainpiot: The extra character is need to align the behavior between the multiline strings cells and the array cells. The location starts on the first `"`, before the Python code. The function did not account for that situation previously.

ghislainpiot:
I found out about the issue when doing 2063, I was adding the tests and realized that the line was starting in the middle of a bunch of spaces.

joke1196:
I see thank you for the clarification! Is there a way we could add a test that would mirror the issue you found in 2063? This way there would be no doubt on how the locations should be computed.

joke1196 (approved):
LGTM! I can see why it was shifted if we started with the first character being an escaped character. This is the reason it worked most of the time... Good catch!

PR #1922, opened 2024-08-09
SONARPY-2057: Markdown content should not leak in the generated file

joke1196:


ghislainpiot (approved):
This looks great! Thanks for extracting the two parts of the parser.

PR #1923, opened 2024-08-09
SONARPY-2063 Multiline strings should have a correct end position

ghislainpiot:


python-frontend/src/main/java/org/sonar/python/TokenLocation.java
```diff
@@ -36,9 +36,13 @@ public TokenLocation(Token token) {
...
     if (lines.length > 1) {
-      endLine = token.line() + lines.length - 1;
-      endLineOffset = lines[lines.length - 1].length();
-
+      if (token.isCompressed()) {
+        endLine = token.line();
+        endLineOffset = this.startLineOffset + token.valueLength();
+      } else {
+        endLine = token.line() + lines.length - 1;
+        endLineOffset = lines[lines.length - 1].length();
```
joke1196, line 44: Not sure when this happens but shouldn't we just compute as well the `valueLength()` of the last line when uncompressed?

joke1196 (changes requested):
Just a small question when uncompressed. Otherwise looks good thanks!

joke1196 (approved):
Thank you for the clarification!

PR #1926, opened 2024-08-12
SONARPY-2066:IPython help command should be parsed correctly

joke1196:


ghislainpiot (approved):
LGTM! 
The generated token looks a bit odd to me, it is a `DynamicObjectInfoStatementImpl` with no children and no question marks after. If we want to make a rule targeting the help command, we will need to change the token.

joke1196:
Yes agreed. I will add a test to clarify the intention.

PR #1927, opened 2024-08-12
SONARPY-2070 Fix NPE when type inference visit is missing

maksim-grebeniuk-sonarsource:


python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -2329,6 +2330,20 @@ def foo():
...
+      def foo():
+          subscription[call('goal_{}'.format(1))] = 1
+      """);
+
+    var literal = TreeUtils.firstChild(fileInput, StringLiteral.class::isInstance)
+      .map(StringLiteral.class::cast)
+      .get();
+
+    var type = literal.typeV2();
+    Assertions.assertThat(type).isNotNull();
```
guillaume-dequenne-sonarsource, line 2345: I think you could simply assert that it's a string type now that the NPE is fixed.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #1934, opened 2024-08-16
SONARPY-2074: Fix count of line of codes for single line notebooks

joke1196:


ghislainpiot (approved):
LGTM once the SQ analysis is ok !

PR #1936, opened 2024-08-19
SONARPY-2082 Jupyter notebooks should not have their fileLinesContext filled

ghislainpiot:


joke1196 (approved):
This looks good to me. I am not sure exactly what the NCLOC_DATA_KEY is used for, but it seems it could be removed.

ghislainpiot:
The documentation says : 
```Information about lines of code in file. Key-value pairs, where key - is a number of line, and value - is an indicator of whether line contains code (1) or not (0). If a line number is missing in the map it is equivalent to the default value (0).```

I don't think this works in the context of virtual files and especially Jupyter notebooks, and the LOC appears to still register correctly.

PR #1948, opened 2024-08-27
SONARPY-2087 Formatted raw strings with double backslashes should be parsed correctly

ghislainpiot:


python-frontend/src/main/java/org/sonar/python/lexer/FStringChannel.java
```diff
@@ -154,7 +157,7 @@ private static boolean canConsumeFStringPrefix(StringBuilder sb, CodeReader code
   }
 
   private static boolean isRawStringBackSlash(CodeReader code, FStringState state) {
```
joke1196, line 159: Maybe for future us and for clarity we could be super explicit on the function name to only target single backslashes.
```suggestion
  private static boolean isRawStringSingleBackSlash(CodeReader code, FStringState state) {
```

joke1196 (approved):
LGTM. Thank for diving into the F-string world. I left a small suggestion for naming.

PR #1953, opened 2024-09-06
SONARPY-1902 Rule S6979: "torch.tensor" should be used instead of "torch.autograd.Variable"

Seppli11:
When merging this PR, the corresponding RSPEC [PR](https://github.com/SonarSource/rspec/pull/3959) should be merged as well.


joke1196 (approved):
This looks really good! Thanks a lot for the work!

PR #1954, opened 2024-09-09
SONARPY-2115 Migrate S1244 FloatingPointEqualityCheck to the V2 type …

guillaume-dequenne-sonarsource:
…model


ghislainpiot (approved):
LGTM!

I am surprised to see that we do have quickfixes that create imports after all.

guillaume-dequenne-sonarsource:
> LGTM!
> 
> I am surprised to see that we do have quickfixes that create imports after all.

Indeed, we have some, but the ordering is not really made cleanly...So that may be a problem we still want to address.

PR #1955, opened 2024-09-09
SONARPY-1907 implement Rule S6983 : The nb_workers parameter should be specified for torch.utils.data.DataLoader

Seppli11:


python-checks/src/main/java/org/sonar/python/checks/PyTorchDataLoaderNumWorkersCheck.java
```diff
@@ -0,0 +1,63 @@
...
+        && !checkIfUnpackArgIsPresent(arguments)) {
+
+        ctx.addIssue(callExpression.callee(), MESSAGE);
+      }
+    });
+  }
+
+  private static boolean checkIfNumWorkersArgIsPresent(List<Argument> arguments) {
+    RegularArgument numWorkersArg = TreeUtils.nthArgumentOrKeyword(NUM_WORKERS_ARG_POSITION, NUM_WORKERS_ARG_NAME, arguments);
+    return numWorkersArg == null;
```
joke1196, line 57: not sure if it is more readable, but maybe we don't need the assignment here.

joke1196 (changes requested):
This looks good. I added a few comments and suggestions mainly for naming conventions.

joke1196 (approved):
LGTM! One small detail before you merge I think you can make it that the coverage reaches 100%. I know we did it before already but it will not require a lot of time.

PR #1956, opened 2024-09-09
SONARPY-1914 Usage of "torch.load" can lead to untrusted code execution

Seppli11:


python-checks/src/main/java/org/sonar/python/checks/TorchLoadLeadsToUntrustedCodeExecutionCheck.java
```diff
@@ -0,0 +1,44 @@
...
+
+  public static final String TORCH_LOAD = "torch.load";
+  public static final String MESSAGE = "Replace this call with a safe alternative";
+
+  @Override
+  public void initialize(Context context) {
+    context.registerSyntaxNodeConsumer(Tree.Kind.CALL_EXPR, ctx -> {
+      CallExpression callExpression = (CallExpression) ctx.syntaxNode();
+      Symbol calleeSymbol = callExpression.calleeSymbol();
+      if(calleeSymbol != null && TORCH_LOAD.equals(calleeSymbol.fullyQualifiedName())) {
```
Seppli11, line 39: `calleeSymbol != null` is missing test coverage. However, this is hard to test for and would complicate the tests unnecessarily

guillaume-dequenne-sonarsource: Are you sure? If you search for `calleeSymbol != null` within the project, this is a construct that is used many times and is most likely tested somewhere else. I think ensuring this is well-tested should not be too complex.

Seppli11: It would require something like `(lambda x: x)()` in the lit file, at least as far as I understand it. Or is there a test case which makes sense test where this can occure?

guillaume-dequenne-sonarsource: To cover this, you only have to have a case that triggers it in the unit test file of the rule (i.e `torchLoadLeadsToUntrustedCodeExecution.py`). Callees can be missing symbols for many reasons, indeed as an immediately called lambda, or probably even something simpler like `something[42]()`. There are most likely many other cases that could trigger this as well.

joke1196 (changes requested):
The implementation looks good. The only issue is that maybe we should refine a bit to not raise when `weights_only` is set to `True`. In the PyTorch documentation it is stated:
```
torch.load() unless weights_only parameter is set to True, uses pickle module implicitly, which is known to be insecure. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling. Never load data that could have come from an untrusted source in an unsafe mode, or that could have been tampered with. Only load data you trust.
``` 
This means we should update as well the RSPEC PR, specifying that this issue will not be raised in such a case.

python-checks/src/test/resources/checks/torchLoadLeadsToUntrustedCodeExecution.py
```diff
@@ -0,0 +1,20 @@
...
+some_path = ...
+model = load(some_path) # Noncompliant
+
+torch.load(model2, 'model.pth', weights_only=False) #Noncompliant
+
+model2 = ...
+safetensors.torch.load_model(model2, 'model.pth')
+
+torch.load(model2, 'model.pth', weights_only=True)
+torch.load(model2, 'model.pth', weights_only=some_value)
```
joke1196, line 17: Here what if `some_value` is set to `False`. You could maybe add a test for that. I think it would be a good idea to raise in such a case. I would suggest to look into `ReachingDefinitionAnalysis` in order to figure out how we can check the assigned value of `some_value`. I think it make sense to go that road but I would not go deeper for example if a function is called for the `weights_only` parameter. I would still add it as a test to better show the intent to the next person who will read this rule. ```torch.load(..., weights_only=foo()) # Compliant```.

python-checks/src/main/java/org/sonar/python/checks/TorchLoadLeadsToUntrustedCodeExecutionCheck.java
```diff
@@ -0,0 +1,63 @@
...
+import org.sonar.plugins.python.api.tree.Name;
+import org.sonar.plugins.python.api.tree.RegularArgument;
+import org.sonar.plugins.python.api.tree.Tree;
+import org.sonar.python.tree.TreeUtils;
+
+@Rule(key = "S6985")
+public class TorchLoadLeadsToUntrustedCodeExecutionCheck extends PythonSubscriptionCheck {
+
+  public static final String TORCH_LOAD = "torch.load";
+  public static final String MESSAGE = "Replace this call with a safe alternative";
```
joke1196, line 38: The period at the end of the message is missing :)

python-checks/src/main/java/org/sonar/python/checks/TorchLoadLeadsToUntrustedCodeExecutionCheck.java
```diff
@@ -0,0 +1,63 @@
...
+    context.registerSyntaxNodeConsumer(Tree.Kind.CALL_EXPR, ctx -> {
+      CallExpression callExpression = (CallExpression) ctx.syntaxNode();
+      Symbol calleeSymbol = callExpression.calleeSymbol();
+      if (calleeSymbol != null && TORCH_LOAD.equals(calleeSymbol.fullyQualifiedName()) && !isWeightsOnlySetToTrue(callExpression.arguments())) {
+        ctx.addIssue(callExpression.callee(), MESSAGE);
+      }
+    });
+  }
+
+  private static boolean isWeightsOnlySetToTrue(List<Argument> arguments) {
```
joke1196, line 53: Currently if `weights_only` has not the kind `NAME`, we will return `false` which will raise the issue. 
I think you could add a test with a function:
```
def foo():
    ...
load(..., weights_only=foo())
```
here we should not raise an issue as we cannot be sure of the return value of `foo()`.

joke1196 (changes requested):
Thanks for the work! I have added a few comments as this version could raise some FPs.

joke1196 (approved):
Looks good thank. One small detail here, it would have been great to split the code from TreeUtils in another PR. Here I find it is fine as the code
is quite concise, but for bigger change having a dedicated PR would make it easier to review.

PR #1957, opened 2024-09-10
SONARPY-2118 Ensure function parameter types are correctly mapped in V2 type model

ghislainpiot:


python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -581,6 +584,71 @@ def foo3(a: int = "123"):
...
+      """);
+
+    var functionDef = (FunctionDef) root.statements().statements().get(0);
+    assertThat(((FunctionType) functionDef.name().typeV2()).parameters().get(0).declaredType().unwrappedType()).isEqualTo(INT_TYPE);
+    assertThat(((FunctionType) functionDef.name().typeV2()).parameters().get(1).declaredType().unwrappedType()).isEqualTo(TUPLE_TYPE);
+    assertThat(((FunctionType) functionDef.name().typeV2()).parameters().get(2).declaredType().unwrappedType()).isEqualTo(DICT_TYPE);
+  }
+
+  @Test
+  void inferFunctionParameterTypes2() {
```
guillaume-dequenne-sonarsource, line 601: I'm not sure I understand this test. I guess `param: A` is unknown because the class `A` is defined twice, and only the trivial type inference has run at this point. Is that correct?

Do we need to have the `foo` methods in `A` for that behavior to show (if not, I would suggest to remove them - especially as their name clashes with the function `foo`, which adds confusion).

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -581,6 +584,71 @@ def foo3(a: int = "123"):
...
+    var functionDef = (FunctionDef) root.statements().statements().get(2);
+    assertThat(((FunctionType) functionDef.name().typeV2()).parameters().get(0).declaredType().unwrappedType()).isEqualTo(PythonType.UNKNOWN);
+    assertThat(((FunctionType) functionDef.name().typeV2()).returnType().unwrappedType()).isEqualTo(PythonType.UNKNOWN);
+  }
+
+  @Test
+  void inferFunctionParameterTypes3() {
+    FileInput root = inferTypes("""
+      class A:
+        def foo():
```
guillaume-dequenne-sonarsource, line 622: Here probably the method `foo` is not strictly needed, although it's nice to see the type is correctly mapped despite the name "conflict", so I don't mind keeping that.

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -581,6 +584,71 @@ def foo3(a: int = "123"):
...
+      """);
+
+    var functionDef = (FunctionDef) root.statements().statements().get(2);
+    var patternType = ((Name) ((ExpressionStatementImpl) root.statements().statements().get(1)).expressions().get(0)).typeV2();
+
+    assertThat(((FunctionType) functionDef.name().typeV2()).parameters().get(0).declaredType().unwrappedType()).isEqualTo(patternType);
+    assertThat(((FunctionType) functionDef.name().typeV2()).returnType().unwrappedType()).isEqualTo(patternType);
+  }
+
+
```
guillaume-dequenne-sonarsource, line 651: Another test could be with type aliases, or anything else that may rely on dataflow,  such as:

```
my_alias = int
def foo(param: my_alias): ...
```

I would expect `param` not to be mapped to `int`, but since we've relied for long on the heuristic of "single assigned names", it may actually work...Having a test would be good to make sure what the behavior is.

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -648,6 +644,16 @@ def foo(param: Pattern) -> Pattern:
...
 
+  @Test
+  void inferFunctionParameterTypes5() {
+    FileInput root = inferTypes("""
+      my_alias = int
+      def foo(param: my_alias): ...
+      """);
+
+    var functionDef = (FunctionDef) root.statements().statements().get(1);
+    assertThat(((FunctionType) functionDef.name().typeV2()).parameters().get(0).declaredType().unwrappedType()).isEqualTo(INT_TYPE);
```
guillaume-dequenne-sonarsource, line 655: I'm surprised this works, that's pretty nice!
Now the question would be, what if it's assigned multiple times?
```
my_alias = int
my_alias = str
def foo(param: my_alias): ...
```
or even:
```
if cond:
    my_alias = int
else:
    my_alias = str
def foo(param: my_alias): ...
```

guillaume-dequenne-sonarsource (approved):
Thanks for the added test cases!
I left a suggestion to have a further test. Sorry, I initially did not request it because I didn't think the type alias one would actually be covered. 
Looks good though!

PR #1961, opened 2024-09-10
SONARPY-2114 Migrate S5707 ExceptionCauseTypeCheck to the V2 type model

guillaume-dequenne-sonarsource:
Notes:
- If this approach of using a `TypeWrapper` is accepted, I think it would make sense to refactor the implementation of `FunctionType#returnType` to use it as well, instead of special casing it. I didn't make this refactor here, to keep this PR simple. However, I think we could then create another ticket/PR to harmonize our implementations.
- I am not fully certain of the terminology around `TypeWrapper`, given we already have `ObjectType#unwrappedType` that serves a different purpose. I wouldn't want this to be confusing, and I am open to suggestions in case another name makes more sense.
- I also have doubts about whether `TypeWrapper` should be exposed in the various APIs (as it currently is) or if it should be made transparent by converting them on the fly to `PythonType.`. I chose the first approach for simplicity and also to avoid converting collections such as  `List<TypeWrapper>` to `List<PythonType>` on each call, but I'm open to revisit it.


maksim-grebeniuk-sonarsource (approved):
LGTM

PR #1962, opened 2024-09-11
SONARPY-1915 implement S6929 for PyTorch

Seppli11:


Seppli11:
How can I force the symbol to not be null, but not have a fully qualified name?

guillaume-dequenne-sonarsource:
> How can I force the symbol to not be null, but not have a fully qualified name?

I think you can trigger this case if you import the same name twice from different places:
```
from something import foo
from something_else import foo

foo() # FQN: null
```

In this case, because the FQN of the symbol that is actually used can either be `something.foo` or `something_else.foo`, the FQN is set to null to avoid possible FPs.

python-checks/src/main/java/org/sonar/python/checks/TfPyTorchSpecifyReductionAxisCheck.java
```diff
@@ -0,0 +1,122 @@
...
+        createTfQuickFix(callExpression).ifPresent(issue::addQuickFix);
+      }
+
+      if (isPyTorchReductionMissingDimArg(symbol, callExpression)) {
+        context.addIssue(callExpression.callee(), PY_TORCH_MESSAGE);
+      }
+    }
+  }
+
+  private static boolean hasUnpackArgument(List<Argument> arguments) {
```
joke1196, line 101: I was not sure but did you already migrated this method to TreeUtils. I think I saw this one several times now.

python-checks/src/test/java/org/sonar/python/checks/TfPyTorchSpecifyReductionAxisCheckTest.java
```diff
@@ -0,0 +1,65 @@
...
+    PythonQuickFixVerifier.verifyNoQuickFixes(new TfPyTorchSpecifyReductionAxisCheck(),
+        """
+        from tensorflow import math
+        math.reduce_all()
+        """);
+  }
+
+  @Test
+  void testPyTorch() {
+    PythonCheckVerifier.verify("src/test/resources/checks/pyTorchSpecifyReductionDim.py", new TfPyTorchSpecifyReductionAxisCheck());
```
joke1196, line 63: Here you could create a static variable with the `new TfPyTorchSpecifyReductionAxis()` that you can pass to the verifier.

joke1196 (approved):
Looks good! I have added a small comment but I will approve and you can merge once you made the change.

PR #1969, opened 2024-09-16
SONARPY-2127 Implement ClassDescriptorToPythonTypeConverter

maksim-grebeniuk-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/v2/converter/ClassDescriptorToPythonTypeConverter.java
```diff
@@ -0,0 +1,58 @@
...
+public class ClassDescriptorToPythonTypeConverter implements DescriptorToPythonTypeConverter {
+
+  private static PythonType convert(ConversionContext ctx, ClassDescriptor from) {
+    var typeBuilder = new ClassTypeBuilder()
+      .withName(from.name())
+      .withDefinitionLocation(from.definitionLocation());
+
+    from.superClasses().stream()
+      .map(fqn -> ctx.lazyTypesContext().getOrCreateLazyType(fqn))
+      .map(LazyTypeWrapper::new)
```
ghislainpiot, line 38:
```suggestion
      .map(fqn -> ctx.lazyTypesContext().getOrCreateLazyTypeWrapper(fqn))
```

I would use the wrapper instead

ghislainpiot (approved):
I have a small suggestion on using the helper function. Otherwise, LGTM!

PR #1974, opened 2024-09-17
SONARPY-1898: Subclasses of "torch.nn.Module" should call the initializer

Seppli11:


Seppli11:
> ## [Quality Gate failed](https://next.sonarqube.com/sonarqube/dashboard?id=org.sonarsource.python%3Apython&pullRequest=1974) **Quality Gate failed**
> Failed conditions [](https://next.sonarqube.com/sonarqube/component_measures?id=org.sonarsource.python%3Apython&pullRequest=1974&metric=new_coverage&view=list) [94.0% Coverage on New Code](https://next.sonarqube.com/sonarqube/component_measures?id=org.sonarsource.python%3Apython&pullRequest=1974&metric=new_coverage&view=list) (required ≥ 95%)
> 
> [See analysis details on SonarQube](https://next.sonarqube.com/sonarqube/dashboard?id=org.sonarsource.python%3Apython&pullRequest=1974)

I'm not sure if it is possible to get code coverage up to 100%

python-checks/src/main/java/org/sonar/python/checks/TorchModuleShouldCallInitCheck.java
```diff
@@ -0,0 +1,106 @@
...
+import org.sonar.plugins.python.api.tree.RegularArgument;
+import org.sonar.plugins.python.api.tree.Tree;
+import org.sonar.python.checks.utils.CheckUtils;
+import org.sonar.python.quickfix.TextEditUtils;
+import org.sonar.python.tree.TreeUtils;
+
+@Rule(key = "S6978")
+public class TorchModuleShouldCallInitCheck extends PythonSubscriptionCheck {
+  private static final String TORCH_NN_MODULE = "torch.nn.Module";
+  private static final String MESSAGE = "Add a call to super().__init__()";
```
joke1196, line 45: The period at the end of the message is missing.

python-checks/src/main/java/org/sonar/python/checks/TorchModuleShouldCallInitCheck.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  private static boolean isSuperConstructorCall(CallExpression callExpr) {
+    return callExpr.callee() instanceof QualifiedExpression qualifiedCallee && isSuperCall(qualifiedCallee.qualifier()) && "__init__".equals(qualifiedCallee.name().name());
+  }
+
+  private static boolean isSuperCall(Expression qualifier) {
+    if (qualifier instanceof CallExpression callExpression) {
+      Symbol superSymbol = callExpression.calleeSymbol();
+      return superSymbol != null && "super".equals(superSymbol.name());
```
joke1196, line 91: Here if you import a function called super or pass it as a parameter of the `__init__` method wouldn't you get a null symbol for the super call expression?

python-checks/src/test/resources/checks/torchModuleShouldCallInit.py
```diff
@@ -0,0 +1,48 @@
+import torch.nn as nn
+
+class NonCompliantModule(nn.Module):
+     #^^^^^^^^^^^^^^^^^^> {{Inheritance happens here}}
+    def __init__(self): #Noncompliant {{Add a call to super().__init__()}}
+       #^^^^^^^^
+        ...
+
+class NonCompliantModule(nn.Module):
+    def __init__(self): #Noncompliant
```
joke1196, line 10: Are this and the next example FPs? Are we sure we want to raise in such a case?

joke1196 (changes requested):
Looks good! I have added a few comments and questions.

python-checks/src/test/resources/checks/torchModuleShouldCallInit.py
Seppli11: Hmm, in a way yes. Can we easily check if some where in the call graph super().__init__() is called?

joke1196: I think to minimize the FPs we could check if a call is made to `super().__init__()` anywhere in the class body. This would result in FNs that I think are acceptable.
```
class NonCompliantModule(OtherModule, nn.Module):
    def __init__(self): #FN
        not_call_super()

    def call_super(self):
        super().__init__()

class NonCompliantModule2(OtherModule, nn.Module):
    def __init__(self): #FN
        do_something()

    class Nested(Other):
        def __init__(self):
            super().__init__()
```

joke1196 (approved):
Great job!

PR #1975, opened 2024-09-18
SONARPY-2146 Add owner to FunctionType in type inference V2

ghislainpiot:


python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
```diff
@@ -287,11 +287,12 @@ private FunctionType buildFunctionType(FunctionDef functionDef) {
       functionTypeBuilder.withTypeOrigin(TypeOrigin.LOCAL);
     }
     FunctionType functionType = functionTypeBuilder.build();
-    if (owner != null) {
+
+    if (owner instanceof ClassType classTypeOwner) {
```
maksim-grebeniuk-sonarsource, line 291: as discussed: this condition excludes adding members to the owner if it is `ModuleType` so probably this needs to be changed

maksim-grebeniuk-sonarsource (changes requested):
Looks good, need to investigate members for module types (see the comment)

PR #1976, opened 2024-09-18
Rule S6982: module mode should be set after load_state_dict

Seppli11:


python-checks/src/main/java/org/sonar/python/checks/TorchModuleModeShouldBeSetAfterLoadingCheck.java
```diff
@@ -0,0 +1,114 @@
...
+      .filter(usage -> usage.tree().firstToken().line() > callExpr.firstToken().line())
+      .toList();
+  }
+
+  private static Optional<Name> getFunctionCallReceiverName(CallExpression callExpr) {
+    Expression calleeExpr = callExpr.callee();
+    if (calleeExpr instanceof QualifiedExpression qualifiedExpr) {
+      return Optional.of(qualifiedExpr.qualifier()).flatMap(TreeUtils.toOptionalInstanceOfMapper(Name.class));
+    }
+    return Optional.empty();
```
joke1196, line 110: Here maybe it would be preferable to directly start the check with an optional. I find it makes it easier to read.
```
Optional.of(callExpr.callee())
  .flatMap(toOptionalInstanceOfMapper(QualifiedExpression)
...
```

joke1196 (approved):
This looks good! Thank you! Just added a minor comment on the coding style.

PR #1982, opened 2024-09-23
SONARPY-2148 add typeshed types for torch.nn.Module and pyvision modules

Seppli11:


python-checks/src/test/resources/checks/torchModuleModeShouldBeSetAfterLoadingCheck.py
```diff
@@ -7,26 +7,23 @@ def noncompliant():
...
 def noncompliant():
-    get_model().load_state_dict(torch.load('model_weights.pth')) # Noncompliant
-
-def noncompliant(model):
-    weights = torch.load('model_weights.pth')
-    weights2 = weights
-    model.load_state_dict(weights2) # Noncompliant
+    model = CustomModule(...)
+    model.load_state_dict(torch.load('model_weights.pth')) # false negative
+    ...
```
joke1196, line 16: I did not notice first, but we are missing a few tests.
```
load = torch.load(...)
model.load_state_dict(load)
```
as well as a check on the recursion. I think we could limit the recursion to 3.

Seppli11: I removed the recursion entirely. Checking if the parameter is torch.load doesn't make sense anymore since the type of the model can be properly resolved

joke1196 (approved):
LGTM. Before merging we should clean up the commit history, this way we can rebase the commits properly. The changes to S6982 should be part of another ticket.

PR #1983, opened 2024-09-23
NO-JIRA: Updated protobuf library version

joke1196:
This update should fix the ws alert without any changes to the outcome of typeshed serialization


Seppli11 (approved):
lgtm

PR #1984, opened 2024-09-23
SONARPY-2150: Fix FP on S6982 when method is used on optimizers

Seppli11:


joke1196 (approved):
LGTM! It is way cleaner when using the typeshed stubs. Thank you for your work!

PR #1986, opened 2024-09-24
SONARPY-2151: Fix fn when import torch is used instead of the full import name torch.nn

Seppli11:


joke1196 (approved):
LGTM. Thanks for making the change!

PR #1999, opened 2024-09-25
SONARPY-2159 Implement VarSymbolToDescriptorConverter

maksim-grebeniuk-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/v2/typeshed/VarSymbolToDescriptorConverter.java
```diff
@@ -0,0 +1,37 @@
...
+import org.sonar.python.types.protobuf.SymbolsProtos;
+
+public class VarSymbolToDescriptorConverter {
+
+  public Descriptor convert(SymbolsProtos.VarSymbol varSymbol) {
+    var typeAnnotation = Optional.of(varSymbol)
+      .map(SymbolsProtos.VarSymbol::getTypeAnnotation)
+      .map(SymbolsProtos.Type::getFullyQualifiedName)
+      .orElse(null);
+    return new VariableDescriptor(varSymbol.getName(), varSymbol.getFullyQualifiedName(), typeAnnotation);
```
guillaume-dequenne-sonarsource, line 34: When looking at what is done in `SymbolImpl`, there is a step of normalization of FQNs that removes the `builtins` prefix.

I'm personally in favor of keeping this prefix, however I have the feeling the change of behavior here is not intentional (and if we decide to add back the `builtins` prefix we need to do it everywhere anyway).

On top of that, there is some logic about FQNs when they are from a class, which I'm not sure whether we need to take care here or not (it depends on what is returned by the protobuf for some sample variables - which would benefit from being tested).

python-frontend/src/test/java/org/sonar/python/semantic/v2/typeshed/VarSymbolToDescriptorConverterTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.sonar.python.index.VariableDescriptor;
+import org.sonar.python.types.protobuf.SymbolsProtos;
+
+class VarSymbolToDescriptorConverterTest {
+
+  @Test
+  void test() {
```
guillaume-dequenne-sonarsource, line 30: As mentioned in the other comment, I guess we'd probably like to have a test for something that is from the `builtins` module. And maybe a test of a "real symbol" that comes from a class, to see if that is the output we want (that latter test might be one that is a bit more functional in nature, and integrated with an actual conversion - maybe the converter needs some additional context...)

guillaume-dequenne-sonarsource (changes requested):
This looks good, but as mentioned in the comment there is some special logic done with FQNs that we should ensure is not going to be problematic.

python-frontend/src/test/java/org/sonar/python/semantic/v2/typeshed/VarSymbolToDescriptorConverterTest.java
maksim-grebeniuk-sonarsource: as discussed - the test with the `ProtoSymbol` data from real proto files will be added as a part of [SONARPY-2160 PR](https://github.com/SonarSource/sonar-python/pull/2005)

PR #2001, opened 2024-09-25
SONARPY-2155 Implement FunctionSymbolToDescriptorConverter

maksim-grebeniuk-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/v2/typeshed/FunctionSymbolToDescriptorConverter.java
```diff
@@ -0,0 +1,64 @@
...
+    parameterConverter = new ParameterSymbolToDescriptorConverter();
+  }
+
+  public FunctionDescriptor convert(SymbolsProtos.FunctionSymbol functionSymbol) {
+    return convert(functionSymbol, false);
+  }
+
+  public FunctionDescriptor convert(SymbolsProtos.FunctionSymbol functionSymbol, boolean isParentIsAClass) {
+    var fullyQualifiedName = TypeShedUtils.normalizedFqn(functionSymbol.getFullyQualifiedName());
+    var returnType = TypeShedUtils.getTypesNormalizedFqn(functionSymbol.getReturnAnnotation());
```
ghislainpiot, line 41: If the kind of the `getReturnAnnotation()` is not `INSTANCE`, then we cannot rely on the fully qualified name. In `InferredTypes#fromTypeshedProtobuf`, the logic is more complicated than this.

python-frontend/src/main/java/org/sonar/python/semantic/v2/typeshed/ParameterSymbolToDescriptorConverter.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+package org.sonar.python.semantic.v2.typeshed;
+
+import org.sonar.python.index.FunctionDescriptor;
+import org.sonar.python.types.protobuf.SymbolsProtos;
+
+public class ParameterSymbolToDescriptorConverter {
+
+  FunctionDescriptor.Parameter convert(SymbolsProtos.ParameterSymbol parameter) {
+    var annotatedType = TypeShedUtils.getTypesNormalizedFqn(parameter.getTypeAnnotation());
```
ghislainpiot, line 28: Same thing here

PR #2005, opened 2024-09-25
SONARPY-2160 Implement TypeShedDescriptorsProvider

maksim-grebeniuk-sonarsource:


maksim-grebeniuk-sonarsource:
As discussed - in this PR need to add a test of proper `builtins` fully qualified name processing for the `ProtoSymbol` data read from proto files

python-frontend/src/main/java/org/sonar/python/semantic/v2/typeshed/TypeShedDescriptorsProvider.java
```diff
@@ -0,0 +1,143 @@
...
+    this.projectBasePackages = projectBasePackages;
+  }
+
+  //================================================================================
+  // Public methods
+  //================================================================================
+
+  public Map<String, Descriptor> builtinSymbols() {
+    if (builtins == null) {
+      supportedPythonVersions();
```
ghislainpiot, line 76: I think the name of the function should be changed if the return value is never used.

PR #2006, opened 2024-09-25
SONARPY-633: FP on rule S5719 when abc.abstractstaticmethod is used

Seppli11:


guillaume-dequenne-sonarsource (approved):
LGTM.
Please do not hesitate to squash commits when the separation is not meaningful (in this case, the PR is simple enough that separating the tests and check update is not really needed).

PR #2010, opened 2024-09-26
SONARPY-2164: Update sonar-plugin-api to version 10.11.0.2468

joke1196:


pom.xml
```diff
@@ -95,7 +95,7 @@
       at SonarSource/sonarqube/gradle.properties
     -->
     <sonar.version>10.6.0.92116</sonar.version>
-    <sonar.api.version>10.7.0.2191</sonar.api.version>
+    <sonar.api.version>10.11.0.2468</sonar.api.version>
```
guillaume-dequenne-sonarsource, line 98: Small comment: It would be good to use the actual version number instead of "latest version" in the ticket name/commit, so that things are clearer.

Seppli11 (approved):
lgtm

PR #2012, opened 2024-09-26
SONARPY-1018: S5727 should not raise when comparing xml ElementTree.getroot() to None

joke1196:


joke1196:
I can see the difference in the typeshed stubs that are unrelated to the change (sqlalchemy,xlib, etc...), are basically a revert of the changes made during fae4aa6e19a3d193849a2eefd1aabd9470a37161. We should investigate why there are differences even though the stubs were created in the "same" docker environment.

joke1196:
No, I don't think there is such a rule.

PR #2017, opened 2024-09-27
SONARPY-2161 Replace v2 TypeShed with TypeShedDescriptorsProvider

ghislainpiot:


python-frontend/src/main/java/org/sonar/python/types/v2/TypeUtils.java
```diff
@@ -31,4 +31,11 @@ static PythonType resolved(PythonType pythonType) {
     }
     return pythonType;
   }
+
+  public static PythonType ensureClassTypeWrapped(PythonType pythonType) {
```
guillaume-dequenne-sonarsource, line 35: Note: I think this will further the ambiguity between the technical "wrapper" that is used to resolve `LazyType`, and the functional "wrapper" that is used to signify that we're dealing with an instance of a type rather than the raw type itself.
We probably want a method name that conveys the fact that we're just ensuring we have an `ObjectType`.

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -2355,6 +2355,15 @@ void basic_imported_symbols() {
     assertThat(((UnionType) mathModule.resolveMember("acos").get()).candidates()).allMatch(FunctionType.class::isInstance);
   }
 
+  @Test
+  void import_conflict_between_member_and_submodule() {
```
guillaume-dequenne-sonarsource, line 2359: I think it could be nice to reference the related ticket hre.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #2020, opened 2024-09-27
SONARPY-1727: S5905: Make sure the quick fix removes trailing commas

Seppli11:


python-checks/src/main/java/org/sonar/python/checks/tests/AssertOnTupleLiteralCheck.java
```diff
@@ -43,12 +43,18 @@ public void initialize(Context context) {
...
-            .addTextEdit(TextEditUtils.remove(tuple.rightParenthesis()))
-            .build());
+          var quickfixBuilder = PythonQuickFix.newQuickFix(QUICK_FIX_MESSAGE)
+            .addTextEdit(TextEditUtils.remove(tuple.leftParenthesis()));
+
+          if(tuple.elements().size() == 1 && !tuple.commas().isEmpty()) {
+            quickfixBuilder.addTextEdit(TextEditUtils.replaceRange(tuple.commas().get(0), tuple.rightParenthesis(), ""));
+          } else {
+            quickfixBuilder.addTextEdit(TextEditUtils.remove(tuple.rightParenthesis()));
+          }
```
joke1196, line 55: I am not sure this would work with `assert (a,b,c,)` as this only check for single element tuples. Is that on purpose?

python-checks/src/main/java/org/sonar/python/checks/tests/AssertOnTupleLiteralCheck.java
```diff
@@ -43,12 +43,18 @@ public void initialize(Context context) {
 
         var issue = ctx.addIssue(tuple, MESSAGE);
 
-        if (tuple.leftParenthesis() != null && tuple.rightParenthesis() != null) {
+        if (tuple.leftParenthesis() != null && tuple.rightParenthesis() != null && !tuple.elements().isEmpty()) {
```
joke1196, line 46: How do we raise on this test `assert () # Noncompliant` when the `tuple.elements` is not empty?

joke1196 (changes requested):
I have two questions. Not sure I get which should be handled in the scope of the ticket.

python-checks/src/main/java/org/sonar/python/checks/tests/AssertOnTupleLiteralCheck.java
Seppli11: Oh, ups. No it's not intentional. I wasn't aware that trailing commas are allowed in general :sweat:

python-checks/src/main/java/org/sonar/python/checks/tests/AssertOnTupleLiteralCheck.java
Seppli11: I was thinking that it wouldn't make sense to offer a quick fix for `assert ()` as this would result in just `assert `. Or is that a sensible quick fix?

joke1196: Oh yes, this makes sense. I thought the check was at the level of the issue detection. And did not get why this would raise! Thanks for the clarification!

python-checks/src/main/java/org/sonar/python/checks/tests/AssertOnTupleLiteralCheck.java
Seppli11: Now that I'm thinking about this, I'm not sure if the quick fix actually makes sense. For an assert like `assert (a,b)` there are two ways to correct this with different semantics. it could be `assert a, b` (where `b` is the error message), or it could be `assert a and b`. The current quick fix assumes that it is the first one. 
However, given something like `assert (a,b,c)` the only sensible way to correct that is to generate `assert a and b and c`. But the current quick fix would generate `assert a, b, c` which isn't syntactically correct.

Seppli11: @joke1196, what do you think? Would it make more sense to always join the different elements with `and`?

joke1196: So as discussed we should provide the quick fix only for the single element tuple, as in other cases we risk breaking the code.

python-checks/src/main/java/org/sonar/python/checks/tests/AssertOnTupleLiteralCheck.java
```diff
@@ -43,17 +44,23 @@ public void initialize(Context context) {
...
-            .addTextEdit(TextEditUtils.remove(tuple.rightParenthesis()))
-            .build());
+        if (isSingletonTupleWithParenthesis(tuple)) {
+          var quickfixBuilder = PythonQuickFix.newQuickFix(QUICK_FIX_MESSAGE)
+            .addTextEdit(TextEditUtils.remove(tuple.leftParenthesis()));
+
+          Token lastComma =  tuple.commas().get(tuple.commas().size() - 1);
+          quickfixBuilder.addTextEdit(TextEditUtils.replaceRange(lastComma, tuple.rightParenthesis(), ""));
+
+          issue.addQuickFix(quickfixBuilder.build());
```
joke1196, line 54: Nitpick: maybe we can group the quickfixes statement together
```suggestion
          Token lastComma =  tuple.commas().get(tuple.commas().size() - 1);
          var quickfixBuilder = PythonQuickFix.newQuickFix(QUICK_FIX_MESSAGE)
            .addTextEdit(TextEditUtils.remove(tuple.leftParenthesis()));
            .addTextEdit(TextEditUtils.replaceRange(lastComma, tuple.rightParenthesis(), ""));

          issue.addQuickFix(quickfixBuilder.build());
```

joke1196 (approved):
LGTM. I have just added a small nitpick

PR #2021, opened 2024-09-30
SONARPY-1002: [Second Attempt] Fix FP on S2275 for nested replacement fields with format specifiers

Seppli11:


joke1196 (approved):
I find this a bit confusing. One thing that we could improve was the naming of the methods, as the difference between `tryParseField` and `tryParsingField`
is not super obvious. But also I am not sure what a better name we could give maybe one could be `validateField`. I will approve as I do not think we should spend more time on this issue, as this code resolves the ticket, 
and is still better organized than it was before.

PR #2022, opened 2024-09-30
SONARPY-1572: Fix parse error for challenging unicode characters

joke1196:


Seppli11 (approved):
It might be more readable if the line 106 would be split up into multiple lines.
But apart from that it lgtm

PR #2023, opened 2024-09-30
SONARPY-1756 exclude `function` from the built-in shadowing check

Seppli11:


joke1196 (approved):
LGTM! There were a lot of changes on the expectation of our ITs, from what I can see there were all FPs, that's great.

PR #2024, opened 2024-10-01
SONARPY-2178 Add awareness of the Python version & disambiguation to the DescriptorToPythonTypeConverters

ghislainpiot:


python-frontend/src/test/java/org/sonar/python/semantic/v2/typeshed/ModuleSymbolToDescriptorConverterTest.java
```diff
@@ -34,7 +34,7 @@ class ModuleSymbolToDescriptorConverterTest {
 
   @Test
   void test() {
-    var converter = new ModuleSymbolToDescriptorConverter();
+    var converter = new ModuleSymbolToDescriptorConverter(ProjectPythonVersion.currentVersions());
```
guillaume-dequenne-sonarsource, line 37: Couldn't that be source of nondeterminism if `ProjectPythonVersion#setCurrentVersions` is called in another test? Isn't it better to call `PythonVersionUtils#allVersions` explicitly?

python-frontend/src/main/java/org/sonar/python/semantic/v2/typeshed/ProtoUtils.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  private static Descriptor disambiguateSymbolsWithSameName(Set<Descriptor> descriptors) {
+    if (descriptors.size() > 1) {
+      return AmbiguousDescriptor.create(descriptors);
+    }
+    return descriptors.iterator().next();
+  }
+
+  static boolean isValidForPythonVersion(List<String> validForPythonVersions, Set<String> supportedPythonVersions) {
```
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: You mean because we're creating this intersection set for all tested symbol? Do you think this is significant enough that we should rethink the approach already?

python-frontend/src/main/java/org/sonar/python/semantic/v2/typeshed/ProtoUtils.java
```diff
@@ -0,0 +1,65 @@
...
+    if (validForPythonVersions.isEmpty()) {
+      return true;
+    }
+    // TODO: SONARPY-1522 - remove this workaround when we will have all the stubs for Python 3.12.
```
guillaume-dequenne-sonarsource, line 56: I'm surprised to see we haven't dealt with this yet despite 3.13 being around the corner :(

guillaume-dequenne-sonarsource (commented):
Nice job! I like the approach.
I just have one question regarding risk of nondeterminism, and would like to know more about your concern with regard to the number of sets we create before proceeding.

python-frontend/src/main/java/org/sonar/python/semantic/v2/typeshed/ProtoUtils.java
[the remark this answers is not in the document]
ghislainpiot: This is one of the problem. As discussed, we should create a new ticket to hoist the TypeShedDescriptorsProvider up to the ProjectLevelSymbolTable

python-frontend/src/test/java/org/sonar/python/semantic/v2/typeshed/TypeShedDescriptorsProviderTest.java
```diff
@@ -41,14 +39,18 @@ class TypeShedDescriptorsProviderTest {
   @RegisterExtension
   public LogTesterJUnit5 logTester = new LogTesterJUnit5().setLevel(Level.DEBUG);
 
-  @BeforeEach
-  void setup() {
-    ProjectPythonVersion.setCurrentVersions(PythonVersionUtils.allVersions());
+  private static TypeShedDescriptorsProvider cleanProvider() {
```
guillaume-dequenne-sonarsource, line 42: _Minor:_ I think this could just be called `typeshedProvider` for clarity.
It took me a minute trying to understand what `var provider = cleanProvider();` meant without looking at this method (note how the `var` adds to the confusion here 😅 )

PR #2025, opened 2024-10-01
 SONARPY-1974: fix false positive for unused variable (assignment expression in generator expression)

Seppli11:


python-checks/src/test/java/org/sonar/python/checks/UnusedLocalVariableCheckTest.java
```diff
@@ -42,178 +46,250 @@ void custom() {
...
+              return 
+          """,
+        "Remove assignment target"
+      ),
+      new QuickFixTestData("assignmentExpression2",
+        """
+          def foo():
+            if any((i := j*2) % 2 == 1 for j in range(3)):
+              return
+          """,
```
joke1196, line 184: We could maybe check that the quick fix does not break anything even with set/dict comprehension.

joke1196 (approved):
LGTM. I have just added a small comment.

joke1196:
Before merging could verify that Peter is properly added to the contributors list. (I think it is already the case in the individual commits, but should verify that is it still the case after the squash).

PR #2028, opened 2024-10-02
SONARPY-2170 Support detailed return types in Descriptor model

guillaume-dequenne-sonarsource:
This change is a bit rough, in the sense that the `FunctionDescriptor` model can now hold 2 representations of return type. One is `annotatedReturnTypeName`, as a `String`, which historically has been used for locally defined functions. The other is `TypeAnnotationDescriptor` which is a representation that is closer to our protobuf model used for Typeshed symbols. 

Note: only return types are taken care of in the PR. SONARPY-2179 has been created to handle parameters as well.

### DIFF
--- a/python-frontend/src/main/java/org/sonar/python/types/v2/ResolvableType.java
+++ b/python-frontend/src/main/java/org/sonar/python/types/v2/ResolvableType.java
@@ -19,23 +19,6 @@
  */
 package org.sonar.python.types.v2;
 
-public class TypeUtils {
-
-  private TypeUtils() {
-
-  }
-
-  static PythonType resolved(PythonType pythonType) {
-    if (pythonType instanceof ResolvableType resolvableType) {
-      return resolvableType.resolve();
-    }
-    return pythonType;
-  }
-
-  public static PythonType ensureWrappedObjectType(PythonType pythonType) {
-    if (!(pythonType instanceof ObjectType)) {
-      return new ObjectType(pythonType);
-    }
-    return pythonType;
-  }
+public interface ResolvableType {
+  PythonType resolve();
 }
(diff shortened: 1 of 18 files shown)


python-frontend/src/main/java/org/sonar/python/types/v2/ResolvableType.java
```diff
@@ -0,0 +1,24 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.python.types.v2;
+
+public interface ResolvableType {
```
guillaume-dequenne-sonarsource, line 22: Not sure about naming, open to change.
`LazyType` itself would have been nice, but I didn't want to rename the class itself...

ghislainpiot:
I don't know if adding a check to `UnionType` to prevent creating one with LazyType would be useful to avoid problems in the future

python-frontend/src/main/java/org/sonar/python/types/v2/ResolvableType.java
ghislainpiot: I can't think of a better name than that or LazyType

ghislainpiot (approved):
It looks good to me in terms of functionality. I think one issue is that we are creating more and more invisible contracts ( don't put LazyType in UnionType here) that are not enforced at construction or through the Java type system. However, breaking the contract crashes the analysis at runtime

PR #2034, opened 2024-10-03
SONARPY-1834 check if pipeline is used by sklearn.compose estimator

Seppli11:


python-checks/src/main/java/org/sonar/python/checks/SklearnPipelineSpecifyMemoryArgumentCheck.java
```diff
@@ -72,26 +76,33 @@ private static void createIssue(SubscriptionContext subscriptionContext, CallExp
...
         }
         callExpression = (CallExpression) TreeUtils.firstAncestorOfKind(callExpression, Tree.Kind.CALL_EXPR);
       }
       return false;
     });
   }
+
+  private static boolean isPipelineCreation(CallExpression callExpression) {
+    return Optional.ofNullable(callExpression.calleeSymbol())
+      .map(Symbol::fullyQualifiedName)
```
joke1196, line 96: This code looks quite good but we are checking twice if the symbol is null and if the qualified name is null as well. We could reactor this, and just directly check the FQN against `sklearn.compose`  and `Pipeline`,  etc...

joke1196 (approved):
Looks good, I have added a suggestion to refactor a part of the check.

PR #2037, opened 2024-10-04
SONARPY-1793 Fix long stack trace on lexing error

Seppli11:


sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonScanner.java
```diff
@@ -122,21 +123,26 @@ protected void scanFile(PythonInputFile inputFile) throws IOException {
...
-      context.newAnalysisError()
+
+      NewAnalysisError analysisError = context.newAnalysisError()
         .onFile(inputFile.wrappedFile())
-        .at(inputFile.wrappedFile().newPointer(line, 0))
-        .message(newMessage)
-        .save();
+        .message(newMessage);
+
+      if (line > 0) {
```
Seppli11, line 137: If a lexer error is caught, the exception is wrapped in a `RecognitionException` with the `line` set to `0`. Subsequently, this will cause an `IllegalStateException` in the `at` method below.

That being said, I'm not sure if this is 100% kosher...

Seppli11: Also, this line is currently not really tested. I'm not sure if it makes sense to test it.

joke1196: The issue I see is with the Notebooks, I am not sure that in the JSON file the first line we refer to is 1. We should be careful about that.

joke1196 (commented):
I have added a few comments hopefully this helps a bit. In the PythonSensor we already have tests on parse error, we could make them a bit more complete with more assertions on the line etc... And for the notebooks we have similar tests in the IPynbSensorTest file

PR #2041, opened 2024-10-04
SONARPY-1711: Rule S5659: do not raise under certain use of get_unverified_header()

joke1196:


python-checks/src/main/java/org/sonar/python/checks/JwtVerificationCheck.java
```diff
@@ -175,4 +186,83 @@ private static boolean isCallToVerifyJwt(Tree t) {
...
+  private static Stream<StringLiteral> argumentsAccessedByGet(Symbol symbol, CallExpression call) {
+    var usages = getHeaderDictUsages(symbol, call);
+    var parentOfUsages = usages.map(Usage::tree).map(Tree::parent);
+    var callExpressionsFromUsages = getCallExprWhereDictIsAccessedWithGet(parentOfUsages);
+    var callExpressionFromGetUnverifiedHeaders = getCallExprWhereDictIsAccessedWithGet(Stream.of(call.parent()));
+    var arguments = Stream.concat(callExpressionsFromUsages, callExpressionFromGetUnverifiedHeaders).map(CallExpression::arguments).flatMap(Collection::stream);
+    return getStringLiteralArguments(arguments);
+  }
+
+  private static Stream<Usage> getHeaderDictUsages(Symbol symbol, CallExpression call) {
```
Seppli11, line 222: I think a name, which indicates that only usages after the definition are returned, might be better. E.g. `getForwardUsages`

python-checks/src/main/java/org/sonar/python/checks/JwtVerificationCheck.java
```diff
@@ -175,4 +186,83 @@ private static boolean isCallToVerifyJwt(Tree t) {
...
+    return symbol.usages().stream()
+      .filter(usage -> usage.tree().firstToken().line() > call.callee().firstToken().line());
+  }
+
+  private static Stream<CallExpression> getCallExprWhereDictIsAccessedWithGet(Stream<Tree> parentQualifiedExpr) {
+    return parentQualifiedExpr
+      .filter(parent -> parent.is(Tree.Kind.QUALIFIED_EXPR))
+      .map(TreeUtils.toInstanceOfMapper(QualifiedExpression.class))
```
Seppli11, line 230: Can be replaced with the following:
```suggestion
      .flatMap(TreeUtils.toStreamInstanceOfMapper(QualifiedExpression.class))
```

python-checks/src/main/java/org/sonar/python/checks/JwtVerificationCheck.java
```diff
@@ -175,4 +186,83 @@ private static boolean isCallToVerifyJwt(Tree t) {
...
+      .filter(Objects::nonNull)
+      .filter(expr -> "get".equals(expr.name().name()))
+      .filter(expr -> expr.parent().is(Kind.CALL_EXPR))
+      .map(QualifiedExpression::parent)
+      .map(TreeUtils.toInstanceOfMapper(CallExpression.class));
```
Seppli11, line 235: Could be written as
```suggestion
      .map(QualifiedExpression::parent)
      .flatMap(TreeUtils.toStreamInstanceOfMapper(CallExpression.class));
```

python-checks/src/main/java/org/sonar/python/checks/JwtVerificationCheck.java
```diff
@@ -175,4 +186,83 @@ private static boolean isCallToVerifyJwt(Tree t) {
...
+  }
+
+  private static Stream<StringLiteral> getStringLiteralArguments(Stream<Argument> arguments) {
+    return arguments.filter(arg -> arg.is(Tree.Kind.REGULAR_ARGUMENT))
+      .map(TreeUtils.toInstanceOfMapper(RegularArgument.class))
```
Seppli11, line 240:
```suggestion
    return arguments
      .flatMap(TreeUtils.toStreamInstanceOfMapper(RegularArgument.class))
```

python-checks/src/main/java/org/sonar/python/checks/JwtVerificationCheck.java
```diff
@@ -175,4 +186,83 @@ private static boolean isCallToVerifyJwt(Tree t) {
...
+      .map(RegularArgument::expression)
+      .map(TreeUtils.toInstanceOfMapper(StringLiteral.class))
+      .filter(Objects::nonNull);
```
Seppli11, line 243:
```suggestion
      .flatMap(TreeUtils.toStreamInstanceOfMapper(StringLiteral.class));
```

Seppli11 (changes requested):
Is it on purpose that something like `foo(jwt.get_unverified_header(token).get("x5u"))` or `return jwt.get_unverified_header(token).get("x5u")` is not compliant? 
If yes, I think, a test case should be added to document this.  

Apart from this I've made a comment when `TreeUtils.toStreamInstanceOfMapper` can be used instead of first filtering and then mapping.

Otherwise, it looks good

EDIT: fixed the second example

joke1196:
So the first is not specified in the ticket but I would think it should be compliant, I will fix this. The second is documented in the ticket and in test cases as the token passed to the function does not affect how we would raise issues.

Seppli11 (approved):
lgtm!

PR #2045, opened 2024-10-07
SONARPY-1913: Update Microsoft stubs

Seppli11:


joke1196 (approved):
Looks good to me. We are still going back and forth with the Xlib stubs. We should really try and fix that, PRs will be easier to review.

PR #2049, opened 2024-10-08
Refactor Python version comparison to PythonVersionsUtils

joke1196:


python-frontend/src/main/java/org/sonar/plugins/python/api/PythonVersionUtils.java
```diff
@@ -168,6 +169,12 @@ private static boolean guessPythonVersion(Set<Version> pythonVersions, String ve
     return true;
   }
 
+  public static boolean isPythonVersionGreaterOrEqualThan(Set<Version> sourcePythonVersions, Version required) {
```
Seppli11, line 172: Maybe it would be worth reordering the parameters. 
Right now, one could read it as `(Set<Version> sourcePythonVersions) isGreatherOrEqualThan (Version required)`. 
If writing as 
```suggestion
  public static boolean isPythonVersionGreaterOrEqualThan(Version required, Set<Version> sourcePythonVersions) {
```
doesn't violate some convention, this might slightly improve readability

Seppli11 (approved):
Apart from the one, admittedly very nit-picky comment, this looks good to me

python-frontend/src/main/java/org/sonar/plugins/python/api/PythonVersionUtils.java
joke1196: I see, but I guess reordering would mean `Version required` isGreaterOrEqualThan `Set<Version>` which is not true as it should be all the elements of the sets should be greater or equal than the required version. I would change then the name if the method to be `areSourcePythonVersionsGreaterOrEqualThan`

PR #2050, opened 2024-10-08
SONARPY-2068: Tabulations (\t) should be handled correctly as an escape character

Seppli11:


sonar-python-plugin/src/test/java/org/sonar/plugins/python/TestUtils.java
```diff
@@ -59,4 +62,11 @@ public static PythonInputFile createInputFile(File baseDir, String name, InputFi
       .build());
   }
 
+  public static List<ColumnMapping> mapToColumnMappingList(Map<Integer, Integer> map) {
```
Seppli11, line 65: Do you have a `TestUtils` that is accessible in all tests?

sonar-python-plugin/src/test/java/org/sonar/plugins/python/PythonHighlighterTest.java
```diff
@@ -211,14 +212,20 @@ void number() {
 
   @Test
   void highlightingNotebooks() {
-    String pythonContent = "def foo():\n    pass\na = \"test\" # comment\n# test \\n \\\\n test\nb = 3J\n#SONAR_PYTHON_NOTEBOOK_CELL_DELIMITER";
+    String pythonContent = """
```
joke1196, line 215: Maybe in the highlighter we could add a `\t` to one of the test. To be sure the impact of the changes you made are propagated here.

joke1196 (changes requested):
Looks really good. Thanks for the clean up! I have added a small suggestion for the naming as well as small change on test. As stated in my comment we should add code for something that we do not support (unicode escape sequences), but I think this is fine like so as it is a way better version than what was there already :) (which already had code for something we were not using)

joke1196 (approved):
LGTM! Thanks a lot for this. This clean up was really needed!

PR #2052, opened 2024-10-09
SONARPY-2095: Update pluginApiMinVersion in pom.xml

Seppli11:


pom.xml
```diff
@@ -78,7 +78,7 @@
     <!-- Release: enable publication to Bintray -->
     <artifactsToPublish>${project.groupId}:sonar-python-plugin:jar</artifactsToPublish>
     <!-- we depend on API ${sonar.version} but we keep backward compatibility with LTS -->
-    <pluginApiMinVersion>9.9</pluginApiMinVersion>
+    <pluginApiMinVersion>10.11</pluginApiMinVersion>
```
Seppli11, line 81: The comment above makes me a bit unsure if we're allowed to change this to 10.11, while still being compatible with the LTA.
The current LTA, 9.9, uses the plugin API version `9.13.0.360` (see [here](https://github.com/SonarSource/sonarqube/blob/branch-9.8/build.gradle#L207)).

Seppli11: Currently, on peach, version 10.7 has been installed, which uses the plugin API version `10.11.0.2468` (see [here](https://github.com/SonarSource/sonarqube/blob/branch-10.7/gradle.properties#L9)). Thus sonarqube 10.7 should be able to load the python-analyzer

guillaume-dequenne-sonarsource: I think we generally update this after LTA releases, from what I remember. It's probably best to verify this, though.

Seppli11: Looking at the history, it looks like the `pluginApiMinVersion`/`sonarQubeMinVersion` were upgraded to the latest LTA version (or lower) but not above that.

joke1196: I think we can close this PR as discussed, and if the need to update arises we would then act on it.

PR #2055, opened 2024-10-09
SONARPY-2081: Array index in hex form should be handled correctly

Seppli11:


joke1196 (approved):
LGTM. Is there a file in our Ruling (that we are currently excluding) that contains such a case? It would be good to add it back if it is the case. We could also check on peach projects if we could include files that fail the parsing.

PR #2057, opened 2024-10-10
SONARPY-2153: S6978: Fix False Negative related to resolving `torch.nn.Module`

Seppli11:


joke1196 (approved):
Thank you for taking care of this. It would be easier for me to review if you have dedicated commits for the code you wrote, and a commit for the generated files. Because here in both commits there are changes on `torchvision` protobufs, it is not easy to see what exactly changed. No need to change anything here, the PR looks good like this!

PR #2062, opened 2024-10-11
SONARPY-2197: Update protobuf.version to v4 (major)

joke1196:


joke1196:
The updates on the binaries are only this change that will happen when generating the stubs on my machine compared to other squad member machines. These changes are not the result of the updated version (except for the checksum).

Seppli11 (approved):
lgtm!

Updating the typeshed binary feels so risky... I hope this doesn't introduce FN left and right

PR #2063, opened 2024-10-11
SONARPY-2198: Update mockito monorepo to v5 (major)

joke1196:


Seppli11 (approved):
LGTM!

I'm not quite sure if mockito-inline is still required. From a quick google search, it looks like, mockito-inline was moved into mockito-core. However, I only found articles/SO answer mentioning this being the case, no proper announcement.

PR #2064, opened 2024-10-11
SONARPY-862 S117 (LocalVariableAndParameterNameConventionCheck) shouldn't raise on type aliases

thomas-serre-sonarsource:


python-checks/src/main/java/org/sonar/python/checks/LocalVariableAndParameterNameConventionCheck.java
```diff
@@ -78,7 +83,42 @@ private void checkName(Symbol symbol, SubscriptionContext ctx) {
   }
 
   private static boolean isType(Symbol symbol) {
-    return symbol.usages().stream().map(Usage::tree).filter(Expression.class::isInstance).map(Expression.class::cast).anyMatch(e -> e.type().mustBeOrExtend("type"));
+    return isExtendingType(symbol) || isAssignedFromSpecialForm(symbol);
+  }
+
+  private static boolean isExtendingType(Symbol symbol) {
+    return symbol.usages().stream().map(Usage::tree).filter(Expression.class::isInstance).map(Expression.class::cast).anyMatch(e -> e.type().mustBeOrExtend("type")) ||
+      Objects.equals(symbol.annotatedTypeName(), "typing._SpecialForm");
```
guillaume-dequenne-sonarsource, line 91: _Nitpick:_ to avoid the `null` case, you can also write `"typing._SpecialForm".equals(symbol.annotatedTypeName())` which is a bit more concise.

Another more conservative approach could've been to simply exclude any symbol whose annotated type name starts with `typing`, as the likelihood of a type alias is high in such case, and it would avoid to single out `_SpecialForm` explicitly.

python-checks/src/main/java/org/sonar/python/checks/LocalVariableAndParameterNameConventionCheck.java
```diff
@@ -78,7 +83,42 @@ private void checkName(Symbol symbol, SubscriptionContext ctx) {
...
+          if (assignedSymbol != null && isExtendingType(assignedSymbol)) {
+            return true;
+          }
+        }
+      }
+    }
+    return false;
+  }
+
+  private static Expression getAssignedValue(Tree assignmentTree) {
```
guillaume-dequenne-sonarsource, line 114: _Nitpick:_ Here, (and in line 95), what you have is a name that is part of an assignment rather than the assignment itself. To reflect this, I'd suggest renaming to `assignedName` (and `assignedNames` respectively).

guillaume-dequenne-sonarsource (approved):
Left some minor comments, but LGTM otherwise!

python-checks/src/main/java/org/sonar/python/checks/LocalVariableAndParameterNameConventionCheck.java
thomas-serre-sonarsource: I follow your idea to exclude any symbol whose annotated type starts with `typing` :smile:

PR #2068, opened 2024-10-13
SONARPY-2193: Update sonarlint-core.version to v10 (major)

joke1196:


its/plugin/it-python-plugin-test/src/test/java/com/sonar/python/it/plugin/SonarLintTest.java
```diff
@@ -59,50 +67,64 @@ class SonarLintTest {
...
+        new ActiveRule("python:BackticksUsage", SonarLanguage.PYTHON.name()))
+      .build();
+
+    Map<Level, List<String>> logsByLevel = new HashMap<>();
+    LogOutput logOutput = new LogOutput() {
+      @Override
+      public void log(String formattedMessage, Level level, @Nullable String stacktrace) {
+        List<String> logs = logsByLevel.getOrDefault(level, new ArrayList<>());
+        logs.add(formattedMessage);
+        logsByLevel.putIfAbsent(level, logs);
```
Seppli11, line 126: I find the version above a bit more concise
```suggestion
        logsByLevel.computeIfAbsent(level, k -> new ArrayList<>()).add(formattedMessage);
```

its/plugin/it-python-plugin-test/src/test/java/com/sonar/python/it/plugin/IPythonTest.java
Seppli11, line 1: Would it maybe make sense to change this class's name to `SonarLintIPythonTest` or similar, to reflect that it does the something as `SonarLint` but for ipython?

Seppli11 (approved):
I left two comments about some small stuff.
Apart from that, LGTM!

its/plugin/it-python-plugin-test/src/test/java/com/sonar/python/it/plugin/IPythonTest.java
joke1196: Yes I think it makes a lot of sense!

PR #2072, opened 2024-10-14
NO-JIRA Fix quality issues on Next

guillaume-dequenne-sonarsource:


Seppli11 (approved):
Looks good to me!

One thing, 
`SQLQueriesCheck#getExpression(@Nullable Expression expr)` still passes the `expr` argument directly to `Expression#removeParettheses`. I don't think it is ever called with a nullable value, but the `@Nullable` annotation could be misleading.

guillaume-dequenne-sonarsource:
> SQLQueriesCheck

That's a very good point! I believe the `@Nullable` argument explicitly came from the fact that the function was calling itself recursively, using the `Nullable` return from `removeParentheses`. If that return value is no longer nullable, neither is the parameter to `getExpression`.
I updated this accordingly, thanks for the review!

PR #1871, opened 2024-07-24
SONARPY-2013 Make it possible to OR type predicates

guillaume-dequenne-sonarsource:
(This is a draft, requesting review to gather your opinion and possibly discuss it - However I would be fine not merging it in its current state, as it may simply not be needed yet)

PR #2074, opened 2024-10-15
SONARPY-2203 Migrate S2201 IgnoredPureOperationsCheck to the new type model

guillaume-dequenne-sonarsource:
Note: this PR includes two tickets (SONARPY-2203 and its prerequisite, SONARPY-2201) that can actually be reviewed independently. The limitation in re-exported stub symbols (e.g `flask.Response`) will be tackled in SONARPY-2209.

### DIFF
--- a/python-frontend/src/test/java/org/sonar/python/types/v2/TypesTestUtils.java
+++ b/python-frontend/src/test/java/org/sonar/python/types/v2/TypesTestUtils.java
@@ -50,10 +50,13 @@ public class TypesTestUtils {
   }
 
   public static FileInput parseAndInferTypes(PythonFile pythonFile, String... code) {
+    return parseAndInferTypes(PROJECT_LEVEL_TYPE_TABLE, pythonFile, code);
+  }
+
+  public static FileInput parseAndInferTypes(ProjectLevelTypeTable typeTable, PythonFile pythonFile, String... code) {
     FileInput fileInput = PythonTestUtils.parseWithoutSymbols(code);
-    var symbolTable = new SymbolTableBuilderV2(fileInput)
-      .build();
-    new TypeInferenceV2(PROJECT_LEVEL_TYPE_TABLE, pythonFile, symbolTable).inferTypes(fileInput);
+    var symbolTable = new SymbolTableBuilderV2(fileInput).build();
+    new TypeInferenceV2(typeTable, pythonFile, symbolTable).inferTypes(fileInput);
     return fileInput;
   }
 }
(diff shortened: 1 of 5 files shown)


python-frontend/src/test/java/org/sonar/python/types/v2/TypesTestUtils.java
```diff
@@ -54,6 +54,13 @@ public static FileInput parseAndInferTypes(PythonFile pythonFile, String... code
     var symbolTable = new SymbolTableBuilderV2(fileInput)
       .build();
     new TypeInferenceV2(PROJECT_LEVEL_TYPE_TABLE, pythonFile, symbolTable).inferTypes(fileInput);
+    return parseAndInferTypes(PROJECT_LEVEL_TYPE_TABLE, pythonFile, code);
```
Seppli11, line 57: I don't quite understand why the code needs to be parsed and run through type-inference twice

Seppli11 (approved):
I only have a small comment in the TypeTestUtils, where I'm unsure what the code does. If the code is correct, maybe the naming could be improved to help with understanding.
Apart from that, it looks good to me.

One thing, that is outside of the scope of this PR, is that I find it a bit confusing, that the `TypeChecker` creates a `TypeCheckerBuilder`, since the name would suggest that `TypeCheckerBuilder` builds `TypeChecker`. 

Otherwise, it is interesting to see the new API in action.

python-frontend/src/test/java/org/sonar/python/types/v2/TypesTestUtils.java
guillaume-dequenne-sonarsource: That's a very good point, thanks! It's actually a refactoring mistake on my end. I fixed it (and squashed it in the 1st commit).

guillaume-dequenne-sonarsource:
> I only have a small comment in the TypeTestUtils, where I'm unsure what the code does. If the code is correct, maybe the naming could be improved to help with understanding. Apart from that, it looks good to me.
> 
> One thing, that is outside of the scope of this PR, is that I find it a bit confusing, that the `TypeChecker` creates a `TypeCheckerBuilder`, since the name would suggest that `TypeCheckerBuilder` builds `TypeChecker`.
> 
> Otherwise, it is interesting to see the new API in action.

Thanks! About the builder, note that this is actually a `TypeCheckBuilder`, so the `TypeChecker` allows you to manipulate a builder to create the type checks. I was actually looking at ways to make this builder more flexible by making it possible to combine AND and OR predicates (see [here](https://github.com/SonarSource/sonar-python/pull/1871/commits/2591de1e288004647440327df7dd733ea1442e98)), but I didn't pursue this, because of other possible refactorings we discussed with Maksim. I think it would make sense to revisit this discussion when we start discussing exposing this in the public API, and inviting security and DBD to use it.

PR #2075, opened 2024-10-15
SONARPY-2187 Extract the SymbolV2 / PythonType state map at the end of a module execution 

thomas-serre-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/v2/types/PropagationVisitor.java
```diff
@@ -75,6 +76,16 @@ public void visitFunctionDef(FunctionDef functionDef) {
     propagationsByLhs.computeIfAbsent(symbol, s -> new HashSet<>()).add(definition);
   }
 
+  @Override
+  public void visitClassDef(ClassDef classDef) {
+    super.visitClassDef(classDef);
+    Name name = classDef.name();
+    var symbol = name.symbolV2();
+    Definition definition = new Definition(symbol, name);
```
maksim-grebeniuk-sonarsource, line 84: super minor, I think from the code style perspective it will look better if all 3 variables there will be defined with `var` of with a type predicate, to keep similarity

python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeInferenceV2.java
```diff
@@ -147,7 +154,7 @@ private void flowSensitiveTypeInference(ControlFlowGraph cfg, Set<SymbolV2> trac
       parameterTypes);
 
     flowSensitiveTypeInference.compute(cfg);
-    flowSensitiveTypeInference.compute(cfg);
+    return ((TypeInferenceProgramState) flowSensitiveTypeInference.compute(cfg)).typesBySymbol();
```
maksim-grebeniuk-sonarsource, line 157: I'd get rid of this class cast here by overloading the `compute` method for the `FlowSensitiveTypeInference`
```java
public class FlowSensitiveTypeInference extends ForwardAnalysis {
  ...
  @Override
  public TypeInferenceProgramState compute(ControlFlowGraph cfg) {
    return (TypeInferenceProgramState) super.compute(cfg);
  }
  ...
```

maksim-grebeniuk-sonarsource (approved):
Looks good, just a couple of minor comments

python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeInferenceV2.java
```diff
@@ -92,7 +96,7 @@ private void inferTypesAndMemberAccessSymbols(FunctionDef functionDef) {
       .map(Parameter::name)
       .collect(Collectors.toSet());
     Set<SymbolV2> localVariables = symbolTable.getSymbolsByRootTree(functionDef);
-    inferTypesAndMemberAccessSymbols(
+    typesBySymbol = inferTypesAndMemberAccessSymbols(
```
guillaume-dequenne-sonarsource, line 99: This is not the right place to set `typeBySymbol` for what we want to do.
This method exists for `FileInput` and for `FunctionDef`. To simulate an import, we need to consider the module-level type inference only (the `FileInput` method), as this one will override `typesBySymbol` for each function.

guillaume-dequenne-sonarsource (changes requested):
I believe `typeBySymbols` is set in the wrong place.
Because it's never used and not tested, it's not visible, but I suspect it would produce the wrong output for a module containing a few functions.
I suggest adding a way to return the `typeBySymbols` map and to add a test that ensures it produces the right output.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #2076, opened 2024-10-15
NO-JIRA: Updated doc for protoc command

joke1196:


python-frontend/typeshed_serializer/README.md
```diff
@@ -1,5 +1,6 @@
 To run:
 
-in `python-frontend/src/main/protobuf` directory, run: ```protoc -I=. --python_out=../../../typeshed_serializer/serializer/proto_out ./symbols.proto```
+in `python-frontend/src/main/protobuf` directory, run:
```
guillaume-dequenne-sonarsource, line 3: What does this do? Do we need it?
On my side I had some issues with Protobuf but eventually manage to sort them, and I didn't need to run a different command.

joke1196: I had an issue with the proto3_optional on my side, but it could be related to my version of the protobuf compiler. If you can generate the Python files without errors and without the flag then we should not bother changing the command.

guillaume-dequenne-sonarsource: Yeah I had issues as well and after reinstalling it, it went fine. I suggest we close the PR if it's fine for you as well (unless the added command has intrinsic benefits you identified).

PR #2077, opened 2024-10-15
SONARPY-2188 set UnresolvedImportType in TrivialTypeInferenceVisitor

Seppli11:


python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
```diff
@@ -297,53 +300,75 @@ private FunctionType buildFunctionType(FunctionDef functionDef) {
...
+
+  private static void generateNames(PythonType resolvedType, List<Name> names, List<String> fqn) {
+    if (resolvedType instanceof ModuleType module) {
+      for (int i = names.size() - 1; i >= 0; i--) {
+        setTypeToName(names.get(i), module);
+        module = Optional.ofNullable(module)
+          .map(ModuleType::parent)
+          .orElse(null);
+      }
+    } else if (resolvedType instanceof UnknownType) {
```
ghislainpiot, line 328: No coverage, but same behavior as before

python-frontend/src/main/java/org/sonar/python/types/v2/UnresolvedImportType.java
```diff
@@ -0,0 +1,23 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.python.types.v2;
+
+public record UnresolvedImportType(String importPath) implements PythonType {
```
guillaume-dequenne-sonarsource, line 22: One thing that I think could be problematic is the fact that this is now different from `PythonType.UNKNOWN`.

We have pieces of logic that check for `PythonType.UNKNOWN` in order to ensure a more conservative behavior (typically, in the `TypeChecker`, having `PythonType.UNKNOWN` somewhere would make a predicate test return `TriBool.UNKNOWN` instead of `TriBool.FALSE`).

Since we now have `UnresolvedImportType` that is different from `PythonType.UNKNOWN`, I have the feeling we should adapt these bits of logic to stay intentional (perhaps no longer relying on equality checks against `PythonType.UNKNOWN`?).

guillaume-dequenne-sonarsource (changes requested):
This looks good!
A minor comment about test util methods.

And another about the possible risk of impacting logic relying on equality comparisons with`PythonType.UNKNOWN`. This should not pose a problem for indexing., however if we're implicitly going to use this for rule execution, it may have some unintended impact. 

Note that I'm not necessarily asking to change that in this PR, however we should at least acknowledge the possible impact through tests (and decide accordingly what we want to do afterwards).

python-frontend/src/main/java/org/sonar/python/types/v2/UnresolvedImportType.java
Seppli11: Could a possible solution be to introduce a `Unkown` interface with both a `UnkownImpl` and the `UnresolvedImportType` as implementations? 
I could imagine that having multiple types, which all act as `Unknown`, could lead to subtle bugs since everybody has to remember to check for every possible Unknown type or remember to always use the appropriate helper function.

Just a thought though. Maybe this is would introduce too much complexity for what we gain from this

thomas-serre-sonarsource (approved):
LGTM ! 
I left a very minor comment that you can ignore if you wish

python-frontend/src/main/java/org/sonar/python/types/v2/UnionType.java
```diff
@@ -91,7 +91,7 @@ public static PythonType or(@Nullable PythonType type1, @Nullable PythonType typ
     if (type2 == null) {
       return type1;
     }
-    if (type1.equals(PythonType.UNKNOWN) || type2.equals(PythonType.UNKNOWN)) {
+    if (type1 == PythonType.UNKNOWN || type2 == PythonType.UNKNOWN) {
```
ghislainpiot, line 94: We want to keep unions of UnresolvedImportType for indexing

python-frontend/src/main/java/org/sonar/python/types/v2/TypeCheckBuilder.java
```diff
@@ -139,7 +139,7 @@ public TriBool test(PythonType pythonType) {
       if (pythonType instanceof ObjectType objectType) {
         pythonType = objectType.unwrappedType();
       }
-      if (pythonType == PythonType.UNKNOWN || expectedType == PythonType.UNKNOWN) {
+      if (pythonType instanceof UnknownType || expectedType instanceof UnknownType) {
```
guillaume-dequenne-sonarsource, line 142: I would expect to see a test that breaks without this change.

As I mentioned in my first comment, my concern with the initial implementation was that we might break some logic when we assume `expectedType == PythonType.UNKNOWN` is enough to test that a type is unknown (since it no longer is the case).

In a TDD approach I think it would have made more sense to create such a test to prove the concern right before changing the implementation.

guillaume-dequenne-sonarsource (changes requested):
The changes look good!
I still think we need a test that shows they were necessary though.

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
```diff
@@ -279,7 +281,7 @@ private FunctionType buildFunctionType(FunctionDef functionDef) {
     TypeAnnotation typeAnnotation = functionDef.returnTypeAnnotation();
     if (typeAnnotation != null) {
       PythonType returnType = typeAnnotation.expression().typeV2();
-      functionTypeBuilder.withReturnType(returnType == PythonType.UNKNOWN ? returnType : new ObjectType(returnType, TypeSource.TYPE_HINT));
+      functionTypeBuilder.withReturnType(returnType instanceof UnknownType ? returnType : new ObjectType(returnType, TypeSource.TYPE_HINT));
       functionTypeBuilder.withTypeOrigin(TypeOrigin.LOCAL);
```
ghislainpiot, line 285: Cannot be hit because of SONARPY-2234

python-frontend/src/main/java/org/sonar/python/types/v2/TypeCheckBuilder.java
```diff
@@ -233,5 +233,8 @@ private static Set<PythonType> collectTypes(PythonType type) {
     }
   }
 
+  private static boolean containsUnknown(Set<PythonType> types) {
+    return types.stream().anyMatch(UnknownType.class::isInstance);
+  }
```
ghislainpiot, line 241: This is covered now

python-frontend/src/main/java/org/sonar/python/types/v2/TypeCheckBuilder.java
ghislainpiot: I have added the test.
I tried to add tests, but I don't think some scenarios are reachable with the current type inference.

python-frontend/src/test/java/org/sonar/python/types/v2/UnresolvedImportTypeTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+
+  @Test
+  void imported_unknown() {
+    FileInput fileInput = inferTypesWithNoResolution("""
+      import unknown
+      unknown
+      """);
+    var unknownType = ((UnresolvedImportType) ((ExpressionStatement) fileInput.statements().statements().get(1)).expressions().get(0).typeV2());
+    assertThat(unknownType.importPath()).isEqualTo("unknown");
```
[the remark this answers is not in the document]
ghislainpiot: I think this is a pretty common pattern in the V2 tests, but I think a deeper refactoring is overdue.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #2078, opened 2024-10-16
SONARPY-2209 Avoid duplicating re-exported stub symbols when deserializing stubs

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/v2/converter/AnyDescriptorToPythonTypeConverter.java
```diff
@@ -43,6 +47,26 @@ Descriptor.Kind.AMBIGUOUS, new AmbiguousDescriptorToPythonTypeConverter()
...
   }
 
+  public Optional<ModuleType> convertModuleType(String moduleName, String moduleFqn,
+    ModuleType parent, Map<String, Descriptor> stringDescriptorMap, boolean registerAsSubmodule) {
+    Map<String, TypeWrapper> moduleMembers = new HashMap<>();
+    for (var entry : stringDescriptorMap.entrySet()) {
+      var descriptor = entry.getValue();
+      var name = entry.getKey();
+      PythonType result;
+      String fullyQualifiedName = descriptor.fullyQualifiedName();
```
ghislainpiot, line 57: If I understand correctly, this is the fullname from Mypy

python-frontend/src/main/java/org/sonar/python/semantic/v2/ProjectLevelTypeTable.java
```diff
@@ -66,16 +68,26 @@ public PythonType getType(List<String> typeFqnParts) {
...
       var part = typeFqnParts.get(i);
+      var moduleFqnParts = IntStream.rangeClosed(0, i)
+        .mapToObj(typeFqnParts::get)
+        .toList();
       if (parent instanceof ObjectType) {
         return PythonType.UNKNOWN;
       }
+      if (parent instanceof ModuleType moduleType) {
+        TypeWrapper typeWrapper = moduleType.members().get(part);
+        if (typeWrapper instanceof LazyTypeWrapper lazyTypeWrapper && !lazyTypeWrapper.isResolved()) {
```
ghislainpiot, line 79: Does this replace the idea of having a `submodules` field from [SONARPY-2176](https://sonarsource.atlassian.net/browse/SONARPY-2176?atlOrigin=eyJpIjoiMjgwMjE2NWVhZThiNDk3Mjg3MzkzY2Y5N2ExMGI0OGUiLCJwIjoiaiJ9)

ghislainpiot, line 79: Manually checking if a LazyType is not resolved makes it feel like a bandaid to me. I think properly representing submodules would be better, but perhaps outside the scope of the PR.

python-frontend/src/main/java/org/sonar/python/types/v2/ModuleType.java
```diff
@@ -60,7 +79,11 @@ public ModuleType(@Nullable String name, @Nullable ModuleType parent) {
 
   @Override
   public Optional<PythonType> resolveMember(String memberName) {
-    return Optional.ofNullable(members.get(memberName));
+    return Optional.ofNullable(members.get(memberName)).map(TypeWrapper::type);
+  }
+
+  public Optional<PythonType> resolveSubmodule(String submoduleName) {
+    return Optional.ofNullable(subModules.get(submoduleName)).map(TypeWrapper::type);
```
ghislainpiot, line 86: Is it expected that `subModules` is only used in test code? Not counting adding and retrieving from the parents, it never leaves the module types.

python-frontend/src/test/java/org/sonar/python/types/v2/ModuleTypeTest.java
```diff
@@ -39,12 +40,43 @@ void resolveMemberTest() {
   @Test
   void replaceMemberIfUnknown() {
     var a = new ModuleType("a");
-    a.members().put("b", PythonType.UNKNOWN);
+    a.members().put("b", TypeWrapper.of(PythonType.UNKNOWN));
```
ghislainpiot, line 43: nitpick: We have `TypeWrapper.UNKNOWN_TYPE_WRAPPER`

ghislainpiot (approved):
I think this is a good start towards improving resolution, there are still quite a few edge cases from what is in the tests

python-frontend/src/main/java/org/sonar/python/types/v2/ModuleType.java
guillaume-dequenne-sonarsource: This is a good point. I've added 13d2386c59f360c387ac80f302f0c6ad1a723995 to instead consistently register submodules in the `subModule` field.
I've updated the strategy of `resolveMember` to first resolve actual members, then sub modules if no member is present. The behavior is not entirely accurate depending on the context, but I believe it is a decent enough approximation before we tackle SONARPY-2176.

python-frontend/src/main/java/org/sonar/python/semantic/v2/ProjectLevelTypeTable.java
guillaume-dequenne-sonarsource: That makes sense. I've added 13d2386c59f360c387ac80f302f0c6ad1a723995 which adds a bit more intentionality, although the "LazyType is not resolved" heuristics is still used.

Ultimately, I think SONARPY-2176 should model a bit better the cases where a submodule or a member should be looked at in priority. I do think that the "LazyType is not resolved" is a case where resolving the submodule first is necessary (because the LazyType is apparently pointing to it - not resolving it would lead to a stack overflow), and we only tackle this one due to the crashes this avoids.

python-frontend/src/main/java/org/sonar/python/semantic/v2/converter/AnyDescriptorToPythonTypeConverter.java
guillaume-dequenne-sonarsource: Yes, which represents the actual fully qualified name of the symbol (where it's defined, rather than where it's re-exposed).
Do you think it could be better named?

ghislainpiot: It was just to be sure, I think it is strange that we didn't keep the name of the filed from Mypy

ghislainpiot (approved):
Looks good to me now.

python-frontend/src/main/java/org/sonar/python/types/v2/ModuleType.java
```diff
@@ -30,23 +30,23 @@
 public final class ModuleType implements PythonType {
   private final String name;
   private final ModuleType parent;
-  private final Map<String, PythonType> members;
+  private final Map<String, TypeWrapper> members;
+  private final Map<String, TypeWrapper> subModules = new HashMap<>();
```
maksim-grebeniuk-sonarsource, line 34: I'd suggest to assign the field value in the same place where other fields are initialized, to have single place to look at

maksim-grebeniuk-sonarsource (approved):
LGTM

PR #2085, opened 2024-10-18
SONARPY-2202: Migrate S5727 ComparisonToNoneCheck to the new type model

Seppli11:


python-frontend/src/main/java/org/sonar/python/types/v2/TypeCheckBuilder.java
```diff
@@ -227,5 +232,40 @@ private static Set<PythonType> collectTypes(PythonType type) {
     }
   }
 
+  record IsIdentityComparableWith(PythonType expectedType) implements TypePredicate {
+    public IsIdentityComparableWith {
+      expectedType = expectedType.unwrappedType();
+    }
 
+    @Override
+    public TriBool test(PythonType otherMaybeWrappedType) {
```
Seppli11, line 241: I'm not entirely sure that the behaviour of this method and `InferredType.isIdentityComparableWith` is the same.

My understanding of this check is that it checks if two types are the same, or, if one is a union, one of its candidates is the same as the other type. Additionally, if one of the two types is unknown, the result is always `TriBool.UNKNOWN`.

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeCheckerBuilderTest.java
Seppli11, line 1: was moved to types/v2

ghislainpiot (changes requested):
The utility methods seem to be handling a lot of edge cases that should never happen in the first place.
I think it would be good to create tickets to investigate (and eliminate) the creation of those impossible types.

I didn't review the TypeChecker changes as discussed, I don't think handling the identify case is worth the added complexity on the main predicate.

PR #2088, opened 2024-10-21
SONARPY-2229 Fix FP on S2201 when instantiating the result of a type call

guillaume-dequenne-sonarsource:
This is admittedly a very naive approach to fix this FP.

However, I believe that this deserves a broader discussion, considering SONARPY-2013 to allow for easier manipulation of type predicates.
I also considered adding a proper `PythonType#kind` API to avoid checking for instances. However, I feel this can still be done later if the need proves real.

### DIFF
--- a/python-checks/src/main/java/org/sonar/python/checks/IgnoredPureOperationsCheck.java
+++ b/python-checks/src/main/java/org/sonar/python/checks/IgnoredPureOperationsCheck.java
@@ -298,23 +298,21 @@ public class IgnoredPureOperationsCheck extends PythonSubscriptionCheck {
     if (expression.is(Tree.Kind.CALL_EXPR)) {
       CallExpression callExpression = (CallExpression) expression;
       PythonType pythonType = callExpression.callee().typeV2();
-      if (typeChecker.typeCheckBuilder().isObjectType().check(pythonType) == TriBool.FALSE) {
-        PURE_FUNCTIONS.stream()
-          .filter(f -> typeChecker.typeCheckBuilder().isTypeWithName(f).check(pythonType).equals(TriBool.TRUE))
-          .findFirst()
-          .ifPresent(result -> ctx.addIssue(callExpression.callee(), String.format(MESSAGE_FORMAT, result)));
-      }
+      PURE_FUNCTIONS.stream()
+        .filter(f -> typeChecker.typeCheckBuilder().isTypeWithName(f).check(pythonType).equals(TriBool.TRUE))
+        .findFirst()
+        .ifPresent(result -> ctx.addIssue(callExpression.callee(), String.format(MESSAGE_FORMAT, result)));
     } else if (expression.is(Tree.Kind.SUBSCRIPTION)) {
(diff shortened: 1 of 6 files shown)


python-checks/src/main/java/org/sonar/python/checks/IgnoredPureOperationsCheck.java
```diff
@@ -298,10 +298,12 @@ private static void checkExpression(SubscriptionContext ctx, Expression expressi
     if (expression.is(Tree.Kind.CALL_EXPR)) {
       CallExpression callExpression = (CallExpression) expression;
       PythonType pythonType = callExpression.callee().typeV2();
-      PURE_FUNCTIONS.stream()
-        .filter(f -> typeChecker.typeCheckBuilder().isTypeWithName(f).check(pythonType).equals(TriBool.TRUE))
-        .findFirst()
-        .ifPresent(result -> ctx.addIssue(callExpression.callee(), String.format(MESSAGE_FORMAT, result)));
+      if (typeChecker.typeCheckBuilder().isObjectType().check(pythonType) == TriBool.FALSE) {
```
Seppli11, line 301: I think extracting this condition into its own method would help readability as it can be named

Seppli11 (changes requested):
According to our discussion, I think it would make sense to change the type checks in `TypeCheckBuilder` to either check if a type is an instance of `a` or *is* `a`. 
Since this PR is using `isTypeWithName` introducing this change for this method would make sense. 

Maybe it also would make sense to align the behaviour of `isInstanceOf`, so that checking with `isTypeWithName` checks for the type and `isInstanceOf` if it is an instance of a type. Having (somewhat) uniform behaviour for the type checkers when it comes to union type would make the API easier to use. But this should probably be its separate PR.

python-checks/src/main/java/org/sonar/python/checks/IgnoredPureOperationsCheck.java
guillaume-dequenne-sonarsource: (I removed the condition altogether)

guillaume-dequenne-sonarsource:
> According to our discussion, I think it would make sense to change the type checks in `TypeCheckBuilder` to either check if a type is an instance of `a` or _is_ `a`. Since this PR is using `isTypeWithName` introducing this change for this method would make sense.
> 
> Maybe it also would make sense to align the behaviour of `isInstanceOf`, so that checking with `isTypeWithName` checks for the type and `isInstanceOf` if it is an instance of a type. Having (somewhat) uniform behaviour for the type checkers when it comes to union type would make the API easier to use. But this should probably be its separate PR.

Following our discussion, I instead added the `isStrict` option to the `isSameTypeAsPredicate`. I feel that we should indeed have a more global approach to dealing with "in an instance of" vs "is the exact type", but I feel it should probably be part of a more global refactoring (we'll have to clarify the public type checking API we want to expose eventually, I think it can be part of that).

Seppli11 (approved):
lgtm

PR #2090, opened 2024-10-21
SONARPY-2219 Pass the result of the type inference for a Python file to ProjectLevelTypesTable

thomas-serre-sonarsource:


python-frontend/src/test/java/org/sonar/python/semantic/v2/converter/PythonTypeToDescriptorConverterTest.java
```diff
@@ -0,0 +1,226 @@
...
+
+    UnionType unionType = new UnionType(Set.of(classType, anotherClassType));
+    Descriptor descriptor = converter.convert("foo", new SymbolV2("myUnionType"), Set.of(unionType, classType));
+
+    assertThat(descriptor).isInstanceOf(AmbiguousDescriptor.class);
+    AmbiguousDescriptor ambiguousDescriptor = (AmbiguousDescriptor) descriptor;
+    assertThat(ambiguousDescriptor.name()).isEqualTo("myUnionType");
+    assertThat(ambiguousDescriptor.fullyQualifiedName()).isEqualTo("foo.myUnionType");
+    assertThat(ambiguousDescriptor.kind()).isEqualTo(Descriptor.Kind.AMBIGUOUS);
+    // TODO SONARPY-2225 the two class types in the union are rigorously the same but the converter creates an ambigouous descriptor
```
guillaume-dequenne-sonarsource, line 220: I believe it is to be expected since we don't implement `equals` for `PythonType` and expect the unicity of PythonType objects for each available type.

I guess that means that we'd have an ambiguous descriptor in the case where there are 2 definitions that are strictly identical:
```
if x:
  class A: ...
else:
  class A: ...
```

However, I think it's okay to accept this behavior for now. If this prevents proper resolution in some cases, we can always implement something that would merge types that are essentially identical.

PR #2094, opened 2024-10-22
SONARPY-2242 TypeInferenceV2 should resolve stubs with variable descriptors pointing to submodules

ghislainpiot:
[SONARPY-2242](https://sonarsource.atlassian.net/browse/SONARPY-2242)

[SONARPY-2242]: https://sonarsource.atlassian.net/browse/SONARPY-2242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python-frontend/src/main/java/org/sonar/python/semantic/v2/ProjectLevelTypeTable.java
+++ b/python-frontend/src/main/java/org/sonar/python/semantic/v2/ProjectLevelTypeTable.java
@@ -67,10 +67,11 @@ public class ProjectLevelTypeTable {
       if (parent instanceof ModuleType moduleType) {
         TypeWrapper typeWrapper = moduleType.members().get(part);
         if (typeWrapper instanceof LazyTypeWrapper lazyTypeWrapper && !lazyTypeWrapper.isResolved()) {
-          if (i == typeFqnParts.size() - 1) {
+          if (i == typeFqnParts.size() - 1 && !(lazyTypeWrapper.hasImportPath(String.join(".", typeFqnParts)))) {
             // this is the name we are looking for, resolve it
             return typeWrapper.type();
           }
+
           // The member of the module is a LazyType, which means it's a re-exported type from a submodule
           // We try to resolve the submodule instead
           Optional<PythonType> subModule = moduleType.resolveSubmodule(part);
--- a/python-frontend/src/main/java/org/sonar/python/semantic/v2/ProjectLevelTypeTable.java
+++ b/python-frontend/src/main/java/org/sonar/python/semantic/v2/ProjectLevelTypeTable.java
@@ -67,8 +67,8 @@ public class ProjectLevelTypeTable {
       if (parent instanceof ModuleType moduleType) {
         TypeWrapper typeWrapper = moduleType.members().get(part);
         if (typeWrapper instanceof LazyTypeWrapper lazyTypeWrapper && !lazyTypeWrapper.isResolved()) {
-          if (i == typeFqnParts.size() - 1 && !(lazyTypeWrapper.hasImportPath(String.join(".", typeFqnParts)))) {
-            // this is the name we are looking for, resolve it
+          if (i == typeFqnParts.size() - 1 && !fqnSameAsImportedPath(typeFqnParts, lazyTypeWrapper)) {
+            // this is the name we are looking for and its imported path is not the one we are trying to find, resolve it
             return typeWrapper.type();
           }
 
@@ -89,6 +89,10 @@ public class ProjectLevelTypeTable {
       }
     }
     return parent;
+  }
+
+  private static boolean fqnSameAsImportedPath(List<String> typeFqnParts, LazyTypeWrapper lazyTypeWrapper) {
+    return lazyTypeWrapper.hasImportPath(String.join(".", typeFqnParts));
   }
 
   /**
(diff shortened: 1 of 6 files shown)


python-frontend/src/main/java/org/sonar/python/semantic/v2/ProjectLevelTypeTable.java
```diff
@@ -67,10 +67,11 @@ public PythonType getType(List<String> typeFqnParts) {
       if (parent instanceof ModuleType moduleType) {
         TypeWrapper typeWrapper = moduleType.members().get(part);
         if (typeWrapper instanceof LazyTypeWrapper lazyTypeWrapper && !lazyTypeWrapper.isResolved()) {
-          if (i == typeFqnParts.size() - 1) {
+          if (i == typeFqnParts.size() - 1 && !(lazyTypeWrapper.hasImportPath(String.join(".", typeFqnParts)))) {
```
guillaume-dequenne-sonarsource, line 70: Acknowledging the confusion when looked at this together, I'm thinking maybe we should extract this to another method, or at least add a tiny bit more info on the purpose of this in a comment (maybe updating the one on line 71).

python-frontend/src/test/java/org/sonar/python/semantic/v2/ProjectLevelTypeTableTest.java
```diff
@@ -253,4 +253,14 @@ class lib: ...
     // SONARPY-2176 lib should be resolved as the renamed class "A" here
     assertThat(aType.name()).isEqualTo("A");
   }
+
+  @Test
+  void resolveCustomStub() {
```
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: (Note: I took the liberty to rename the ticket/PR title consistently with this, however feel free to further update it)

python-frontend/src/main/java/org/sonar/python/semantic/v2/converter/VariableDescriptorToPythonTypeConverter.java
```diff
@@ -29,6 +29,12 @@
 public class VariableDescriptorToPythonTypeConverter implements DescriptorToPythonTypeConverter {
 
   public PythonType convert(ConversionContext ctx, VariableDescriptor from) {
+    if (from.isImportedModule()) {
+      var fqn = from.fullyQualifiedName();
+      if (fqn != null) {
+        return ctx.lazyTypesContext().getOrCreateLazyType(fqn);
```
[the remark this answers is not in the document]
ghislainpiot: I tried the implementation with `LazyModuleType`, and it opened up some problematic discussions. 
One example is if there is a `LazyType` and a `ModuleLazyType` with the same imported path, what happens ?
It also didn't help in the code, and the change was bigger for no real benefit (I didn't do profiling)

python-frontend/src/main/java/org/sonar/python/semantic/v2/converter/VariableDescriptorToPythonTypeConverter.java
```diff
@@ -29,6 +29,12 @@
 public class VariableDescriptorToPythonTypeConverter implements DescriptorToPythonTypeConverter {
 
   public PythonType convert(ConversionContext ctx, VariableDescriptor from) {
+    if (from.isImportedModule()) {
+      var fqn = from.fullyQualifiedName();
+      if (fqn != null) {
```
ghislainpiot, line 34: I can do a mock to cover it, but I am not sure of the value

ghislainpiot, line 34: I can do a mock to cover it, but I am not sure of the value.

python-frontend/src/main/java/org/sonar/python/semantic/v2/ProjectLevelTypeTable.java
```diff
@@ -67,8 +67,8 @@ public PythonType getType(List<String> typeFqnParts) {
       if (parent instanceof ModuleType moduleType) {
         TypeWrapper typeWrapper = moduleType.members().get(part);
         if (typeWrapper instanceof LazyTypeWrapper lazyTypeWrapper && !lazyTypeWrapper.isResolved()) {
-          if (i == typeFqnParts.size() - 1 && !(lazyTypeWrapper.hasImportPath(String.join(".", typeFqnParts)))) {
-            // this is the name we are looking for, resolve it
+          if (i == typeFqnParts.size() - 1 && !fqnSameAsImportedPath(typeFqnParts, lazyTypeWrapper)) {
```
guillaume-dequenne-sonarsource, line 70: I must admit, this is not really what I had in mind, `fqnSameAsImportedPath` and `lazyTypeWrapper.hasImportPath(String.join(".", typeFqnParts))` read very similar and don't provide the extra clarity I expected.

What I had in mind was extracting the full condition (both parts of the &&) and expanding a bit more on the cases that can lead us to this condition. Basically explaining that we try to resolve a lazy type only if it's pointing us to a different module, and if it's going to point us on the same module again, we instead try to resolve a submodule of the same name instead of the member.
(Sorry, I realize my suggestion is a bit unclear, and this is probably why we need one in the first place)

guillaume-dequenne-sonarsource (approved):
Looks good!
I had in mind something a bit more detailed for the explanation and the method extraction, but I would be okay to merge as-is for the sake of not blocking this and other tickets depending on this.

PR #2097, opened 2024-10-23
SONARPY-2204 Migrate S5795 IdentityComparisonWithCachedTypesCheck to the new type model

Seppli11:
[SONARPY-2204](https://sonarsource.atlassian.net/browse/SONARPY-2204)

Part of 

[SONARPY-2204]: https://sonarsource.atlassian.net/browse/SONARPY-2204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python-checks/src/main/java/org/sonar/python/checks/IdentityComparisonWithCachedTypesCheck.java
+++ b/python-checks/src/main/java/org/sonar/python/checks/IdentityComparisonWithCachedTypesCheck.java
@@ -19,24 +19,25 @@
  */
 package org.sonar.python.checks;
 
-import java.util.HashSet;
 import java.util.List;
-import java.util.Set;
 import org.sonar.check.Rule;
 import org.sonar.plugins.python.api.PythonSubscriptionCheck;
 import org.sonar.plugins.python.api.SubscriptionContext;
-import org.sonar.plugins.python.api.symbols.Symbol;
-import org.sonar.plugins.python.api.symbols.Usage;
+import org.sonar.plugins.python.api.quickfix.PythonQuickFix;
 import org.sonar.plugins.python.api.tree.CallExpression;
 import org.sonar.plugins.python.api.tree.Expression;
 import org.sonar.plugins.python.api.tree.IsExpression;
 import org.sonar.plugins.python.api.tree.Name;
 import org.sonar.plugins.python.api.tree.Tree;
-import org.sonar.plugins.python.api.types.InferredType;
-import org.sonar.plugins.python.api.quickfix.PythonQuickFix;
-import org.sonar.python.checks.utils.CheckUtils;
+import org.sonar.plugins.python.api.types.BuiltinTypes;
 import org.sonar.python.quickfix.TextEditUtils;
+import org.sonar.python.semantic.v2.SymbolV2;
+import org.sonar.python.semantic.v2.UsageV2;
 import org.sonar.python.tree.TreeUtils;
+import org.sonar.python.types.v2.PythonType;
+import org.sonar.python.types.v2.TriBool;
+import org.sonar.python.types.v2.TypeCheckBuilder;
+import org.sonar.python.types.v2.TypeChecker;
 
 import static java.util.Arrays.asList;
 
@@ -49,12 +50,40 @@ public class IdentityComparisonWithCachedTypesCheck extends PythonSubscriptionCh
   public static final String IS_QUICK_FIX_MESSAGE = "Replace with \"==\"";
   public static final String IS_NOT_QUICK_FIX_MESSAGE = "Replace with \"!=\"";
 
+  /**
+   * Fully qualified names of constructors and functions that are guaranteed to create fresh objects with
+   * references not shared anywhere else.
+   * <p>
+   * If a reference arises from a call to one of these functions, and it does not escape anywhere else, then an
+   * <code>is</code>-comparison with such a reference will always return <code>False</code>.
+   * <p>
+   * Note that these are fully qualified names of (value-level) expressions, not types.
+   */
+  private static final List<String> FQNS_CONSTRUCTORS_RETURNING_UNIQUE_REF =
+    asList("frozenset", "bytes", "int", "float", "str", "tuple", "hash");
+
+  /**
+   * Names of types that usually should not be compared with <code>is</code>.
+   * <p>
+   * Note that these are names of types, not `fqn`s of expressions.
+   */
+  private static final List<String> NAMES_OF_TYPES_UNSUITABLE_FOR_COMPARISON =
+    asList("frozenset", "bytes", "int", "float", "tuple");
+
+  private TypeChecker typeChecker;
+  private TypeCheckBuilder isNoneTypeChecker;
+
   @Override
   public void initialize(Context context) {
-    context.registerSyntaxNodeConsumer(Tree.Kind.IS, IdentityComparisonWithCachedTypesCheck::checkIsComparison);
+    context.registerSyntaxNodeConsumer(Tree.Kind.FILE_INPUT, ctx -> {
+      typeChecker = ctx.typeChecker();
+      isNoneTypeChecker = ctx.typeChecker().typeCheckBuilder().isBuiltinWithName(BuiltinTypes.NONE_TYPE);
+    });
+
+    context.registerSyntaxNodeConsumer(Tree.Kind.IS, this::checkIsComparison);
   }
 
-  private static void checkIsComparison(SubscriptionContext ctx) {
+  private void checkIsComparison(SubscriptionContext ctx) {
     IsExpression isExpr = (IsExpression) ctx.syntaxNode();
 
     // Comparison to none is checked by S5727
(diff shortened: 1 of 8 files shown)


python-checks/src/main/java/org/sonar/python/checks/IdentityComparisonWithCachedTypesCheck.java
```diff
@@ -49,12 +52,40 @@ public class IdentityComparisonWithCachedTypesCheck extends PythonSubscriptionCh
   public static final String IS_QUICK_FIX_MESSAGE = "Replace with \"==\"";
   public static final String IS_NOT_QUICK_FIX_MESSAGE = "Replace with \"!=\"";
 
+  /**
+   * Fully qualified names of constructors and functions that would are guaranteed to create fresh objects with
```
guillaume-dequenne-sonarsource, line 54: I know you didn't add this, but there's a typo here (`that would are guaranteed` -> `that are guaranteed`). Can you take the opportunity to fix it?

python-checks/src/main/java/org/sonar/python/checks/IdentityComparisonWithCachedTypesCheck.java
```diff
@@ -56,21 +54,21 @@ public class IdentityComparisonWithCachedTypesCheck extends PythonSubscriptionCh
...
    * <p>
-   * If a reference arises from an call to one of these functions, and it does not escape anywhere else, then an
+   * If a reference arises from a call to one of these functions, and it does not escape anywhere else, then an
    * <code>is</code>-comparison with such a reference will always return <code>False</code>.
    * <p>
    * Note that these are fully qualified names of (value-level) expressions, not types.
    */
-  private static final Set<String> FQNS_CONSTRUCTORS_RETURNING_UNIQUE_REF = new HashSet<>(
-    asList("frozenset", "bytes", "int", "float", "str", "tuple", "hash"));
+  private static final List<String> FQNS_CONSTRUCTORS_RETURNING_UNIQUE_REF =
```
guillaume-dequenne-sonarsource, line 62: Note: I'm confused to see this being further changed in this `propagate unary expressions` commit. I feel it would have been cleaner to squash this with the first commit.

python-checks/src/test/resources/checks/identityComparisonWithCachedTypes.py
```diff
@@ -18,6 +18,9 @@ def literal_comparison(param):
     (4, 5) is param # Noncompliant {{Replace this "is" operator with "=="; identity operator is not reliable here.}}
     #      ^^
 
+    param is -1 #Noncompliant
```
guillaume-dequenne-sonarsource, line 21: _Nitpick:_ We usually add a space `# Noncompliant`.
Also, just out of curiosity did you add these cases because of regressions on ruling?

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
```diff
@@ -152,6 +156,23 @@ public void visitNumericLiteral(NumericLiteral numericLiteral) {
     numericLiteralImpl.typeV2(new ObjectType(pythonType, new ArrayList<>(), new ArrayList<>()));
   }
 
+  @Override
+  public void visitUnaryExpression(UnaryExpression unaryExpr) {
```
guillaume-dequenne-sonarsource, line 162: I think by having this in the trivial type inference visitor rather than the flow-sensitive one, you are actually not covering cases such as:

```python
x = 42
y = -x
```

The trivial type inference visitor is really meant to evaluate the type of trivial expressions that do not depend on control flow, such as builtins or the names of class and function definitions. It is only used for propagation in some cases where "normal" propagation would not lead to the desired result (see SONARPY-2244 and SONARPY-2245 for an example).

Here, the expression `-42` has a trivial type only because the expression `42` has a trivial type (`INT`), but you don't have a guarantee that the expression for a unary operation will be trivial.

python-checks/src/main/java/org/sonar/python/checks/IdentityComparisonWithCachedTypesCheck.java
```diff
@@ -83,40 +114,24 @@ private static void checkIsComparison(SubscriptionContext ctx) {
...
-  private static boolean isUnsuitableOperand(Expression expr) {
-    InferredType tpe = expr.type();
-    if (isUnsuitableType(tpe)) {
-      if (expr.is(Tree.Kind.NAME)) {
-        Symbol symb = ((Name) expr).symbol();
+  private boolean isUnsuitableOperand(Expression expr) {
+    PythonType type = expr.typeV2();
+    if (isUnsuitableType(type)) {
+      if (expr instanceof Name name) {
+        SymbolV2 symb = name.symbolV2();
```
guillaume-dequenne-sonarsource, line 132: Does the comment and condition below still make sense? I don't think `Name#symbolV2` can return null.

guillaume-dequenne-sonarsource (changes requested):
The changes on the rule look good.
I'm not fully certain on the changes for the propagation of unary expression as I believe it should probably be done in `FlowSensitiveTypeInference`. I also think it would have deserved an explicitly separate ticket.
Small nitpick on the commit history that is not as clean as it could be.

python-checks/src/test/resources/checks/identityComparisonWithCachedTypes.py
Seppli11: Yes, it was from a ruling

Seppli11:
As discussed, the way this PR implements the propagation of unary expressions means that only the type of unary literal expressions are properly propagated (e.g. `-1`, but not `-x` or `-(1 + 1)`). However, as a baby step, this moves us in the right direction. Therefore, a disabled test was added, which can be enabled as part of SONARPY-2257, which should implement the proper propagation of unary expressions.

python-checks/src/test/resources/checks/identityComparisonWithCachedTypes.py
```diff
@@ -152,3 +152,7 @@ def comparison_to_none():
     "str" is None
     123 is None
     None is "str"
+
+def potential_null_symbols():
+    type(arr) is tuple # see SONARPY-2259
+    some_thing is other_thing # see SONARPY-2260
```
guillaume-dequenne-sonarsource, line 158: I'm not sure it makes sense to reference these other tickets here.
I think there is a distinction to be made between:
- A test case added to highlight an unwanted behavior of the rule (e.g FP/FN).
- A test case added to ensure proper test coverage, because of an unrelated limitation.

In the first case, it is normal to reference the ticket that would fix the unwanted behavior. In the second, fixing the ticket will not change the test result anyway, and the reference to the ticket is likely to stay in the code because of that (c.f Michael's comments on our weekly about stale ticket references in our code). Simply having the case for coverage is enough.

That being said, I think it would make sense to have actual frontend tests for our symbol tables (most likely in `SymbolTableBuilderV2Test`) that highlight these behaviors and link to a ticket.

guillaume-dequenne-sonarsource (changes requested):
The overall changes look good.
However, I think it would make sense to make intentions a bit clearer.
I'd like you to split this in 3 commits:
1. Infer types of unary expressions for literals
2. Implement the changes of the rule (keeping the null check for symbols & the test that covers it, but not the ticket references to SONARPY-2259 and SONARPY-2260)
3. Document missing symbols for builtins and unassigned variables. (add unit tests in `SymbolTableBuilderV2Test` highlighting SONARPY-2259 and SONARPY-2260).

(Normally, I'd split the 3rd commit if implementing the migration from scratch, but I think the cost of adding it at this point of the PR is rather low)

python-checks/src/test/resources/checks/identityComparisonWithCachedTypes.py
```diff
@@ -18,6 +18,9 @@ def literal_comparison(param):
     (4, 5) is param # Noncompliant {{Replace this "is" operator with "=="; identity operator is not reliable here.}}
     #      ^^
 
+    param is -1 # Noncompliant
+    param is 1 + 1 #Noncompliant
```
guillaume-dequenne-sonarsource, line 22: Nitpick: Missing space.

python-checks/src/main/java/org/sonar/python/checks/IdentityComparisonWithCachedTypesCheck.java
```diff
@@ -83,75 +112,71 @@ private static void checkIsComparison(SubscriptionContext ctx) {
...
-    if (isUnsuitableType(tpe)) {
-      if (expr.is(Tree.Kind.NAME)) {
-        Symbol symb = ((Name) expr).symbol();
-        // Impossible to cover. If the type could be inferred, then the `symb` cannot be null.
-        return symb != null && !isVariableThatCanEscape(symb);
+  private boolean isUnsuitableOperand(Expression expr) {
+    PythonType type = expr.typeV2();
+    if (isUnsuitableType(type)) {
+      if (expr instanceof Name name) {
+        SymbolV2 symbol = name.symbolV2();
```
guillaume-dequenne-sonarsource, line 132: `Name#symbolV2` should probably be annotated `@CheckForNull` since it can return `null`

guillaume-dequenne-sonarsource (approved):
LGTM. When SONARPY-2275 is merged, I believe this can be rebased, which will make ruling green then this can be merged.

PR #2099, opened 2024-10-23
SONARPY-2245 Fix resolving type for a name if its type is UnresolvedImportType

maksim-grebeniuk-sonarsource:
[SONARPY-2245](https://sonarsource.atlassian.net/browse/SONARPY-2245)

[SONARPY-2245]: https://sonarsource.atlassian.net/browse/SONARPY-2245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
+++ b/python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
@@ -471,7 +471,7 @@ public class TrivialTypeInferenceVisitor extends BaseTreeVisitor {
       .map(Expression.class::cast)
       .map(Expression::typeV2)
       // TODO: classes (SONARPY-1829) and functions should be propagated like other types
-      .filter(t -> (t instanceof ClassType) || (t instanceof FunctionType) || (t instanceof ModuleType))
+      .filter(t -> (t instanceof ClassType) || (t instanceof FunctionType) || (t instanceof ModuleType) || (t instanceof UnknownType.UnresolvedImportType))
       .ifPresent(type -> setTypeToName(name, type));
   }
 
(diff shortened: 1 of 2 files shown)


python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
```diff
@@ -471,7 +471,7 @@ public void visitName(Name name) {
       .map(Expression.class::cast)
       .map(Expression::typeV2)
       // TODO: classes (SONARPY-1829) and functions should be propagated like other types
-      .filter(t -> (t instanceof ClassType) || (t instanceof FunctionType) || (t instanceof ModuleType))
+      .filter(t -> (t instanceof ClassType) || (t instanceof FunctionType) || (t instanceof ModuleType) || (t instanceof UnknownType.UnresolvedImportType))
```
guillaume-dequenne-sonarsource, line 474: I'm wondering if we should extract this to a dedicated method that could be named something like `shouldTypeBeEagerlyPropagated`, to make this heuristic more intentional. 

I have the feeling that we'll never fully remove this heuristic (even as we fix SONARPY-1829), because of the reasons mentioned in SONARPY-2245 (basically, we sometimes want to circumvent the actual Python propagations to ensure classes, functions and imports referenced in inner scopes are correctly resolved, even if they could technically be shadowed)

guillaume-dequenne-sonarsource (commented):
Left some comments.
I would expect the `// FAILS` comment to be removed, and am tempted to say we should extract the custom logic in a dedicated method. I leave @ghislainpiot with the decision to approve as-is or not, though.

ghislainpiot (changes requested):
I think putting the filter condition in a method would make sense, it would help understanding the code in my opinion.

ghislainpiot (approved):
LGTM!

PR #2112, opened 2024-10-29
SONARPY-2275 Fix FP on S5756 for calls on class variables defined as classmethods

guillaume-dequenne-sonarsource:


Seppli11 (approved):
LGTM

PR #2114, opened 2024-10-30
SONARPY-2276 Fix FPs on S930 when an object method is assigned to a variable

guillaume-dequenne-sonarsource:
[SONARPY-2276](https://sonarsource.atlassian.net/browse/SONARPY-2276)

[SONARPY-2276]: https://sonarsource.atlassian.net/browse/SONARPY-2276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

SONARPY-2285 has been created to handle bound/unbound methods in a more deliberate way. For now, we just ignore them when handling with re-exports.

It is to be noted that S930 has not been migrated to the new type model yet (only the cross-file analysis was impacted through descriptors). I've still added tests to document the current behavior (no issues raised on bound/unbound methods assigned to variables) that hopefully will help avoid problems when performing the migration.

Ruling tests are still red, due to some other yet unfixed issues.

Note: I suggest merging this PR in 2 commits, so that commits on the converter implementation (still not on master) can be squashed together.

### DIFF
--- a/python-checks/src/test/resources/checks/argumentNumber.py
+++ b/python-checks/src/test/resources/checks/argumentNumber.py
@@ -214,3 +214,20 @@ def logging_api():
 def foo(day, tz):
     b = datetime.date.fromordinal(day).replace(tzinfo=tz) # FN SONARPY-1472
     a = datetime.datetime.fromordinal(day).replace(tzinfo=tz) # OK
+
+
+def bound_and_unbound_methods():
+    class ClassWithMethod:
+        def some_method(self, param): ...
+
+    unbound_method = ClassWithMethod.some_method
+    bound_method = ClassWithMethod().some_method
+
+    unbound_method(1, 2) # OK
+    bound_method(1) # OK
+
+    unbound_method() # FN SONARPY-2285
+    unbound_method(1) # FN SONARPY-2285
+    unbound_method(1, 2, 3) # FN SONARPY-2285
+    bound_method() # FN SONARPY-2285
+    bound_method(1, 2) # FN SONARPY-2285
--- a/python-checks/src/test/resources/checks/argumentNumberWithImport.py
+++ b/python-checks/src/test/resources/checks/argumentNumberWithImport.py
@@ -1,5 +1,7 @@
 from argumentNumberImported import fn, A
 import argumentNumberImported
+from argumentNumberImported import reassigned_unbound_meth as aliased_unbound
+from argumentNumberImported import reassigned_bound_meth as aliased_bound
 
 fn(1, 2) # Noncompliant
 fn(1) # OK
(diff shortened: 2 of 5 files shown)


python-frontend/src/main/java/org/sonar/python/semantic/v2/converter/PythonTypeToDescriptorConverter.java
```diff
@@ -42,7 +44,7 @@ public class PythonTypeToDescriptorConverter {
 
   public Descriptor convert(String moduleFqn, SymbolV2 symbol, Set<PythonType> types) {
     var candidates = types.stream()
-      .map(type -> convert(moduleFqn, moduleFqn, symbol.name(), type))
+      .map(type -> convert(moduleFqn, moduleFqn, symbol.name(), type, symbol))
```
thomas-serre-sonarsource, line 47: _personal preference_ I find it a bit weird to have `symbol` and `symbolName` as parameter of convert. A symbol will contain the symbolName so I would guess that we don't need to pass the `symbolName` parameter. 
I would be in favor of changing the signature of `convert` to 
``` java
private static Descriptor convert(String moduleFqn, String parentFqn, String symbolName, List<UsageV2> symbolUsages, PythonType type)
```
Then it will also simplify the logic since you won't have to check for null at line 66. WDYT?

thomas-serre-sonarsource (approved):
LGTM ! :muscle: 
I left very minor differences. Feel free to ignore them.

python-frontend/src/main/java/org/sonar/python/semantic/v2/converter/PythonTypeToDescriptorConverter.java
guillaume-dequenne-sonarsource: I actually started with this approach. However, class members don't have symbols and because of that, I was losing access to the necessary `symbolName` (for which `Member#name` is used instead on line 107).

I ended up duplicating the method for top-level types that have symbols and members, and it became a bit of a mess. I decided to just add the parameter instead.

python-checks/src/test/resources/checks/argumentNumberWithImport.py
```diff
@@ -12,3 +14,26 @@ def stub_functions():
...
+    argumentNumberImported.reassigned_unbound_meth()
+    argumentNumberImported.reassigned_unbound_meth(42, 43)
+    argumentNumberImported.reassigned_bound_meth()
+    argumentNumberImported.reassigned_bound_meth(42, 43)
+
+    aliased_unbound()
+    aliased_unbound(1, 2, 3)
+
+    aliased_bound()
+    aliased_bound(42, 43)
```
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: (I committed the change)

python-frontend/src/main/java/org/sonar/python/semantic/v2/converter/PythonTypeToDescriptorConverter.java
thomas-serre-sonarsource: I think I agree with you but I was not explicit enough in my proposition: 
The idea was not to pass only `symbol` as a parameter and not passing `symbolName`. The idea was to replace the `symbol` parameter by `symbolUsages` keeping the `symbolName` parameter. I think this way it shouldn't be a problem even for members.

python-checks/src/test/resources/checks/argumentNumber.py
```diff
@@ -214,3 +214,20 @@ def logging_api():
...
+    bound_method = ClassWithMethod().some_method
+
+    unbound_method(1, 2) # OK
+    bound_method(1) # OK
+
+    unbound_method() # FN
+    unbound_method(1) # FN
+    unbound_method(1, 2, 3) # FN
+    bound_method() # FN
+    bound_method(1, 2) # FN
```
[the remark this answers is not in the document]
thomas-serre-sonarsource: IMO It is still worth it to add the reference. We can still update the issue number when we tackle SONARPY-2285 to point to the next ticket. I assume that SONARPY-2285 and the other required ticket will be linked so it will be clear enough.
I'm not super strong on it so feel free to keep it like this ;)

python-frontend/src/main/java/org/sonar/python/semantic/v2/converter/PythonTypeToDescriptorConverter.java
guillaume-dequenne-sonarsource: Oh, sorry, I read the suggested code a bit too quickly.
Indeed, it makes a lot of sense, I made the change!

python-checks/src/test/resources/checks/argumentNumber.py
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: Fair enough, I made the change then!

PR #2118, opened 2024-10-30
Update Jira automation to avoid creation of unwanted tickets

guillaume-dequenne-sonarsource:


.github/workflows/PullRequestCreated.yml
```diff
@@ -13,6 +13,8 @@ jobs:
     # For external PR, ticket should be created manually
     if: |
         github.event.pull_request.head.repo.full_name == github.repository
+        && github.event.sender.type != 'Bot'
+        && ${{ !startsWith(github.event.pull_request.title, 'NO-JIRA') }}
```
ghislainpiot, line 17: Should we also check if it is the `dependencies` tag to avoid Renovate PRs ?

pavel-mikula-sonarsource, line 17: I disagree with NO-JIRA, as that thing should not exist anymore. The sole purpose of this action is to prevent NO-JIRA in the future, as it should simply not be needed anymore.
```suggestion
```

pavel-mikula-sonarsource (changes requested):
The NO-JIRA condition should not exist.

.github/workflows/PullRequestCreated.yml
guillaume-dequenne-sonarsource: I understand your point, and I still believe that some edge cases may either prove noisy in Jira or require knowledge of how the action exactly works to avoid noise.
I believe that in most cases, the action will simply trigger and fulfill its goal to accurately reflect work in Jira.

In the rare cases where a ticket is NOT desired, I want to keep things simple and be able to use this simple flag. I believe we'll still have the opportunity to evaluate if it is misused/overused, and react accordingly.

ghislainpiot (approved):
The code looks good to me, but I can't speak for the spirit of the JIRA synchronization

PR #2120, opened 2024-10-31
[DO NOT MERGE] SONARPY-2295 Support relative imports in type inference V2

guillaume-dequenne-sonarsource:
[SONARPY-2295](https://sonarsource.atlassian.net/browse/SONARPY-2295)

[SONARPY-2295]: https://sonarsource.atlassian.net/browse/SONARPY-2295?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
```diff
@@ -96,13 +98,15 @@ public class TrivialTypeInferenceVisitor extends BaseTreeVisitor {
 
   private final TypeTable projectLevelTypeTable;
   private final String fileId;
+  private final String fullyQualifiedModuleName;
```
maksim-grebeniuk-sonarsource, line 101: We could store it as a list of strings

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
```diff
@@ -381,10 +385,23 @@ private static void generateNames(PythonType resolvedType, List<Name> names, Lis
...
-      .map(ImportFrom::module)
+    List<String> fromModuleFqn;
+    List<Token> dotPrefixTokens = importFrom.dottedPrefixForModule();
+
+    List<String> dottedPart = Optional.ofNullable(importFrom.module())
       .map(TrivialTypeInferenceVisitor::dottedNameToPartFqn)
-      .ifPresent(fqn -> setTypeToImportFromStatement(importFrom, fqn));
+      .orElse(new ArrayList<>());
+
+    List<String> moduleFqnElements = new ArrayList<>(Arrays.stream(fullyQualifiedModuleName.split("\\.")).toList());
```
maksim-grebeniuk-sonarsource, line 395: then this operation makes no sense, but if you prefer to store module name as a single string then this expression could be simpler:
```java
var moduleFqnElements =  List.of(fullyQualifiedModuleName.split("\\."))
```

maksim-grebeniuk-sonarsource: also, does it make sense to define this variable outside of the `if` condition?

maksim-grebeniuk-sonarsource: also if you'll get rid of `dotPrefixTokens` variable and simply assign the `fromModuleFqn` - it will allow to get rid of the `else` part
```java
    var fromModuleFqn = Optional.ofNullable(importFrom.module())
      .map(TrivialTypeInferenceVisitor::dottedNameToPartFqn)
      .orElse(new ArrayList<>());
    var dotPrefixTokens = importFrom.dottedPrefixForModule();
    if (!dotPrefixTokens.isEmpty()) {
      var moduleFqnElements = List.of(fullyQualifiedModuleName.split("\\."));
      // Relative import: we start from the current module FQN and go up as many levels as there are dots in the import statement
      int sizeLimit = Math.max(0, moduleFqnElements.size() - dotPrefixTokens.size());
      fromModuleFqn = Stream.concat(moduleFqnElements.stream().limit(sizeLimit), fromModuleFqn.stream()).toList();
    }
    setTypeToImportFromStatement(importFrom, fromModuleFqn);
```

maksim-grebeniuk-sonarsource (changes requested):
Looks good, just a couple of small changes could be applied

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
guillaume-dequenne-sonarsource: I actually use the string in my other PR to introduce FQNs for `FunctionType`, so I'd rather keep the string here and split it when needed in the import resolution.

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypeInferenceVisitor.java
guillaume-dequenne-sonarsource: That makes sense, thanks!

maksim-grebeniuk-sonarsource (approved):
LGTM

guillaume-dequenne-sonarsource:
Note: I re-created https://github.com/SonarSource/sonar-python/pull/2122 to target master instead of the feature branch.

guillaume-dequenne-sonarsource:
Duplicate.

PR #2121, opened 2024-11-01
SONARPY-2189

guillaume-dequenne-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/ProjectLevelSymbolTable.java
```diff
@@ -206,20 +210,35 @@ private void addModuleV2(FileInput astRoot, String packageName, PythonFile pytho
...
-      if ("django.urls.conf.path".equals(calleeSymbol.fullyQualifiedName())) {
+      super.visitCallExpression(callExpression);
+      if (isCallRegisteringDjangoView(callExpression)) {
         RegularArgument viewArgument = nthArgumentOrKeyword(1, "view", callExpression.arguments());
         if (viewArgument != null) {
-          getSymbolFromTree(viewArgument.expression())
-            .filter(symbol -> symbol.fullyQualifiedName() != null)
-            .ifPresent(symbol -> djangoViewsFQN.add(symbol.fullyQualifiedName()));
+          PythonType pythonType = viewArgument.expression().typeV2();
+          if (pythonType instanceof UnknownType.UnresolvedImportType unresolvedImportType) {
```
maksim-grebeniuk-sonarsource, line 227: Since we will need to add FQN for other types (at least for `ClassType`) I suggest to introduce an interface like `HasFullyQualifiedName`, and implement it for the `UnresolvedImportType` and for the `FunctionType`

python-frontend/src/test/java/org/sonar/python/semantic/ProjectLevelSymbolTableTest.java
```diff
@@ -987,6 +987,59 @@ void django_views() {
...
+
+  @Test
+  void django_views_ambiguous() {
+    String content = """
+      from django.urls import path
+      if x:
+        def ambiguous(): ...
+      else:
+        def ambiguous(): ...
+      urlpatterns.append(path('bar', ambiguous, name='bar'))
```
maksim-grebeniuk-sonarsource, line 1024: Minor: here is definition of the `urlpatterns` is missed which could confuse reading this test case

python-frontend/src/test/java/org/sonar/python/semantic/ProjectLevelSymbolTableTest.java
```diff
@@ -987,6 +987,59 @@ void django_views() {
...
+    projectSymbolTable.addModule(parseWithoutSymbols(content), "my_package", pythonFile("urls.py"));
+    assertThat(projectSymbolTable.isDjangoView("my_package.urls.ambiguous")).isFalse();
+  }
+
+  @Test
+  void django_views_conf_import() {
+    String content = """
+      from django.urls import conf
+      import views
+      urlpatterns = [conf.path('foo', views.foo, name='foo'), conf.path('baz')]
```
maksim-grebeniuk-sonarsource, line 1036: Same here

python-frontend/src/main/java/org/sonar/python/types/v2/FunctionType.java
```diff
@@ -135,4 +138,8 @@ public PythonType owner() {
   public TypeOrigin typeOrigin() {
     return typeOrigin;
   }
+
+  public String fullyQualifiedName() {
```
maksim-grebeniuk-sonarsource, line 142: Since now we have let's say two types having `fqn` (the `UnresolvedImportType` and the `FunctionType` and in nearest future we may want to add it to more types - I think it makes sense to introduce an interface like:
```java
public interface HasFullyQualifiedName {
  String fullyQualifiedName();
}
```

maksim-grebeniuk-sonarsource (changes requested):
Looks good, left a couple of comments.
Also, for the moment we have its own way of construction of the `FQN` in the `PythonTypeToDescriptorConverter`, maybe it make sense to change it for the FunctionType by using new method in this PR

python-frontend/src/main/java/org/sonar/python/semantic/ProjectLevelSymbolTable.java
guillaume-dequenne-sonarsource: As discussed, `UnresolvedImportType` does not have a fully qualified name, because we said fully qualified name is strictly the name where a type is defined, and not just a re-exported name. Because the type is unknown, it has an `importPath` instead, which I think is good to keep as a clearly distinct concept from `fullyQualifiedName`.

In this case, I think what is misleading is probably the fact that I kept `djangoViewsFqns` when it probably should be `djangoViewsImportPaths` (`fullyQualifiedName` being seen here as a more precise import path).

Wdyt?

If we agree on that, I can rename the variable for `djangoViewsFqns`. As for the `HasFullyQualifiedName` interface, I agree it will make sense when introduced for `ClassType`, though. I'd probably introduce it at the same time as `ClassType` however.

guillaume-dequenne-sonarsource: As discussed, `UnresolvedImportType` does not have a fully qualified name, because we said fully qualified name is strictly the name where a type is defined, and not just a re-exported name. Because the type is unknown, it has an `importPath` instead, which I think is good to keep as a clearly distinct concept from `fullyQualifiedName`.

In this case, I think what is misleading is probably the fact that I kept `djangoViewsFqns` when it probably should be `djangoViewsImportPaths` (`fullyQualifiedName` being seen here as a more precise import path).

Wdyt?

If we agree on that, I can rename the variable for `djangoViewsFqns`. As for the `HasFullyQualifiedName` interface, I agree it will make sense when introduced for `ClassType`, though. I'd probably introduce it at the same time as `ClassType` however.

python-frontend/src/main/java/org/sonar/python/types/v2/FunctionType.java
guillaume-dequenne-sonarsource: As mentioned in the previous comment, `UnresolvedImportType` does not have a `fqn`.

We could have another interface that returns `importPath` (which would be the FQN for `FunctionType` and `ClassType`). I'm not yet sure about that though, as I'm afraid it could be misused.

python-frontend/src/test/java/org/sonar/python/semantic/ProjectLevelSymbolTableTest.java
guillaume-dequenne-sonarsource: Here `urlpatterns` is defined though? Are you referring to something else?

maksim-grebeniuk-sonarsource (approved):
LGTM

PR #2123, opened 2024-11-01
MMF-3945 Remove unneeded tests

guillaume-dequenne-sonarsource:
As part of cleaning up MMF-3945 branch, I'm removing these tests that were used to initially ensure V2-created descriptors were equivalent to V1 descriptors.
Now that we've moved to only using V2 descriptors, these tests duplicate existing tests for project symbol building.


maksim-grebeniuk-sonarsource (approved):
LGTM

PR #2125, opened 2024-11-01
SONARPY-2290 Support decorators for FunctionType to descriptors converter

maksim-grebeniuk-sonarsource:
[SONARPY-2290](https://sonarsource.atlassian.net/browse/SONARPY-2290)

[SONARPY-2290]: https://sonarsource.atlassian.net/browse/SONARPY-2290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python-checks/src/main/java/org/sonar/python/checks/ChangeMethodContractCheck.java
+++ b/python-checks/src/main/java/org/sonar/python/checks/ChangeMethodContractCheck.java
@@ -43,6 +43,7 @@ import static org.sonar.plugins.python.api.tree.Tree.Kind.FUNCDEF;
 public class ChangeMethodContractCheck extends PythonSubscriptionCheck {
 
   private static final Set<String> IGNORING_DECORATORS = Set.of(
+    "abc.abstractmethod",
     "abstractmethod",
     "overload"
   );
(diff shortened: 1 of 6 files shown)


python-checks/src/main/java/org/sonar/python/checks/ChangeMethodContractCheck.java
```diff
@@ -43,6 +43,7 @@
 public class ChangeMethodContractCheck extends PythonSubscriptionCheck {
 
   private static final Set<String> IGNORING_DECORATORS = Set.of(
+    "abc.abstractmethod",
     "abstractmethod",
```
guillaume-dequenne-sonarsource, line 47: Can you add (either in `ProjectLevelTypeTableTest` or `ProjectLevelSymbolTableTest`, or any relevant place) a test that highlights the discrepancies between decorator properly resolved from Typeshed (where I assume we only get a local name) and the case where a proper FQN is resolved?
We should also have a ticket to track this.

guillaume-dequenne-sonarsource (changes requested):
Looks good!
I think it would be good to add a test to document the inconsistent behavior between decorators coming from Protobuf and the others, as well as having a ticket for it.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #2128, opened 2024-11-01
Sonarpy 2293

guillaume-dequenne-sonarsource:


python-frontend/src/test/java/org/sonar/python/semantic/ProjectLevelSymbolTableTest.java
```diff
@@ -526,9 +526,9 @@ void test_imported_modules() {
     assertThat(projectLevelSymbolTable.importsByModule()).containsOnly(
       Map.entry("mod", Set.of("A")),
       Map.entry("mod2", Set.of("A", "B")),
-      Map.entry("mod3", Set.of("C.D.foo", "C.D.bar")),
+      Map.entry("mod3", Set.of("C.D")),
       Map.entry("mod4", Set.of("E")),
-      Map.entry("mod5", Collections.emptySet())
+      Map.entry("mod5", Set.of("F"))
```
guillaume-dequenne-sonarsource, line 531: Note: If a relative import appears to go above the root of the project, its resolution will default to the project root and not go higher.

thomas-serre-sonarsource (approved):
LGTM! 
Missing coverage is unrelated to your ticket, so it is good like this.
There are no new failing ITs from what I check.

My only comment is that I need to figure out what is the impact of changing the behavior of FQN. : 
> Only modules referenced from imports are added to importedModulesFQN. That  means that from a.b.c import d, e will only record a.b.c rather than a.b.c.d and a.b.c.e.  This is more consistent with the actual semantics of Python.

Anyway, It seems harmless!

guillaume-dequenne-sonarsource:
> LGTM! Missing coverage is unrelated to your ticket, so it is good like this. There are no new failing ITs from what I check.
> 
> My only comment is that I need to figure out what is the impact of changing the behavior of FQN. :
> 
> > Only modules referenced from imports are added to importedModulesFQN. That  means that from a.b.c import d, e will only record a.b.c rather than a.b.c.d and a.b.c.e.  This is more consistent with the actual semantics of Python.
> 
> Anyway, It seems harmless!

Indeed, about the coverage I think we should have a look at the MMF branch in its entirety before merging to double check for quality issues and missing coverage. In this case it felt unrelated to the task at hand so indeed I didn't look at it too much.

About the impact of the change of behavior for `from` import, it should hopefully not have any concrete impact. In `from a.b.c import d, e`, only `a.b.c` should refer to a proper module. I think that before, we had both entries that eventually resolved to the proper `a.b.c` module. Now, we resolve `a.b.c` directly.

PR #2131, opened 2024-11-05
SONARPY-2303 Stop running V1 type inference during the indexing phase

guillaume-dequenne-sonarsource:
[SONARPY-2303](https://sonarsource.atlassian.net/browse/SONARPY-2303)

[SONARPY-2303]: https://sonarsource.atlassian.net/browse/SONARPY-2303?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


python-frontend/src/main/java/org/sonar/plugins/python/api/PythonInputFileContext.java
```diff
@@ -59,7 +64,7 @@ public CacheContext cacheContext() {
 
   @Beta
   public Collection<Symbol> stubFilesSymbols() {
-    return TypeShed.stubFilesSymbols();
+    return projectLevelSymbolTable.typeShedDescriptorsProvider().stubFilesSymbols(projectLevelSymbolTable);
```
guillaume-dequenne-sonarsource, line 67: Note: I notice it's a bit weird to pass the `projectLevelSymbolTable` to the `typeShedDescriptorsProvider` when it is itself a part of the `projectLevelSymbolTable`. However, given it is code meant to disappear after we finish the removal of the V1 engine, I think it's okay to keep as it is.

maksim-grebeniuk-sonarsource: why not delegate this method inside the `ProjectLevelSymbolTable` to hide it behind the scenes?

python-frontend/src/main/java/org/sonar/python/semantic/v2/typeshed/TypeShedDescriptorsProvider.java
```diff
@@ -128,4 +139,35 @@ static ModuleSymbol deserializedModule(String moduleName, InputStream resource)
...
+
+    for (Descriptor descriptor : allDescriptors) {
+      if (descriptor.fullyQualifiedName() != null) {
+        Pair pair = symbolsByFqn.computeIfAbsent(descriptor.fullyQualifiedName(), k ->
+          new Pair(DescriptorUtils.symbolFromDescriptor(descriptor, projectLevelSymbolTable, null, new HashMap<>(), new HashMap<>()), descriptor));
+        cachedSymbols.add(pair.symbol());
+      }
+    }
+    return cachedSymbols;
+  }
```
thomas-serre-sonarsource, line 166: Why do you have to introduce a record here? 
```suggestion
  public Collection<Symbol> stubFilesSymbols(ProjectLevelSymbolTable projectLevelSymbolTable) {
    if (cachedSymbols != null) {
      return cachedSymbols;
    }
    Collection<Descriptor> allDescriptors = new ArrayList<>(stubFilesDescriptors());
    allDescriptors.addAll(builtinDescriptors().values());

    Map<String, Symbol> symbolsByFqn = new HashMap<>();
    cachedSymbols = new HashSet<>();

    for (Descriptor descriptor : allDescriptors) {
      if (descriptor.fullyQualifiedName() != null) {
        Symbol symbol = symbolsByFqn.computeIfAbsent(descriptor.fullyQualifiedName(), k -> DescriptorUtils.symbolFromDescriptor(descriptor, projectLevelSymbolTable, null, new HashMap<>(), new HashMap<>()));
        cachedSymbols.add(symbol);
      }
    }
    return cachedSymbols;
  }
```

python-frontend/src/main/java/org/sonar/python/semantic/v2/typeshed/TypeShedDescriptorsProvider.java
```diff
@@ -128,4 +139,35 @@ static ModuleSymbol deserializedModule(String moduleName, InputStream resource)
     }
   }
 
+
+  /**
+   * Returns stub symbols to be used by SonarSecurity.
+   * Ambiguous symbols that only contain class symbols are disambiguated with latest Python version.
+   */
+  public Collection<Symbol> stubFilesSymbols(ProjectLevelSymbolTable projectLevelSymbolTable) {
```
maksim-grebeniuk-sonarsource, line 147: I think that the `Descriptors` provider should not operate with the `Symbol` model, it should just provide descriptors. The conversion from the descriptor to the symbol model is on the `projectLevelSymbolTable` side

sonar-python-plugin/src/main/java/org/sonar/plugins/python/indexer/SonarQubePythonIndexer.java
```diff
@@ -172,10 +183,7 @@ public void computeGlobalSymbols(List<PythonInputFile> files, SensorContext cont
     if (caching.isCacheEnabled()) {
       saveGlobalSymbolsInCache(files);
       saveMainFilesListInCache(new HashSet<>(inputFileToFQN.values()));
-      Set<String> stubModules = TypeShed.stubModules();
-      if (!stubModules.isEmpty()) {
-        caching.writeTypeshedModules(stubModules);
-      }
+      // Information on used Typeshed stubs needs to be done at the end of the analysis, as it is not computed during indexing anymore
```
thomas-serre-sonarsource, line 186: I appreciate this comment :smile: I wouldn't have understand really what s going on without it :+1:

python-frontend/src/main/java/org/sonar/python/semantic/ProjectLevelSymbolTable.java
```diff
@@ -226,8 +226,8 @@ public void visitCallExpression(CallExpression callExpression) {
       }
     }
 
-    private static boolean isCallRegisteringDjangoView(CallExpression callExpression) {
-      TypeChecker typeChecker = new TypeChecker(new BasicTypeTable());
+    private boolean isCallRegisteringDjangoView(CallExpression callExpression) {
+      TypeChecker typeChecker = new TypeChecker(new BasicTypeTable(new ProjectLevelTypeTable(ProjectLevelSymbolTable.this)));
```
thomas-serre-sonarsource, line 230: _nit_ This `TypeChecker` construction is quite unclear to me (3 constructors on the same line, `ProjectLevelTypeTable` and `ProjectLevelSymbolTable` have so similar naming) but I guess it is another battle :wink: Ok like this!

python-frontend/src/main/java/org/sonar/python/index/DescriptorUtils.java
```diff
@@ -67,11 +67,57 @@ public static Symbol symbolFromDescriptor(Descriptor descriptor, ProjectLevelSym
...
+      case ALIAS:
+        Descriptor recreatedDescriptor = recreateDescriptorFromAlias((AliasDescriptor) descriptor);
+        return symbolFromDescriptor(recreatedDescriptor, projectLevelSymbolTable, symbolName, createdSymbolsByDescriptor, createdSymbolsByFqn);
       default:
         throw new IllegalStateException(String.format("Error while creating a Symbol from a Descriptor: Unexpected descriptor kind: %s", descriptor.kind()));
     }
   }
 
+  private static Descriptor recreateDescriptorFromAlias(AliasDescriptor aliasDescriptor) {
+    return switch (aliasDescriptor.originalDescriptor().kind()) {
```
maksim-grebeniuk-sonarsource, line 79: if replace this `kind` `switch-case` with just `instanceof` `if-else` - it will allow to get rid of unsafe class casts in the `recreateFunctionDescriptor` and the `recreateClassDescriptor` methods like:
```java
private static Descriptor recreateDescriptorFromAlias(AliasDescriptor aliasDescriptor) {
  if (aliasDescriptor.originalDescriptor() instanceof FunctionDescriptor functionDescriptor) {
    return recreateFunctionDescriptor(aliasDescriptor, functionDescriptor);
  } else if (aliasDescriptor.originalDescriptor() instanceof ClassDescriptor classDescriptor) {
    return recreateClassDescriptor(aliasDescriptor, classDescriptor);
  }
  throw new IllegalStateException(String.format("Error while recreating a descriptor from an alias: Unexpected alias kind: %s", aliasDescriptor.kind()));
}
```

python-frontend/src/main/java/org/sonar/python/index/DescriptorUtils.java
```diff
@@ -67,11 +67,57 @@ public static Symbol symbolFromDescriptor(Descriptor descriptor, ProjectLevelSym
...
         return ambiguousSymbol;
+      case ALIAS:
+        Descriptor recreatedDescriptor = recreateDescriptorFromAlias((AliasDescriptor) descriptor);
+        return symbolFromDescriptor(recreatedDescriptor, projectLevelSymbolTable, symbolName, createdSymbolsByDescriptor, createdSymbolsByFqn);
       default:
         throw new IllegalStateException(String.format("Error while creating a Symbol from a Descriptor: Unexpected descriptor kind: %s", descriptor.kind()));
     }
   }
 
+  private static Descriptor recreateDescriptorFromAlias(AliasDescriptor aliasDescriptor) {
```
thomas-serre-sonarsource, line 78: I think there are no unit tests for this function, especially in the case that the original descriptor is not a class or a function.

thomas-serre-sonarsource (changes requested):
LGTM! Great job!! 
Thanks for rewriting the git history; it was appreciated for the review.

I left mainly questions but no show-stopper.
I put it back in progress for you to answer them.

python-frontend/src/main/java/org/sonar/python/semantic/v2/converter/AliasDescriptorToPythonTypeConverter.java
```diff
@@ -0,0 +1,29 @@
...
+package org.sonar.python.semantic.v2.converter;
+
+import org.sonar.python.index.AliasDescriptor;
+import org.sonar.python.index.Descriptor;
+import org.sonar.python.types.v2.PythonType;
+
+public class AliasDescriptorToPythonTypeConverter implements DescriptorToPythonTypeConverter {
+  @Override
+  public PythonType convert(ConversionContext ctx, Descriptor from) {
+    // We should try to retrieve the original type if possible, instead of recreating it
```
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: Initially my goal was not to perform a conversion here, but to return a `LazyType` pointing to the original type. Then the lazy type resolution would fetch the original type in the relevant symbol table entry instead of performing the conversion.
The problem is that the resolution of lazy types is done in a `TypeWrapper` and eager resolution of them can trigger stack overflows.
To make this work, we should either allow `TypeWrapper` to be returned here, or to broaden the possibility of using `LazyType` in some new ways. I don't think it is technically difficult, but it involves some more thought and refactoring that is not really needed for now.

python-frontend/src/main/java/org/sonar/python/semantic/v2/typeshed/TypeShedDescriptorsProvider.java
guillaume-dequenne-sonarsource: I don't really remember why I needed that in a previous iteration...but certainly I don't anymore, good catch!

python-frontend/src/main/java/org/sonar/python/semantic/v2/BasicTypeTable.java
```diff
@@ -21,28 +21,41 @@
 import org.sonar.python.types.v2.UnknownType;
 
 public class BasicTypeTable implements TypeTable {
+
+  private final ProjectLevelTypeTable projectLevelTypeTable;
```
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: I was thinking of something of the sort, but I decided to go for the simplest approach. Eventually, I'd like to revisit this whole concept of `ProjectLevelTypeTable` / `BasicTypeTable` as I find them to be slightly confusing. If that's okay for you I think I'd prefer not touch it for now, and refactor later.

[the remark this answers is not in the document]
thomas-serre-sonarsource: makes sense! keep it like this for now

thomas-serre-sonarsource (approved):
LGTM! Well done :muscle: :rocket:

PR #2139, opened 2024-11-07
SONARPY-2311 Fix setting null items to ClassDescriptor.superClasses field

maksim-grebeniuk-sonarsource:


python-frontend/src/main/java/org/sonar/python/semantic/v2/converter/PythonTypeToDescriptorConverter.java
```diff
@@ -120,9 +120,8 @@ private static Descriptor convert(String moduleFqn, String parentFqn, String sym
     var hasSuperClassWithoutDescriptor = false;
     var superClasses = new ArrayList<String>();
     for (var superClassWrapper : type.superClasses()) {
-      var superClass = superClassWrapper.type();
-      if (superClass != PythonType.UNKNOWN) {
-        var superClassFqn = typeFqn(moduleFqn, superClass);
+      var superClassFqn = typeFqn(moduleFqn, superClassWrapper.type());
+      if (superClassFqn != null) {
```
guillaume-dequenne-sonarsource, line 124: I have a question that is not strictly related to this PR, but to the overall functional impact of what this piece of code is doing.

Before, we used the `hasSuperClassWithoutDescriptor` as a flag to determine whether a type hierarchy was unknown or not. (This was directly mapped to the `ClassSymbol#hasSuperClassWithoutDescriptor` field used in `ClassSymbol#hasUnresolvedTypeHierarchy`).

Now that we can resolve super class FQNs for `UnresolvedImportType`, I'm wondering how we guarantee the behavior of `ClassSymbol#hasUnresolvedTypeHierarchy` in cases like where a class inherits from another that is external to the project (and which we ultimately can't resolve).

Is there any specific measure for that?

guillaume-dequenne-sonarsource (commented):
The change looks good!

While this may be addressed outside of this PR (through another ticket/PR with tests - or maybe it already was and I missed it), I have a question regarding the overall behavior of `hasSuperClassWithoutDescriptor`. I feel this could have negative impacts on rule results (FPs). I'd like to know a bit better about the state of it before we proceed.

python-frontend/src/main/java/org/sonar/python/semantic/v2/converter/PythonTypeToDescriptorConverter.java
maksim-grebeniuk-sonarsource: Added assertions on symbol model

guillaume-dequenne-sonarsource (approved):
LGTM, thanks for the added tests!

PR #2143, opened 2024-11-08
SONARPY-2315 Add unit test case on S5890 when annotated type is imported and has field members only

maksim-grebeniuk-sonarsource:
[SONARPY-2315](https://sonarsource.atlassian.net/browse/SONARPY-2315)

[SONARPY-2315]: https://sonarsource.atlassian.net/browse/SONARPY-2315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python-checks/src/test/resources/checks/inconsistentTypeHintImported.py
+++ b/python-checks/src/test/resources/checks/inconsistentTypeHintImported.py
@@ -1,0 +1,5 @@
+class A:
+  v : int = 10
+
+class B:
+    def foo(self): ...
--- a/python-checks/src/test/resources/checks/inconsistentTypeHintImported.py
+++ b/python-checks/src/test/resources/checks/inconsistentTypeHintImported.py
@@ -1,5 +1,5 @@
-class A:
+class ClassWithFieldOnly:
   v : int = 10
 
-class B:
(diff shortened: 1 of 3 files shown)


python-checks/src/test/resources/checks/inconsistentTypeHintImported.py
```diff
@@ -0,0 +1,5 @@
+class A:
+  v : int = 10
+
+class B:
+    def foo(self): ...
```
thomas-serre-sonarsource, line 5: _nit_ You can add an empty line here.

thomas-serre-sonarsource, line 5: _minor_ To be more explicit about the intention of the test, I suggest to rename the classes
```suggestion
class ClassWithFieldOnly:
  v : int = 10

class ClassWithMethodOnly:
    def foo(self): ...
```

thomas-serre-sonarsource (approved):
LGTM! :+1: 
I left two minor comments

PR #2145, opened 2024-11-08
Fix quality gate issues

thomas-serre-sonarsource:


python-frontend/src/main/java/org/sonar/python/index/FunctionDescriptor.java
```diff
@@ -43,12 +43,12 @@ public class FunctionDescriptor implements Descriptor {
   private final TypeAnnotationDescriptor typeAnnotationDescriptor;
 
 
-  public FunctionDescriptor(String name, String fullyQualifiedName, List<Parameter> parameters, boolean isAsynchronous,
+  public FunctionDescriptor(String name, @Nullable String fullyQualifiedName, List<Parameter> parameters, boolean isAsynchronous,
```
guillaume-dequenne-sonarsource, line 46: After second look, I think fully qualified names for `FunctionDescriptor` should never be null again.

In `DescriptorsToProtobuf`, we're making null checks during deserialization from the cache because it used to be nullable (thus serialized with an empty value).
Due to our `cacheVersion` checks, we'd disregard any cached value that is null, so we should be safe to remove this check.
Note that if we actually do that:
- It breaks `ClassDescriptorTest` which are based on a V1->Descriptor converison.
- It is no longer consistent with the `descriptors.proto` model that says that `FunctionDescriptor#fullyQualifiedName` is an optional field.

Fixing this would significantly expand the scope of this PR, so I'd suggest creating a hardening ticket to ensure `FunctionDescriptor#fullyQualifiedName` is not longer nullable and adapt the related tests there, wdyt?

guillaume-dequenne-sonarsource (approved):
Looks good, thanks for handling this!

About `fullyQualifiedName`, I think we should ensure it's no longer `Nullable`, but I'd recommend doing it in a separate ticket. Wdyt?

python-frontend/src/main/java/org/sonar/python/index/FunctionDescriptor.java
thomas-serre-sonarsource: As discussed live, we leave the current PR _as is_ and create two tickets:
- One to remove the nullability of `fullyqualifiedname` (SONARPY-2335) 
- One to remove the conversion logic from V1 symbol to descriptor (SONARPY-2336)

PR #2147, opened 2024-11-12
SONARPY-2289 Add tests that document new issues that are raised by S1172 

thomas-serre-sonarsource:


python-frontend/src/test/java/org/sonar/python/semantic/ProjectLevelSymbolTableTest.java
```diff
@@ -679,6 +681,30 @@ void class_symbol_inheritance_from_nested_class() {
...
+      "class B(A): ",
+      "  def foo(self):",
+      "    pass"
+    );
+    projectLevelSymbolTable.addModule(importFileInput, "packageName", pythonFile("mod2.py"));
+    projectLevelSymbolTable.addModule(importedFileInput, "packageName", pythonFile("mod.py"));
+    Set<Symbol> globalSymbols = projectLevelSymbolTable.getSymbolsFromModule("packageName.mod");
+    Optional<ClassSymbol> classA = globalSymbols.stream().filter(s -> s.name().equals("A")).map(ClassSymbol.class::cast).findFirst();
+    assertThat(classA).isPresent();
+    // SONARPY-2327 The method call to foo() in class A is not considered as a member of A, while A was defined in the same file as B, it will be.
```
guillaume-dequenne-sonarsource, line 703: I assume you mean "..., if A was defined in the same file as B, it would be" ?

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -349,6 +351,25 @@ def foo(a, b): ...
...
+    FileInput fileInput = inferTypes("""
+      class A:
+        def meth(self):
+          return self.foo()
+      class B(A):
+        def foo(self):
+          ...
+      """
+    );
+    Optional<ClassSymbol> classA = fileInput.globalVariables().stream().filter(s -> s.name().equals("A")).map(ClassSymbol.class::cast).findFirst();
```
guillaume-dequenne-sonarsource, line 365: `ClassSymbol` is type inference V1. The assertions made in this test are essentially the same as in `TypeInferenceTest` (where they belong).

In `TypeInferenceV2Test`, we assert on the result of type inference V2, which means the result of `typeV2` calls that would provide an instance of `ClassType`.

That said, I guess we also could have the results of assertions on both `ClassSymbol` and `ClassType` to highlight the possible inconsistency (rather than looking both here and in the `TypeInferenceTest`), that's up to you.

python-frontend/src/test/java/org/sonar/python/types/TypeInferenceTest.java
```diff
@@ -803,6 +806,21 @@ void user_defined_attributes_reassigned() {
     ).type()).isEqualTo(DECL_INT);
   }
 
+  @Test
+  void child_class_method_call_is_not_a_member_of_parent_class() {
```
guillaume-dequenne-sonarsource, line 810: I guess it is, according to the test results?

guillaume-dequenne-sonarsource (changes requested):
This looks good!
I left small nitpick comments, as well as a more general comment on using `TypeInferenceV2Test` to assert on V2 types instead of V1.

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -362,12 +360,9 @@ def foo(self):
...
     );
-    Optional<ClassSymbol> classA = fileInput.globalVariables().stream().filter(s -> s.name().equals("A")).map(ClassSymbol.class::cast).findFirst();
-    assertThat(classA).isPresent();
-    // SONARPY-2327 The method call to foo() in class A is considered as a member of A, while if A was defined in another file than B, it won't be.
-    assertThat(classA.get().canHaveMember("foo")).isTrue();
-    assertThat(classA.get().declaredMembers()).extracting("kind", "name")
-      .containsExactlyInAnyOrder(Tuple.tuple(Symbol.Kind.FUNCTION, "meth"), Tuple.tuple(Symbol.Kind.OTHER, "foo"));
+    ClassType classA = (ClassType) ((ClassDef) fileInput.statements().statements().get(0)).name().typeV2();
+    assertThat(classA.members().stream().anyMatch(m -> m.name().equals("foo"))).isFalse();
+    assertThat(classA.members().stream().anyMatch(m -> m.name().equals("meth"))).isTrue();
```
thomas-serre-sonarsource, line 365: Using the V2 type inference functions, `foo` is no longer a member of class A. Thus, I removed the comment referring to ticket SONARPY-2327: It is not a problem of inconsistency between declaring the parent class and the child class in the same file or in two separate files, but a problem of migrating the rule to the V2 type. 
If you agree, I'll close the ticket 2327. WDYT?

python-frontend/src/test/java/org/sonar/python/semantic/ProjectLevelSymbolTableTest.java
```diff
@@ -700,7 +700,6 @@ void child_class_method_call_is_not_a_member_of_parent_class() {
     Set<Symbol> globalSymbols = projectLevelSymbolTable.getSymbolsFromModule("packageName.mod");
     Optional<ClassSymbol> classA = globalSymbols.stream().filter(s -> s.name().equals("A")).map(ClassSymbol.class::cast).findFirst();
     assertThat(classA).isPresent();
-    // SONARPY-2327 The method call to foo() in class A is not considered as a member of A, while A was defined in the same file as B, it will be.
```
thomas-serre-sonarsource: See comment below on the other comment I ve deleted.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #2152, opened 2024-11-13
SONARPY-2330 Add unit test case on S6542 when a class is inherited from an imported class with a metaclass

maksim-grebeniuk-sonarsource:
[SONARPY-2330](https://sonarsource.atlassian.net/browse/SONARPY-2330)

[SONARPY-2330]: https://sonarsource.atlassian.net/browse/SONARPY-2330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


python-checks/src/test/resources/checks/useOfAnyAsTypeHintImporting.py
```diff
@@ -0,0 +1,15 @@
...
+class LocalParent: ...
+
+class LocalParentWithMetaClass(metaclass=ABCMeta): ...
+
+class LocalWithMetaClassInherited(LocalParentWithMetaClass):
+    def local_inherited_foo(self) -> Any: # Noncompliant
+        ...
+
+class ImportedWithMetaClassInherited(ImportedParentWithMetaClass):
+    def imported_inherited_foo(self) -> Any: # FN SONARPY-2331
```
guillaume-dequenne-sonarsource, line 14: For my understanding, is this because losing the metaclass information makes the `classSymbol.hasUnresolvedTypeHierarchy()` to be true? Or is it something else?

python-checks/src/test/java/org/sonar/python/checks/UseOfAnyAsTypeHintCheckTest.java
```diff
@@ -42,4 +43,15 @@ void useOfUserDefinedTypeCalledAny() {
   void useOfOverrideOrOverloadDecorator() {
     PythonCheckVerifier.verify(List.of("src/test/resources/checks/useOfOverrideOrOverloadDecorator.py", "src/test/resources/checks/reexport_typing_overload_override.py"), new UseOfAnyAsTypeHintCheck());
   }
+
+  @Test
+  void useOfAnyImpored() {
```
thomas-serre-sonarsource, line 48:
```suggestion
  void useOfAnyImported() {
```

python-checks/src/test/java/org/sonar/python/checks/UseOfAnyAsTypeHintCheckTest.java
```diff
@@ -42,4 +43,15 @@ void useOfUserDefinedTypeCalledAny() {
...
+    PythonCheckVerifier.verify(
+      List.of(
+        "src/test/resources/checks/useOfAnyAsTypeHintImported.py",
+        "src/test/resources/checks/useOfAnyAsTypeHintImporting.py"
```
thomas-serre-sonarsource, line 52: __minor__ Since there are many existing files used to test this rule, I suggest regrouping them in a folder, for example in `src/test/resources/checks/useOfAnyAsTypeHint`

python-checks/src/test/resources/checks/useOfAnyAsTypeHintImported.py
```diff
@@ -0,0 +1,3 @@
+from abc import ABCMeta
+
+class ImportedParentWithMetaClass(metaclass=ABCMeta): ...
```
thomas-serre-sonarsource, line 3: _minor_ I suggest adding a class that has no metaclass in this file. Then, you can add a child class in the other file to confirm that the metaclass is really the root cause.

python-checks/src/test/resources/checks/useOfAnyAsTypeHintImporting.py
```diff
@@ -0,0 +1,15 @@
+from typing import Any
+from abc import ABCMeta
+from useOfAnyAsTypeHintImported import ImportedParentWithMetaClass
+
+class LocalParent: ...
+
```
thomas-serre-sonarsource, line 6: _minor_ useless statement?

python-checks/src/test/resources/checks/useOfAnyAsTypeHintImporting.py
maksim-grebeniuk-sonarsource: for this specific case, it leads to the incorrect result of the `ClassSymbolImpl.hasSuperClassWithUnknownMetaClass` method

thomas-serre-sonarsource (approved):
LGTM! :muscle:

PR #2153, opened 2024-11-13
SONARPY-2331 Implement setting metaclassFQN field for ClassType to ClassDescriptor conversion

maksim-grebeniuk-sonarsource:
[SONARPY-2331](https://sonarsource.atlassian.net/browse/SONARPY-2331)

[SONARPY-2331]: https://sonarsource.atlassian.net/browse/SONARPY-2331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python-frontend/src/main/java/org/sonar/python/semantic/v2/converter/PythonTypeToDescriptorConverter.java
+++ b/python-frontend/src/main/java/org/sonar/python/semantic/v2/converter/PythonTypeToDescriptorConverter.java
@@ -129,6 +129,12 @@ public class PythonTypeToDescriptorConverter {
       }
     }
 
+    var metaclassFQN = type.metaClasses()
+      .stream()
+      .map(metaClass -> typeFqn(moduleFqn, metaClass))
+      .findFirst()
+      .orElse(null);
+
     return new ClassDescriptor(symbolName, symbolFqn,
       superClasses,
       memberDescriptors,
@@ -136,7 +142,7 @@ public class PythonTypeToDescriptorConverter {
       type.definitionLocation().orElse(null),
       hasSuperClassWithoutDescriptor,
       type.hasMetaClass(),
-      null,
+      metaclassFQN,
       false
     );
   }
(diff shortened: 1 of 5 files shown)


python-frontend/src/main/java/org/sonar/python/semantic/v2/converter/PythonTypeToDescriptorConverter.java
```diff
@@ -129,14 +129,20 @@ private static Descriptor convert(String moduleFqn, String parentFqn, String sym
...
+
     return new ClassDescriptor(symbolName, symbolFqn,
       superClasses,
       memberDescriptors,
       type.hasDecorators(),
       type.definitionLocation().orElse(null),
       hasSuperClassWithoutDescriptor,
       type.hasMetaClass(),
-      null,
+      metaclassFQN,
```
guillaume-dequenne-sonarsource, line 145: Not strictly related to the PR, but I see the `metaclassFQN` method is not annotated as `CheckForNull` which is probably misleading.

Also, I'm a bit concerned that we're going to cement the usage of the `metaclassFQN` somewhat inconsistently (because `importType.importPath()` does not produce a fully qualified name according to our definition). I think it's okay for now, but we should keep in mind that this probably need to change eventually (like for decorators).

python-frontend/src/test/java/org/sonar/python/semantic/ProjectLevelSymbolTableTest.java
```diff
@@ -1125,6 +1125,18 @@ class Field(MetaField): ...
     assertThat(symbol.hasUnresolvedTypeHierarchy()).isTrue();
   }
 
+  @Test
+  void class_wth_metaclass() {
```
guillaume-dequenne-sonarsource, line 1129: It would probably make sense to have some other tests where the metaclass is local or unresolved. I know it's basically testing `typeFqn` again, but since it's not a fully valid implementation (cf above comment), we probably want to have it documented.

guillaume-dequenne-sonarsource (changes requested):
Looks good!
Left a few comments, but they're quite open to discussion - not sure we need to make changes as part of this PR.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #2159, opened 2024-11-15
SONARPY-2329 Add tests documenting new S2638 FNs

thomas-serre-sonarsource:
[SONARPY-2329](https://sonarsource.atlassian.net/browse/SONARPY-2329)

Part of 

[SONARPY-2329]: https://sonarsource.atlassian.net/browse/SONARPY-2329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


python-checks/src/test/resources/checks/changeMethodContract/changeMethodContractIntermediate.py
```diff
@@ -0,0 +1,8 @@
+from changeMethodContractParent import ParentClass
+from typing import Any
+
+class IntermediateClass(ParentClass):
+  ...
+
+class IntermediateWithAnyClass(ParentClass[Any]):
```
maksim-grebeniuk-sonarsource, line 7: does it make a difference in case we have an extra file here compared to just having these classes declared in the `changeMethodContractParent.py`?

maksim-grebeniuk-sonarsource, line 7: The real reason of the FN is not related to the `Any` type, but related to inheritance from types with generic type. So, I guess it will be easier to understand what's going on if the class name will be like `IntermediateWithGenericTypeParent` or something like that

maksim-grebeniuk-sonarsource (changes requested):
Overall, it looks good, I have two comments whiches need to be clarified

python-checks/src/test/resources/checks/changeMethodContract/changeMethodContractIntermediate.py
thomas-serre-sonarsource: Makes sense!

maksim-grebeniuk-sonarsource (approved):
LGTM

PR #2160, opened 2024-11-15
SONARPY-2342 Fix NPE when metaclassFQN resolved to null during a PythonType to Descriptor conversion

maksim-grebeniuk-sonarsource:
[SONARPY-2342](https://sonarsource.atlassian.net/browse/SONARPY-2342)

[SONARPY-2342]: https://sonarsource.atlassian.net/browse/SONARPY-2342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python-frontend/src/main/java/org/sonar/python/semantic/v2/converter/PythonTypeToDescriptorConverter.java
+++ b/python-frontend/src/main/java/org/sonar/python/semantic/v2/converter/PythonTypeToDescriptorConverter.java
@@ -132,6 +132,7 @@ public class PythonTypeToDescriptorConverter {
     var metaclassFQN = type.metaClasses()
       .stream()
       .map(metaClass -> typeFqn(moduleFqn, metaClass))
+      .filter(Objects::nonNull)
       .findFirst()
       .orElse(null);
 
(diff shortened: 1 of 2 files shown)


python-frontend/src/main/java/org/sonar/python/semantic/v2/converter/PythonTypeToDescriptorConverter.java
```diff
@@ -132,6 +132,7 @@ private static Descriptor convert(String moduleFqn, String parentFqn, String sym
     var metaclassFQN = type.metaClasses()
       .stream()
       .map(metaClass -> typeFqn(moduleFqn, metaClass))
+      .filter(Objects::nonNull)
```
guillaume-dequenne-sonarsource, line 135: This here is yet again an example of pitfalls when using streams this way.
You missed this edge case during implementation, and I missed it during review. The coverage was 100%, yet we experienced an NPE after merging.
I think we need to make progress and I don't want to refactor just for the sake of it (maybe `typeFqn` should return Optional?), but I must admit I'm a bit annoyed that we fall into similar traps so often.

guillaume-dequenne-sonarsource (approved):
LGTM. As mentioned, I'm a bit annoyed that we keep getting surprised by these cases. It should not happen IMO.

PR #2162, opened 2024-11-18
SONARPY-2313 Fix FN on S6543 when a generic class is defined in another file

guillaume-dequenne-sonarsource:
This PR introduces the notion of a `SpecialFormType`.
This kind of type is a special typing internal type used to mark typing constructs (such as `typing.Union`, `typing.Optional`, etc...).

In `typing.pyi`, these constructs are defined as:

```
Union: _SpecialForm
Generic: _SpecialForm
...
```
Representing these constructs as instances of the `ClassType` `SpecialForm` is problematic for two reasons:
- It introduces the risk of FPs by considering these like "normal" types
- They would be represented as `ObjectType` instances of `SpecialForm`. Disambiguating them would be difficult (relying on `attributes`, which is initially not suited to store simple names).

Therefore, I believe introducing this special kind of type is simpler to circumvent these problems.


python-checks/src/test/resources/checks/changeMethodContract/changeMethodContractParent.py
```diff
@@ -1,6 +1,6 @@
 from typing import Any
 
-class ParentClass(object):
+class ParentClass[T](object):
```
guillaume-dequenne-sonarsource, line 3: Note: the issues were initially not raised due to `ParentClass` not being a proper generic class.

python-frontend/src/main/java/org/sonar/python/semantic/v2/converter/VariableDescriptorToPythonTypeConverter.java
```diff
@@ -19,25 +19,30 @@
...
-      if (fqn != null) {
-        return ctx.lazyTypesContext().getOrCreateLazyType(fqn);
+    String fullyQualifiedName = from.fullyQualifiedName();
+    if (from.isImportedModule() && fullyQualifiedName != null) {
+      return ctx.lazyTypesContext().getOrCreateLazyType(fullyQualifiedName);
+    }
+    String annotatedType = from.annotatedType();
+    if (annotatedType != null) {
+      if ("typing._SpecialForm".equals(annotatedType) && fullyQualifiedName != null) {
+        // Defensive null check on fullyQualifiedName: it should never be null for SpecialForm
```
thomas-serre-sonarsource, line 39: :+1:

python-frontend/src/main/java/org/sonar/python/types/v2/TypeCheckBuilder.java
```diff
@@ -269,6 +274,27 @@ public TriBool test(PythonType pythonType) {
...
+      if (pythonType instanceof ClassType classType && classType.isGeneric()) {
+        return TriBool.TRUE;
+      }
+      if (pythonType instanceof SpecialFormType specialFormType) {
+        return "typing.Generic".equals(specialFormType.fullyQualifiedName()) ? TriBool.TRUE : TriBool.UNKNOWN;
+      }
+      if (pythonType instanceof UnknownType.UnresolvedImportType unresolvedImportType && "typing.Generic".equals(unresolvedImportType.importPath())) {
+        return TriBool.TRUE;
+      }
+      return TriBool.UNKNOWN;
```
thomas-serre-sonarsource, line 294: this function never returns false. Is it a conservative choice? For example, It seems paradoxical for `ClassType` to have `isGeneric` false and to return `UNKNOWN` here.

thomas-serre-sonarsource (approved):
Nice job, well done! Super clean implementation :muscle:

python-frontend/src/main/java/org/sonar/python/types/v2/TypeCheckBuilder.java
guillaume-dequenne-sonarsource: Yes, it's conservative. I think that `isGeneric` in `ClassType` should probably be a `TriBool` in fact, as what I want to express is that `isGeneric=true` means that it's for sure a generic class, and `false` just means there's no sign that it is.
I don't think it should have an impact until we actually want to raise issues on explicitly non generic classes, where this heuristic will have to be thought about more deeply.

python-checks/src/test/resources/checks/genericTypeWithoutArgumentImported.py
```diff
@@ -0,0 +1,9 @@
+from typing import Generic
+class SomeGeneric(Generic[T]):
+  ...
+
+class SomeGenericWithTypeParam[T](): ...
+
+class MyImportedChild(SomeGeneric[T]): ...
+class MyImportedChild2(SomeGeneric): ...
+class MyImportedChild3(SomeGeneric[str]): ...
```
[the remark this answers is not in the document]
guillaume-dequenne-sonarsource: I renamed `MyImportedChild2` to `MyImportedNonGenericChild` (negated) to be more explicit about the fact it's no longer a generic class.

python-checks/src/test/resources/checks/genericTypeWithoutArgumentImporting.py
```diff
@@ -1,5 +1,11 @@
-# mod.py
-from genericTypeWithoutArgumentImported import SomeGeneric, SomeGenericWithTypeParam, MyImportedChild, MyImportedChild2, MyImportedChild3
+from genericTypeWithoutArgumentImported import (
+    SomeGeneric,
+    SomeGenericWithTypeParam,
+    MyImportedGenericTypeVarChild,
+    MyImportedNonGenericChild,
+    MyImportedConcreteChild,
+    SomeGenericIncorrectlyDefined
+)
```
thomas-serre-sonarsource, line 8: may be?
```suggestion
from genericTypeWithoutArgumentImported import *
```

thomas-serre-sonarsource (approved):
LGTM! It is better with this last commit in my opinion. Good job

python-checks/src/test/resources/checks/genericTypeWithoutArgumentImporting.py
guillaume-dequenne-sonarsource: Since support of wildcard import is brittle at the moment, I don't want to add confusion to this test: I prefer to keep the explicit import.

PR #2164, opened 2024-11-18
SONARPY-2345 Document new S1172 FN when super class is ambiguous

thomas-serre-sonarsource:
[SONARPY-2345](https://sonarsource.atlassian.net/browse/SONARPY-2345)

Part of 

[SONARPY-2345]: https://sonarsource.atlassian.net/browse/SONARPY-2345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python-checks/src/test/resources/checks/unusedFunctionParameter/unusedFunctionParameterImport.py
+++ b/python-checks/src/test/resources/checks/unusedFunctionParameter/unusedFunctionParameterImport.py
@@ -1,4 +1,5 @@
-from unusedFunctionParameterImported import ImportedParent
+from unusedFunctionParameterImported import ImportedParent, ParentWithDuplicatedParent
+
 class ChildFromImported(ImportedParent):
 
     # SONARPY-2327 `method_defined_in_child_class_only` is not considered a member of ImportedParent class, thus S1172 is raised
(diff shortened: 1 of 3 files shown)


python-checks/src/test/resources/checks/unusedFunctionParameter/unusedFunctionParameterImport.py
```diff
@@ -9,3 +10,9 @@ def method_defined_in_child_class_only(self, a): # Noncompliant
     def method_defined_in_child_class_only_and_not_used(self, a): # Noncompliant
         #                                                     ^
         return compute()
+
+class ChildFromDuplicated(ParentWithDuplicatedParent):
+
+    def do_something(self, a): # FN SONARPY-2347 ChildFromDuplicated has an unresolved type hierarchy, because of the duplicated parent classes
```
guillaume-dequenne-sonarsource, line 16: I think this is a consequence of SONARPY-1829: `DuplicatedParent`, being a class, is not propagated like other types. In `unusedFunctionParameterImported.py`, the hierarchy of `ParentWithDuplicatedParent` should be clearly defined, as the second definition of `DuplicatedParent` always overrides the first one.
I'm not sure we need SONARPY-2347 as a result. Now that it's here, we can probably mark it as blocked by SONARPY-1829 though (or just add the example in SONARPY-1829 and archive SONARPY-2347 as duplicate).

guillaume-dequenne-sonarsource (approved):
Looks good!
As mentioned in the comment, the root cause of the problem already has a dedicated ticket, so I'm not entirely sure of the value of this new one.

python-checks/src/test/resources/checks/unusedFunctionParameter/unusedFunctionParameterImport.py
thomas-serre-sonarsource: Alright! I canceled the SONARPY-2347 and changed the reference in the unit test to SONARPY-1829. Thanks for noticing it!

PR #2167, opened 2024-11-19
SONARPY-2310: Partly support wildcard import

joke1196:
[SONARPY-2310](https://sonarsource.atlassian.net/browse/SONARPY-2310)

[SONARPY-2310]: https://sonarsource.atlassian.net/browse/SONARPY-2310?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
+++ b/python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
@@ -596,7 +596,7 @@ public class TypeInferenceV2Test {
       """);
 
     var functionDef = (FunctionDef) root.statements().statements().get(1);
-    var lastExpressionStatement = (ExpressionStatement) functionDef.body().statements().get(functionDef.body().statements().size() -1);
+    var lastExpressionStatement = (ExpressionStatement) functionDef.body().statements().get(functionDef.body().statements().size() - 1);
     assertThat(lastExpressionStatement.expressions().get(0).typeV2().unwrappedType()).isEqualTo(PythonType.UNKNOWN);
   }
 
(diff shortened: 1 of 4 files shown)


python-frontend/src/main/java/org/sonar/python/semantic/v2/TypeInferenceV2.java
```diff
@@ -61,10 +62,15 @@ public TypeInferenceV2(TypeTable projectLevelTypeTable, PythonFile pythonFile, S
   }
 
   public Map<SymbolV2, Set<PythonType>> inferTypes(FileInput fileInput) {
-    TrivialTypeInferenceVisitor trivialTypeInferenceVisitor = new TrivialTypeInferenceVisitor(projectLevelTypeTable, pythonFile, fullyQualifiedModuleName);
+    WildcardImportVisitor wildcardImportVisitor = new WildcardImportVisitor(projectLevelTypeTable);
```
guillaume-dequenne-sonarsource, line 65: Why would this need to be its dedicated visitor?
I know we're going for a naive implementation, but I think it's still sensible to assume that we don't want the following resolution:
```
foo()
from lib import * # contains foo
```

I'm under the impression that relying on the main visit of `TrivialTypeInferenceVisitor` should be enough here, no?

Seppli11: The `TrivialTypeInferenceVisistor` is enough (initially that's what we did). Our thought was that if it is moved to its own visitor, it is easier to rip it out when we eventually will do that. But we didn't think of the example you provided

python-frontend/src/test/java/org/sonar/python/semantic/v2/TypeInferenceV2Test.java
```diff
@@ -3124,4 +3262,25 @@ private static Statement lastStatement(StatementList statementList) {
     List<Statement> statements = statementList.statements();
     return statements.get(statements.size() - 1);
   }
+
+  private static class TestProject {
```
guillaume-dequenne-sonarsource, line 3272: That's a pretty neat construct!
I'm wondering if it wouldn't make more sense for this to be in a different test category though (either `ProjectLevelSymbolTableTest` or `ProjectLevelTypeTableTest`, or even a new `CrossFileTypeInferenceTest` dedicated to tests focused on import resolution.
We probably can keep this here for now though.

guillaume-dequenne-sonarsource (approved):
Kudos on the clear tests (well documenting the limitations) and the clear separation of commits (even if I didn't notice it initially, sorry about that!).
LGTM.

PR #2170, opened 2024-11-20
SONARPY-2332 Fix FP on S5845 when comparing enum objects with types

thomas-serre-sonarsource:
[SONARPY-2332](https://sonarsource.atlassian.net/browse/SONARPY-2332)

Part of 

[SONARPY-2332]: https://sonarsource.atlassian.net/browse/SONARPY-2332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python-checks/src/test/resources/checks/tests/assertOnDissimilarTypes.py
+++ b/python-checks/src/test/resources/checks/tests/assertOnDissimilarTypes.py
@@ -198,8 +198,8 @@ class ComparingTypeAndEnum(unittest.TestCase):
     def test_type_of_enum_and_enum_class(self):
         EnumType = Enum("EnumType", names=["one", "two"])
         enum_member = EnumType(1)
-        # FP SONARPY-2332: EnumType is a proper "type" object since the Enum constructor with argument creates a new Enum type
-        self.assertIs(type(enum_member), EnumType) # Noncompliant
+        # EnumType is a proper "type" object since the Enum constructor with argument creates a new Enum type
+        self.assertIs(type(enum_member), EnumType)
 
 class DerivedEnumWithMembers(Enum):
     ONE = 1
@@ -208,7 +208,7 @@ class DerivedEnumWithMembers(Enum):
 class ComparingTypeAndDerivedEnumWithMembers(unittest.TestCase):
     def test_type_of_derived_enum_and_derived_enum_class(self):
         derived_enum_member = DerivedEnumWithMembers(1)
-        self.assertIs(type(enum_member), EnumType)
+        self.assertIs(type(derived_enum_member), EnumType)
 
 class DerivedEnumWithoutMembers(Enum):
     ...
@@ -218,12 +218,12 @@ class ComparingTypeAndDerivedEnumWithoutMembers(unittest.TestCase):
     def test_type_of_derived_enum_and_derived_enum_class(self):
         DerivedEnumType = DerivedEnumWithoutMembers("DerivedEnumType", names=["one", "two"])
         derived_enum_member = DerivedEnumType(1)
-        # FP SONARPY-2332: DerivedEnumType is a proper "type" object because `DerivedEnumWithoutMembers(str, list)` will create a new enum,
+        # DerivedEnumType is a proper "type" object because `DerivedEnumWithoutMembers(str, list)` will create a new enum,
         # just like `Enum(str, list)` would
-        self.assertIs(type(derived_enum_member), DerivedEnumType) # Noncompliant
+        self.assertIs(type(derived_enum_member), DerivedEnumType)
 
         DerivedEnumTypeFromDict = DerivedEnumWithoutMembers("DerivedEnumType", OrderedDict([("one", 1), ("two", 2)]))
         derived_enum_member_from_dict = DerivedEnumTypeFromDict(1)
-        # FP SONARPY-2332: DerivedEnumTypeFromDict is a proper "type" object because `DerivedEnumWithoutMembers(str, list)` will create a new enum,
+        # DerivedEnumTypeFromDict is a proper "type" object because `DerivedEnumWithoutMembers(str, list)` will create a new enum,
         # just like `Enum(str, list)`
-        self.assertIs(type(derived_enum_member_from_dict), DerivedEnumTypeFromDict) # Noncompliant
+        self.assertIs(type(derived_enum_member_from_dict), DerivedEnumTypeFromDict)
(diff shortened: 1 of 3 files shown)


python-checks/src/test/resources/checks/tests/assertOnDissimilarTypes.py
```diff
@@ -208,7 +207,7 @@ class DerivedEnumWithMembers(Enum):
 class ComparingTypeAndDerivedEnumWithMembers(unittest.TestCase):
     def test_type_of_derived_enum_and_derived_enum_class(self):
         derived_enum_member = DerivedEnumWithMembers(1)
-        self.assertIs(type(enum_member), EnumType)
+        self.assertIs(type(derived_enum_member), EnumType)
```
thomas-serre-sonarsource, line 210: Mistake that was present in the previous ticket. I corrected it here.

python-frontend/src/main/java/org/sonar/python/types/RuntimeType.java
```diff
@@ -31,6 +31,9 @@
 
 public class RuntimeType implements InferredType {
 
+  // Name of the type returned by the type() function
+  private static final String TYPE_CLASS_NAME = "type";
```
maksim-grebeniuk-sonarsource, line 35: Did you try to use `org.sonar.python.types.InferredTypes#TYPE`?
It looks like that for the v1 type inference we have this constant representing the `type` class.

python-frontend/src/main/java/org/sonar/python/types/RuntimeType.java
```diff
@@ -52,9 +55,27 @@ public boolean isIdentityComparableWith(InferredType other) {
...
+      boolean hasThisMetaClass = hasMetaclassInHierarchy(this);
+      return (TYPE_CLASS_NAME.equals(getTypeClass().name()) && hasOtherMetaClass)
+        || (hasThisMetaClass && TYPE_CLASS_NAME.equals(otherRuntimeType.getTypeClass().name()));
+    }
+    return false;
+  }
+
+  private static boolean hasMetaclassInHierarchy(RuntimeType runtimeType) {
+    return runtimeType.getTypeClass().hasMetaClass()
+      || runtimeType.getTypeClass().superClasses().stream().filter(ClassSymbol.class::isInstance).anyMatch(c -> ((ClassSymbol) c).hasMetaClass());
```
maksim-grebeniuk-sonarsource, line 76: should we look only for current type parents for a `metaclass`, or should we go deeper in the types hierarchy?

python-frontend/src/main/java/org/sonar/python/types/RuntimeType.java
```diff
@@ -52,9 +55,27 @@ public boolean isIdentityComparableWith(InferredType other) {
...
+      return true;
+    }
     return this.equals(other);
   }
 
+  private boolean isComparingTypeWithMetaclass(InferredType other) {
+    if (other instanceof RuntimeType otherRuntimeType) {
+      boolean hasOtherMetaClass = hasMetaclassInHierarchy(otherRuntimeType);
+      boolean hasThisMetaClass = hasMetaclassInHierarchy(this);
+      return (TYPE_CLASS_NAME.equals(getTypeClass().name()) && hasOtherMetaClass)
```
maksim-grebeniuk-sonarsource, line 68: I guess there could be just a comparison with the `org.sonar.python.types.InferredTypes#TYPE` constant.

python-frontend/src/test/java/org/sonar/python/types/RuntimeTypeTest.java
```diff
@@ -64,6 +64,34 @@ void isIdentityComparableWith() {
     assertThat(aType.isIdentityComparableWith(new DeclaredType(b))).isTrue();
   }
 
+  @Test
+  void isIdentityComparableWithMetaclass() {
+    RuntimeType typeType = new RuntimeType(new ClassSymbolImpl("type", "type"));
```
maksim-grebeniuk-sonarsource, line 69: here you can use `org.sonar.python.types.InferredTypes#TYPE`

python-frontend/src/main/java/org/sonar/python/types/RuntimeType.java
thomas-serre-sonarsource: good catch, I ll use it. thanks

python-frontend/src/main/java/org/sonar/python/types/RuntimeType.java
thomas-serre-sonarsource: I intend to go deeper as the original issue contained a child class from `enum.Enum`

python-frontend/src/main/java/org/sonar/python/types/RuntimeType.java
```diff
@@ -52,7 +52,23 @@ public boolean isIdentityComparableWith(InferredType other) {
     if (other instanceof UnionType) {
       return other.isIdentityComparableWith(this);
     }
-    return this.equals(other);
+    return isComparingTypeWithMetaclass(other) || this.equals(other);
+  }
+
+  private boolean isComparingTypeWithMetaclass(InferredType other) {
```
maksim-grebeniuk-sonarsource, line 58: as discussed, could be simplified like this :
```java
private boolean isComparingTypeWithMetaclass(InferredType other) {
    if (other instanceof RuntimeType otherRuntimeType) {
      boolean hasOtherMetaClass = hasMetaclassInHierarchy(otherRuntimeType);
      boolean hasThisMetaClass = hasMetaclassInHierarchy(this);
      return (InferredTypes.TYPE.equals(this) && hasOtherMetaClass)
        || (hasThisMetaClass && InferredTypes.TYPE.equals(otherRuntimeType));
    }
    return false;
  }
```

python-frontend/src/main/java/org/sonar/python/types/RuntimeType.java
maksim-grebeniuk-sonarsource: as discussed - we need to check whole hierarchy, not nearest parents

maksim-grebeniuk-sonarsource (changes requested):
After a discussion some changes requested

maksim-grebeniuk-sonarsource (approved):
LGTM, thanks for the fixes

PR #2198, opened 2024-11-29
SONARPY-2382 fix unary propagation problem

Seppli11:


python-checks/src/test/resources/checks/nonCallableCalledNoIssues.py
```diff
@@ -1,3 +1,6 @@
 
 def param_with_type_hint(a: int):
-    a() # OK
\ No newline at end of file
+    a() # OK
+
+def call_unknown_value():
+    (~s)(X) # Ok
```
thomas-serre-sonarsource, line 6: _minor_ I find it not explicit what you want to test here. What matters is not really the fact that `s` is unknown but that there is a unary ~ operator applied on `s` right? 
I suggest adding a test in `nonCallableCalled` called `def callable_with_unary_operator` and checking that not issue are raised when `+,-,~...` are applied

python-frontend/src/main/java/org/sonar/python/semantic/v2/types/TrivialTypePropagationVisitor.java
```diff
@@ -52,22 +51,32 @@ public TrivialTypePropagationVisitor(TypeTable typeTable) {
...
+      case "+", "-" -> mapUnaryPlusMinusType(type);
+      default -> PythonType.UNKNOWN;
+    };
+  }
+
+  private PythonType mapInvertExprType(PythonType type) {
+    if(isIntTypeCheck.check(type) == TriBool.TRUE || isBooleanTypeCheck.check(type) == TriBool.TRUE) {
+      return intType;
+    }
+    return PythonType.UNKNOWN;
```
thomas-serre-sonarsource, line 79: clean implementation, well done :muscle:

python-frontend/src/test/java/org/sonar/python/semantic/v2/types/TrivialTypePropagationVisitorTest.java
```diff
@@ -105,4 +127,15 @@ void testNotOfCustomClass() {
...
+  void testUnknownOperator() {
+    var operator = mock(TokenImpl.class);
+    when(operator.value()).thenReturn("invalid_operator");
+    UnaryExpressionImpl expr = new UnaryExpressionImpl(operator, lastExpression("1"));
+    expr.typeV2(TypesTestUtils.INT_TYPE);
+
+    expr.accept(trivialTypePropagationVisitor);
+    assertThat(expr.typeV2()).isEqualTo(PythonType.UNKNOWN);
+  }
 }
```
thomas-serre-sonarsource, line 141: _nit_ add the extra line there

python-checks/src/test/resources/checks/nonCallableCalledNoIssues.py
thomas-serre-sonarsource: Now that I read the all PR I understand why you called the test function like this. Still, I think it will be clearer for someone who hasn't the context to rewrite this test to be more explicit

thomas-serre-sonarsource (approved):
LGTM, well done :muscle: 

I left a minor comment on the test

PR #2215, opened 2024-12-04
SONARPY-2385 Add test for S5756 about uncallable lambda

thomas-serre-sonarsource:


maksim-grebeniuk-sonarsource (approved):
LGTM

maksim-grebeniuk-sonarsource (changes requested):
need to discuss the naming and probably add more test cases

python-checks/src/test/resources/checks/nonCallableCalled.py
```diff
@@ -407,3 +407,6 @@ def function_type_is_callable():
 def object_typevar():
     scheduled = []
     scheduled.pop()()  # OK
+
+def uncallable_lambda_are_non_compliant():
```
maksim-grebeniuk-sonarsource, line 411: rename to `non-callable inside lamda`

maksim-grebeniuk-sonarsource (approved):
LGTM

PR #2218, opened 2024-12-05
SONARPY-2407 Refactor Token#line and pythonLine to avoid mixing the two

ghislainpiot:
[SONARPY-2407](https://sonarsource.atlassian.net/browse/SONARPY-2407)

[SONARPY-2407]: https://sonarsource.atlassian.net/browse/SONARPY-2407?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python-checks/src/main/java/org/sonar/python/checks/FloatingPointEqualityCheck.java
+++ b/python-checks/src/main/java/org/sonar/python/checks/FloatingPointEqualityCheck.java
@@ -20,6 +20,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Set;
 import org.sonar.check.Rule;
+import org.sonar.plugins.python.api.PythonLine;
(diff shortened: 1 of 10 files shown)


joke1196 (approved):
LGTM! I was wondering if there was a way to test the quick fixes with the notebooks, but I think our test framework does not allow for that. Maybe for safety, as we should release on Monday, we could, when merged to master, redeploy peach to see if anything broke before going on weekend, WDYT?

ghislainpiot:
> LGTM! I was wondering if there was a way to test the quick fixes with the notebooks, but I think our test framework does not allow for that. Maybe for safety, as we should release on Monday, we could, when merged to master, redeploy peach to see if anything broke before going on weekend, WDYT?

I can put it on the dogfood and redeploy peach if needed, it would make me feel better too

PR #2225, opened 2024-12-09
Revert checking for FQN instead of a heuristic

Seppli11:
Part of SONARPY-2367


python-checks/src/test/resources/checks/hotspots/csrfDisabledCheck/flask/flaskExempt.py
```diff
@@ -11,3 +12,8 @@ def exemptExample():
   #            ^^^^^^
   def csrftestpost():
       pass
+
+  @app.route('/csrftest1/', methods=['POST'])
```
guillaume-dequenne-sonarsource, line 16: It could be nice to have a compliant test showing that the `exempt` is what triggers the issue, such as:
```
  @app.route('/csrftest1/', methods=['POST'])
  @exported_csrf_protect.something_else
  def compliant_from_reexport():
    pass
```

python-checks/src/test/resources/checks/hotspots/unsafeHttpMethods/flask/views.py
```diff
@@ -8,6 +9,21 @@
...
+# 127.0.0.1:5000/methods/sensitive1
+@methods.route('/sensitive1', methods=['GET', 'POST'])  # Noncompliant
+def sensitive1():
+    return Response("Method: " + request.method, 200)
+
+# 127.0.0.1:5000/methods/sensitive1
+@other_routes.route('/sensitive1', methods=['GET', 'POST'])  # Noncompliant
+def sensitive1_other_routes():
+    return Response("Method: " + request.method, 200)
+
```
guillaume-dequenne-sonarsource, line 26: I looks like the first and last tests are duplicate of each other, and the second one duplicates an existing test, and may not be valuable in the current context.

guillaume-dequenne-sonarsource (changes requested):
This looks good!
There are small things to fix with the tests, but it's good that they document the reliance on the existing heuristics.

PR #2229, opened 2024-12-09
Update rules metadata

guillaume-dequenne-sonarsource:


ghislainpiot (approved):
LGTM! 
Are you sure you have the latest API rule jar ? It feels like some HTML files are not stable

PR #2232, opened 2024-12-10
SONARPY-2429: Merge the ruling and the extended ruling

Seppli11:
[SONARPY-2429](https://sonarsource.atlassian.net/browse/SONARPY-2429)

This PR is based on [SONARPY-2430](https://github.com/SonarSource/sonar-python/tree/SONARPY-2430). Therefore, SONARPY-2430 has to be merged before this PR can be merged into [ruling-parallelization-master](https://github.com/SonarSource/sonar-python/tree/ruling-parallelization-master)

[SONARPY-2429]: https://sonarsource.atlassian.net/browse/SONARPY-2429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SONARPY-2430]: https://sonarsource.atlassian.net/browse/SONARPY-2430?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,10 +1,13 @@
 [submodule "its/sources_ruling"]
 	path = its/sources_ruling
 	url = https://github.com/SonarCommunity/python-test-sources.git
-	branch = extended
+	branch = master-temp
+	shallow = true
 [submodule "tools/typeshed_serializer/resources/typeshed"]
 	path = python-frontend/typeshed_serializer/resources/typeshed
 	url = https://github.com/python/typeshed.git
+	shallow = true
 [submodule "tools/typeshed_serializer/resources/python-type-stubs"]
 	path = python-frontend/typeshed_serializer/resources/python-type-stubs
 	url = https://github.com/microsoft/python-type-stubs
+	shallow = true


.gitmodules
```diff
@@ -1,8 +1,5 @@
-[submodule "its/sources"]
-	path = its/sources
-	url = https://github.com/SonarCommunity/python-test-sources.git
-[submodule "its/sources_extended"]
-	path = its/sources_extended
+[submodule "its/sources_ruling"]
+	path = its/sources_ruling
 	url = https://github.com/SonarCommunity/python-test-sources.git
 	branch = extended
```
guillaume-dequenne-sonarsource, line 4: I'm a bit confused here, how are we actually retaining the original ruling sources (from `master`)? It seems the only submodule defined is for the `extended` branch and that branch hasn't changed?
I'd have expected to see the `master-temp` branch here?

guillaume-dequenne-sonarsource (commented):
This looks good! I just have one question, because I don't understand how the current submodule setup works, but this looks very promising!

.gitmodules
Seppli11: On the python-test-sources side, I've essentially merged the two branches into `master-temp`, which is checked out in the `its/sources_ruling` submodule. 
According to [this StackOverflow answer](https://stackoverflow.com/questions/5033441/where-does-git-store-the-sha1-of-the-commit-for-a-submodule), git keeps track of the hash of the submodules directly in its database. That's probably the reason for the following diff
image

Seppli11: I've changed the .gitmodules and added `shallow = true` (see [here](https://git-scm.com/docs/gitmodules#Documentation/gitmodules.txt-submoduleltnamegtshallow) for documentation)

guillaume-dequenne-sonarsource (approved):
LGTM!
Please address the issue on Next before merging (I think we can simply accept it tbh).

PR #2235, opened 2024-12-11
SONARPY-2451 SONARPY-2459 SONARPY-2464 Collect data for the Jupyter notebooks

ghislainpiot:
[SONARPY-2451](https://sonarsource.atlassian.net/browse/SONARPY-2451)

SONARPY-2459

[SONARPY-2451]: https://sonarsource.atlassian.net/browse/SONARPY-2451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPynbSensor.java
+++ b/sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPynbSensor.java
@@ -47,6 +47,7 @@ public final class IPynbSensor implements Sensor {
   private final NoSonarFilter noSonarFilter;
   private final PythonIndexer indexer;
   private static final String FAIL_FAST_PROPERTY_NAME = "sonar.internal.analysis.failFast";
+  private final SensorTelemetryStorage sensorTelemetryStorage;
 
   public IPynbSensor(FileLinesContextFactory fileLinesContextFactory, CheckFactory checkFactory, NoSonarFilter noSonarFilter) {
     this(fileLinesContextFactory, checkFactory, noSonarFilter, null);
@@ -58,6 +59,7 @@ public final class IPynbSensor implements Sensor {
     this.fileLinesContextFactory = fileLinesContextFactory;
     this.noSonarFilter = noSonarFilter;
     this.indexer = indexer;
+    this.sensorTelemetryStorage = new SensorTelemetryStorage();
   }
 
   @Override
@@ -80,6 +82,7 @@ public final class IPynbSensor implements Sensor {
     } else {
       processNotebooksFiles(pythonFiles, context);
     }
+    sensorTelemetryStorage.send(context);
   }
 
   private void processNotebooksFiles(List<PythonInputFile> pythonFiles, SensorContext context) {
@@ -89,20 +92,29 @@ public final class IPynbSensor implements Sensor {
     PythonIndexer pythonIndexer = new SonarQubePythonIndexer(pythonFiles, cacheContext, context);
     PythonScanner scanner = new PythonScanner(context, checks, fileLinesContextFactory, noSonarFilter, PythonParser.createIPythonParser(), pythonIndexer);
     scanner.execute(pythonFiles, context);
+    sensorTelemetryStorage.updateMetric(SensorTelemetryStorage.MetricKey.NOTEBOOK_RECOGNITION_ERROR_KEY, String.valueOf(scanner.getRecognitionErrorCount()));
   }
 
-  private static List<PythonInputFile> parseNotebooks(List<PythonInputFile> pythonFiles, SensorContext context) {
+  private List<PythonInputFile> parseNotebooks(List<PythonInputFile> pythonFiles, SensorContext context) {
     List<PythonInputFile> generatedIPythonFiles = new ArrayList<>();
+
+    sensorTelemetryStorage.updateMetric(SensorTelemetryStorage.MetricKey.NOTEBOOK_TOTAL_KEY, String.valueOf(pythonFiles.size()));
+    var numberOfExceptions = 0;
+
     for (PythonInputFile inputFile : pythonFiles) {
       try {
+        sensorTelemetryStorage.updateMetric(SensorTelemetryStorage.MetricKey.NOTEBOOK_PRESENT_KEY, "1");
         var result = IpynbNotebookParser.parseNotebook(inputFile);
         result.ifPresent(generatedIPythonFiles::add);
       } catch (Exception e) {
+        numberOfExceptions++;
         if (context.config().getBoolean(FAIL_FAST_PROPERTY_NAME).orElse(false) && !isErrorOnTestFile(inputFile)) {
           throw new IllegalStateException("Exception when parsing " + inputFile, e);
         }
       }
     }
+
+    sensorTelemetryStorage.updateMetric(SensorTelemetryStorage.MetricKey.NOTEBOOK_EXCEPTION_KEY, String.valueOf(numberOfExceptions));
     return generatedIPythonFiles;
   }
 
--- a/sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPynbSensor.java
+++ b/sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPynbSensor.java
@@ -92,18 +92,18 @@ public final class IPynbSensor implements Sensor {
     PythonIndexer pythonIndexer = new SonarQubePythonIndexer(pythonFiles, cacheContext, context);
     PythonScanner scanner = new PythonScanner(context, checks, fileLinesContextFactory, noSonarFilter, PythonParser.createIPythonParser(), pythonIndexer);
     scanner.execute(pythonFiles, context);
-    sensorTelemetryStorage.updateMetric(SensorTelemetryStorage.MetricKey.NOTEBOOK_RECOGNITION_ERROR_KEY, String.valueOf(scanner.getRecognitionErrorCount()));
+    sensorTelemetryStorage.updateMetric(TelemetryMetricKey.NOTEBOOK_RECOGNITION_ERROR_KEY, scanner.getRecognitionErrorCount());
   }
 
   private List<PythonInputFile> parseNotebooks(List<PythonInputFile> pythonFiles, SensorContext context) {
     List<PythonInputFile> generatedIPythonFiles = new ArrayList<>();
 
-    sensorTelemetryStorage.updateMetric(SensorTelemetryStorage.MetricKey.NOTEBOOK_TOTAL_KEY, String.valueOf(pythonFiles.size()));
+    sensorTelemetryStorage.updateMetric(TelemetryMetricKey.NOTEBOOK_TOTAL_KEY, pythonFiles.size());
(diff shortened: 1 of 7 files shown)


sonar-python-plugin/src/main/java/org/sonar/plugins/python/SensorTelemetryStorage.java
```diff
@@ -0,0 +1,44 @@
...
+  public static final String NOTEBOOK_PRESENT_KEY = "python.notebook.present";
+  public static final String NOTEBOOK_TOTAL_KEY = "python.notebook.total";
+  public static final String NOTEBOOK_PARSE_ERROR_KEY = "python.notebook.parse_error";
+  public static final String NOTEBOOK_RECOGNITION_ERROR_KEY = "python.notebook.recognition_error";
+  public static final String NOTEBOOK_EXCEPTION_KEY = "python.notebook.exceptions";
+  final Map<String, String> data;
+
+  public SensorTelemetryStorage() {
+    data = new HashMap<>();
+    data.put(NOTEBOOK_PRESENT_KEY, "false");
```
joke1196, line 37: I think we could hide the implementation detail of the `Map` by making it private and accessing it through an API that is more domain oriented, i.e. `updateMetrics`. Following this direction We could also provide a specific type for the metrics key, i.e.  `TelemetryKey` so that user to not pass any string to the Telemetry but only predefined key. I think it is good to be strict here so that we do not have a mess later on when other devs try to use the telemetry API

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPynbSensor.java
```diff
@@ -80,29 +87,42 @@ public void execute(SensorContext context) {
...
     scanner.execute(pythonFiles, context);
+    sensorTelemetryStorage.data.put(SensorTelemetryStorage.NOTEBOOK_RECOGNITION_ERROR_KEY, String.valueOf(scanner.getRecognitionErrorCount()));
   }
 
-  private static List<PythonInputFile> parseNotebooks(List<PythonInputFile> pythonFiles, SensorContext context) {
+  private List<PythonInputFile> parseNotebooks(List<PythonInputFile> pythonFiles, SensorContext context) {
     List<PythonInputFile> generatedIPythonFiles = new ArrayList<>();
+
+    sensorTelemetryStorage.data.put(SensorTelemetryStorage.NOTEBOOK_TOTAL_KEY, String.valueOf(pythonFiles.size()));
+    final int[] numberOfFailedParsing = {0};
```
joke1196, line 107: I guess this is a workaround as you need a final var for the lambda? I think it would be more readable to an if 
```suggestion
int numberOfFailedParsing = 0;
if(result.isPresent()){
  generatedIPythonFiles.add(result.get());
}else {
  numberOfFailedParsing++;
}
```
instead of trying to work against java.

its/plugin/it-python-plugin-test/src/test/java/com/sonar/python/it/plugin/TestsUtils.java
```diff
@@ -50,6 +50,8 @@ public final class TestsUtils {
     .useDefaultAdminCredentialsForBuilds(true)
     .setSonarVersion(System.getProperty(SQ_VERSION_PROPERTY, DEFAULT_SQ_VERSION))
     .addPlugin(PLUGIN_LOCATION)
+    .setServerProperty("sonar.telemetry.enable", "false") // Disable telemetry waiting for ORCH-497
```
joke1196, line 53: I don't think this should be part of this PR. Maybe a merge issue?

joke1196 (changes requested):
This looks quite good. I added suggestions for the shape of the API, and the workaround the functional API of java.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/SensorTelemetryStorage.java
```diff
@@ -25,20 +25,36 @@
...
-  public static final String NOTEBOOK_TOTAL_KEY = "python.notebook.total";
-  public static final String NOTEBOOK_PARSE_ERROR_KEY = "python.notebook.parse_error";
-  public static final String NOTEBOOK_RECOGNITION_ERROR_KEY = "python.notebook.recognition_error";
-  public static final String NOTEBOOK_EXCEPTION_KEY = "python.notebook.exceptions";
+  public static final MetricKey NOTEBOOK_PRESENT_KEY = new MetricKey("python.notebook.present");
+  public static final MetricKey NOTEBOOK_TOTAL_KEY = new MetricKey("python.notebook.total");
+  public static final MetricKey NOTEBOOK_PARSE_ERROR_KEY = new MetricKey("python.notebook.parse_error");
+  public static final MetricKey NOTEBOOK_RECOGNITION_ERROR_KEY = new MetricKey("python.notebook.recognition_error");
+  public static final MetricKey NOTEBOOK_EXCEPTION_KEY = new MetricKey("python.notebook.exceptions");
   final Map<String, String> data;
```
joke1196, line 32: I think we could make this private

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPynbSensor.java
```diff
@@ -80,29 +87,40 @@ public void execute(SensorContext context) {
     } else {
       processNotebooksFiles(pythonFiles, context);
     }
+    sensorTelemetryStorage.send(context);
   }
 
   private void processNotebooksFiles(List<PythonInputFile> pythonFiles, SensorContext context) {
-    pythonFiles = parseNotebooks(pythonFiles, context);
+    pythonFiles = this.parseNotebooks(pythonFiles, context);
```
maksim-grebeniuk-sonarsource, line 94: is this `this` needed?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/SensorTelemetryStorage.java
```diff
@@ -0,0 +1,79 @@
...
+    } catch (Exception e) {
+      LOG.error("Failed to send metrics", e);
+    }
+  }
+
+  public void updateMetric(MetricKey key, String value) {
+    data.put(key.key(), value);
+  }
+
+  public enum MetricKey {
```
maksim-grebeniuk-sonarsource, line 63: Since this enum massively used outside of the `SensorTelemetryStorage` - probably it makes sense to make it a top-level class

sonar-python-plugin/src/main/java/org/sonar/plugins/python/SensorTelemetryStorage.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.sonar.api.batch.sensor.SensorContext;
+import org.sonar.api.utils.Version;
+
+public class SensorTelemetryStorage {
+  private static final Logger LOG = LoggerFactory.getLogger(SensorTelemetryStorage.class);
+
+  private final Map<String, String> data;
```
maksim-grebeniuk-sonarsource, line 30: why the key of the `data` map can't be of `MetricKey` type?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/SensorTelemetryStorage.java
```diff
@@ -0,0 +1,79 @@
...
+
+  public SensorTelemetryStorage() {
+    data = new HashMap<>();
+  }
+
+  public Map<String, String> data() {
```
maksim-grebeniuk-sonarsource, line 36: it seems that this method is used for test purposes only, so probably it should not be public.

maksim-grebeniuk-sonarsource (approved):
LGTM, as discussed some small changes required but no need to revisit again

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPynbSensor.java
```diff
@@ -80,29 +82,39 @@ public void execute(SensorContext context) {
...
-  private static List<PythonInputFile> parseNotebooks(List<PythonInputFile> pythonFiles, SensorContext context) {
+  private List<PythonInputFile> parseNotebooks(List<PythonInputFile> pythonFiles, SensorContext context) {
     List<PythonInputFile> generatedIPythonFiles = new ArrayList<>();
+
+    sensorTelemetryStorage.updateMetric(TelemetryMetricKey.NOTEBOOK_TOTAL_KEY, String.valueOf(pythonFiles.size()));
+    var numberOfExceptions = 0;
+
     for (PythonInputFile inputFile : pythonFiles) {
       try {
+        sensorTelemetryStorage.updateMetric(TelemetryMetricKey.NOTEBOOK_PRESENT_KEY, "1");
```
Seppli11, line 106: what happens when the key is not present?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/SensorTelemetryStorage.java
```diff
@@ -0,0 +1,58 @@
...
+
+      } else {
+        LOG.info("Skipping sending metrics because the plugin API version is {}", apiVersion);
+      }
+    } catch (Exception e) {
+      LOG.error("Failed to send metrics", e);
+    }
+  }
+
+  public void updateMetric(TelemetryMetricKey key, String value) {
```
Seppli11, line 54: There are a few calls to `updateMetric` where `value` is an `int` and `String.valueOf` is called. Would it maybe make sense to have an overload?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/SensorTelemetryStorage.java
```diff
@@ -0,0 +1,58 @@
...
+import org.sonar.api.batch.sensor.SensorContext;
+import org.sonar.api.utils.Version;
+
+public class SensorTelemetryStorage {
+  private static final Logger LOG = LoggerFactory.getLogger(SensorTelemetryStorage.class);
+
+  private final Map<TelemetryMetricKey, String> data;
+
+  public SensorTelemetryStorage() {
+    data = new EnumMap<>(TelemetryMetricKey.class);
```
Seppli11, line 32: nitpick: the assignment could be moved to the field declaration and the constructor can be removed

Seppli11 (changes requested):
It looks good to me. However, I had a question during review and there are some minor nitpicks.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/IPynbSensor.java
ghislainpiot: The field is just not present. I don't think it is possible to have a default value

Seppli11: That doesn't pose an issue to analyze the data afterward?

PR #2236, opened 2024-12-11
SONARPY-2432: Split build into building and testing

Seppli11:


.cirrus.yml
```diff
@@ -94,10 +94,29 @@ build_task:
     - git checkout
     - cd -
     - source cirrus-env BUILD
-    - regular_mvn_build_deploy_analyze -DfailStubGenerationFast=true
+    - regular_mvn_build_deploy_analyze -DfailStubGenerationFast=true -DskipTypshed -DskipTests=true -Dsonar.skip=true
```
guillaume-dequenne-sonarsource, line 97: If we're not executing the tests, could we even go one step further and not initialize the submodules?
Or do we still expect things to fail here in case of checksum mismatch?

Seppli11: I think you're right, we don't need it here. I moved it down to the analyze_task, where we probably do need it.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #2243, opened 2024-12-13
SONARPY-2464 Add safety around the telemetry mechanism

ghislainpiot:
[SONARPY-2464](https://sonarsource.atlassian.net/browse/SONARPY-2464)

[SONARPY-2464]: https://sonarsource.atlassian.net/browse/SONARPY-2464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sonar-python-plugin/src/main/java/org/sonar/plugins/python/SensorTelemetryStorage.java
+++ b/sonar-python-plugin/src/main/java/org/sonar/plugins/python/SensorTelemetryStorage.java
@@ -35,7 +35,6 @@ public class SensorTelemetryStorage {
 
   public SensorTelemetryStorage() {
     data = new HashMap<>();
-    data.put(NOTEBOOK_PRESENT_KEY.key(), "false");
   }
 
   public Map<String, String> data() {
--- a/sonar-python-plugin/src/main/java/org/sonar/plugins/python/SensorTelemetryStorage.java
+++ b/sonar-python-plugin/src/main/java/org/sonar/plugins/python/SensorTelemetryStorage.java
@@ -22,6 +22,7 @@ import java.util.Map;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.sonar.api.batch.sensor.SensorContext;
+import org.sonar.api.utils.Version;
 
 public class SensorTelemetryStorage {
   private static final Logger LOG = LoggerFactory.getLogger(SensorTelemetryStorage.class);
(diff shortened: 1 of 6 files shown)


sonar-python-plugin/src/main/java/org/sonar/plugins/python/SensorTelemetryStorage.java
```diff
@@ -42,8 +43,19 @@ public Map<String, String> data() {
   }
 
   public void send(SensorContext sensorContext) {
-    data.forEach(sensorContext::addTelemetryProperty);
     data.forEach((k, v) -> LOG.info("Metrics property: {}={}", k, v));
+    var apiVersion = sensorContext.runtime().getApiVersion();
+    if (apiVersion.isGreaterThanOrEqual(Version.create(10, 9))) {
+      // This try/catch block should be useless, as in the worst case it should be a no-op depending on the SensorContext implementation
+      // It exists to be extra sure for the LTA
```
maksim-grebeniuk-sonarsource, line 50: If you want to be super-sure that no exception will be thrown up then probably the `try` statement should wrap entire method like:
```java
try {
  data.forEach((k, v) -> LOG.info("Metrics property: {}={}", k, v));
  var apiVersion = sensorContext.runtime().getApiVersion();
  if (apiVersion.isGreaterThanOrEqual(Version.create(10, 9))) {
    data.forEach(sensorContext::addTelemetryProperty);
  } else  {
    LOG.info("Skipping sending metrics because the plugin API version is {}", apiVersion);
  }
} catch (Exception e) {
  LOG.error("Failed to send metrics", e);
}
```

sonar-python-plugin/src/main/java/org/sonar/plugins/python/SensorTelemetryStorage.java
```diff
@@ -42,8 +43,19 @@ public Map<String, String> data() {
...
+      try {
+        data.forEach(sensorContext::addTelemetryProperty);
```
maksim-grebeniuk-sonarsource, line 52: also, why do we need to iterate over the `data` map twice?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/SensorTelemetryStorage.java
```diff
@@ -42,8 +43,19 @@ public Map<String, String> data() {
...
+      } catch (Exception e) {
+        LOG.error("Failed to send metrics", e);
+      }
+    } else {
+      LOG.info("Skipping sending metrics because the plugin API version is {}", apiVersion);
+    }
   }
 
   public void updateMetric(MetricKey key, String value) {
```
maksim-grebeniuk-sonarsource, line 61: I have a question: why MetricKey is not an Enum?

ghislainpiot:
I have applied all comments, I will put everything in the same branch

PR #2248, opened 2024-12-13
SONARPY-2462: Parallelize the qa_plugin tests

Seppli11:
[SONARPY-2462](https://sonarsource.atlassian.net/browse/SONARPY-2462)

[SONARPY-2462]: https://sonarsource.atlassian.net/browse/SONARPY-2462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


its/plugin/it-python-plugin-test/src/test/java/com/sonar/python/it/plugin/NotebookPluginTest.java
```diff
@@ -17,38 +17,34 @@
...
-import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
 import org.sonarqube.ws.Issues;
 
 import static com.sonar.python.it.plugin.TestsUtils.issues;
 import static org.assertj.core.api.Assertions.assertThat;
 
+@Disabled
```
guillaume-dequenne-sonarsource, line 30: Was that test not run before? And it's actually failing?

its/plugin/it-python-plugin-test/src/test/java/com/sonar/python/it/plugin/NotebookPluginTest.java
```diff
@@ -17,38 +17,34 @@
...
   private static final String PROJECT_KEY = "ipynb_json_project";
 
   @RegisterExtension
-  public static final OrchestratorExtension ORCHESTRATOR = TestsUtils.ORCHESTRATOR;
+  public static final ConcurrentOrchestratorExtension ORCHESTRATOR = TestsUtils.ORCHESTRATOR;
 
-  @BeforeAll
-  static void startServer() {
+  @Test
+  void test() {
```
guillaume-dequenne-sonarsource, line 39: Running the test locally, the issues do look legit. We could update the assertions (I believe asserting an empty list of issue was a mistake).
We can do that here or in a separate PR, as you prefer.

Seppli11: I'll change the assertion. I think it should be doable in this PR

its/plugin/it-python-plugin-test/src/test/java/com/sonar/python/it/plugin/NotebookPluginTest.java
Seppli11: As far as I can tell, it wasn't run before. I've tried running this test on master and it failed as well. So, it seems that it failing is not connected to changes made in this PR.

its/plugin/it-python-plugin-test/src/test/java/com/sonar/python/it/plugin/ConcurrentOrchestratorExtension.java
```diff
@@ -0,0 +1,82 @@
...
+
+  public SonarScanner createSonarScanner() {
+    return SonarScanner.create();
+  }
+
+  public static ConcurrentOrchestratorExtensionBuilder builderEnv() {
+    return new ConcurrentOrchestratorExtensionBuilder(Configuration.createEnv());
+  }
+
+  public static ConcurrentOrchestratorExtensionBuilder builder(Configuration config) {
```
guillaume-dequenne-sonarsource, line 64: This seems unused, do we need it?

guillaume-dequenne-sonarsource (approved):
Small comment on unsued method.
LGTM otherwise, great work!

its/plugin/it-python-plugin-test/src/test/java/com/sonar/python/it/plugin/ConcurrentOrchestratorExtension.java
Seppli11: Good catch. You're right, we don't need it. thx

PR #2252, opened 2024-12-16
SONARPY-2457 Collect data for the Python version

ghislainpiot:
[SONARPY-2457](https://sonarsource.atlassian.net/browse/SONARPY-2457)

[SONARPY-2457]: https://sonarsource.atlassian.net/browse/SONARPY-2457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonSensor.java
+++ b/sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonSensor.java
@@ -68,17 +68,14 @@ public final class PythonSensor implements Sensor {
   static final String UNSET_VERSION_WARNING = "Your code is analyzed as compatible with all Python 3 versions by default." +
     " You can get a more precise analysis by setting the exact Python version in your configuration via the parameter \"sonar.python.version\"";
 
+  private final SensorTelemetryStorage sensorTelemetryStorage;
+
   /**
    * Constructor to be used by pico if neither PythonCustomRuleRepository nor PythonIndexer are to be found and injected.
    */
   public PythonSensor(FileLinesContextFactory fileLinesContextFactory, CheckFactory checkFactory,
     NoSonarFilter noSonarFilter, AnalysisWarningsWrapper analysisWarnings) {
-    this(fileLinesContextFactory, checkFactory, noSonarFilter, null, null, null, analysisWarnings);
-  }
-
-  public PythonSensor(FileLinesContextFactory fileLinesContextFactory, CheckFactory checkFactory, NoSonarFilter noSonarFilter,
-    PythonCustomRuleRepository[] customRuleRepositories, AnalysisWarningsWrapper analysisWarnings) {
-    this(fileLinesContextFactory, checkFactory, noSonarFilter, customRuleRepositories, null, null, analysisWarnings);
+    this(fileLinesContextFactory, checkFactory, noSonarFilter, null, null, null, analysisWarnings, new SensorTelemetryStorage());
   }
 
   public PythonSensor(FileLinesContextFactory fileLinesContextFactory, CheckFactory checkFactory, NoSonarFilter noSonarFilter,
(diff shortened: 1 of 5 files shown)


sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonSensor.java
```diff
@@ -68,30 +68,33 @@ public final class PythonSensor implements Sensor {
...
 
   public PythonSensor(FileLinesContextFactory fileLinesContextFactory, CheckFactory checkFactory, NoSonarFilter noSonarFilter,
     @Nullable PythonCustomRuleRepository[] customRuleRepositories, @Nullable PythonIndexer indexer,
     @Nullable SonarLintCache sonarLintCache, AnalysisWarningsWrapper analysisWarnings) {
+    this(fileLinesContextFactory, checkFactory, noSonarFilter, customRuleRepositories, indexer, sonarLintCache, analysisWarnings, new SensorTelemetryStorage());
+  }
+
+  public PythonSensor(FileLinesContextFactory fileLinesContextFactory, CheckFactory checkFactory, NoSonarFilter noSonarFilter,
+    @Nullable PythonCustomRuleRepository[] customRuleRepositories, @Nullable PythonIndexer indexer,
+    @Nullable SonarLintCache sonarLintCache, AnalysisWarningsWrapper analysisWarnings, SensorTelemetryStorage sensorTelemetryStorage) {
```
guillaume-dequenne-sonarsource, line 97: I'm not sure I understand in which case we should expect this `sensorTelemetryStorage` to be injected and in which case we should expect to instantiate it ourselves, and what difference it makes, can you explain?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonSensor.java
```diff
@@ -121,15 +130,25 @@ public void execute(SensorContext context) {
...
     pythonIndexer.setSonarLintCache(sonarLintCache);
     TypeShed.setProjectLevelSymbolTable(pythonIndexer.projectLevelSymbolTable());
     PythonScanner scanner = new PythonScanner(context, checks, fileLinesContextFactory, noSonarFilter, PythonParser.create(), pythonIndexer);
     scanner.execute(pythonFiles, context);
+    sensorTelemetryStorage.send(context);
     durationReport.stop();
   }
 
+  private void setTelemetry(SensorContext context, String[] pythonVersionParameter) {
+    var isVersionSet = pythonVersionParameter.length != 0 || context.runtime().getProduct() == SonarProduct.SONARLINT;
```
guillaume-dequenne-sonarsource, line 145: Why do we consider the version to be set in SonarLint context? Wouldn't that skew the data and imply more people actively set the version than what actually happens?

sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonSensor.java
```diff
@@ -121,15 +130,25 @@ public void execute(SensorContext context) {
...
+    if (pythonVersionParameter.length != 0) {
+      sensorTelemetryStorage.updateMetric(TelemetryMetricKey.PYTHON_VERSION_KEY, String.join(",", pythonVersionParameter));
```
guillaume-dequenne-sonarsource, line 147: So that means we can set more than just booleans for the telemetry?

guillaume-dequenne-sonarsource (commented):
I'm missing some context on this topic, so I might miss things, and I'm taking what was already implemented as reference.

This looks good though! I left some questions for my own comprehension. My only concern is about the strategy for the `isVersionSet` key that I could see skewed when counting SonarLint.

sonar-python-plugin/src/main/java/org/sonar/plugins/python/PythonSensor.java
ghislainpiot: Yes, CFamily is sending strings.

Seppli11 (approved):
LGTM!
Out of curiosity, why don't we send telemetry from sonarlint? Would it just screw up the data too much?

ghislainpiot:
> LGTM! Out of curiosity, why don't we send telemetry from sonarlint? Would it just screw up the data too much?

I don't think telemetry is sent right now on their side anyways.
It would also probably screw up the stats

PR #2256, opened 2024-12-17
SONARPY-2475 split ruling into two

Seppli11:


Seppli11:
This PR nets us ~30 seconds increase for the ruling. However, it is quite a hacky way of doing it. Since 30 seconds isn't that much, I'll close the PR

PR #2293, opened 2026-07-15
SONARPY-998 Add regression coverage for Mapping compatibility

erwan-leforestier-sonarsource:
## Summary
Adds regression coverage for the SONARPY-998 false-positive scenario so S5655 and S5890 keep accepting `dict` values for `Mapping` annotations when `sonar.python.version` is unset. It also locks the reported `3.9` workaround path with explicit versioned tests.

## Changes
- add small dedicated S5655 and S5890 fixtures for `Mapping`-annotated `dict` values
- verify the scenario explicitly when the Python version is unspecified
- verify the same scenario explicitly when the Python version is pinned to `3.9`

:warning::warning: This is not ready for review :warning::warning:

### DIFF
--- a/python-checks/src/test/java/org/sonar/python/checks/ArgumentTypeCheckTest.java
+++ b/python-checks/src/test/java/org/sonar/python/checks/ArgumentTypeCheckTest.java
@@ -16,11 +16,22 @@
  */
 package org.sonar.python.checks;
 
+import java.util.EnumSet;
 import java.util.List;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
+import org.sonar.plugins.python.api.ProjectPythonVersion;
+import org.sonar.plugins.python.api.PythonVersionUtils;
 import org.sonar.python.checks.utils.PythonCheckVerifier;
+import org.sonar.python.types.TypeShed;
 
 class ArgumentTypeCheckTest {
+
+  @AfterEach
+  void reset_python_version() {
+    ProjectPythonVersion.setCurrentVersions(PythonVersionUtils.allVersions());
+    TypeShed.resetBuiltinSymbols();
+  }
 
   @Test
   void test() {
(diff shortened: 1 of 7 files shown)


erwan-leforestier-sonarsource:
Closing this PR because the reported FP is already fixed in the current analyzer and this branch only adds regression tests on top of that existing behavior.

### Why this is not an issue anymore

- The S5655 and S5890 scenarios from SONARPY-998 no longer reproduce on current `master`.
- In practice, `dict` is already accepted as compatible with `Mapping` in the reported cases, including when `sonar.python.version` is left unset.
- The branch behind this PR contains test-only changes; there was no production change left to make.

### How this was verified

- I used the exact repro from the Jira ticket and the community thread and added focused local regression checks for:
  - unspecified Python version
  - explicit `sonar.python.version=3.9`
- I ran:

```bash
IS_COMMUNITY=true mvn -pl python-checks -am \
  -Dtest=org.sonar.python.checks.ArgumentTypeCheckTest,org.sonar.python.checks.InconsistentTypeHintCheckTest \
  -Dsurefire.failIfNoSpecifiedTests=false test

IS_COMMUNITY=true mvn -pl python-frontend \
  -Dtest=org.sonar.python.types.InferredTypesTest#dict_is_compatible_with_mapping_annotations test
```

- Both checks passed.
- I also confirmed that the diff in this PR is test-only.

### Timeline

- 2022-04-06: community report opened:
  https://community.sonarsource.com/t/s5655-s5890-fp-dict-literal-not-recognized-as-mapping/61197
- 2022-04-06: `SONARPY-998` was created from that report. In the same community thread, `sonar.python.version=3.9` was suggested as a workaround.
- 2023-06-05: the underlying fix first landed in the typeshed refresh work in PR #1477:
  https://github.com/SonarSource/sonar-python/pull/1477
- 2023-06-19: the broader follow-up stub-generation work in PR #1478 was merged:
  https://github.com/SonarSource/sonar-python/pull/1478

The short version is that SONARPY-998 stayed open, but the underlying bug was fixed indirectly by the 2023 typeshed/stub refresh work, so there is no analyzer change left to merge here.

